PR #3, opened 2014-07-28
7.0 account invoice unique customer ref

mcassuto:


mcassuto:
@rvalyi @pedrobaeza @bwrsandman @max3903 : here the PR in the  correct branch.

A great thanks to all

max3903:
:+1:

rvalyi:
:+1:

pedrobaeza:
Hi, I've been thinking what @rvalyi said about ids, and the constraint you are using is processed via ORM, not via Postgres. If you were using _sql_constraints_, then Postgres would process it. So, I think you need to check every id for duplication, not only the first one.

Regards.

mcassuto:
Hi @pedrobaeza ,

these modules are only tested against 7.0; not against 8.0.

I really not sure about the use case: how can you have the same record modified by the same user or 2 different users at the exactly same time ?? 
I did the modification asked anyway ;-)

pedrobaeza:
Thanks :+1:

mcassuto:
@bwrsandman: .pot/.po files are now correct, containing all the needed strings

PR #4, opened 2014-08-13
7.0 metafiles

pedrobaeza:


rvalyi:
:+1:  LGTM

yvaucher:
:-1: 

A dependency is missing:

except_orm: (u'Error', u"You try to install module 'account_invoice_template' that depends on module 'account_move_template'.\nBut the latter module is not available in your system.")

pedrobaeza:
Remaining error to be catched in another PR.

PR #8, opened 2014-08-16
sale_order_partial_invoice - Some fixes and enhancements:

pedrobaeza:
[IMP] Make readonly wizard non-modifiable fields.
[FIX] Check if the quantity to invoice is greater than the remaining one.
[FIX] Made a little hack to avoid the creation of counterpart invoice lines in subsequent invoices.
[FIX] Write 'progress' state manually when order is fully invoiced, because workflow doesn't do it.
[IMP] Open the resulting invoice after the wizard.
[FIX] Change type of name field to text to avoid format loosing.

### DIFF
--- a/sale_order_partial_invoice/i18n/es.po
+++ b/sale_order_partial_invoice/i18n/es.po
@@ -6,108 +6,14 @@ msgid ""
 msgstr ""
 "Project-Id-Version: OpenERP Server 7.0\n"
 "Report-Msgid-Bugs-To: \n"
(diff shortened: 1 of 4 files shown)


gurneyalex:
could you add an automated test for the "little hack"?

pedrobaeza:
@gurneyalex I have extended one test to check the hack. Travis errors are unrelated (pylint issues). Please review it.

eLBati:
:+1: thanks

jgrandguillaume:
:+1:

PR #12, opened 2014-09-04
[add] free invoice line module

scigghia:
Hi all!
This module adds management of Free invoice lines (omaggio in fattura for Italian accounting)

### DIFF
--- a/free_invoice_line/__init__.py
+++ b/free_invoice_line/__init__.py
@@ -1,0 +1,25 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    OpenERP, Open Source Management Solution
+#    Copyright (c) 2012 Andrea Cometa All Rights Reserved.
+#                       www.andreacometa.it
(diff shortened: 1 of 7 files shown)


pedrobaeza:
Hi, Andrea, can you ellaborate the user case for this module?

scigghia:
Hi Pedro,
it is usefull when user needs to invoice one or more "free samples", then the amount for these lines  and tax lines must be 0

pedrobaeza:
Yeah,  but this can be easily achieved filling a 100% discount or emptying the price.

Technically speaking, you're entirely overwriting `_amount_all`, which is a very bad bad practise, so for now :-1:

alexis-via:
I have the same opinion as Pedro. It don't understand why we would need a full module with so many lines of code just to have an invoice line at 0 ; you could set the unit_price to 0 or set the discount to 100%... and you would get the same result, no ?

joaoalf:
:-1: 
Agree with others.

PR #19, opened 2014-10-07
account_invoice_force_number : Port to v8

alexis-via:
I ported the module account_invoice_force_number to v8 with new API

I also:
- updated POT file
- added French translation
- use more flexible xpath for the inherit of invoice form views


tafaRU:
LGTM :+1: 
Except for the key 'active' I think should be removed according to https://github.com/nbessi/openerp-conventions/blob/master/v7/manifest_checker.py @nbessi please could you confirm if still valid in v8? Thank you.

pedrobaeza:
:+1: 

`active` key doesn't hurt, but it doesn't add either any value.

eLBati:
`git diff origin/7.0..akretion/8.0  account_invoice_force_number/` gives me http://pastie.org/9630430

So, :+1:

PR #24, opened 2014-10-21
[FIX] add payment term rounding issue if balance

vrenaville:
In case of a line of payment term set to balance none is return instead of the rounded value amount

### DIFF
--- a/payment_term_rounding/account.py
+++ b/payment_term_rounding/account.py
@@ -65,7 +65,7 @@ class AccountPaymentTermLine(orm.Model):
                 amt = float_round(amt, precision_rounding=line.amount_round)
             return float_round(amt, precision_digits=prec)
         elif line.value == 'balance':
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Please rebase your branch to enable automatic merge, and why are you rounding again the amount. I think that with only this line it's enough:

```
-amt = float_round(remaining_amount, precision_digits=prec)
+return float_round(remaining_amount, precision_digits=prec)
```

vrenaville:
Hello,
I did the mistake, two rounding is not useful ;) , i replace it in my last commit with just a return

pedrobaeza:
Rebase it's still needed. Please do it.

pedrobaeza:
Thanks :+1:

tafaRU:
:+1:

yvaucher:
:+1:

PR #25, opened 2014-10-27
Don't init taxes with False if doesn't exists.

feketemihai:
IF they are init with False, the onchange method raise error on map_tax (the v8 function doesn't allow checking if taxes exists...i created a pull request but Martin said to fix it from the call).


feketemihai:
I agree with you and it was ok with False, but only on 7.0 but since the 8.0 version doesn't provide the existance it's working just like in account/partner.py because taxes are already a list...

pedrobaeza:
OK, I get it. Thanks for the explanation.

:+1:

yvaucher:
:+1:

PR #26, opened 2014-10-27
Update invoice_fiscal_position_update with better use of the new API

alexis-via:
Inspired by the work of @eLBati on purchase_fiscal_position_update


invoice_fiscal_position_update/account_invoice.py
```diff
@@ -28,30 +28,17 @@
...
-            self, fiscal_position_id, type, invoice_line):
-        '''Function executed by the on_change on the fiscal_position field
-        of invoice ; it updates taxes and accounts on all invoice lines'''
-        fp_obj = self.env['account.fiscal.position']
-        res = {'value': {}}
-        line_dict = self.resolve_2many_commands(
-            'invoice_line', invoice_line)
+    @api.onchange('fiscal_position')
+    def fiscal_position_change(self):
+        '''Updates taxes and accounts on all invoice lines'''
```
pedrobaeza: Please use doble quote for docstrings (before http://legacy.python.org/dev/peps/pep-0257/#id16).

invoice_fiscal_position_update/account_invoice.py
```diff
@@ -28,30 +28,17 @@
...
+        res = {}
```
pedrobaeza, line 35: Use self.ensure_one() to avoid call this method without a recordset of more than one record.

alexis-via:
@pedrobaeza Do you have an example of the use of self.ensure_one() in a real module ? I didn't find it in the official addons nor in the banking OCA branch for example.

pedrobaeza:
I have already used it in our MRP project:

https://github.com/odoomrp/odoomrp-wip/blob/8.0/sale_product_variants/models/sale_order.py#L90

And you can see here a little explanation:

https://github.com/nbessi/odoo_new_api_guideline/blob/master/source/decorator.rst#apione

It's a simple call that will lead to an exception if self it's not one record. This way, you prevent that the code is executed without notice for more records (but you haven't taken into account in your code).

alexis-via:
@pedrobaeza 
I implemented your 2 comments of the PR ; do you approve it now ?

pedrobaeza:
Thanks for the changes.

:+1:

bealdav:
:+1:

PR #27, opened 2014-10-28
[ADD] Add new module account_origin_document_on_credit_note

joaoalf:
This module fills in the origin field when creating a credit note from an invoice.

### DIFF
--- a/account_origin_document_on_credit_note/__init__.py
+++ b/account_origin_document_on_credit_note/__init__.py
@@ -1,0 +1,23 @@
+# -*- coding: utf-8 -*-
+###############################################################################
+#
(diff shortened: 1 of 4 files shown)


veloutin:
I agree with @gdgellatly that the upstream odoo code has a lot of code that begs refactoring to allow
easier extending. It would help save a lot of extra queries. In the meantime, we sort of have to live with
the current state.

:+1: for this module.

gdgellatly:
@veloutin that may be true in many cases, but in this case while the refactoring of the wizard has not been done, it has of the refund method so I'd rather see it there.

max3903:
:+1:

PR #28, opened 2014-10-30
account_invoice_merge : little fixes

laetitia-gangloff:
Hello,

I make a litte fix in account_invoice_merge.

uos_id is not a required field on invoice line so I refactored the part where it is use.

I also add a context key (is_merge) to disting invoice from merge invoice. It can be usefull to know this info at invoice creation (in my case, at creation I add some line function of the product, I don't want to get these line duplicate).
And finally I pass, this context to the create method.

### DIFF
--- a/account_invoice_merge/invoice.py
+++ b/account_invoice_merge/invoice.py
@@ -78,6 +78,9 @@ class account_invoice(orm.Model):
 
         """
         wf_service = netsvc.LocalService("workflow")
+        if context is None:
+            context = {}
+        context['is_merge'] = True
 
         def make_key(br, fields):
             list_key = []
@@ -142,10 +145,12 @@ class account_invoice(orm.Model):
                     invoice_line, self._get_invoice_line_key_cols(
                         cr, uid, invoice_line))
                 o_line = invoice_infos['invoice_line'].setdefault(line_key, {})
+                uos_factor = (invoice_line.uos_id and
+                              invoice_line.uos_id.factor or 1.0)
                 if o_line:
                     # merge the line with an existing line
                     o_line['quantity'] += invoice_line.quantity * \
-                        invoice_line.uos_id.factor / o_line['uom_factor']
+                        uos_factor / o_line['uom_factor']
                 else:
                     # append a new "standalone" line
                     for field in ('quantity', 'uos_id'):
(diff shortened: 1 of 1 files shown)


account_invoice_merge/invoice.py
```diff
@@ -78,6 +78,9 @@ def do_merge(self, cr, uid, ids, context=None):
 
         """
         wf_service = netsvc.LocalService("workflow")
+        if context is None:
+            context = {}
+        context['is_merge'] = True
```
pedrobaeza, line 83: Why do you need this?

sbidoul, line 83: @laetitia-gangloff since this is not used anywhere is this module, I think you should do it in an override of do_merge in your customer module.

account_invoice_merge/invoice.py
```diff
@@ -142,10 +145,12 @@ def make_key(br, fields):
                     invoice_line, self._get_invoice_line_key_cols(
                         cr, uid, invoice_line))
                 o_line = invoice_infos['invoice_line'].setdefault(line_key, {})
+                uos_factor = (invoice_line.uos_id and
+                              invoice_line.uos_id.factor or 1.0)
                 if o_line:
                     # merge the line with an existing line
                     o_line['quantity'] += invoice_line.quantity * \
-                        invoice_line.uos_id.factor / o_line['uom_factor']
+                        uos_factor / o_line['uom_factor']
```
pedrobaeza, line 153: Why do you need to replay this operation? Does not `invoice_line.quantity` already have the correct value?

account_invoice_merge/invoice.py
laetitia-gangloff, line 83: @pedrobaeza as I try to explain in the first message, I add this context key to distinguish normal invoice creation from merge invoice creation. In my use case, at creation I add some line function of the product, I don't want to get these line duplicate.

@sbidoul I thought it can be usefull for other. But it is not a problem to move this part in my customer module.

account_invoice_merge/invoice.py
laetitia-gangloff, line 153: @pedrobaeza When two lines concern the same product, we update the quantity. uom can be different from one line to another.

What I don't know is why the module uos instead of uom.

pedrobaeza, line 153: OK, I understand. Thanks for the explanation.

pedrobaeza:
:+1:

guewen:
:+1:

PR #31, opened 2014-11-18
[ADD] new module "partner_refunds_buttons"

anthony-muschang:
This module add new smart buttons in the Partner view to Customer Refunds and Supplier Refunds.


account_refund_partner/__openerp__.py
```diff
@@ -0,0 +1,47 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+# Author: Anthony Muschang
+# Copyright (c) 2014 Acsone SA/NV (http://www.acsone.eu)
+# All Rights Reserved
```
sbidoul: Remove All Rights Reserved

account_refund_partner/__openerp__.py
```diff
@@ -0,0 +1,47 @@
...
+#
+# WARNING: This program as such is intended to be used by professional
```
sbidoul: Remove this non-standard license section

account_refund_partner/__openerp__.py
```diff
@@ -0,0 +1,47 @@
...
+    'category': 'Accounting & Finance',
+    'license': 'AGPL-3',
+    'description': """
+Add shortcuts to Refunds in partner screen
+==========================================
+
+This module add shortcuts (i.e: smart button) in the Partner view to
+ Customer Refunds and Supplier Refunds
+""",
+    'author': 'Anthony Muschang, ACSONE',
```
sbidoul: author: ACSONE SA/NV for consistency with our other modules

account_refund_partner/tests/__init__.py
```diff
@@ -0,0 +1,39 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
+#
+##############################################################################
+
+from . import test_refund_count
+
+fast_suite = [
```
sbidoul: fast_suite and checks are not necessary in 8.0

account_refund_partner/res_partner.py
```diff
@@ -0,0 +1,47 @@
...
+##############################################################################
+from openerp import models, fields, api
+
+
+class res_partner(models.Model):
+    _inherit = "res.partner"
+
+    supplier_refund_count = fields.Integer(compute='_refund_count',
+                                           string="# Supplier Refund",
+                                           groups='account.' +
```
pedrobaeza: You don't need to put `+` sign. Consecutive strings are concatenated automatically, but it's better to use this indentation (also PEP8 compliant) in these cases:

```
supplier_refund_count = fields.Integer(
    compute='_refund_count', string="# Supplier Refund", 
    groups='account.group_account_invoice')
```

account_refund_partner/res_partner.py
```diff
@@ -0,0 +1,47 @@
...
+                                           'group_account_invoice')
+
+    customer_refund_count = fields.Integer(compute='_refund_count',
+                                           string="# Supplier Refund",
+                                           groups='account.' +
+                                           'group_account_invoice')
+
+    @api.multi
```
pedrobaeza: Put @api.one to avoid one extra indentation level and an unneeded `for`.

sbidoul:
I'm not easy with the module name. Is account_partner_refund_buttons more intuitive?

pedrobaeza:
I would say _partner_refunds_buttons_ to follow convention about first word is the area affected.

sbidoul:
@pedrobaeza good.

anthony-muschang:
Thanks for the feedback. I renamed the module and made some improvements.

pedrobaeza:
Thanks for the changes. One doubt: standard smart button 'Invoices' includes also refunds in its domain? If so, maybe we should modify it to extract that part, because these buttons make the trick.

sbidoul:
@pedrobaeza you are right. The Invoiced button includes refunds. But when you click on it, it shows a mix of invoices and refunds, with a wrong total. It even mixes supplier and customer invoices if the partner has both! This looks like a bug to me.

@anthony-muschang I suggest the following:
- put the Sup. Refund button next to the Sup. Invoices button
- add a Cust. Invoices button with a count before your Cust. Refund button.

In parallel, we must see with Odoo what to do with this weird Invoiced button.

anthony-muschang:
- I cannot put the 'Sup. Refund' button next to the "Sup. Invoices" without adding the module purchase to this module dependencies.
- Ok for adding the 'Cust. Invoices' button. 

For information the "Invoiced" button is added by account and the "Sup. Invoices" button is added by purchase.

pedrobaeza:
You can put it without depending on purchase, but not modifying the other button.

sbidoul:
@pedrobaeza do you think this module makes sense for OCA?

We'll discuss internally, but I think I've been too optimistic. Since the standard buttons are not very coherent, it's difficult to make something clean.

pedrobaeza:
Yeah, maybe you're right. We should push Odoo to include it in core.

legalsylvain:
I close this PR, due to last old comments.

PR #33, opened 2014-12-12
[7.0] [ADD] account invoice default salesteam

veloutin:
This module set the Sales Team (section_id field) in the 2 following scenarios:
- when creating an invoice, set on partner_id.onchange() from the partner's
  Sales Team
- when invoicing from sales_order, set from the Sales Order's Sales Team.


max3903:
:+1:

account_invoice_default_salesteam/account_invoice.py
```diff
@@ -0,0 +1,57 @@
...
+###############################################################################
+
+from openerp.osv import orm
+
+
+class AccountInvoice(orm.Model):
+    _inherit = "account.invoice"
+
+    def create(self, cr, uid, values, context=None):
+        if "partner_id" in values and not values.get("section_id"):
```
pedrobaeza, line 30: Why do you need this patch at creation time?

veloutin, line 30: To use the partner's sales team if we did not supply one.

pedrobaeza:
:+1:

PR #35, opened 2015-01-04
[7.0] [FIX] account_invoice_merge [needs review]

llacroix:
Missing plus operator.


pedrobaeza:
Although it seems a bit odd, Python allows this concatenation without `+`, so there's no issue.

:-1:

sbidoul:
@pedrobaeza I'd say @llacroix is right. Without the + it is equivalent to `"[('id','in',[,".join(map(str, allinvoices.keys()))` wich is not what we want.

So :+1: from my side.

pedrobaeza:
OK, I see. You're right.

Sorry for the mistake.

:+1:

llacroix:
Yes, that's kind of disturbing. The order of operator has some importance here.

llacroix:
:+1:

PR #43, opened 2015-02-02
8.0 account invoice merge 7.0 forward port

Cedric-Pigeon:
Forward port from 7.0 branch to 8.0

### DIFF
--- a/account_invoice_merge/invoice.py
+++ b/account_invoice_merge/invoice.py
@@ -76,9 +76,6 @@ class account_invoice(models.Model):
          @return: new account invoice id
 
         """
-        context = self.env.context.copy()
-        context['is_merge'] = True
-
         def make_key(br, fields):
             list_key = []
             for field in fields:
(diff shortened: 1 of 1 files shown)


account_invoice_merge/invoice.py
```diff
@@ -166,7 +170,7 @@ def make_key(br, fields):
                 (0, 0, value) for value in
                 invoice_data['invoice_line'].itervalues()]
             # create the new invoice
-            newinvoice = self.create(invoice_data)
+            newinvoice = self.with_context(context).create(invoice_data)
```
pedrobaeza: Better to put: `....with_context(is_merge=True)....` and you don't have to copy the context.

Cedric-Pigeon: Ok, Done. 

I misunderstood the behavior of with_context. I was not aware that it does an extend.

Thanks.

pedrobaeza: It can be used in both ways: passing a dictionary to replace the current one, or only passing keyword arguments to extend current one.

pedrobaeza:
:+1:

yvaucher:
:+1:

sbidoul:
:+1:

PR #44, opened 2015-02-03
No need to re-browse after search in new API

Cedric-Pigeon:
Because search method already returns a recorset, there is no need to re-browse on it.

### DIFF
--- a/account_invoice_merge/invoice.py
+++ b/account_invoice_merge/invoice.py
@@ -190,8 +190,8 @@ class account_invoice(models.Model):
             if so_obj:
                 todos = so_obj.search(
                     [('invoice_ids', 'in', invoices_info[new_invoice_id])])
+                todos.write({'invoice_ids': [(4, new_invoice_id)]})
                 for org_so in todos:
-                    org_so.write({'invoice_ids': [(4, new_invoice_id)]})
                     for so_line in org_so.order_line:
                         invoice_line_ids = invoice_line_obj.search(
                             [('product_id', '=', so_line.product_id.id),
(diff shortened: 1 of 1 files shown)


pedrobaeza:
You can even write directly in return recordset without making a for.

account_invoice_merge/invoice.py
```diff
@@ -188,9 +188,9 @@ def make_key(br, fields):
             if 'purchase.order' in self.env.registry else False
         for new_invoice_id in invoices_info:
             if so_obj:
-                todo_ids = so_obj.search(
+                todos = so_obj.search(
                     [('invoice_ids', 'in', invoices_info[new_invoice_id])])
-                for org_so in so_obj.browse(todo_ids):
+                for org_so in todos:
                     org_so.write({'invoice_ids': [(4, new_invoice_id)]})
```
yvaucher: Following @pedrobaeza comment, you can extract write for loop and do

```
todos.write({'invoice_ids': [(4, new_invoice_id)]})
```

account_invoice_merge/invoice.py
```diff
@@ -200,17 +200,16 @@ def make_key(br, fields):
                             so_line.write(
                                 {'invoice_lines': [(6, 0, invoice_line_ids)]})
             if po_obj:
-                todo_ids = po_obj.search(
+                todos = po_obj.search(
                     [('invoice_ids', 'in', invoices_info[new_invoice_id])])
-                for org_po in po_obj.browse(todo_ids):
+                for org_po in todos:
```
yvaucher: Remove for loop and do

```
todos.write({'invoice_ids': [(4, new_invoice_id)]})
```

Cedric-Pigeon:
@pedrobaeza @yvaucher Thanks for review. Code is now fixed.

yvaucher:
@Cedric-Pigeon Thanks for the changes :+1:

sbidoul:
:+1: 

The needs fixing tag can be removed.

PR #45, opened 2015-02-04
[IMP] Replace tuples with column names to list to be more flexible to add some fields

adrienpeiffer:
Replacement of tuples with column names to list to be more flexible to add custom fields in the key


pedrobaeza:
:+1:

sbidoul:
:+1:

Cedric-Pigeon:
:+1:

PR #49, opened 2015-02-17
[FIX]open form view corresponding to the type of invoice

Cedric-Pigeon:
After merging all selected invoices, the wizard shows the result in a tree view. 
Unfortunately, when you want to see it on the form view the supplier invoice form view is always loaded. Even if you work on customer invoices.

This PR suggest a fix for this bug.

### DIFF
--- a/account_invoice_merge/wizard/invoice_merge.py
+++ b/account_invoice_merge/wizard/invoice_merge.py
@@ -89,22 +89,21 @@ class invoice_merge(models.TransientModel):
              @param ids: the ID or list of IDs
              @param context: A standard dictionary
 
-             @return: account invoice view
+             @return: account invoice action
         """
         inv_obj = self.env['account.invoice']
-        mod_obj = self.env['ir.model.data']
-        result = mod_obj._get_id('account', 'invoice_form')
-        data = mod_obj.search_read([('id', '=', result)], ['res_id'])[0]\
-            or False
-        invoices = inv_obj.browse(self.env.context.get('active_ids', []))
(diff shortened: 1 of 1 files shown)


account_invoice_merge/wizard/invoice_merge.py
```diff
@@ -89,22 +89,21 @@ def merge_invoices(self):
...
-        }
+        xid = {
+            'out_invoice': 'action_invoice_tree1',
+            'out_refund': 'action_invoice_tree3',
+            'in_invoice': 'action_invoice_tree2',
+            'in_refund': 'action_invoice_tree4',
+        }[invoices[0].type]
+        action = aw_obj.for_xml_id('account', xid)
+        action.update({
+            'domain': [('id', 'in', ids + allinvoices.keys())],
```
pedrobaeza, line 107: Why are you getting ids? You only need `allinvoices.keys()`. You can use direct value writing instead an update on the dictionary.

Cedric-Pigeon, line 107: @pedrobaeza I am getting ids to be able to display in the tree view the new invoice and all original invoices cancelled in place of only the new one.

pedrobaeza, line 107: OK, I get it

pedrobaeza:
:+1:

Olivier-LAURENT:
:+1: Nice, user has now a quick visual validation on the result of its merge action ... !

lmignon:
:+1:

PR #50, opened 2015-02-17
[8.0]Let the choice to the user to keep or not the reference from original invoices

Cedric-Pigeon:
By default the wizard aggregates all values from the field name of all original invoices in the field name of the new invoice.
Depending of the number of invoices to merge, this name could be very long and difficult to print on report.

So this PR add a new field on wizard to let the use choose to keep or not this reference.

### DIFF
--- a/account_invoice_merge/invoice.py
+++ b/account_invoice_merge/invoice.py
@@ -56,7 +56,7 @@ class account_invoice(models.Model):
         }
 
     @api.multi
-    def do_merge(self):
+    def do_merge(self, keep_references=True):
         """
         To merge similar type of account invoices.
         Invoices will only be merged if:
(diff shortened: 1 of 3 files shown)


pedrobaeza:
:+1: if Travis goes OK

Cedric-Pigeon:
I don't understand why Travis goes wrong ?

pedrobaeza:
I have restarted it because sometimes pip fails. Now it's OK.

Cedric-Pigeon:
@pedrobaeza Thanks.

adrienpeiffer:
:+1: , but Could you rebase this one ?

Cedric-Pigeon:
@adrienpeiffer rebase done.

lmignon:
@Cedric-Pigeon It remains one pep8 error

Cedric-Pigeon:
@lmignon Now fixed.

lmignon:
:+1:

yvaucher:
:+1:

PR #51, opened 2015-02-18
[8.0][ADD] add new module to manage order of invoice lines

Cedric-Pigeon:
This PR suggest a new module to manage order of customer invoice lines.

You can sort invoice lines by sequence (default), name, price_unit or price_subtotal, either ascending or descending.

The sort options can be set on an invoice or stored customer by customer.

### DIFF
--- a/account_invoice_line_sort/models/account_invoice.py
+++ b/account_invoice_line_sort/models/account_invoice.py
@@ -23,7 +23,7 @@
 ##############################################################################
 
 from openerp import models, fields, api
-from operator import attrgetter 
+from operator import attrgetter
 
 AVAILABLE_SORT_OPTIONS = [
     ('sequence', 'Sequence'),
--- a/account_invoice_line_sort/models/account_invoice.py
+++ b/account_invoice_line_sort/models/account_invoice.py
@@ -74,12 +74,11 @@ class account_invoice(models.Model):
 
     @api.one
     def _sort_account_invoice_line(self):
-        line_model = self.env['account.invoice.line']
         if self.invoice_line:
-            order = "%s %s" % (self.line_order, self.line_order_direction)
-            domain = [('id', 'in', self.invoice_line.ids)]
             sequence = 0
-            for line in line_model.search(domain, order=order):
+            key = attrgetter(self.line_order)
+            reverse = self.line_order_direction == 'desc'
+            for line in self.invoice_line.sorted(key=key, reverse=reverse):
                 sequence += 10
                 line.sequence = sequence
 
--- a/account_invoice_line_sort/models/account_invoice.py
+++ b/account_invoice_line_sort/models/account_invoice.py
@@ -23,6 +23,7 @@
 ##############################################################################
 
 from openerp import models, fields, api
+from operator import attrgetter 
 
 AVAILABLE_SORT_OPTIONS = [
     ('sequence', 'Sequence'),
--- a/account_invoice_line_sort/models/account_invoice.py
+++ b/account_invoice_line_sort/models/account_invoice.py
@@ -1,0 +1,130 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+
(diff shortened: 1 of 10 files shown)


account_invoice_line_sort/models/account_invoice.py
```diff
@@ -0,0 +1,130 @@
...
+        return res
+
+    @api.one
+    def _sort_account_invoice_line(self):
+        line_model = self.env['account.invoice.line']
+        if self.invoice_line:
+            order = "%s %s" % (self.line_order, self.line_order_direction)
+            domain = [('id', 'in', self.invoice_line.ids)]
+            sequence = 0
+            for line in line_model.search(domain, order=order):
```
pedrobaeza: You don't need to make a search. You can use `self.invoice_line.ordered(...)`.

Cedric-Pigeon: @pedrobaeza Can you give me more information about `.ordered(...)`? I can't find any sample in Odoo.

pedrobaeza: Sorry, I made a mistake. It's sorted, not ordered: https://www.odoo.com/documentation/8.0/reference/orm.html

lmignon: It would be nice to also support dot notation in AVAILABLE_SORT_OPTIONS to also be able to sort by product category. 'product_id.categ_id.name'

Cedric-Pigeon: @pedrobaeza Ok I see, Thanks.

@lmignon I suggest to support it later as an improvement of this base module. It is not as simple as it looks like.

lmignon: @Cedric-Pigeon something like this:

``` python
for line in self.invoice_line.sorted(key=attrgetter(self.line_order), reverse=self.line_order_direction=='asc'):
    sequence += 10
    line.sequence = sequence
```

The standard operator 'attrgetter' supports the dot notation. https://docs.python.org/2/library/operator.html#operator.attrgetter

Cedric-Pigeon: @lmignon Great ! But what if you encounter a null value in your dot path ? For example product_id is null ?

lmignon: @Cedric-Pigeon it's supported by odoo. product_id.categ_id.name will return False even if product_id is False.

Cedric-Pigeon: @lmignon OK for technical aspect but in a functional point of view I don't see the interest to sort on list on a criteria that is not visible for the user. In any case I will still do it. Thanks.

lmignon: @Cedric-Pigeon e.g. products from 'Food' category before 'Non-Food' products. It's an example. IMO, it would be easier to extend the module for others specific needs if the dot notation is natively supported even if the module doesn't provide this kind of predefined sort option in AVAILABLE_SORT_OPTIONS

Cedric-Pigeon: @pedrobaeza I use now sorted function. Thanks for your help
@lmignon You are right. Code integrated in last commit. Thanks.

pedrobaeza:
Please correct pep8 (that it's a code error in this case, not style)

lmignon:
./account_invoice_line_sort/models/account_invoice.py:26:32: W291 trailing whitespace

Cedric-Pigeon:
@pedrobaeza Indeed I forgot an import of library. Fixed now
@lmignon Trailing whitespace removed

lmignon:
:+1: Once Travis will be OK. Thank you for the contribution.

pedrobaeza:
:+1:

laetitia-gangloff:
:+1:

PR #56, opened 2015-03-03
add account_invoice_sequency_by_period

luc-demeyer:
This module enables period based invoice numbering. 
It's a very small module created this afternoon in a couple of hours. 

I am looking forward to the feedback from the OCA experts.
There are a few improvements that would be nice (e.g. adding some test scenarios, adding a wizard to facilitate the creation of the period specific sequences), but these improvements would take more time than what is paid for by the sponsoring customer project.

### DIFF
--- a/account_invoice_sequence_by_period/README.rst
+++ b/account_invoice_sequence_by_period/README.rst
@@ -1,0 +1,5 @@
+Description
+===========
+
+This module extends the sequence object to enable separate sequences by Accounting Period for Invoice Numbering.
+For this purpose, the sequence legend has been extended with the %(p)s key.
--- a/account_invoice_sequence_by_period/__init__.py
+++ b/account_invoice_sequence_by_period/__init__.py
@@ -1,0 +1,24 @@
+# -*- encoding: utf-8 -*-
+##############################################################################
+#
+#    OpenERP, Open Source Management Solution
+#
+#    Copyright (c) 2015 Noviat nv/sa (www.noviat.com).
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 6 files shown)


pedrobaeza:
I find this module very similar to this one: https://github.com/OCA/account-financial-tools/tree/8.0/account_auto_fy_sequence, so I don't think it's needed.

luc-demeyer:
@pedrobaeza 
Pedro, it's impressive to see a reaction within minutes after submission of my PR.
This module is different from the 'account_auto_fy_sequence'. The requirement I cover are separated invoice numbering schemes per fiscal period..
E.g. for period 01/2015 you can configure the system with a sequence like EXJ15-01-0001, whereas invoices in Period 02/2015 will start  with EXJ15-02-0001. This facilitates allows encoding in two fiscal periods at the same time while keeping correctly sequenced numbers.

Installing this module does require the configuration of a sequence per period, hence 12 times more work when a new fiscalyear starts and therefor it's probably a good idea to make a next version of this module dependent on the 'account_auto_fy_sequence' and extend the account._journal create_sequence method to generate sequences per period.

pedrobaeza:
Yeah, I see that it's not exactly the same, but I don't know any company that wants a different sequence by period. In fact, in Spain it's forbidden by law. Numbering must consecutive, progressive and without holes. That's why we have the module _l10n_es_account_invoice_sequence_.

Well, as a first version, can serve, but better to put in the same repo, and specify in known issues section of the README this question you mention.

luc-demeyer:
@pedrobaeza
in Belgium Invoice numbering must be without holes per fiscal period.

According to feedback from a french customer, this is the same in France (@alexis-via can you confirm this ?).

Concerning repo: this one is really invoice specific hence my PR to account-invoice but a next version wiould best  take advantage of / depend on  account_auto_fy_sequence hence easier to put in the 'account-financial-tools'. 

Have we ever considered to merge these two repos into one ?

pedrobaeza:
Well, the target is not exactly the same, because this repo is supposed to serve as the target for modules that aids invoicing processes (from other places in the system), but as in other occassions, there is some overlapping with another repos.

luc-demeyer:
Close PR since replaced by new PR to account-financial-tools

PR #64, opened 2015-04-02
[FIX] [account_invoice_force_number] duplicated field

dcorio:
the internal_number fields was already present on the original views (hiddent) then the module was behaving weirdly because of this.

### DIFF
--- a/account_invoice_force_number/invoice_view.xml
+++ b/account_invoice_force_number/invoice_view.xml
@@ -7,11 +7,11 @@
             <field name="model">account.invoice</field>
             <field name="inherit_id" ref="account.invoice_form"/>
             <field name="arch" type="xml">
-                <field name="internal_number" position="replace">
-                    <field name="internal_number"
-                        attrs="{'invisible': [('state', '!=', 'draft')]}"
-                        string="Force Number"
-                        help="Force invoice number. Use this field if you don't want to use the default numbering"/>
+                <field name="internal_number" position="attributes">
+                    <attribute name="attrs">{'invisible': [('state', '!=', 'draft')]}</attribute>
+                    <attribute name="string">Force Number</attribute>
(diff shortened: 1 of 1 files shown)


eLBati:
LGTM :+1: 
thanks

account_invoice_force_number/invoice_view.xml
```diff
@@ -7,12 +7,12 @@
             <field name="model">account.invoice</field>
             <field name="inherit_id" ref="account.invoice_form"/>
             <field name="arch" type="xml">
-                <xpath expr="//field[@name='fiscal_position']" position="after">
+                <field name="internal_number" position="replace">
```
adrienpeiffer: Attributes could be used directly instead of replace

account_invoice_force_number/invoice_view.xml
```diff
@@ -21,12 +21,12 @@
             <field name="model">account.invoice</field>
             <field name="inherit_id" ref="account.invoice_supplier_form"/>
             <field name="arch" type="xml">
-                <xpath expr="//field[@name='fiscal_position']" position="after">
+                <field name="internal_number" position="replace">
```
adrienpeiffer: Same remark as top of file

dcorio:
@adrienpeiffer that would require to change 3 attributes instead of one field. is it better anyway?

adrienpeiffer:
@dcorio yes, i think that's better. Let me take an example, if someone inherits the standard view to add an other attribute on this field. This replace may break other changes.

eLBati:
@dcorio @adrienpeiffer here it is https://github.com/abstract-open-solutions/account-invoicing/pull/1

dcorio:
thank you so much @eLBati. i haven't found the time to fix it myself.
i just tested your PR and it works as expected.

:+1:

adrienpeiffer:
:+1:  Thanks !

PR #66, opened 2015-04-08
[ADD] add module to force currency rate on account invoice

bguillot:
Hello, 

This PR adds a module : account_invoice_currency_rate. 
It allows to manually set the currency rate on account invoice and it will be used on the move creation.

Thanks for reviews

### DIFF
--- a/account_invoice_currency_rate/__init__.py
+++ b/account_invoice_currency_rate/__init__.py
@@ -1,0 +1,25 @@
+# -*- coding: utf-8 -*-
+###############################################################################
+#
+#   Module for OpenERP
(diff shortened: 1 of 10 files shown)


chafique-delli:
Hello Benoit,

I tested your module, it is good but i will hide the button 'Force currency rate' on invoice when currency_id.rate=1 that is the currency of the company otherwise it is confusing.

Best Regards,

bguillot:
Hello Chafique,

Thank you for the review, I think you are right.

It was a bit tricky to hide the button with those conditions, so I had to add a function field to decide when to display the button. If someone has a better idea I can change the code.

chafique-delli:
It's good for me now. :+1:

max3903:
:+1:

PR #67, opened 2015-04-17
[ADD] Add module for splitting invoices.

ecino:
This module adds the ability to split an existing invoice into
two different invoices by adding a new wizard on invoices allowing
the user to select which lines are to be moved in a new invoice.

Works only on unpaid invoices.


pedrobaeza:
The name can be shorten to _account_invoice_split_

account_invoice_split_invoice/__openerp__.py
```diff
@@ -0,0 +1,56 @@
+# -*- encoding: utf-8 -*-
+##############################################################################
+#
+#       ______ Releasing children from poverty      _
+#      / ____/___  ____ ___  ____  ____ ___________(_)___  ____
```
pedrobaeza: Don't use this kind of signatures.

account_invoice_split_invoice/__openerp__.py
```diff
@@ -0,0 +1,56 @@
...
+    'website': 'http://www.compassion.ch',
+    'category': 'Accounting',
+    'depends': ['account'],
+    'external_dependencies': {},
+    'data': [
+        'view/account_invoice_split_wizard_view.xml',
+    ],
+    'demo': [],
+    'description': """
+=====================================================
```
pedrobaeza: Please use OCA's template: https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst

ecino:
I renamed the module and copied OCA template. I'm not 100% sure on the description folder, is it necessary to have a copy of the README in html format ?

pedrobaeza:
Thanks for the changing.

index.html is not needed for the normal working. It was intended for a better visualization on apps.odoo.com, but now the render of the description (for v7) or README (for v8) is acceptable.

Anyways, please put the same content of the README file inside description key of **openperp**.py, because in v7, README.rst file is still not parsed.

ecino:
Will reopen for v8.

PR #72, opened 2015-04-29
Migration of account_invoice_line_no_picking_name

tafaRU:


pedrobaeza:
:+1: 

You can add OCA's logo for completion.

account_invoice_line_no_picking_name/stock_move.py
```diff
@@ -0,0 +1,44 @@
...
+            user = self.pool['res.users'].browse(
+                cr, uid, uid, context=context)
+            user_groups = [g.id for g in user.groups_id]
+            ref = self.pool['ir.model.data'].get_object_reference(
+                cr, uid, 'account_invoice_line_no_picking_name',
+                'group_not_use_picking_name_per_invoice_line'
+            )
+            if ref and len(ref) > 1 and ref[1]:
+                group_id = ref[1]
+                if group_id in user_groups:
```
guewen: You should use `self.user_has_groups(cr, uid, 'account_invoice_line_no_picking_name.group_not_use_picking_name_per_invoice_line', context=context` (see https://github.com/odoo/odoo/blob/8.0/openerp/models.py#L1363-L1374). It calls `ResUsers.has_groups()` which uses a cache for the result.

tafaRU:
@guewen just applied your tip, very useful, thanks!

eLBati:
:+1:

tafaRU:
@guewen could you please check if PR looks good now?

Thank you in advance.

account_invoice_line_no_picking_name/stock_move.py
```diff
@@ -0,0 +1,40 @@
...
+        res = super(StockMove, self)._get_invoice_line_vals(
+            cr, uid, move, partner, inv_type, context=context)
+        if move:
+            if self.user_has_groups(
+                    cr, uid,
+                    'account_invoice_line_no_picking_name.'
+                    'group_not_use_picking_name_per_invoice_line',
+                    context=context
+            ):
+                res['name'] = move.name
```
adrienpeiffer, line 39: This key and this value isn't already present in res ?
https://github.com/odoo/odoo/blob/8.0/addons/stock_account/stock.py#L158

pedrobaeza:
@tafaRU, any news about this?

tafaRU:
@adrienpeiffer, you're right so this module is superfluous since 8.0.
I close this PR then.

PR #62, opened 2015-03-30
[MIG] Migration of stock_invoice_picking to 8.0

mileo:

PR #74, opened 2015-05-05
Migration of stock_invoice_picking_incoterm

tafaRU:
This PR depends on https://github.com/OCA/account-invoicing/pull/62


eLBati:
Thanks :+1:

archetipo:
:+1:

eLBati:
@tafaRU After renaming of `stock_picking_invoicing`, theoretically this module should now be called `stock_picking_invoicing_incoterm`? What do you think?

tafaRU:
@eLBati I agree with the renaming of the module.

About the migration, I'm wondering if I can handle it simply updating https://github.com/OCA/OpenUpgrade/blob/8.0/openerp/addons/openupgrade_records/lib/apriori.py#L5 cc @pedrobaeza

pedrobaeza:
For the v7 > v8 module migration, indeed you only need to add an entry in https://github.com/OCA/OpenUpgrade/blob/8.0/openerp/addons/openupgrade_records/lib/apriori.py#L5 as we have done with another renamed module.

tafaRU:
@pedrobaeza, thanks! Please, one more thing: in case the module v8 is already installed, what's the best way to replace it by that one with the new name avoiding data loss?

Thank you in advance.

pedrobaeza:
For that things, I use the shell module and start a shell:

``` shell
odoo.py --addons-path=<path> shell --config=<config_path> -d <dbname>
```

and execute this code:

``` python
from openupgradelib import openupgrade
openupgrade.update_module_names([('old_name', 'new_name')])
self.env.cr.commit()
```

And you have the module renamed.

tafaRU:
@pedrobaeza, thanks a lot, you know a bit more than the devil:exclamation:

tafaRU:
@eLBati @pedrobaeza I renamed the module as suggested at https://github.com/OCA/account-invoicing/pull/74#issuecomment-149563028

Now the module is ready to be merged, please review it.

Thank you in advance.

gurneyalex:
:+1: (code review, no test)

hurrinico:
:+1: (no test)

PR #88, opened 2015-07-07
[FIX] stock_picking_invoicing: using wrong partner for invoice (+ tests)

eLBati:
when delivering to a company address

### DIFF
--- a/stock_picking_invoicing/stock.py
+++ b/stock_picking_invoicing/stock.py
@@ -49,3 +49,19 @@ class StockPicking(models.Model):
                                   (picking.name, picking.invoice_id.number))
                 picking.invoice_state = '2binvoiced'
         return True
+
+
+class StockMove(models.Model):
+    _inherit = 'stock.move'
+
+    @api.model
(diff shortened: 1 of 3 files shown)


stock_picking_invoicing/stock.py
```diff
@@ -49,3 +49,19 @@ def set_to_be_invoiced(self):
...
+class StockMove(models.Model):
+    _inherit = 'stock.move'
+
+    @api.model
+    def _get_master_data(self, move, company):
+        data = super(StockMove, self)._get_master_data(move, company)
+        if move.picking_id.partner_id.id != data[0].id:
+            # if someone else modified invoice partner, I use it
+            return data
+        partner_invoie_id = move.picking_id.partner_id.address_get(
```
tafaRU: It would be better use `partner_invoice_id`. Could you please rename it?

archetipo:
:+1:

pedrobaeza:
Thanks for the test :+1:

PR #90, opened 2015-07-13
Add invoice number view tree

archetipo:
This PR contains the porting form [addons-vauxoo/invoice_number_view_tree](https://github.com/Vauxoo/addons-vauxoo/tree/8.0/invoice_number_view_tree),
and it was preserved the history of commits


invoice_number_view_tree/__openerp__.py
```diff
@@ -0,0 +1,49 @@
...
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+#
+{
+    "name": "Invoice number",
+    "version": "1.0",
+    "author": "Vauxoo",
+    "category": "Localization",
+    "description": """""",
```
tafaRU: Please remove this key

invoice_number_view_tree/__openerp__.py
```diff
@@ -0,0 +1,49 @@
...
+        "account"
+    ],
+    "demo": [],
+    "data": [
+        "view/invoice_number.xml"
+    ],
+    "test": [],
+    "js": [],
+    "css": [],
+    "qweb": [],
```
tafaRU: Please remove empty keys.

invoice_number_view_tree/README.rst
```diff
@@ -0,0 +1,32 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Invoice number:
+===============
+
+supplier_invoice_number field added in the tree and search view by
+inheritance
+
```
tafaRU, line 9: Please add `Bug Tracker` section, see https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst as reference.

invoice_number_view_tree/__init__.py
```diff
@@ -0,0 +1,25 @@
+# -*- encoding: utf-8 -*-
```
tafaRU: This file is empty so it can be deleted.

invoice_number_view_tree/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+#
+#    Module Writen to OpenERP, Open Source Management Solution
```
tafaRU: s/writen/written

invoice_number_view_tree/__openerp__.py
```diff
@@ -0,0 +1,43 @@
+# -*- encoding: utf-8 -*-
+#
+#    Module Writen to OpenERP, Open Source Management Solution
```
tafaRU: Same as above

pedrobaeza:
Modules should be named _account_invoice_supplier_number_info_ or similar, but it's a bit light to be a module itself. I would suggest to propose this on OCB (it's only a view change, not DB layout change), and in Odoo master to be included in next version.

:-1:

archetipo:
thanks @pedrobaeza.
ok then I close it and proceed as you suggest .

eLBati:
> I would suggest to propose this on OCB (it's only a view change, not DB layout change), and in Odoo master to be included in next version

Hi @pedrobaeza 
It seems `supplier_invoice_number` is not present in master: https://github.com/odoo/odoo/blob/master/addons/account/models/account_invoice.py

Anyway, as this functionality is not a bug fix, I think it is more suitable for a small module, rather than an OCB patch

PR #93, opened 2015-08-18
[8.0][FIX] Merge invoices from purchase order

adrienpeiffer:
Currently in account_invoice_merge, in case of merge of invoices genereted from the purchase order, when the new invoice is paid, the related purchase order don't go to the 'done' state.

I take this opportunity to create a new module to make the link between account_invoice_merge and purchase module.
- [x] Depends on #98

### DIFF
--- a/account_invoice_merge_purchase/README.rst
+++ b/account_invoice_merge_purchase/README.rst
@@ -1,0 +1,46 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Merge invoices from purchase order
+==================================
+
+This module provides a compatibility between purchase order workflow and account invoice merge module
+
+
+Usage
+=====
+
+* Click on install button
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/95/8.0
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues <https://github.com/OCA/account-invoicing/issues>`_.
+In case of trouble, please check there if your issue has already been reported.
+If you spotted it first, help us smashing it by providing a detailed and welcomed feedback
+`here <https://github.com/OCA/account-invoicing/issues/new?body=module:%20account_invoice_merge_purchase%0Aversion:%208.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
+
+Credits
+=======
+
+Contributors
+------------
+
+* Adrien Peiffer <adrien.peiffer@acsone.eu>
+
+Maintainer
+----------
+
+.. image:: http://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: http://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose mission is to support the collaborative development of Odoo features and promote its widespread use.
+
+To contribute to this module, please visit http://odoo-community.org.
--- a/account_invoice_merge_purchase/README.rst
+++ b/account_invoice_merge_purchase/README.rst
@@ -10,8 +10,6 @@ This module provides a compatibility between purchase order workflow and account
 Usage
 =====
 
-* Click on install button
-
 .. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
    :alt: Try me on Runbot
    :target: https://runbot.odoo-community.org/runbot/95/8.0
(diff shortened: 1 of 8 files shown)


account_invoice_merge_purchase/README.rst
```diff
@@ -0,0 +1,46 @@
...
+Merge invoices from purchase order
+==================================
+
+This module provides a compatibility between purchase order workflow and account invoice merge module
+
+
+Usage
+=====
+
+* Click on install button
```
pedrobaeza: Remove this line (this should be in installation section - BTW, the module is auto-install)

account_invoice_merge_purchase/__openerp__.py
```diff
@@ -0,0 +1,39 @@
...
+#
+##############################################################################
+{
+    'name': "Account Invoice Merge Purchase",
+
+    'summary': """Compatibility between purchase and account invoice merge""",
+    'author': "ACSONE SA/NV,Odoo Community Association (OCA)",
+    'website': "http://acsone.eu",
+    'category': 'Finance',
+    'version': '0.1',
```
pedrobaeza: Use current version numbering scheme: 8.0.1.0.0

account_invoice_merge_purchase/models/account_invoice.py
```diff
@@ -0,0 +1,54 @@
...
+        po_obj = self.env['purchase.order']
+        invoice_line_obj = self.env['account.invoice.line']
+        for new_invoice_id in invoices_info:
+            todos = po_obj.search(
+                [('invoice_ids', 'in', invoices_info[new_invoice_id])])
+            todos.write({'invoice_ids': [(4, new_invoice_id)]})
+            for org_po in todos:
+                for po_line in org_po.order_line:
+                    invoice_line_ids = invoice_line_obj.search(
+                        [('invoice_id.state', '!=', 'cancel'),
```
pedrobaeza, line 51: Shouldn't this be written in any state (although cancel)?

account_invoice_merge_purchase/tests/test_account_invoice_merge_purchase.py
```diff
@@ -0,0 +1,106 @@
...
+        workflow.trg_validate(self.uid, 'purchase.order',
+                              purchase_order01.id, 'purchase_confirm',
+                              self.cr)
+        # I check if the purchase order is confirmed
+        purchase_order01.invalidate_cache()
+        self.assertEqual(purchase_order01.state, 'approved',
+                         "Purchase order's state isn't correct")
+        invoice_ids = purchase_order01.invoice_ids.ids
+        purchase_order02 = purchase_order01.copy()
+        # I confirm the purchase order
```
pedrobaeza: I confirm the second purchase order

account_invoice_merge_purchase/tests/test_account_invoice_merge_purchase.py
```diff
@@ -0,0 +1,106 @@
...
+        workflow.trg_validate(self.uid, 'purchase.order',
+                              purchase_order02.id, 'purchase_confirm',
+                              self.cr)
+        # I check if the purchase order is confirmed
```
pedrobaeza, line 93: second purchase order

pedrobaeza:
For making the test suite more complete, the flows with the other 2 invoicing methods should be tested.

account_invoice_merge_purchase/README.rst
adrienpeiffer: Of course

account_invoice_merge_purchase/models/account_invoice.py
adrienpeiffer, line 51: what do you mean exactly ?

pedrobaeza, line 51: I mean that you shouldn't filter by state. All invoice lines, independently from their state, should be relinked to the new invoice.

adrienpeiffer:
Hi @pedrobaeza,
I'll add a test for 'picking' invoice method.
For 'manual' method, it's a little more complicated. Indeed, puchase order line invoiced flag is directly set when the invoice is created. So, normally invoice merge has no impact on the status of the order which is computed from the invoice flag of purchase order line. But with purchase_partial_invoicing module for OCA/purchase-workflow, the problem is present. but I think I can't test it here.

account_invoice_merge_purchase/models/account_invoice.py
adrienpeiffer, line 51: Ok, I can remove cancel from the domain. But canceled invoices will be present in the list of invoice's lines on purchase order line

adrienpeiffer:
Rebased to get changes in #98

sbidoul:
Tested on runbot:
- confirm two PO for same supplier
- receive products for one
- receive invoices for both
- merge invoices
- pay merged invoice
- => PO with products received goes to Done
- receive products for the other PO => It goes to Done two

Works fine.

:+1: (functional test and code review)

account_invoice_merge_purchase/models/account_invoice.py
pedrobaeza, line 51: Thinking twice, I see this is only the link, so there's no data loss. :+1:

pedrobaeza:
:+1: Merging

PR #94, opened 2015-08-21
[FIX] sale_order_partial_invoice: Precision on computed fields

pedrobaeza:
There's a rounding problem sometimes when summing quantities for the partial invoices. This patch solves it.

cc @gurneyalex

### DIFF
--- a/sale_order_partial_invoice/model/sale_order.py
+++ b/sale_order_partial_invoice/model/sale_order.py
@@ -45,6 +45,7 @@ invoicing "based on delivery" that will look at those values instead of looking
 in picking.
 """
 from openerp.osv import orm, fields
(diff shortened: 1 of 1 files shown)


gurneyalex:
:+1: (code review only, no test)

gurneyalex:
can you describe a scenario exhibiting the issue?

pedrobaeza:
I have a sale order that was invoiced in 3 times: 0.35, 0.35 and 0.30, and the result of the sum was 0,9999999! In that case, the sale order was not given as done, and the line is still presented to be invoiced.

eLBati:
LGTM :+1:

pedrobaeza:
Please merge

PR #97, opened 2015-08-27
[8.0] Add account invoice split modules

adrienpeiffer:
This PR adds invoice split functionality.
First, account_invoice_split module which is the main module that allow to split invoice.
After that, there are several auto_install modules that provide a compatibility with other Odoo/OCA modules :
- account_invoice_split_purchase adds compatibility with purchase order workflow
- account_invoice_split_sale adds compatibility with sale order workflow
- account_invoice_split_sale_stock adds compatibility with sale order workflow but specifics for sale_stock module
- account_invoice_split_payment that consider fields on account_payment_partner from OCA/bank-payment

I propose all modules directly because they are all necessary for efficiency with an installation of all modules


.travis.yml
```diff
@@ -23,10 +23,12 @@ env:
   matrix:
   - LINT_CHECK="1"
   - TRANSIFEX="1"
-  - TESTS="1" ODOO_REPO="odoo/odoo" EXCLUDE="account_invoice_validation_workflow"
-  - TESTS="1" ODOO_REPO="OCA/OCB"   EXCLUDE="account_invoice_validation_workflow"
+  - TESTS="1" ODOO_REPO="odoo/odoo" EXCLUDE="account_invoice_validation_workflow,account_invoice_split_sale"
+  - TESTS="1" ODOO_REPO="OCA/OCB"   EXCLUDE="account_invoice_validation_workflow,account_invoice_split_sale"
   - TESTS="1" ODOO_REPO="odoo/odoo" INCLUDE="account_invoice_validation_workflow"
   - TESTS="1" ODOO_REPO="OCA/OCB"   INCLUDE="account_invoice_validation_workflow"
+  - TESTS="1" ODOO_REPO="odoo/odoo" INCLUDE="account_invoice_split_sale,account_invoice_split"
```
pedrobaeza, line 30: Why this can't be tested together with the rest?

account_invoice_split/tests/test_account_invoice_split.py
```diff
@@ -0,0 +1,87 @@
...
+        wiz = self.split_obj.with_context(active_ids=[self.invoice_01.id])\
+            .create({})
+        wiz_line = wiz.line_ids\
+            .filtered(lambda r: (r.origin_invoice_line_id == line_to_split))
+        quantity_to_split = 1
+        wiz_line.quantity_to_split = quantity_to_split
+        wiz._split_invoice()
+        # I check that the line to split is deleted
+        invoice_line = \
+            self.invoice_line_obj.search([('id', '=', line_to_split.id)])
```
pedrobaeza: You can use `line_to_split.exists()` method to check it

account_invoice_split/wizard/account_invoice_split.py
```diff
@@ -0,0 +1,192 @@
...
+                'product_id': line.product_id.id,
+                'name': line.name,
+                'total_quantity': line.quantity,
+            }
+            split_lines.append((0, 0, current_line_values))
+        res['line_ids'] = split_lines
+        return res
+
+    @api.model
+    def _get_invoice_values(self, invoice):
```
pedrobaeza, line 71: Why don't you just copy the invoice record and change what needed (and the same for lines)? Probably this way you avoid some glue modules

sbidoul:
I did a functional test for a purchase and sale workflow on runbot. It works well.

In the split wizard, the notebook is not necessary: a label would be nicer.

Otherwise can you look ater @pedrobaeza 's comments and check why travis is red?

account_invoice_split/wizard/account_invoice_split.py
```diff
@@ -0,0 +1,194 @@
...
+            'date_due': invoice.date_due,
+        }
+
+    @api.model
+    def _create_invoice(self, vals):
+        new_invoice = False
+        if vals['invoice_line']:
+            new_invoice = self.env['account.invoice'].create(vals)
+        if not new_invoice:
+            raise exceptions.Warning(_('Nothing to split'))
```
sbidoul: Can you write: "There is nothing to split. Please fill the 'quantities to split' column." This would be more explicit for the user.

account_invoice_split/wizard/account_invoice_split.py
```diff
@@ -0,0 +1,194 @@
...
+    origin_invoice_line_id = fields.Many2one(
+        comodel_name='account.invoice.line', string='Origin Invoice Line')
+    product_id = fields.Many2one(
+        comodel_name='product.product', string='Product')
+    name = fields.Text(string='Description')
+    total_quantity = fields.Float(
+        string='Total Quantity',
+        digits=dp.get_precision('Product Unit of Measure'))
+    quantity_to_split = fields.Float(
+        string='Quantity To split in a new invoice',
```
sbidoul: s/To/to/

account_invoice_split/i18n/account_invoice_split.pot
```diff
@@ -0,0 +1,116 @@
...
+msgstr ""
+
+#. module: account_invoice_split
+#: field:account.invoice.split.line,product_id:0
+msgid "Product"
+msgstr ""
+
+#. module: account_invoice_split
+#: field:account.invoice.split.line,quantity_to_split:0
+msgid "Quantity To split in a new invoice"
```
sbidoul: s/To/to/

account_invoice_split/i18n/fr.po
```diff
@@ -0,0 +1,116 @@
...
+msgstr "Ligne de facture d'origine"
+
+#. module: account_invoice_split
+#: field:account.invoice.split.line,product_id:0
+msgid "Product"
+msgstr "Produit"
+
+#. module: account_invoice_split
+#: field:account.invoice.split.line,quantity_to_split:0
+msgid "Quantity To split in a new invoice"
```
sbidoul: s/To/to/

account_invoice_split/wizard/account_invoice_split.py
```diff
@@ -0,0 +1,166 @@
...
+#
+##############################################################################
+
+from openerp import models, fields, api, exceptions, _
+import openerp.addons.decimal_precision as dp
+
+FIELD_MAPPING = [('id', 'origin_invoice_line_id'),
+                 ('product_id', 'product_id'),
+                 ('name', 'name'),
+                 ('quantity', 'total_quantity')]
```
sbidoul, line 32: It's better to have globals immutables. Make `FIELD_MAPPING` a tuple instead of a list.

account_invoice_split_sale/models/sale_order.py
```diff
@@ -0,0 +1,46 @@
...
+
+from openerp import models, api
+
+
+class SaleOrder(models.Model):
+    _inherit = 'sale.order'
+
+    @api.cr_context
+    def _auto_init(self, cr, context=None):
+        """ hack: pre-create and initialize the picking_policy column so that
```
sbidoul, line 31: Is this hack still necessary?

sbidoul:
Travis is red because of https://github.com/odoo/odoo/issues/8593

account_invoice_split/wizard/account_invoice_split.py
pedrobaeza, line 32: If you make it inmutable, then you can't extend it, which is the purpose I think

sbidoul, line 32: https://github.com/OCA/account-invoicing/pull/98#issuecomment-136292302 wisely taught us that its not a good way to extend. So making it immutable will force us to resist the temptation.

pedrobaeza, line 32: OK, I remember. Well then

eLBati:
Hi @adrienpeiffer what's the state of this?

elicoidal:
@adrienpeiffer Can you have a look? 
Otherwise we might close this PR.

PR #92, opened 2015-07-27
Port account_invoice_supplier_ref_unique to v8 and new API

alexis-via:

PR #98, opened 2015-08-31
[ADD] support analytic plans

hbrunn:
The small refactoring should make it simpler to add more things like that or to change the matching logic completely in an inheriting module

### DIFF
--- a/account_invoice_merge/invoice.py
+++ b/account_invoice_merge/invoice.py
@@ -22,18 +22,27 @@ from openerp import models, api
 from openerp import workflow
 from openerp.osv.orm import browse_record, browse_null
 
-INVOICE_KEY_COLS = ['partner_id', 'user_id', 'type',
-                    'account_id', 'currency_id',
-                    'journal_id', 'company_id', 'partner_bank_id']
-
-INVOICE_LINE_KEY_COLS = ['name', 'origin', 'discount',
-                         'invoice_line_tax_id', 'price_unit',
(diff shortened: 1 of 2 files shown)


pedrobaeza:
You can already inherit the list and add elements. _account_invoice_merge_ module already do it in https://github.com/OCA/account-invoicing/blob/8.0/account_invoice_merge_payment/models/account_invoice.py#L30, and you break it with you change.

hbrunn:
@pedrobaeza doing it this way is a very bad idea. As soon as account_invoice_merge_payment has a static folder, this key is added when the addon is in the addons path without it being installed. At least, this should be protected by means of `_register_hook` or such, and we should encourage using extension mechanisms that respect the difference between "module is loaded python wise" and "module is installed in the database"

hbrunn:
@pedrobaeza I you agree, I'll include fixing this module in the PR

pedrobaeza:
OK, then please change the extension mechanism to the function then.

hbrunn:
@pedrobaeza done

legalsylvain:
:+1: Thanks.

pedrobaeza:
:+1:

sbidoul:
@adrienpeiffer can you have a look at this one?

adrienpeiffer:
LGTM (Code review and functionnal test). I'll update #92 according changes in this one. :+1:

sbidoul:
:+1:

PR #99, opened 2015-08-31
[8.0][ADD] account invoice pricelist

legalsylvain:
This module add a pricelist field on account invoices.
For the end-users,
- It works like in sale / purchase order. Change partner changes pricelist. 
- Pricelist field is readonly because "on change" feature is not implemented; (See roadmap section);
- search view updated to group by pricelist ;

For developers, it's possible to write reporting depending of this field;

Thanks for your review. 
Regards.


account_invoice_pricelist/README.md
```diff
@@ -0,0 +1,86 @@
...
+Usage
+=====
+
+To use this module, you need to:
+
+* go to ...
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/account-invoicing/8.0
```
max3903: The link is not working. It should be:
https://runbot.odoo-community.org/runbot/95/8.0

max3903:
Can you fix Runbot warnings ?

```
WARNING 3114040-99-a99b6e-all openerp.addons.account_invoice_pricelist.model.account_invoice: Can not compute Pricelist for invoices with type 'in_invoice' because 'purchase' module is not installed.
```

legalsylvain:
@max3903 

> Can you fix Runbot warnings ?

No it is a "functional warning". The pricelist of the demo purchase invoice have not been computed, because 'purchase' module is not installed.

1/ we let this warning;
2/ I can change it to INFO level;
3/ remove the logger, as proposed by @pedrobaeza; (maybe the better solution);

pedrobaeza:
README.md should be README.rst

account_invoice_pricelist/README.rst
```diff
@@ -0,0 +1,86 @@
...
+Pricelist on Invoices
+=====================
+
+* Add a stored field pricelist on invoices, related to the partner pricelist;
+* Possibility to group by pricelist on account.invoice view;
+
+.. image:: static/src/description/screenshot_group_by.png
+
+
+This module doesn't add real feature by it self for end-users, but is usefull
```
pedrobaeza: s/usefull/useful

account_invoice_pricelist/README.rst
```diff
@@ -0,0 +1,86 @@
...
+
+Configuration
+=============
+
+Nothing to do.
+
+Note
+====
+
+* The field will not be stored for purchase invoices as long as 'purchase'
```
pedrobaeza: Why not? You only need to filter by pricelist type. This support is already present in product module, so no need for purchase.

account_invoice_pricelist/__openerp__.py
```diff
@@ -0,0 +1,40 @@
...
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+{
+    'name': 'Account - Pricelist on Invoices',
+    'version': '8.0.1.0.0',
+    'summary': 'Add partner pricelist on invoices',
+    'category': 'account',
```
pedrobaeza: "Accounting & Finance"

account_invoice_pricelist/model/account_invoice.py
```diff
@@ -0,0 +1,76 @@
...
+            else:
+                partner = item.partner_id
+
+            if item.type in ('out_invoice', 'out_refund'):
+                # Customer Invoices
+                item.pricelist_id =\
+                    partner.property_product_pricelist.id
+            elif item.type in ('in_invoice', 'in_refund'):
+                # Supplier Invoices
+                if item.partner_id._model._columns.get(
```
pedrobaeza: Don't go low level with `_columns`. Use standard attrgetter instead.

account_invoice_pricelist/model/account_invoice.py
```diff
@@ -0,0 +1,76 @@
...
+                        'property_product_pricelist_purchase', False):
+                    item.pricelist_id =\
+                        partner.property_product_pricelist_purchase.id
+                else:
+                    _logger.warning(_(
```
pedrobaeza: We shouldn't log an entry each time a supplier invoice is issued, because it can happen lots of times without installing purchase module (using only accounting).

account_invoice_pricelist/model/account_invoice.py
```diff
@@ -0,0 +1,76 @@
...
+        help="The pricelist of the partner, when the invoice is created"
+                " or the partner has changed. This is a technical field used"
+                " to reporting.")
+
+    # Compute Section
+    @api.multi
+    @api.depends('partner_id')
+    def compute_pricelist_id(self):
+        partner_obj = self.env['res.partner']
+        for item in self:
```
pedrobaeza: Do it an @api.one and you avoid this loop

account_invoice_pricelist/model/account_invoice.py
```diff
@@ -0,0 +1,76 @@
...
+                    item.pricelist_id =\
+                        partner.property_product_pricelist_purchase.id
+                else:
+                    _logger.warning(_(
+                        "Can not compute Pricelist for invoices with"
+                        " type '%s' because 'purchase' module is not"
+                        " installed.") % (item.type))
+            else:
+                raise ValidationError(_(
+                    "Can not compute Pricelist for invoices with"
```
pedrobaeza: This shouldn't happen, so the warning can go out

account_invoice_pricelist/README.rst
```diff
@@ -0,0 +1,86 @@
...
+   :target: https://runbot.odoo-community.org/runbot/95/8.0
+
+For further information, please visit:
+
+* https://www.odoo.com/forum/help-1
+
+Known issues / Roadmap
+======================
+
+* The pricelist field is readonly, so it's not possible to change it and to
```
pedrobaeza: It's very simple to change this putting the field as regular one and acting on onchange_partner_id.

account_invoice_pricelist/README.rst
legalsylvain: The new pricelist_id field set in account invoice take the value of : 
- invoice.partner_id.property_product_pricelist.id, if it's a sale invoice; (defined in 'product' module)
- invoice.partner_id.property_product_pricelist_purchase.id, if it's a purchase invoice; (defined in 'purchase' module)
  It's not symetric and I guess it's problem of Odoo core (purchase pricelist should be defined in product module) and I don't want to force people who want to use this module, to install purchase. That's why it's optional, with a warning in the log.
  Regards.

account_invoice_pricelist/README.rst
pedrobaeza: You should include here that the computation of prices according pricelists is not included.

legalsylvain: No so easy. we have to change all the prices / taxes / ... of each line. Otherwise, it is useless.

It is in the scope of this module, but not of this PR.

account_invoice_pricelist/model/account_invoice.py
legalsylvain: thanks !

account_invoice_pricelist/README.rst
pedrobaeza: But this note then is not needed, because you were not seeing any purchase pricelist in partner if purchase module is not installed.

account_invoice_pricelist/README.rst
pedrobaeza: To be not-readonly doesn't mean that the computation should be done. Anyway, you can make it with an onchange and put it readonly on views, because when the other change comes, you'll have to change it anyhow.

legalsylvain: I'm not agree with you on that point. I'm not sure that the aim of OCA module is to describe Odoo feature. (or bugs).
In Odoo when we select a partner with a specific pricelist : 
- It will apply **special price** on sale module (sale.order);
- It will apply **special price** on purchase module (purchase.order);
- It will apply **standard price** on account module (account.invoice);

It's not consistent, but this is Odoo mechanism, and this module doesn't change it (for the time being).

pedrobaeza: Please read second part of my last comment.

account_invoice_pricelist/README.rst
legalsylvain: - If you install this module then purchase module, purchase pricelist will not be computed.
- If you install purchase, then this module, all purchase invoice are initialized with purchase partner pricelist.

Maybe it's not clear, I can change the sentences, but I consider it as an important point. Don't you ?

Thanks for you review.

pedrobaeza: Ah, OK, I didn't understand what you mean. Of course, you should rephrase it and put it "Installation" section something like that:

```
Installation
============

Current invoices will be filled with the corresponding partner pricelist when this module is installed, but supplier invoices and refund don't if you don't have already installed purchase module previously, as there is no supplier pricelist on the partner if you don't install it.
```

legalsylvain:
@max3903 : finally fixed.

legalsylvain:
@pedrobaeza : I believe I understood what you suggested. 
I changed the field from a functional to a regular. Is it now ok for you ?
Thanks.

account_invoice_pricelist/model/account_invoice.py
```diff
@@ -0,0 +1,58 @@
...
+
+class AccountInvoice(models.Model):
+    _inherit = 'account.invoice'
+
+    # Column Section
+    pricelist_id = fields.Many2one(
+        comodel_name='product.pricelist', string='Pricelist',
+        help="The pricelist of the partner, when the invoice is created"
+        " or the partner has changed. This is a technical field used"
+        " to reporting.")
```
pedrobaeza: s/to/for

pedrobaeza:
A little nitpicking, but :+1: 

Thanks!

legalsylvain:
discussion about possible improvement on this PR.
https://www.odoo.com/groups/community-59/community-13834569

regards.

account_invoice_pricelist/model/account_invoice.py
```diff
@@ -0,0 +1,58 @@
...
+
+    # Column Section
+    pricelist_id = fields.Many2one(
+        comodel_name='product.pricelist', string='Pricelist',
+        help="The pricelist of the partner, when the invoice is created"
+        " or the partner has changed. This is a technical field used"
+        " for reporting.")
+
+    @api.multi
+    def onchange_partner_id(
```
moylop260, line 37: In v8 you can use: `@api.onchange('field1', 'field2', ...)`

moylop260, line 37: You can see examples [here](https://github.com/odoo/odoo/blob/411204cad5ef5b86c0645b9d860bbc01e3a68722/addons/account/models/account_invoice.py#L410)

legalsylvain, line 37: Hi @moylop260. I don't think that @api.onchange will works in that case, because, I'm overloading a function, that is not @api.onchange decorated. (see line 41). I have to follow the same technic than the function I'm overloading.
kind regards.

moylop260, line 37: You are right.
Then you can add a new method with `@api.onchange`
You don't need original one, because you are'nt change original values.

pedrobaeza, line 37: @moylop260, only one onchange per field is allowed, and old style one has priority over new one.

legalsylvain, line 37: Thanks @pedrobaeza, I didn't know.

moylop260, line 37: @pedrobaeza 

> only one onchange per field is allowed

Are you sure?

Can you review next cases, please?:
[account_asset/account_asset_invoice.py::_onchange_product_id](https://github.com/odoo/odoo/blob/3dc5a3885582392b498cf6c6c116ed1ec6e70a7f/addons/account_asset/account_asset_invoice.py#L67-L73)
[account/models/account_invoice.py::onchange_product_id](https://github.com/odoo/odoo/blob/3dc5a3885582392b498cf6c6c116ed1ec6e70a7f/addons/account/models/account_invoice.py#L1078)

legalsylvain, line 37: @moylop260 : 
I'm not sure you're pointing a stable version and with the same framework.
in V8 (last stable version) : 
account module: product_id_change : https://github.com/odoo/odoo/blob/8.0/addons/account/account_invoice.py#L1308
account_asset : (no overload) : https://github.com/odoo/odoo/blob/8.0/addons/account_asset/account_asset_invoice.py

Do you have some example in V8 ?
kind regards.

moylop260, line 37: @legalsylvain 
You are right!
My bad.
I had a checkout with branch 9.0

moylop260:
:+1:

PR #100, opened 2015-09-08
[ADD] module 'stock_picking_invoice_product_group'

tafaRU:
# 
# Invoices created from picking grouped by product

This module allows you to group invoices generated from picking by product or by product category.
This is possible by selecting the related option in the wizard used for creating draft invoices.
## Example

Created the following pickings:
1. picking_1 - partner_1
   1. product_A - category_1
   2. product_B - category_2
2. picking_2 - partner_2
   1. product_C - category_2
   2. product_D - category_3
3. picking_3 - partner_1
   1. product_C - category_2
   2. product_D - category_3
   3. product_A - category_1

Selecting option 'Group by product category' I get the following invoices:
1. invoice_1 - partner_1
   1. product_A - category_1 --> picking_1
   2. product_A - category_1 --> picking_3
2. invoice_2 - partner_1
   1. product_B - category_2 --> picking_1
   2. product_C - category_2 --> picking_3
3. invoice_3 - partner_1
   1. product_D - category_3 --> picking_3
4. invoice_4 - partner_2
   1. product_C - category_2 --> picking_2
5. invoice_5 - partner_2
   1. product_D - category_3 --> picking_2

On the contrary if I select option 'Group by product' I get the following invoices:
1. invoice_1 - partner_1
   1. product_A - category_1 --> picking_1
   2. product_A - category_1 --> picking_3
2. invoice_2 - partner_1
   1. product_B - category_2 --> picking_1
3. invoice_3 - partner_1
   1. product_C - category_2 --> picking_3
4. invoice_4 - partner_1
   1. product_D - category_3 --> picking_3
5. invoice_5 - partner_2
   1. product_C - category_2 --> picking_2
6. invoice_6 - partner_2
   1. product_D - category_3 --> picking_2


stock_picking_invoice_product_group/models/account_invoice.py
```diff
@@ -0,0 +1,21 @@
...
+
+from openerp import models, fields
+
+
+class AccountInvoice(models.Model):
+    _inherit = 'account.invoice'
+
+    related_picking_ids = fields.Many2many(
+        comodel_name='stock.picking',
+        relation='account_invoice_stock_picking_rel',
```
moylop260: v8 don't need `relation, column1, column2` params
You can use
`related_picking_ids = fields.Many2many('stock.picking', string='Related Pickings'`

tafaRU: @moylop260 thanks for your review.
Yes I know, the official documentation says:

> The attributes relation, column1 and column2 are optional. If not given, names are automatically generated from model names, provided model_name and comodel_name are different!

but It is not entirely clear to me how table and columns names are generated so that I can use them in post_init_hook.

Any insight would be very useful and highly appreciated.

Thanks.

moylop260: @tafaRU 
I got it
Thank you

moylop260:
:+1:

tafaRU:
@pedrobaeza, I would appreciate it if you could review this PR.
Thank you in advance.

stock_picking_invoice_product_group/README.rst
```diff
@@ -0,0 +1,99 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+================================================
+Invoices created from picking grouped by product
+================================================
+
+This module allows you to group invoices generated from picking by product
```
pedrobaeza: s/picking/pickings

stock_picking_invoice_product_group/README.rst
```diff
@@ -0,0 +1,99 @@
...
+or by product category.
+This is possible by selecting the related option in the wizard used
```
pedrobaeza, line 15: Put how to do it in the usage section

stock_picking_invoice_product_group/models/account_invoice.py
```diff
@@ -0,0 +1,21 @@
...
+# (c) 2015 Alex Comba - Agile Business Group
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+from openerp import models, fields
+
+
+class AccountInvoice(models.Model):
+    _inherit = 'account.invoice'
+
+    related_picking_ids = fields.Many2many(
```
pedrobaeza: Why don't you change stock_picking_invoice_link to support a many2many instead?

stock_picking_invoice_product_group/models/stock_move.py
```diff
@@ -0,0 +1,25 @@
...
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+from openerp import models, api
+
+
+class StockMove(models.Model):
+    _inherit = 'stock.move'
+
+    @api.model
+    def _get_partner_id(self):
```
pedrobaeza: Why are you creating these methods? With mapped function you can retrieve them very easily

stock_picking_invoice_product_group/wizard/stock_invoice_onshipping.py
```diff
@@ -0,0 +1,58 @@
...
+        string='Group Type',
+    )
+
+    @api.model
+    def _group_moves_by_group_type(self, moves, group_type):
+        """Return moves grouped by group_type"""
+        grouped_moves = {}
+        for move in moves:
+            if group_type == 'group_by_product':
+                key = (move._get_partner_id(), move._get_product_id())
```
pedrobaeza: You don't need an specific function to be called here. You can use attrgetter. Example: https://github.com/odoo/odoo/blob/8.0/addons/crm/base_partner_merge.py#L463

stock_picking_invoice_product_group/wizard/stock_invoice_onshipping.py
```diff
@@ -0,0 +1,58 @@
...
+                'sale': 'out_invoice',
+                'purchase': 'in_invoice',
+                'sale_refund': 'out_refund',
+                'purchase_refund': 'in_refund'}
+            inv_type = journal2type.get(self.journal_type) or 'out_invoice'
+            picking_ids = self.env.context['active_ids']
+            picking_model = self.env['stock.picking']
+            pickings = picking_model.browse(picking_ids)
+            moves = []
+            for picking in pickings:
```
pedrobaeza: Use `moves = picking.mapped('move_lines')` and use directly a recordset to avoid lists that forbid you some operations.

stock_picking_invoice_product_group/models/account_invoice.py
tafaRU: Because we cannot modify the type of https://github.com/OCA/stock-logistics-workflow/blob/8.0/stock_picking_invoice_link/stock.py#L60 without breaking the compatibility of those modules which depend on it. Moreover, only with this module has sense to manage this use case. Do you agree?

pedrobaeza: Well, having this field serves for all purposes, so duplication doesn't make change. Even more, the module is broken when you merge some invoices (with the corresponding module), because you can merge several invoices from different pickings, so I consider the change is needed (although it requires migration scripts like your post-init-hook and adapt dependent modules).

eLBati: @pedrobaeza `stock_picking_invoice_link` only depends on `stock_account` and in the standard case the relation between invoices and pickings can only be one2many: 1 picking can only refer to  1 invoice.

Modules that change this relation, like this one or `account_invoice_merge` should handle themselves the relation, if needed. Or "bridge" module should be created (one could be named  `account_invoice_merge_picking_link`, for instance)

I would not change `stock_picking_invoice_link` structure, having to change DB data and modules that depend on it, to support modules which could be not used with it

pedrobaeza: I don't agree. Having each module solving by their own, you will end up with three fields for the same purpose. The base module must handle all possible cases from Odoo core + OCA. Now our ecosystem should be expanded not only for the core. The DB change is something that we can perfectly deal with migration scripts. We have done it previously on lot of OCA modules.

eLBati: Ok, I am going to ask the contributor list for other opinions about this

eLBati: No answers :sweat_smile:

@pedrobaeza coming back to this case, I think the odoo rule "do not change database structure for stable addons" is generally a good rule and should be also valid for OCA addons, when possible.

The change you propose should be postponed to 9.0 version of the modules

pedrobaeza: That rule doesn't apply to OCA in general, or we would be doomed by the first module implementation that appears in the repositories (not always the best for several reasons). Migration scripts are enough for handling that changes.

tafaRU: @pedrobaeza, do you have any reference/example about migration scripts and build on them?
Are they also able to handle the case where the `stock_picking_invoice_link` is already installed and it simply needs an upgrade?

Any help would be appreciated.

pedrobaeza: Yes, migration script is for upgrades (you only need to bump version number). If any other task has to be made on first install, then you need to put a post_init_hook. See the migration scripts we have done for handling l10n_es evolution: https://github.com/OCA/l10n-spain/tree/8.0/l10n_es/migrations

tafaRU:
Travis build errors are related to https://github.com/OCA/account-invoicing/issues/185

stock_picking_invoice_product_group/wizards/stock_invoice_onshipping.py
```diff
@@ -0,0 +1,58 @@
...
+            else:
+                key = (partner.id, categ.id)
+            if key not in grouped_moves:
+                grouped_moves[key] = []
+            grouped_moves[key].append(move)
+        return grouped_moves
+
+    @api.multi
+    def create_invoice(self):
+        if self.group_type:
```
espo-tony, line 43: This method implements the api.multi decorator, which makes theoretically possible to have more than one record in self (also if I understand that in a transient model this is very unlikely to happen). If this happens, the syntax self.group_type will surely raise an error.

I think that a better way to write this is either change the decorator from 'api.multi' to 'api.one' or to use the 'self.ensure_one()' method in order to be sure that self always contain just one record, or to manually iterate the use of this module into the recordset using a syntax like 'for record in self'.

stock_picking_invoice_product_group/wizards/stock_invoice_onshipping.py
```diff
@@ -0,0 +1,58 @@
...
+    def _group_moves_by_group_type(self, moves, group_type):
+        """Return moves grouped by group_type"""
+        grouped_moves = {}
+        for move in moves:
+            partner = move.picking_id.partner_id
+            product = move.product_id
+            categ = product.categ_id
+            if group_type == 'group_by_product':
+                key = (partner.id, product.id)
+            else:
```
espo-tony: Here, in the else statement, is basically developed an implicit "default" for the field group_type: if group_type has not the value 'group_by_product', then the system should suppose that it has the value 'group_by_product_category'. This is actually not correct (the field is not declared as required, therefore it could already assume the None value) and not safe in the perspective of further developments.

I think that a better way to achieve the same result could be to explicitly set a default for the field 'group_type' and to declare it as required (either on the model definition or on the view definition).

In any case, I believe it's better to use an elif branch in order to handle the 'group_by_product_category' case, reserving the else statement to eventually handle any unexpected value.

espo-tony (commented):
This module seems to work without any trouble and implements a very useful functionality.

In the code review, though, I'd like to suggest some lesser improvements.

stock_picking_invoice_product_group/wizards/stock_invoice_onshipping.py
tafaRU: I'd prefer to leave the possibility to choose 'Group by partner' option, this is the reason for that kind of implementation.

stock_picking_invoice_product_group/wizards/stock_invoice_onshipping.py
tafaRU: Fixed, thanks!

stock_picking_invoice_product_group/wizards/stock_invoice_onshipping.py
espo-tony: Of course, I'm not suggesting to prevent the user from choosing an option instead of the other. I'm just suggesting to handle the two different cases ('group_by_product' and 'group_by_product_category') in an if/elif statement rather than in an if/else statement. Something like:

```

if group_type == 'group_by_product':
    key = (partner.id, product.id)
elif group_type == 'group_by_product_category':
    key = (partner.id, categ.id)
else:
    #Some kind of exception raising/handling
```

That's because an else statement in this case would implicitly define a default: let's assume the user doesn't select any group type in the wizard (which is currently possible, being the field not required), the system would behave exactly like if he selected the option 'group_by_product_category', regardless if he really did it or not.

If this is the desired behavior, it would be better to explicitly define that option as a default for the field. Moreover, I believe that the system should force the user to choose an option.

eLBati (approved):

tafaRU:
@espo-tony, could you please have a final review?
Thank you in advance,

espo-tony:
@tafaRU, 
It seems to me that the Travis error is unrelated to this PR, being rather due to the module account_invoice_merge_purchase which is not fully migrated yet.
Therefore in my opinion this PR is ready to be merged. :+1:

PR #101, opened 2015-09-13
[8.0][ADD] new module account_billing

kittiu:


account_billing/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+.. image:: http://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: http://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose mission is to support the collaborative development of Odoo features and promote its widespread use.
+
+To contribute to this module, please visit http://odoo-community.org.
```
eLBati: README should be `.rst`

eLBati:
Thanks @kittiu 

I'm testing the module on runbot. Possible enhancement: to add  'Billing Ref' in voucher form after 'customer' field and to fill 'Paid Amount' using the 'Billing Amount'

Anyway, this module would probably better go to https://github.com/OCA/account-payment , what do you think?

kittiu:
Hi eLBati, you mean,
1. Move Billing Ref to below Customer field
2. Default Paid Amount using the Billing Amount from previous steps?
3. Moving to account-payment

I am agree with 2 and 3. But for 1, we why you think it is better to have Billing Ref on the right of form other than the left?

kittiu:
Moved to account-payment project

PR #104, opened 2015-09-23
[FIX] account_invoice_merge - record set used instead of list of ids in do_merge

ghost:
corrected part 'make link between original sale order'


account_invoice_merge/invoice.py
```diff
@@ -207,8 +206,9 @@ def make_key(br, fields):
                             [('product_id', '=', so_line.product_id.id),
                              ('invoice_id', '=', new_invoice_id)])
                         if invoice_line_ids:
+                            invoice_line_ids_list = [x.id for x in invoice_line_ids]
```
yvaucher: Would be better if you renamed that variable `invoice_lines` as search returns a record set. Thus you can do: `invoice_lines.ids` to get a list of ids.

yvaucher:
Seems that was changed a lot from what I reviewed in 2015.

The title of this PR is not reflecting the fix.

I guess your change was dropped.
I let you open an other PR with your new change.

You should also create a dedicated branch per PR to avoid changing state of a PR.

Thus I will close the current PR.

PR #106, opened 2015-10-04
[ENH] Port sale_order_partial_invoice to V8

kittiu:
Most is ported to v8, except function field_qty_invoiced(), which seem only work in v7 (so i remain it, but need review to make it work).

**Problem of field_qty_invoiced()**
- If I use v8 it does not return the live value, causing problem when,
  - Create SO 10 pieces -> Create first Partial Invoice 3 pieces
  - Create next invoice as "create whole invoice"
  - Expect to have 7 pieces, but have 7 pieces and 3 pieces with negative price

**Problem of this version (and in v7)**
- Seem that Partial Invoice can't be used with Advance Invoice % yet ??

**Possible enhancement**
Allow specify partial invoice in percent (which will auto populate the qty)
- I propose we have another option "% order lines", following "Some order lines"
- Selecting it, will show another "%" field on the wizard,


pedrobaeza:
You should start moving the module from __unported__ dir, and then make the changes. I have also almost finished the migration of the module, so if you find troubles, I can continue, but please make what I'm asking you.

kittiu:
Hello Pedro, thank you very much.
So you are saying **unported** was different from the latest 7.0 branch? What is the benefit?
If you almost finished the porting also. I will follow your work later and I can help test and add feature about % (I need)? WDYT?

kittiu:
@pedrobaeza
So, for this one. Are you working on it? You said you are almost done migration.
Or should I continue (with the unported one). Sorry, I am a bit confused. :P

houssine78:
Hi @kittiu  @pedrobaeza 

I've just spot to the mail thread and I just came by to see what is the progress here? Pedro if you migrated the module in your side maybe that we could close this PR?

youring:
Hello, @kittiu. It seems your fork is missing. @pedrobaeza, any plans for the new PR?
One suggestion: to split the qty_delivered and qty_delivered function into separated module to mimic the functionality inside v9.

rubencabrera:
@jcarlosmontoya

rubencabrera:
The original fork is gone, and after cloning and testing, the module seems to work *almost* fine. How should we proceed to make this mergeable?

PR #112, opened 2015-11-27
Migration of account_invoice_sale_origin module

tafaRU:


damdam-s:
:+1: great

eLBati:
thanks :+1:

pedrobaeza:
Isn't better to add a new Many2one field that directly points to the sale order instead of overwriting the origin text field?

tafaRU:
@pedrobaeza, I'm not sure I understood what you mean. Could you please elaborate it a bit more?

Thank you in advance.

pedrobaeza:
This module takes the origin field (a text), and put there the sales order reference, so I ask if isn't better to add a Many2one field pointing to the sale.order object and let the origin field untouched.

tafaRU:
@pedrobaeza thanks for your clarification but I do not think there would be any difficulty.
Workflows **SO --> invoice** and **picking --> invoice** are distinct so I don't see any possible conflicts, maybe there's something that escapes me.

pedrobaeza:
But then I don't see the utility of this module. Odoo already put in the origin field of the invoice the SO number.

tafaRU:
> Odoo already put in the origin field of the invoice the SO number.

Only in case `order_policy` is set as `picking` (https://github.com/odoo/odoo/blob/8.0/addons/stock_account/stock.py#L282), in other cases I cannot see where this happens, I mean on the invoice line not on the invoice. This, indeed, is the reason for this PR.

pedrobaeza:
OK, I see. Don't you think this can be considered as a bug and a PR to Odoo can amend it? If not, please add this explanation to the README to get better comprehension for those who get the module.

tafaRU:
@pedrobaeza, I've just noticed that link above is wrong, it has to be replaced by https://github.com/odoo/odoo/blob/8.0/addons/stock_account/stock.py#L326 :wink:

tafaRU:
@pedrobaeza 

> please add this explanation to the README to get better comprehension for those who get the module.

How can I improve the current description?

> This module displays the origin field of account invoice lines and populate it when a customer invoice is created from a sale order.

Thank you in advance.

pedrobaeza:
Sorry to come again, but the code I'm seeing says that it works correctly:

https://github.com/odoo/odoo/blob/8.0/addons/sale/sale.py#L1011

tafaRU:
@pedrobaeza you are absolutely right, I didn't see it, probably https://github.com/OCA/account-invoicing/blob/7.0/account_invoice_sale_origin/sale.py#L29 got me very confused.

In view of this, do you consider the module useless or not?

pedrobaeza:
Yeah, with this, the module doesn't do anything new.

tafaRU:
@pedrobaeza, and what about https://github.com/tafaRU/account-invoicing/blob/8.0-account_invoice_sale_origin-migr/account_invoice_sale_origin/views/account_invoice_view.xml#L15 ?
By default the orgin field is not shown on the invoice line, it could be useful to show it, don't you think?

pedrobaeza:
Well, I usually see only the header field, and don't find necessary that detail. I see only useful for reporting where you can group by this field, but this view is not the appropriate one.

PR #113, opened 2015-12-07
Add error if invoices not same user_id

lasley:
Invoice merge fails silently if user_id is not the same on all invoices. Added error to accommodate.


pedrobaeza:
I think the idea is not to fail, so it's better to remove where it occurs? Can you identify it?

lasley:
I agree with this, and can easily identify it. Question is - what user should it be switched to? This module is more of a fix-em anyways, so maybe if we just note that it will change to the user of the first invoice?

pedrobaeza:
Yeah, I think it's the better choice. Anyway, I doubt someone uses this field as a significant one.

lasley:
Totally agree. I'll modify the PR here in a bit :smile:

AaronHForgeFlow:
Closing as stale. Feel free to reopen if working again on this.

PR #114, opened 2015-12-09
[ADD] Add account_invoice_partner_timesheet module

adrienpeiffer:
This module extends the usefulness of account_invoice_partner to invoices generated from timesheets.

This is needed because one normally wants to put the customer (company) on the projects/analytic accounts, while the invoices must be sent to the accounting department which is a contact of the customer.


sbidoul:
@adrienpeiffer code looks good. Is the dependency on account_invoice_partner required?

adrienpeiffer:
Hi Stéphane !

Technically not, but without account_invoice_partner, I find the behavior quite strange.

sbidoul:
> Technically not, but without account_invoice_partner, I find the behavior quite strange.

Not necessarily if you consider that the invoice address is correctly used by Odoo when invoicing from SO for instance.

So I'd say this one is complementary to `account_invoice_partner`. But then it probably needs to be renamed `hr_timesheet_invoice_contact`.

PR #127, opened 2016-02-15
[SEC] account_invoice_validation_workflow: fix unsafe eval

gurneyalex:


gurneyalex:
I did not change the version to 9.0 on purpose as the module is still unported to v9

legalsylvain:
:+1: Thanks.

Is there any reason to change the version of a non installable module ? or it's just a good practice ? 

kind regards.

pedrobaeza:
Should we this time leave this without merging and import the commits from 8.0?

antespi:
:+1:

gurneyalex:
@legalsylvain Its a side effect of cherry picking.

I have no strict opinion on this, and leave the decision to the PSC. My goal in submitting a PR is avoiding false positives on unported addons when looking for suspicious code across the OCA repositories (and making sure this bug does not pop back if the developer porting to v9 misses this, since the problem likely won't show up in the diff)

max3903:
:+1:

PR #128, opened 2016-02-19
[8.0] [ADD] stock_picking_invoicing_unified

pedrobaeza:
# 
# Stock Picking Invoicing Unified

Odoo allows to select several pickings and click on "Create Draft Invoices"
option for creating the corresponding invoice(s) depending if you have
selected several partners and if you have marked the "Group by partner"
check.

But it only takes into account the first picking for selecting the type of the
invoice you are going to create (customer/supplier invoice/refund), mixing all
the lines on it. And not only that: if you have returned pickings, the returned
quantities are summed to the rest, instead of decreasing the amount to invoice,
which is the common practise when you have some returns.

This module overpasses this limitation, allowing to invoice them all together
without having to worry about that, and also having the extra feature of
reducing the amount in the invoice when you have delivered or received goods
and some of them (or previous ones) have been returned, instead of needing
to create 2 invoices and to conciliate them for the final/real amount.
# Usage
- Select several pickings from any of the menus that allows it (
  _Warehouse > All Operations_ and click on any of the lines,
  _Purchases > Invoice Control > On Incoming Shipments_, etc).
- Click on _More > Create Draft Invoices_.
- In the resulting dialog, the proper invoices types that are going to be
  created are computed, and you have to select the journals for that types.
- Click on _Create_ button, and the invoices will be correctly created.
- The lines of pickings that are not of the greater picking type are created
  with negative price unit.
# Known issues / Roadmap
- Add tests


clonedagain:
@lbellier maybe you will want to have a look at this module

oihane:
LGTM :+1:

alexis-via:
This is a really great module, I wanted to develop it one day, and you did it before me :)

I made a PR to enhance it:
https://github.com/pedrobaeza/account-invoicing/pull/1

The main change in my PR is that, for a return, the quantity should be < 0, not the price ! For example, you sell 2 units and the customer returns 1. With your implementation, in the invoice stats, you will see that you sold 3 units ! And if you want to do statistics on the average sell price for a particular product, it breaks it too.

My PR also adds default value to journals, to keep the same behavior on that point.

stock_picking_invoicing_unified/wizard/stock_invoice_onshipping.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+class StockInvoiceOnshipping(models.TransientModel):
+
+    _inherit = 'stock.invoice.onshipping'
+
+    journal_id = fields.Many2one(required=False)
+    sale_journal = fields.Many2one(
+        comodel_name='account.journal', string='Sale Journal',
+        domain="[('type', '=', 'sale')]")
```
alexis-via: For curiosity: is there a good reason to use domain as a string, instead of sequence: domain=[('type', '=', 'sale')] ? I usually declare domains as a sequence in my Python code, not as a string... but I don't know if it is equivalent or not.

pedrobaeza: Of course it can be a list of tuples! Maybe a copy/paste from an XML view leads to this.

pedrobaeza:
@alexis-via, I haven't put negative quantities because I remember a constraint that prevents it. Did you check it?

alexis-via:
@pedrobaeza I didn't hit such a constraint during my tests. AFAIK, it is possible to have a negative quantity on invoice line.

alexis-via:
:+1: let's merge it !

alexis-via:
@pedrobaeza In the "models" subdir, the file stock_picking.py should be renamed stock_move.py

pedrobaeza:
@alexis-via, changes done.

andhit-r:
Test steps:
1. Create invoice from 1 delivery order and 1 incoming shipment (same partner)
2. Open customer invoice created from step 1
3. customer invoice data opened through supplier invoice form (see screenshot)

selection_001

It will not break any flow nor produce wrong accounting entry, but IMHO it will make untrained users confuse. Better put explanation on known issue. Other then that :+1:

pedrobaeza:
The form view issue can't be solved easily. The only solution is to avoid the opening in form view. I'll merge this one and noted that problem as an issue. The standard also has that problem.

PR #130, opened 2016-02-23
[8.0][ADD] sale_invoice_plan (and dependencies)

kittiu:


kittiu:
Hello, there are some more things to update, i..e, lint check, tests, which I can add.
However, I am aware that it is using many hook addon (which is really needed in my opinion). Will it still eligible for OCA? Will there be issues?
it depending on order_invoice_line_percentage (this module can be used alone), which also depend on purchase (as it used for both kind of order), is it ok?

PR #135, opened 2016-03-07
[PORT] 9.0 account invoice shipping address

ghost:
This modules adds shipping address to invoice.


account_invoice_shipping_address/__openerp__.py
```diff
@@ -23,9 +23,9 @@
 
 {
     "name": "Invoice Shipping Address",
-    'summary': """
+    'description': """
```
pedrobaeza: Don't put description. The description is in the README.rst file. This is only summary.

pedrobaeza:
Thanks for the PR. I have seen that there's a very similar module on https://github.com/OCA/account-invoice-reporting/tree/9.0/account_invoice_delivery_address.

Can you tell me which one is better?

ghost:
I think the module you mentioned is more complete than this. It controls the delivery address and also provides an RML report. What do you think of moving those to this PR and replacing it with QWeb report?

pedrobaeza:
Yeah, that seems a better option.

account_invoice_shipping_address/__openerp__.py
ghost: Just summary key?

pedrobaeza: From Odoo v8, the description is automatically binded from the README.rst file, so no need to duplicate it.

PR #137, opened 2016-03-09
[IMP] account_invoice_line_description: get product aware of the partner language

tafaRU:


pedrobaeza:
:+1:

PCatinean:
Passing the language object in the context is the way to go or should the code be passed? Or both work?

PCatinean:
It seems standard addons have the same practice so :+1:

hbrunn:
`lang` is a selection field, so it's not an object in the first place: https://github.com/OCA/OCB/blob/8.0/openerp/addons/base/res/res_partner.py#L246

is this PR causing travis to fail?

tafaRU:
> is this PR causing travis to fail?

It seems not.
https://travis-ci.org/OCA/account-invoicing/jobs/114791659#L1046 suggests **account_invoice_merge_purchase** as guilty.

gurneyalex:
:+1:

hbrunn:
@tafaRU I'd have a look at the tests that fail and see if they do something that presupposes a certain language, English probably

gurneyalex:
@hbrunn the test on the 8.0 branch were green 4 days ago. I reran them today and now they are red :crying_cat_face: 

-> something changed (external dependency? Odoo?) which broke the tests here...

pedrobaeza:
It seems that the validation process is interfering. Maybe an extra module is now installed?

gurneyalex:
I'm merging this fix as the breakage is clearly unrelated.

PR #149, opened 2016-04-27
[9.0] account_invoice_force_number

dcorio:
Initial porting to v9


eLBati:
Looking at the diff with version 8
http://pastie.org/10817456
it seems files like i18n/de.po and security/security.xml are removed.

Are you missing some commits from 8.0 ?

If you want, you can follow this https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-9.0#if-the-module-already-exists-in-the-90-branch-but-it-doesnt-contain-all-the-commit-history-or-youre-not-sure-if-it-have-it

dcorio:
really weird, didn't notice.
thanks for the link, i'm checking

PR #150, opened 2016-04-29
[9.0] account_invoice_force_number

dcorio:
module ported to v9
i closed the previous pr to correctly migrate commit history from v8


dcorio:
@eLBati @tafaRU let's try again :)

account_invoice_force_number/__openerp__.py
```diff
@@ -1,27 +1,11 @@
...
+# © 2011-2016 Agile Business Group sagl (<http://www.agilebg.com>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
 {
     'name': "Force Invoice Number",
-    'version': '8.0.0.1.0',
+    'version': '9.0.0.1.0',
     'category': 'Accounting & Finance',
     'summary': "Allows to force invoice numbering on specific invoices",
     'description': """
```
tafaRU: Please remove it as it's already present in README.

account_invoice_force_number/__openerp__.py
```diff
@@ -30,18 +14,13 @@
...
     'license': 'AGPL-3',
-    "depends": [
-        'account'
-    ],
-    "data": [
-        'security/security.xml',
-        'invoice_view.xml'
-    ],
+    'depends': ['base', 'account', 'product'],
+    'data': ['views/invoice_view.xml'],
```
eLBati: @dcorio `security/security.xml` should be kept

account_invoice_force_number/views/invoice_view.xml
```diff
@@ -0,0 +1,30 @@
...
+    </record>
+
+    <record model="ir.ui.view" id="account_invoice_supplier_form_int_number">
+        <field name="name">account.invoice.supplier.form.int_number</field>
+        <field name="model">account.invoice</field>
+        <field name="inherit_id" ref="account.invoice_supplier_form"/>
+        <field name="arch" type="xml">
+            <field name="journal_id" position="after">
+                <field name="internal_number"
+                    attrs="{'invisible': [('state', '!=', 'draft')]}"/>
```
eLBati: I'm looking at the diff with version 8: http://pastie.org/10819678
Please keep old attributes, like `help` and `groups`

eLBati: Sorry, just noticed `help` is in python file now

account_invoice_force_number/__openerp__.py
```diff
@@ -1,46 +1,22 @@
...
-    """,
     'author': "Agile Business Group,Odoo Community Association (OCA)",
     'website': 'http://www.agilebg.com',
     'license': 'AGPL-3',
-    "depends": [
-        'account'
-    ],
-    "data": [
-        'invoice_view.xml'
+    'depends': ['base', 'account', 'product'],
```
tafaRU: Why the dependencies are changed? Shouldn't be enough to use account?

tafaRU:
@dcorio, please see my last comment and please reactivate groups attribute as spotted at https://github.com/OCA/account-invoicing/pull/150#discussion_r61683897

Thank you in advance.

account_invoice_force_number/__openerp__.py
dcorio: @tafaRU i was thinking this was needed because tests depend on base and product too. but i was wrong. thanks

dcorio:
@tafaRU should be ok now. thanks for pointing that out

tafaRU:
@dcorio, it seems you didn't push any changes, could you please check it?

dcorio:
@tafaRU sorry. pushed on the correct branch this time.

tafaRU:
@dcorio, I would have preferred the test to be python rather than yaml.
Apart from this, the rest LGTM, thanks! :+1:

dcorio:
@tafaRU thanks. i can convert the test from yaml to unittest to ease maintenance and debug

tafaRU:
@dcorio, ok, thanks!
Also please take the occasion to remove active key as it is an alias for auto_install, set False as default.

eLBati:
Thanks :+1:

dcorio:
Please, set this PR as "WIP" since i just discovered that the internal_number field isn't necessary and i probably just need to use "move_name" instead

dcorio:
@eLBati thanks for the label, but i've decided that, even if we have a move_name field that we should use for the very same purpose of internal_number, it's better to not use it because we'd net to set it as readonly=False, change the string (which is "Journal Entry" right now), the help, etc...

So IMO, this module can be tested as it is now

tafaRU:
> since i just discovered that the internal_number field isn't necessary and i probably just need to use "move_name" instead

@dcorio, is there any news on this?

dcorio:
@tafaRU i'm still for using the internal_number field for that. Using the move_name would require changing the field string that's not less "dirty" than using a second field for this.

can i ask your opinion on this?

I also need to fix a few other things before we can test it

tafaRU:
@dcorio,
what do you mean by
> Using the move_name would require changing the field string that's not less "dirty" than using a second field for this.  

?
Well, from what I can see, `internal_number` is a fields.Char too, so I think the right way to proceed is using `move_name`.
Just to inform you: I'm going to move this PR to branch 10.0 as at the moment I'm not interested to continue on branch 9.0.

tafaRU:
@dcorio, https://github.com/OCA/account-invoicing/pull/242

eLBati:
> Well, from what I can see, internal_number is a fields.Char too, so I think the right way to proceed is using move_name

Agree: it seems ```move_name``` is the new name of ```internal_number```.
See its help:
```
Technical field holding the number given to the invoice, automatically set when the invoice is validated then stored to set the same number again if the invoice is cancelled, set to draft and re-validated
```

PR #151, opened 2016-05-09
FIX  account_invoice_rounding: tax included in price + discount

eLBati:
See commits

### DIFF
--- a/account_invoice_rounding/account.py
+++ b/account_invoice_rounding/account.py
@@ -57,28 +57,11 @@ class AccountInvoice(models.Model):
         return {'amount_total': rounded_total,
                 'amount_untaxed': amount_untaxed}
 
-    @staticmethod
-    def _all_invoice_tax_line_computed(invoice):
-        """ Check if all taxes have been computed on invoice lines
-        :return boolean True if all tax were computed
-        """
-        tax_ids = set()
-        for line in invoice.invoice_line:
-            # invoice_line_tax_id is a many2many if you wonder about it
-            for tax in line.invoice_line_tax_id:
-                if not tax.price_include:
-                    tax_ids.add(tax.id)
-        computed_tax_ids = [tax.id for tax in invoice.tax_line]
-        return len(tax_ids) == len(computed_tax_ids)
-
     def _swedish_round_globally(self, cr, uid, invoice,
                                 rounded_total, delta, context=None):
         """ Add the diff to the biggest tax line
         This ajustment must be done only after all tax are computed
         """
-        # Here we identify that all taxe lines have been computed
-        if not self._all_invoice_tax_line_computed(invoice):
-            return {}
 
         obj_precision = self.pool.get('decimal.precision')
         prec = obj_precision.precision_get(cr, uid, 'Account')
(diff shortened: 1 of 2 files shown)


account_invoice_rounding/account.py
```diff
@@ -57,28 +57,11 @@ def _swedish_add_invoice_line(self, cr, uid, invoice,
...
-        tax_ids = set()
-        for line in invoice.invoice_line:
-            # invoice_line_tax_id is a many2many if you wonder about it
-            for tax in line.invoice_line_tax_id:
-                if not tax.price_include:
-                    tax_ids.add(tax.id)
-        computed_tax_ids = [tax.id for tax in invoice.tax_line]
-        return len(tax_ids) == len(computed_tax_ids)
-
     def _swedish_round_globally(self, cr, uid, invoice,
```
elicoidal, line 60: Why Swedish? Is there any special rounding that does not apply anywhere else?
If possible I would rename to _round_globally()

account_invoice_rounding/test/test_invoice_rounding.yml
```diff
@@ -17,6 +17,37 @@
     type_tax_use: all
     tax_code_id: tax_case
 -
+  I create a Tax Code for 8%
```
elicoidal, line 20: As far as I know yml tests are obsolete: @pedrobaeza can you confirm?

account_invoice_rounding/account.py
eLBati, line 60: See https://en.wikipedia.org/wiki/Cash_rounding

account_invoice_rounding/test/test_invoice_rounding.yml
eLBati, line 20: It seems 9.0 and master use them:
https://github.com/odoo/odoo/blob/9.0/addons/stock/__openerp__.py#L79
https://github.com/odoo/odoo/blob/master/addons/stock/__openerp__.py#L94

eLBati:
Hi @yvaucher what do you think about this?

tafaRU (approved):

rafaelbn:
This PR sound quite important but alto no reviewers :thinking: 

Well we are not using v8 anymore and I cannot review it technically sorry! :cry:

rafaelbn:
@eLBati could you rebase?

eLBati:
It seems meanwhile yml file was removed and tests refactored to python:
https://github.com/OCA/account-invoicing/blob/8.0/account_invoice_rounding/tests/test_invoice_rounding.py

I will not refactor tests of this PR to python, but I can remove ```test_invoice_rounding.yml``` and leave the bugfix part

account_invoice_rounding/account.py
yvaucher: @elicoidal that is the original name of the module and then this method as it is the "wiki-generic" name for this kind of rounding. Could be renamed `cash rounding` thought.

account_invoice_rounding/account.py
```diff
@@ -57,28 +57,11 @@ def _swedish_add_invoice_line(self, cr, uid, invoice,
...
                                 rounded_total, delta, context=None):
         """ Add the diff to the biggest tax line
         This ajustment must be done only after all tax are computed
```
yvaucher, line 63: This docstring should probably be updated then.

yvaucher (approved):
This module was created on version 7.0, then IIRC in 8.0 taxes are computed differently is that right?

Thus this change would make sense.

Just don't forget to update the docstring of the method, please.

PR #158, opened 2016-06-10
8.0

pedrorgil:


eLBati:
Hi @pedrorgil 
what's the purpose of this change?

pedrorgil:
Fix the  exception that is thrown because in api v8 the method "read" is not receiving the "ids" parameter that is being passed here (it was in api v7), so I repaced it with a "browse"

PR #161, opened 2016-07-01
Migration of account_invoice_supplier_ref_unique to 9.0

ThomasBinsfeld:


ThomasBinsfeld:
@sbidoul

sbidoul:
Thanks!

A few cosmetic remarks
- use short headers (example https://github.com/OCA/maintainer-tools/blob/master/template/module/models/model_name.py)
- update the readme to match the latest OCA template https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst

Could you add a short test? (example here https://github.com/odoo/odoo/blob/9.0/addons/account/tests/test_account_supplier_invoice.py)

account_invoice_supplier_ref_unique/tests/test_account_invoice_supplier_ref_unique.py
```diff
@@ -0,0 +1,33 @@
+# -*- coding: utf-8 -*-
+# Copyright 2015 Akretion
+# Copyright 2010 - 2014 Savoir-Faire-Linux
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
```
sbidoul, line 3: This new file is copyright Acsone, not Akretion/SFL.

account_invoice_supplier_ref_unique/tests/test_account_invoice_supplier_ref_unique.py
```diff
@@ -0,0 +1,33 @@
...
+
+from openerp.addons.account.tests.account_test_classes\
+    import AccountingTestCase
+from openerp.exceptions import ValidationError
+
+
+class TestAccountInvoiceSupplierRefUnique(AccountingTestCase):
```
sbidoul: I don't think it's necessary to inherit from AccountingTestCase, you don't use anything from that, and it makes the test more fragile.

account_invoice_supplier_ref_unique/tests/test_account_invoice_supplier_ref_unique.py
```diff
@@ -0,0 +1,33 @@
...
+
+        # Creation of an invoice instance
+        self.env['account.invoice'].create({
+            'partner_id': self.env.ref('base.res_partner_2').id,
+            'account_id': invoice_account,
+            'type': 'in_invoice',
+            'supplier_invoice_number': 'ABC123'})
+
+        # A new invoice instance with an existing supplier_invoice_number
+        with self.assertRaises(ValidationError), self.cr.savepoint():
```
sbidoul: is the savepoint necessary?

account_invoice_supplier_ref_unique/README.rst
```diff
@@ -8,10 +13,10 @@ This module adds a constraint on supplier invoice/refunds to check that (commerc
 Bug Tracker
 ===========
 
-Bugs are tracked on `GitHub Issues <https://github.com/OCA/account-invoicing/issues>`_.
-In case of trouble, please check there if your issue has already been reported.
-If you spotted it first, help us smashing it by providing a detailed and welcomed feedback
-`here <https://github.com/OCA/account-invoicing/issues/new?body=module:%20account_invoice_supplier_ref_unique%0Aversion:%208.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/{project_repo}/issues>`_. In case of trouble, please
```
sbidoul: replace `{project_repo}`

fclementic2c:
Question : Shouldn't we rather have a unique supplier invoice _per supplier_ ?

sbidoul:
> Question : Shouldn't we rather have a unique supplier invoice per supplier ?

@fclementic2c it should be the case: commercial_partner_id is part of the search key

account_invoice_supplier_ref_unique/tests/test_account_invoice_supplier_ref_unique.py
```diff
@@ -0,0 +1,31 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openerp.exceptions import ValidationError
+from openerp.tests import common
+
+
+class TestAccountInvoiceSupplierRefUnique(common.TransactionCase):
+
+    def test_check_unique_supplier_invoice_number_insensitive(self):
+        invoice_account = self.env['account.account'].search(
```
sbidoul: nitpicking: s/invoice_account/invoice_account_id/

sbidoul:
:+1:

sbidoul:
Hm, retracting my plus one, because while testing on runbot, I noticed this `supplier_invoice_number` field does not exist anymore in v9. So while in v8 this field redefinition merely set copy=False, in v9 this creates a new field.

In v9 we have apparently only one field named "reference", which relates to the reference_type field (None / Free Reference). So, that reference field seems to be intended as a receptacle for the payment reference, which is in many case different from the supplier invoice number.

So we have two possible approaches here:
1. use the `reference` field as part of the unique key
2. reintroduce `supplier_invoice_number` in this module and add it in the supplier invoice view just before the reference field

I favor option 2. What do other think? @alexis-via @pedrobaeza @chrisdec @fclementic2c

pedrobaeza:
I have always needed only one field for this. Why do you need 2?

sbidoul:
@pedrobaeza for instance, when using structured communications for payment, the supplier invoice number is different from the payment reference/communication.

sbidoul:
It's common to have a supplier using the same structured communication as payment reference on all invoices to a given customer. But the customer entering the invoices wants to avoid duplicates and it can only use the supplier invoice number for that, not the payment reference.

pedrobaeza:
Uhm, I don't have that case in any customer (they don't even use a fixed payment reference in suppliers), but if you see that the two fields are needed, then go with them.

alexis-via:
In France, as we don't use structured payment references, I use the "reference" field to store the supplier invoice number, because the supplier invoice number is used as payment reference, so I would be in favor of option 1. But I understand your point and I don't see how we can change it without modifying the datamodel... As you use structured payment reference in Belgium and we have a Belgian editor, it's surprising that Odoo S.A. didn't handle this problem...

sbidoul:
Ok, so I suggest the following changes to this PR:
- make `supplier_invoice_number` a real field, with the same definition as in v8
- add this field to the supplier invoice / credit note form, before the `reference` field (or reference type, to be checked?)
- add a onchange that copies supplier_invoice_number to reference if reference is empty

@ThomasBinsfeld can you do it so we can check how it works on runbot?

ThomasBinsfeld:
@sbidoul

sbidoul:
@ThomasBinsfeld the invoice number field must be _before_ the reference field, because it is not natural to have a onchange modifying a field displayed before.

The invoice number field must have the same readonly properties as the reference field (ie redonly when not draft).

Perhaps the module must be renamed `account_invoice_supplier_num_unique` ?

sbidoul:
Change field label => Vendor Invoice Number for consistency

image

Otherwise :+1: (code review + functional test)

sbidoul:
This one is ready for final review.

account_invoice_supplier_ref_unique/models/account_invoice.py
```diff
@@ -12,15 +12,17 @@ class AccountInvoice(models.Model):
     supplier_invoice_number = fields.Char(
         string='Vendor invoice number',
         readonly=True,
-        states={'draft': [('readonly', False)]},
-        copy=False)
+        states={'draft': [('readonly', False)]})
```
sbidoul: @ThomasBinsfeld I think this copy=False can stay

account_invoice_supplier_ref_unique/models/account_invoice.py
```diff
@@ -41,3 +43,31 @@ def _check_unique_supplier_invoice_number_insensitive(self):
...
+
+        return vals
+
+    @api.multi
+    def copy(self, default=None):
+        '''
+        The unique vendor invoice number is not copied in vendor bills
+        '''
+        if self.type in ['in_invoice', 'in_refund']:
+            default = dict(default or {}, supplier_invoice_number='')
```
sbidoul: it's rather the reference field that must not be copied in the case of supplier invoices, since it is unlikely to have two supplier invoices with the same payment reference, but that is debatable; @chrisdec any opinion on this?

sbidoul: Ok, @ThomasBinsfeld please leave copy=False on supplier_invoice_number and prevent copying the `reference` field for supplier invoices and refunds.

sbidoul:
@chrisdec can you retest this on runbot.odoo-community.org, also in combination with #162 ?

adrienpeiffer:
:+1: Functional test and code review !

Cedric-Pigeon:
:+1:  code review

account_invoice_supplier_ref_unique/__openerp__.py
```diff
@@ -1,34 +1,17 @@
...
-#
-##############################################################################
+# -*- coding: utf-8 -*-
+# Copyright 2016 Acsone
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 {
     'name': 'Unique Supplier Invoice Number in Invoice',
-    'version': '8.0.1.1.0',
+    'version': '9.0.1.1.0',
```
pedrobaeza: Version is 9.0.1.0.0

account_invoice_supplier_ref_unique/__openerp__.py
```diff
@@ -1,34 +1,17 @@
...
     'summary': 'Checks that supplier invoices are not entered twice',
     'author': "Savoir-faire Linux,Odoo Community Association (OCA)",
```
pedrobaeza: You can add yourself here

account_invoice_supplier_ref_unique/__openerp__.py
```diff
@@ -1,34 +1,17 @@
...
     'maintainer': 'Savoir-faire Linux',
-    'website': 'http://www.savoirfairelinux.com',
+    'website': 'https://odoo-community.org',
```
pedrobaeza: But don't change website

account_invoice_supplier_ref_unique/models/account_invoice.py
```diff
@@ -0,0 +1,74 @@
...
+    supplier_invoice_number = fields.Char(
+        string='Vendor invoice number',
+        readonly=True,
+        states={'draft': [('readonly', False)]},
+        copy=False)
+
+    @api.one
+    @api.constrains('supplier_invoice_number')
+    def _check_unique_supplier_invoice_number_insensitive(self):
+        '''
```
pedrobaeza: Use double quota for comments, and first line contains text.

account_invoice_supplier_ref_unique/models/account_invoice.py
```diff
@@ -0,0 +1,74 @@
...
+
+    @api.onchange('supplier_invoice_number')
+    def _onchange_supplier_invoice_number(self):
+        if not self.reference:
+            self.reference = self.supplier_invoice_number
+
+    @api.model
+    def _prepare_refund(self, invoice, date_invoice=None,
+                        date=None, description=None, journal_id=None):
+        '''
```
pedrobaeza: Double quote

account_invoice_supplier_ref_unique/models/account_invoice.py
```diff
@@ -0,0 +1,74 @@
...
+        The unique vendor invoice number cannot be passed to the credit note
+        in vendor bills
+        '''
+        vals = super(AccountInvoice, self)._prepare_refund(invoice,
```
pedrobaeza: You can make this to avoid a lot of lines:

```
vals = super(AccountInvoice, self)._prepare_refund(
    invoice, date_invoice, date, description, journal_id)
```

account_invoice_supplier_ref_unique/models/account_invoice.py
```diff
@@ -0,0 +1,74 @@
...
+                                                           date_invoice,
+                                                           date,
+                                                           description,
+                                                           journal_id)
+
+        if invoice and invoice.type in ['in_invoice', 'in_refund'] and\
+                'reference' in vals:
+            vals['reference'] = ''
```
pedrobaeza, line 60: Should this be the other field (supplier_invoice_number)?

account_invoice_supplier_ref_unique/models/account_invoice.py
```diff
@@ -0,0 +1,74 @@
...
+
+        return vals
+
+    @api.multi
+    def copy(self, default=None):
+        '''
```
pedrobaeza: Double quote

account_invoice_supplier_ref_unique/models/account_invoice.py
```diff
@@ -0,0 +1,74 @@
...
+        The unique vendor invoice number is not copied in vendor bills
+        '''
+        if self.type in ['in_invoice', 'in_refund']:
+            default = dict(default or {}, reference='')
```
pedrobaeza, line 70: Should this be the other field (supplier_invoice_number)?

account_invoice_supplier_ref_unique/tests/test_account_invoice_supplier_ref_unique.py
```diff
@@ -0,0 +1,53 @@
...
+    @classmethod
+    def setUpClass(self):
+        super(TestAccountInvoiceSupplierRefUnique, self).setUpClass()
+
+        # ENVIRONMENTS
+        self.account_account = self.env['account.account']
+        self.account_invoice = self.env['account.invoice']
+
+        # INSTANCES
+        self.partner = self.env.ref('base.res_partner_2')
```
pedrobaeza, line 20: I prefer to not use demo data, as they change from version to version, and this case is very simple to re-create a new one.

account_invoice_supplier_ref_unique/views/account_invoice.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
```
pedrobaeza: Use `<odoo>` and remove `<data>` tag

account_invoice_supplier_ref_unique/views/account_invoice.xml
```diff
@@ -0,0 +1,17 @@
...
+        <field name="inherit_id" ref="account.invoice_supplier_form"/>
+        <field name="arch" type="xml">
+            <field name="reference" position="before">
+                <field name="supplier_invoice_number"/>
+            </field>
+        </field>
+    </record>
+
+</data>
+</openerp>
```
pedrobaeza: Line at the end

pedrobaeza (changes requested):

account_invoice_supplier_ref_unique/models/account_invoice.py
ThomasBinsfeld: No. The field supplier_invoice_number is not copied nor passed to the refund invoice anyway. But we need to cancel the reference passing because it took the unique reference value. And it is a problem if you cannot edit the reference.

pedrobaeza (approved):

PR #105, opened 2015-09-29
[ADD] account_invoice_pricelist: use pricelist when calculating prices

hbrunn:
This adds support for adding invoice lines by hand while still getting the correct price from the chosen pricelist

PR #163, opened 2016-07-12
[ADD] [8.0] Module account_invoice_line_pricelist

espo-tony:
By installing this module, on customer/sale invoice lines the unit price will
be calculated based on the Price List set for the Customer.


astirpe:
@espo-tony you are now an [OCA member](https://odoo-community.org/members/onestein-b-v-antonio-esposito-3769), congratulations!
This PR is ready to be reviewed.
:+1:

ehahouimohssine:
Can you add the functionality (to invoice) that is also present on the sales order? The functionality is: When you select a customer that has a pricelist you can modify that pricelist by a selection box. This way you're not depedent on the selected pricelist.

espo-tony:
Please, someone could explain me why Travis raises an error on my commit, when testing on Branch 8.0? For what I can read in the related log it seems that is another module to fail.

elicoidal:
:+1: Excellent job

elicoidal:
Trravis error seems unrelated (account_invoice_merge_purchase)

ehahouimohssine:
I have one feedback point. If the client is selected on the invoice , the field "price list" need to be automatically filled in with the price list that is active for the customer. (The user can immediately see the price list that is active for the customer)

espo-tony:
Thanks for the feedback: this is exactly the behavior I designed, and for this reason:
If you don't select any pricelist from the invoice, it will use, by default, the pricelist which is selected on the partner (and of course, if you change partner the default pricelist will update as well). The pricelist field on the invoice is only meant to be used to select a different pricelist than what is currently setted on the partner itself.

Do you think this behavior should be changed, just to give to the user an immediate information on what pricelist is currently using?

pedrobaeza:
Please don't start another PR when there's one module that is intended for that (account_invoice_pricelist), and there's a PR which already covers the same topic: https://github.com/OCA/account-invoicing/pull/105. There is still a little problem in it, so maybe you have already fixed it in your module, but what you have to do instead is to complete that PR, making a PR over @hbrunn's branch, or directly superseed his PR taking his commits and adding yours.

Until then, :-1:

espo-tony:
Hi Pedro,

You're right, we didn't notice that there was already a PR on the same subject.
My implementation should fix the problem you mentioned in the other PR about the quantity related pricelist. Also, my implementation takes care about different currencies set on invoice and pricelist. We will create a PR on the branch owned by @hbrunn ASAP.

Thank you for the advice!

espo-tony:
For anyone interested in using this module, you can find it here: [account_invoice_line_pricelist](https://github.com/onesteinbv/addons-onestein/tree/8.0/account_invoice_line_pricelist)

PR #168, opened 2016-07-19
9.0-add-account_invoice_change_currency

susport:
This module replace original odoo wizard for changing currency on an
invoice with serveral
improovements:
- Preview and allow to change the
  rate thats is going to be used:
- Log the currency change on the
  chatter
- Add this functionality to supplier invoices


account_invoice_change_currency/wizard/account_change_currency.py
```diff
@@ -0,0 +1,57 @@
...
+    _name = 'account.change.currency'
+    _description = 'Change Currency'
+
+    currency_id = fields.Many2one('res.currency', string='Change to',
+                                  required=True,
+                                  help="Select a currency to "
+                                       "apply on the invoice")
+    currency_rate = fields.Float('Currency Rate',
+                                 required=True,
+                                 help="Select a currency to "
```
espo-tony, line 16: I suppose the help attribute for this field is incorrect.

espo-tony (commented):
I've performed a test of the functionalities implemented by this module and everything seems to work well.

Still I'd like to suggest a little improvement: I noticed that there isn't any constrains on the currency rate used for the conversion, allowing the user to potentially set a negative or 0 number for that field, which is functionally incorrect, being a currency rate a positive number by definition.

PR #170, opened 2016-07-20
[9.0] [ADD] account_invoice_blocking

ThomasBinsfeld:
Partial migration of https://github.com/OCA/bank-payment/tree/8.0/account_payment_blocking

This module allows the user to set a blocking (No Follow-up) flag directly on the invoice. This facilitates the blocking of the invoice's move lines.


ThomasBinsfeld:
@sbidoul @adrienpeiffer

adrienpeiffer:
Great ! :+1: Code review and functionnal test

sbidoul:
:+1: (code review) 

Very useful module to preserve the blocked flag when cancelling  invoices for correction.

adrienpeiffer:
I think this one is ready to merge.

account_invoice_blocking/models/account_invoice.py
```diff
@@ -0,0 +1,82 @@
...
+            invoice._update_blocked(invoice, invoice.blocked)
+
+    @api.multi
+    def action_move_create(self):
+        '''
+        This method overrides the invoice's move line creation
+        This method calls the update of the invoice's move lines based on
+        the value of the field 'draft_blocked'
+        '''
+        super(AccountInvoice, self).action_move_create()
```
rousseldenis: @ThomasBinsfeld Doesn't super result must be returned at the end to ensure inheritance (as result is used in workflow)?

ThomasBinsfeld: Right, thanks :wink:

rousseldenis:
:+1: (code review)

account_invoice_blocking/views/account_invoice.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2016 Acsone SA/NV
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+
+<odoo>
+
+    <record model="ir.ui.view" id="account_invoice_form_view">
```
rousseldenis, line 7: @ThomasBinsfeld It seems that the field is only visible in supplier invoice form. Should appears in customer form too.

ThomasBinsfeld: @rousseldenis I think that it only makes sense to block supplier invoices so it is right that the field is only visible on supplier invoices.

rousseldenis: @ThomasBinsfeld I looked into code and saw that blocked field is used in reports like report_overdue that take into account customers too.

So, I think it should be generic.

sbidoul: @ThomasBinsfeld @rousseldenis blocking invoice can also mean "do not send overdue reminders". I suggest to write this in the roadmap section of the README as a future possible improvement.

rousseldenis: @sbidoul That's why I wanted to add both fields in customer invoice and just change label for them

rousseldenis (approved):
:+1:

account_invoice_blocking/__openerp__.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 Acsone SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Account Invoice Blocking',
+    'summary': """
+        This module allows the user to set a blocking (No Follow-up)
```
pedrobaeza: Put a shorter summary for not messing up the module view.

account_invoice_blocking/models/account_invoice.py
```diff
@@ -0,0 +1,83 @@
...
+from openerp import api, fields, models
+
+
+class AccountInvoice(models.Model):
+
+    _inherit = 'account.invoice'
+
+    @api.model
+    def _get_move_line(self, invoice):
+        '''
```
pedrobaeza: Change these quotes with double ones according PEP257 (several of this)

account_invoice_blocking/models/account_invoice.py
```diff
@@ -0,0 +1,83 @@
...
+                all(line.blocked for line in move_lines) or False
+
+    blocked = fields.Boolean(
+        string='No Follow-up',
+        states={'draft': [('readonly', True)]},
+        compute='_get_move_blocked',
+        inverse='_set_move_blocked')
+
+    draft_blocked = fields.Boolean(
+        string='No Follow-up')
```
pedrobaeza: Put a help attribute in this field explaining what this means (the same you have in the README can serve)

pedrobaeza (changes requested):

pedrobaeza (approved):

PR #171, opened 2016-07-22
9.0-add-account_invoice_prices_update

susport:
This module adds a wizard to update prices of the products of an invoice
according to the price list is chosen in the wizard

### DIFF
--- a/account_invoice_prices_update/README.rst
+++ b/account_invoice_prices_update/README.rst
@@ -1,0 +1,56 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+=============================
+Account Invoice Prices Update
+=============================
+
+This module adds a wizard to update prices of the products of an invoice according to the price list is chosen in the wizard.
+
+
+Installation
+============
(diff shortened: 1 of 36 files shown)


account_invoice_prices_update/README.rst
```diff
@@ -0,0 +1,56 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+=============================
+Account Invoice Prices Update
+=============================
+
+This module adds a wizard to update prices of the products of an invoice according to the price list is chosen in the wizard.
```
jjscarafia: remove "is"
..according to the pricelist chosen in the...

jjscarafia:
:+1: useful module when you create invoices not from PO or SO, for eg. from debit or credit notes and to update prices

pedrobaeza:
A wizard is not the best option for updating the prices, but a field like in the other objects (sale order and purchase order).  The name also doesn't reflect the real source of the prices (the pricelist). You should migrate the module enhancement that is proposed for v8 here: https://github.com/OCA/account-invoicing/pull/105, or at least:
- Take its name
- Use a field pricelist_id
- Develop some tests

PR #176, opened 2016-08-26
[9.0] migration account_invoice_pricelist + new feature

gaelTorrecillas:
migration account_invoice_pricelist to v9 and adding the feature on invoices


gaelTorrecillas:
bad branch management
to be redone

PR #179, opened 2016-09-13
[9.0][ADD] account_invoice_blocking_payment

ThomasBinsfeld:
Glue module to disable the payment of a blocked invoice. 

Depends on the modules: 
- [x] account_invoice_blocking https://github.com/OCA/account-invoicing/pull/170
- [x] [account_payment_order](https://github.com/OCA/bank-payment/tree/9.0/account_payment_order)


account_invoice_blocking_payment/models/account_payment_order.py
```diff
@@ -0,0 +1,20 @@
...
+    _inherit = 'account.payment.order'
+
+    @api.multi
+    def draft2open(self):
+        for rec in self:
+            for line in rec.payment_line_ids:
+                if line.move_line_id.blocked:
+                    raise UserError(_(
+                        "At least one journal item is blocked."))
+            super(AccountPaymentOrder, self).draft2open()
```
sbidoul: @ThomasBinsfeld I would not do this. `account_payment_order` already has filter to hide blocked invoices. But in the end, if the user wants to pay it anyway, he can do it.

This module must only hide the pay button on the invoice, nothing else I think.

account_invoice_blocking_payment/views/account_payment_line.xml
```diff
@@ -0,0 +1,18 @@
...
+        <field name="model">account.payment.line</field>
+        <field name="inherit_id" ref="account_payment_order.account_payment_line_form"/>
+        <field name="arch" type="xml">
+            <field name="move_line_id" position="attributes">
+                <attribute name="domain">[('blocked', '=', False)]</attribute>
+            </field>
+        </field>
+    </record>
+
+</odoo>
```
sbidoul: Do not change this domain (same reason: manually populating a payment order remains can be seen as an advanced use case where users should be able to pay a blocked invoice if they decide, for instance to do a partial payment but leave it blocked)

sbidoul:
@ThomasBinsfeld can you add bank-payment to oca-dependencies.txt?

I don't understand how travis is green because there are errors (account_payment_order not found).

sbidoul:
This one is red due to #170 awaiting merge

account_invoice_blocking_payment/models/account_invoice.py
```diff
@@ -0,0 +1,26 @@
...
+    @api.multi
+    def create_account_payment_line(self):
+        """
+        A blocked invoice cannot be paid
+        """
+        for rec in self:
+            if rec.draft_blocked or rec.blocked:
+                raise UserError(_(
+                    "A blocked invoice cannot be paid."))
+            super(AccountInvoice, self).create_account_payment_line()
```
rousseldenis: @ThomasBinsfeld Super should not be called there but outside of the loop and must be returned (inheritance break - e.g.: account_payment_order returns action)

account_invoice_blocking_payment/__openerp__.py
```diff
@@ -0,0 +1,21 @@
...
+    'author': 'ACSONE SA/NV,Odoo Community Association (OCA)',
+    'website': 'www.acsone.eu',
+    'depends': [
+        'account_payment_order',
+        'account_invoice_blocking',
+    ],
+    'data': [
+        'views/account_invoice.xml',
+    ],
+    'auto_install': True,
```
rousseldenis, line 20: @ThomasBinsfeld Is that correct?

rousseldenis (changes requested):

rousseldenis (approved):
:+1:

lmignon:
@ThomasBinsfeld can your rebase? #170 is merged

ThomasBinsfeld:
Rebased on 9.0

PR #180, opened 2016-09-13
[ADD][purchase_batch_invoicing] Batch invoicing for purchase orders!

yajo:
This module extends the functionality of purchases to support batch invoicing
purchase orders and to allow you to choose if you want them grouped by purchase
order or by vendor.

@Tecnativa.


purchase_batch_invoicing/README.rst
```diff
@@ -0,0 +1,77 @@
...
+order or by vendor.
+
+Configuration
+=============
+
+An automated task is included to invoice all pending purchase orders every
+week, but it is disabled by default. To enable it:
+
+#. Have *Administration / Settings* permissions.
+#. Go to *Settings > Technical > Automation > Scheduled Actions > Invoice all
```
pedrobaeza: Include `#. Enable Developer mode`

purchase_batch_invoicing/README.rst
```diff
@@ -0,0 +1,77 @@
...
+help us smashing it by providing a detailed and welcomed feedback.
+
+Credits
+=======
+
+Contributors
+------------
+
+* Jairo Llopis <jairo.llopis@tecnativa.com>
+* Pedro Baeza <pedro.baeza@tecnativa.com>
```
pedrobaeza: `Pedro M. Baeza`, please

purchase_batch_invoicing/__openerp__.py
```diff
@@ -0,0 +1,21 @@
...
+# Copyright 2016 Jairo Llopis <jairo.llopis@tecnativa.com>
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+{
+    "name": "Purchase Batch Invoicing",
+    "summary": "Make invoices for all ready purchase orders",
+    "version": "9.0.1.0.0",
+    "category": "Purchases",
+    "website": "https://tecnativa.com/",
+    "author": "Tecnativa, Odoo Community Association (OCA)",
+    "license": "LGPL-3",
```
pedrobaeza: This one should be AGPL, as there's no possibility of using it on website

purchase_batch_invoicing/tests/test_invoicing.py
```diff
@@ -0,0 +1,112 @@
...
+            # Confirm purchase order
+            po.button_confirm()
+            # Receive products
+            cls.env["stock.immediate.transfer"] \
+                .with_context(active_id=po.picking_ids.id) \
+                .create(dict()).process()
+        cls.wizard = cls.env["purchase.batch_invoicing"].with_context(
+            active_ids=cls.pos.ids).create(dict())
+
+    def tearDown(self):
```
pedrobaeza: Why do this in tearDown?

purchase_batch_invoicing/wizards/purchase_batch_invoicing.py
```diff
@@ -0,0 +1,109 @@
...
+            "res_model": "account.invoice",
+            "name": _("Generated Invoices"),
+            "views": [[False, "tree"], [False, "form"]],
+            "domain": [["id", "in", invoices.ids]],
+        }
+
+    @api.model
+    def invoice_all(self, grouping="partner_id"):
+        """Invoice all pending purchase orders."""
+        wizard = self.create({
```
pedrobaeza: Call this method `cron_invoice_all_pending` for signalling the source of the call and the purpose.

purchase_batch_invoicing/wizards/purchase_batch_invoicing.py
```diff
@@ -0,0 +1,109 @@
...
+            "purchase_order_ids": self.env["purchase.order"].search(
+                self._purchase_order_domain()).ids,
+            "grouping": grouping,
+        })
+        return wizard.action_batch_invoice()
```
pedrobaeza: Control before calling this if there is no orders to invoice for avoiding the error message in the log.

purchase_batch_invoicing/README.rst
```diff
@@ -0,0 +1,77 @@
...
+Configuration
+=============
+
+An automated task is included to invoice all pending purchase orders every
+week, but it is disabled by default. To enable it:
+
+#. Have *Administration / Settings* permissions.
+#. Go to *Settings > Technical > Automation > Scheduled Actions > Invoice all
+   pending purchase orders > Edit*.
+#. Enable *Active*.
```
elicoidal: You might want to add:
#. Enable it by clicking on Active and set the date accordingly

purchase_batch_invoicing/README.rst
```diff
@@ -0,0 +1,77 @@
...
+#. Go to *Purchases > Purchase > Purchase Orders > Create* and fill the form.
+#. Press *Confirm*.
+#. Press *Receive Products*.
+#. Press *Validate > Apply*.
+#. Repeat above steps a couple of times.
+#. Go back to *Purchase Orders* and select those you just created.
+#. Press *Action > Batch Invoice*.
+#. Choose the *Grouping* method.
+#. Press *Accept*.
+#. You will get to a screen where you can see all the vendor bills you just
```
elicoidal: great step by step how-to.
Nice-to-have would be to add one or two screenshots

purchase_batch_invoicing/__openerp__.py
elicoidal: Dont forget the headers

elicoidal: and the badge

purchase_batch_invoicing/__openerp__.py
```diff
@@ -0,0 +1,21 @@
...
+    "application": False,
+    "installable": True,
```
elicoidal: needed?

purchase_batch_invoicing/wizards/purchase_batch_invoicing.py
```diff
@@ -0,0 +1,109 @@
...
+        comodel_name="purchase.order",
+        string="Purchase orders",
+        domain="[('invoice_status', '=', 'to invoice')]",
+        required=True,
+        readonly=True,
+        default=lambda self: self._default_purchase_order_ids(),
+    )
+    grouping = fields.Selection(
+        selection=[
+            ("id", "Purchase Order"),
```
elicoidal, line 25: purchase_id (be consistent with the next one)

purchase_batch_invoicing/wizards/purchase_batch_invoicing.py
```diff
@@ -0,0 +1,109 @@
...
+            ("partner_id", "Vendor"),
```
elicoidal, line 26: A possible extension (for the roadmap) would to have the possibility to have one invoice / PO line.

purchase_batch_invoicing/wizards/purchase_batch_invoicing.py
```diff
@@ -0,0 +1,109 @@
...
+                })
+                invoice._onchange_partner_id()
+                for po in pogroup:
+                    invoice.currency_id = po.currency_id
+                    invoice.purchase_id = po
+                    invoice.purchase_order_change()
+                vals = invoice._convert_to_write(invoice._cache)
+            invoices |= invoices.create(vals)
+        if not invoices:
+            raise UserError(_("No ready to invoice purchase orders selected."))
```
elicoidal: "No ready-to-invoice purchase order selected."

purchase_batch_invoicing/wizards/purchase_batch_invoicing_view.xml
```diff
@@ -0,0 +1,43 @@
...
+    <field name="model">purchase.batch_invoicing</field>
+    <field name="arch" type="xml">
+        <form>
+            <sheet>
+                <group attrs="{'invisible': [('purchase_order_ids', '=', False)]}">
+                    <field name="purchase_order_ids"/>
+                    <field name="grouping"/>
+                </group>
+                <div attrs="{'invisible': [('purchase_order_ids', '!=', False)]}">
+                    <p>No ready to invoice purchase orders selected.</p>
```
elicoidal: idem
I am OK with the method but wouldnot it be better to check the condition before and raise a simple warning?

elicoidal: Just not sure of the advantages of one method compared with the other

purchase_batch_invoicing/tests/test_invoicing.py
yajo: Because it has to be run after each test.

purchase_batch_invoicing/wizards/purchase_batch_invoicing.py
yajo: Actually this technical name (that never is shown to the user) is the field name by which we are grouping, [to simplify line 60.](https://github.com/OCA/account-invoicing/pull/180/files/65107aa8f64bd113dda8dc210039c73df1801823#diff-a190f617ba3cd5dd2989cd3590604437R60)

purchase_batch_invoicing/tests/test_invoicing.py
pedrobaeza: Are asserts correctly captured in this method?

purchase_batch_invoicing/wizards/purchase_batch_invoicing.py
yajo: I think this log would be positive. It would tell you that the cron ran, and why it did not succeed. The underlying possible UserError can be due to you not having an account chart too, or any other process involved. After all, logs are there for knowing what happens and why something fails.

purchase_batch_invoicing/wizards/purchase_batch_invoicing_view.xml
yajo: I think it would be quite similar, UX thinking.

purchase_batch_invoicing/tests/test_invoicing.py
yajo: Good point.

purchase_batch_invoicing/wizards/purchase_batch_invoicing.py
yajo: Done

purchase_batch_invoicing/wizards/purchase_batch_invoicing.py
pedrobaeza: It shouldn't be, as it's treated as an error on the log. All monitoring systems will pass this as a problem, alerting without need to sysadmins.

purchase_batch_invoicing/wizards/purchase_batch_invoicing.py
pedrobaeza: This is something that Odoo doesn't allow natively, so it's harder to implement

yajo: And feels somehow useless to me, but a known issues line does not hurt :wink:

purchase_batch_invoicing/wizards/purchase_batch_invoicing.py
```diff
@@ -101,9 +105,15 @@ def action_batch_invoice(self):
...
         wizard = self.create({
             "purchase_order_ids": self.env["purchase.order"].search(
                 self._purchase_order_domain()).ids,
             "grouping": grouping,
         })
-        return wizard.action_batch_invoice()
+        try:
+            wizard.action_batch_invoice()
+        except UserError as error:
+            _logger.warn(error.name)
```
pedrobaeza: I prefer that this is put as info only. Warnings are considered by runbot also, and GitHub CI will show error on this warning.

purchase_batch_invoicing/wizards/purchase_batch_invoicing.py
elicoidal: I know but it helps maintainance in the future and code readibility.
Non-blocking for me

purchase_batch_invoicing/wizards/purchase_batch_invoicing.py
yajo: This cron is disabled by default anyway.

purchase_batch_invoicing/tests/test_invoicing.py
```diff
@@ -0,0 +1,114 @@
...
+        try:
```
pedrobaeza: But with try, you are masking possible incorrect tests, isn't it?

yajo: Not, there's no `except`, so anything will be reraised.

purchase_batch_invoicing/wizards/purchase_batch_invoicing.py
pedrobaeza: Yeah, but it's best to put it as info, and also move the other info to the try block. You get this way one of this option:

```
INFO: Starting to invoice all pending purchase orders.
INFO: Finished invoicing all pending purchase orders.
```

or

```
INFO: Starting to invoice all pending purchase orders.
INFO: No ready-to-invoice purchase orders selected.
```

I think this way is the best.

purchase_batch_invoicing/tests/test_invoicing.py
pedrobaeza: OK, got it

pedrobaeza:
Please check coverage

pedrobaeza:
@Yajo, two things more to do:
- Add a new field in purchase order that is "Invoicing status", working the same as in sales order. This is computed on the wizard to see what purchase orders to invoice, so it's possible to get it.
- Add a button in the purchase order form to create an invoice when it's possible to do it, for a more direct approach.

purchase_batch_invoicing/wizards/purchase_batch_invoicing.py
```diff
@@ -0,0 +1,118 @@
...
+_logger = getLogger(__name__)
+
+
+class PurchaseBatchInvoicing(models.TransientModel):
+    _name = "purchase.batch_invoicing"
+
+    purchase_order_ids = fields.Many2many(
+        comodel_name="purchase.order",
+        string="Purchase orders",
+        domain="[('invoice_status', '=', 'to invoice')]",
```
yajo, line 18: @pedrobaeza, [the field you ask already exists](https://github.com/odoo/odoo/blob/9.0/addons/purchase/purchase.py#L130), but you don't need it because this domain is enforced. All POs you see are ready to invoice.

The button suggestion has landed already :wink:

pedrobaeza: Please put it then on the tree and form views (it's a customer requirement).

pedrobaeza:
Can we avoid the coverage drop?

yajo:
Yes, if we disable Coveralls, that is lying :laughing:

rafaelbn (approved):
tested in runbot

pedrobaeza:
I have tried on runbot and there's a problem. I have tried to invoice PO00010 which is in the proper state and is added on the wizard, but when pressing the button to invoice, it says:

seleccion_006

rafaelbn:
Please @Yajo check https://github.com/OCA/account-invoicing/pull/180#issuecomment-260096119 thanks!

purchase_batch_invoicing/README.rst
```diff
@@ -0,0 +1,89 @@
...
+Usage
+=====
+
+To use this module, you need to:
+
+#. Have *Purchase / User* permissions.
+#. Go to *Purchases > Purchase > Purchase Orders > Create* and fill the form.
+#. Press *Confirm*.
+#. Press *Receive Products*.
+#. Press *Validate > Apply*.
```
yajo, line 35: @pedrobaeza @rafaelbn, the problem described in https://github.com/OCA/account-invoicing/pull/180#issuecomment-260096119 comes from not following these last 2 steps.

So indeed there is an inconsistency problem: the user sees the POs in the GUI but the execution of the wizard discards them because they have not been received and validated.

This happens because [we are checking that there are less items invoiced than delivered](https://github.com/OCA/account-invoicing/pull/180/files#diff-a190f617ba3cd5dd2989cd3590604437R67).

So, we have something to fix here, but which do you consider is the right fix?

1. Remove POs from GUI when they are not available to invoice according to current criteria.
2. Forget about checking the delivered vs invoiced qty, and simply allow to invoice any PO that has "Waiting invoices" status.

Thanks.

pedrobaeza: We need consistency, so better to opt for option 1.

yajo:
https://github.com/OCA/account-invoicing/pull/180#issuecomment-260096119 should be fixed now.

pedrobaeza:
@Yajo, can you check Travis status?

purchase_batch_invoicing/wizards/purchase_batch_invoicing.py
pedrobaeza: Better to not change it for not adding extra code.

pedrobaeza:
@Yajo, why is this failing?

luismontalba (approved):
Functionality tested on runbot

pedrobaeza (approved):
Finally green!

PR #181, opened 2016-09-19
[FIX][7.0] Sale Order Partial Invoice

AaronHForgeFlow:
Make qty to invoice no depending on qty delivered

When creating invoices from not the whole quantity in a sale order line we just need to consider the quantity that have not been invoice yet. Currently we are considering the quantity delivered. So If this quantity delivered is lower than the quantity invoiced, we obtained that the quantity invoiced is negative. IMHO this is wrong and we just need to consider the quantity in the SO line and the already invoiced quantity.

### DIFF
--- a/sale_order_partial_invoice/__openerp__.py
+++ b/sale_order_partial_invoice/__openerp__.py
@@ -21,7 +21,7 @@
 
 {
     'name': 'Sale Partial Invoice',
-    'version': '1.1',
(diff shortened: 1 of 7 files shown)


AaronHForgeFlow:
@gurneyalex can you please give some feedback?

gurneyalex:
@aheficent could you update the test to make your change's scope explicit?

PR #182, opened 2016-09-22
[ADD] Add account_invoice_merge_attachment module

adrienpeiffer:
This module allows to link attachment on merged invoices


pedrobaeza:
I think this should be part of the base module, as ir.attachment is a base model.

adrienpeiffer:
Hi Pedro,
Yes, but the attachement menu on all object comes with document module.
So, I think it's a bit confusing to see "Link attachment" on the merge wizard if there is no attachment menu on invoices and I prefer to keep modularity. it also keeps the standard behavior.

benwillig (approved):

PR #184, opened 2016-09-28
[ADD] [account_invoice_line_sequence] v9

SerpentCS:


account_invoice_line_sequence/views/report_invoice.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+    <data>
```
gmeficent: `<openerp>` can be changed to `<odoo>`
`<data>` tag is dispensable when `noupdate="0"`

account_invoice_line_sequence/views/report_invoice.xml
```diff
@@ -0,0 +1,15 @@
...
+                <th><strong>Sequence</strong></th>
+            </xpath>
+            <xpath expr="//table[@class='table table-condensed']/tbody/tr/td[1]" position="before">
+                 <td>
+                    <span t-field="l.sequence2"/>
+                 </td>
+            </xpath>
+        </template>
+    </data>
+</openerp>
```
gmeficent: Same as above

gmeficent (changes requested):
LGTM :+1: Functional and code review.
Works as described. The sequence on preexisting records are all set to 10, though. I would suggest a post_init_hook

aheficent:
Would you please add a readme file for this one?

AaronHForgeFlow (commented):
@darshan-serpent Would you please add a readme file for this one?

account_invoice_line_sequence/__init__.py
```diff
@@ -0,0 +1,11 @@
+# -*- coding: utf-8 -*-
+#    Author: Alexandre Fayolle
+#    Copyright 2013 Camptocamp SA
+#    Author: Damien Crier
+#    Copyright 2015 Camptocamp SA
+# © 2015 Eficent Business and IT Consulting Services S.L. -
+# Jordi Ballester Alomar
```
LoisRForgeFlow: Can you try to use the same format for all the copyright lines?

account_invoice_line_sequence/__openerp__.py
```diff
@@ -0,0 +1,26 @@
...
+#    Copyright 2015 Camptocamp SA
+# © 2015 Eficent Business and IT Consulting Services S.L. -
+# Jordi Ballester Alomar
+# © 2015 Serpent Consulting Services Pvt. Ltd. - Sudhir Arya
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
+
+{
+    'name': 'Account Invoice line with sequence number',
+    'version': '9.0.1.0.0',
+    'category': 'Accounting',
```
LoisRForgeFlow: Could you add a `summary` here?

account_invoice_line_sequence/models/invoice.py
```diff
@@ -0,0 +1,42 @@
...
+
+    @api.one
+    @api.depends('invoice_line_ids')
+    def compute_max_line_sequence(self):
+        """Allow to know the highest sequence
+        entered in invoice lines.
+        Web add 1 to this value for the next sequence
+        This value is given to the context of the o2m field
+        in the view. So when we create new invoice lines,
+        the sequence is automatically max_sequence + 1
```
LoisRForgeFlow: Can you fix all this text to better exploit the 79 character lines length?

account_invoice_line_sequence/views/report_invoice.xml
LoisRForgeFlow: agree

account_invoice_line_sequence/views/report_invoice.xml
LoisRForgeFlow: agree and add end blank line, please.

LoisRForgeFlow (changes requested):
Style review

gmeficent (approved):

LoisRForgeFlow (approved):
Tested on runbot :+1:

AaronHForgeFlow (commented):
I tested on runbot and it works well. Only one think:
When the invoice is validated it can be confused to have an empty column at the beginning. Is there a way to hide that?

image

JordiBForgeFlow:
@aheficent I think that it cannot be removed.

AaronHForgeFlow (approved):
@jbeficent This is ok to me then :+1:

account_invoice_line_sequence/models/invoice.py
```diff
@@ -0,0 +1,52 @@
...
+# Copyright 2016 Serpent Consulting Services Pvt. Ltd.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from openerp import api, fields, models
+
+
+class AccountInvoiceLine(models.Model):
+    _inherit = 'account.invoice.line'
+
+    sequence = fields.Integer(help="Gives the sequence of this line when "
```
moylop260: Could you add a comment in the code about this re-definition?
(Because we have this same field from [9.0 - account_invoice.py](https://github.com/odoo/odoo/blob/26627391870af2edd8315748929122392a9a5f68/addons/account/models/account_invoice.py#L1129-L1130) and the `store=True` is a default parameter for non-computed values)

account_invoice_line_sequence/models/invoice.py
```diff
@@ -0,0 +1,52 @@
...
+                              "displaying the account invoice.", store=True)
+
+    sequence2 = fields.Integer(help="Shows the sequence of this line in the "
```
moylop260: Could you add a comment in the code about this related field?

account_invoice_line_sequence/models/invoice.py
```diff
@@ -0,0 +1,52 @@
...
+                               " invoice.", related='sequence', readonly=True)
```
moylop260: `store=True`?

account_invoice_line_sequence/models/invoice.py
```diff
@@ -0,0 +1,52 @@
...
+    def _compute_max_line_sequence(self):
+        """Allow to know the highest sequence entered in invoice lines.
+        Then we add 1 to this value for the next sequence.
+        This value is given to the context of the o2m field in the view.
+        So when we create new invoice lines, the sequence is automatically
+        added as :  max_sequence + 1
+        """
+        for rec in self:
+            rec.max_line_sequence =\
+                (max(rec.mapped('invoice_line_ids.sequence') or [0]) + 1)
```
moylop260: Using a search_read with order and limit will be faster:
 - `self.env['account.invoice.line'].search_read([('id', 'in', rec.invoice_line_ids.ids)], ['sequence'], order='sequence DESC', limit=1)`

account_invoice_line_sequence/models/invoice.py
```diff
@@ -0,0 +1,52 @@
...
+
+    max_line_sequence = fields.Integer(string='Max sequence in lines',
+                                       compute='_compute_max_line_sequence')
```
moylop260: What about store=True in order to avoid compute this method by each read?

account_invoice_line_sequence/models/invoice.py
```diff
@@ -0,0 +1,52 @@
...
+
+    @api.multi
+    def _reset_sequence(self):
+        for rec in self:
+            current_sequence = 1
+            for line in rec.invoice_line_ids:
+                line.write({'sequence': current_sequence})
```
moylop260: What about `line.sequence = current_sequence`?

account_invoice_line_sequence/models/invoice.py
```diff
@@ -0,0 +1,52 @@
...
+                current_sequence += 1
+
+    @api.multi
+    def write(self, values):
+        res = super(AccountInvoice, self).write(values)
+        for rec in self:
```
moylop260: `self._reset_sequence()` because supports `multi`

NOTE: Really we want to reset the sequence in the lines by each change in the invoice?

This could be a issue of performance because:

I add a line 1 <- Get all lines to know global current sequence
I add a line 2 <- Get all lines to know the new global current sequence
I add a line 3 <- Get all lines to know the new global current sequence
I save the invoice <- Recompute all lines again...

Maybe you could use `model._cache['sequence']` in order to improvement performance.

account_invoice_line_sequence/models/invoice.py
serpentcs-dev1: Didn't get you, do you mean to use search_read instead of the mapped function?

account_invoice_line_sequence/models/invoice.py
serpentcs-dev1: does it make any difference?

account_invoice_line_sequence/models/invoice.py
serpentcs-dev1: We are resetting the sequence to maintain the order of the invoice lines.

account_invoice_line_sequence/models/invoice.py
JordiBForgeFlow: @darshan-serpent No, he means adding to the field store=True. That would improve the performance. But you would then want to extend the depends to be:
```
@api.depends('invoice_line_ids','invoice_line_ids.sequence')
def _compute_max_line_sequence(self):
```

account_invoice_line_sequence/models/invoice.py
JordiBForgeFlow: @darshan-serpent I think is more elegant `line.sequence = current_sequence`

account_invoice_line_sequence/models/invoice.py
moylop260: What about this comment?

serpentcs-dev1: I tried that solution but then the sequence is set only when the record is saved. Until then it doesn't compute the latest sequence value.

I am trying to upload the screenshot here, but something is wrong with GitHub I cannot upload it for some reason.

moylop260 (approved):

PR #191, opened 2016-10-10
10.0 Port invoice_fiscal_position_update

alexis-via:


pedrobaeza:
Please rename to account_invoice_fiscal...

pedrobaeza (approved):
Remove POT file

account_invoice_fiscal_position_update/models/account_invoice.py
```diff
@@ -27,13 +27,18 @@ def fiscal_position_change(self):
...
+                    # so we need per-company post-filtering
+                    taxes = product.taxes_id.filtered(
+                        lambda tax: tax.company_id == self.company_id)
                 else:
                     account = (
                         product.property_account_expense_id or
                         product.categ_id.property_account_expense_categ_id)
-                    taxes = product.supplier_taxes_id
-                taxes = taxes or account.tax_ids
+                    taxes = product.supplier_taxes_id.filtered(
```
pedrobaeza: Do the filtered only one time for better code readibility

pedrobaeza:
It would be good to start all 10.0 modules with tests.

account_invoice_fiscal_position_update/models/account_invoice.py
alexis-via: If I do the filtering only one time here, the "or" won't work the same way.

alexis-via:
@pedrobaeza Feel free to propose a PR to add tests. I won't have time to do it myself (and, as I explained in my email on the ML, it's not the goal the of OCA v10 Taskforce).

account_invoice_fiscal_position_update/models/account_invoice.py
pedrobaeza: Yes, you can do it at the end, after the or and thus, you don't need to make 3 `filtered`

alexis-via: But if you think about it carefully, you'll see that doing it at the end will modify the behavior: if you have a tax in another company on a product, then

```
taxes = taxes or account.tax_ids
```

will copy taxes on taxes and then it will be filtered and the taxes will be removed. With the current code, it would have taken account.tax_ids.
I'm just saying that to underline the fact that moving the 3 filtered at the end won't give the same result.

pedrobaeza: Yes, but you can reduce to 2 filtered instead of 3:

```
if inv_type in ('out_invoice', 'out_refund'):
    account = (
       product.property_account_income_id or
       product.categ_id.property_account_income_categ_id)
   taxes = product.taxes_id
else:
    account = (
       product.property_account_expense_id or
       product.categ_id.property_account_expense_categ_id)
   taxes = product.supplier_taxes_id
taxes = taxes.filtered(lambda tax: tax.company_id == self.company_id)
taxes = taxes or account.tax_ids.filtered(
    lambda tax: tax.company_id == self.company_id)
```

mourad-ehm:
:+1:  code review & module tested on runbot.

mourad-ehm:
Hi @pedrobaeza, @alexis-via, I added tests.

pedrobaeza:
Thanks for the tests, @mourad-ehm. There's a mix in the branch with other module included in it. Since the moment I said about the tests, they were developed by my colleague on version 9. You can see them here: https://github.com/OCA/account-invoicing/blob/9.0/account_invoice_fiscal_position_update/tests/test_account_invoice.py. It's a pity to duplicate efforts. Sorry for not saying anything here, but I didn't remember my request. You can check if there's something missing in your tests that can be taken from there.

alexis-via:
The tests in the module account_invoice_fiscal_position_update were added in v9 after the creation of this PR for v10.

pedrobaeza:
@alexis-via That's what I'm saying and again sorry for not communicating here for not duplicating efforts.

mourad-ehm:
ping @alexis-via

PR #196, opened 2016-10-14
MIG 9.0 invoice_fiscal_position_update

angelmoya:


invoice_fiscal_position_update/models/account_invoice.py
```diff
@@ -1,60 +1,42 @@
...
 from openerp import models, api, _
 
 
 class account_invoice(models.Model):
     _inherit = "account.invoice"
 
-    @api.onchange('fiscal_position')
-    def fiscal_position_change(self):
+    @api.onchange('fiscal_position_id')
+    def fiscal_position_id_change(self):
```
pedrobaeza: Use standard nomenclature `onchange_fiscal_position_id`

pedrobaeza (changes requested):
You should include tests

angelmoya:
fix to use standard nomenclature and test added

invoice_fiscal_position_update/models/account_invoice.py
```diff
@@ -1,60 +1,42 @@
...
+        for line in self.invoice_line_ids:
             if line.product_id:
                 product = line.product_id
                 if inv_type in ('out_invoice', 'out_refund'):
                     account = (
-                        product.property_account_income or
-                        product.categ_id.property_account_income_categ)
+                        product.property_account_income_id or
+                        product.categ_id.property_account_income_categ_id)
                     taxes = product.taxes_id
```
pedrobaeza: This is not filtered by company

invoice_fiscal_position_update/models/account_invoice.py
```diff
@@ -1,60 +1,42 @@
...
                 else:
                     account = (
-                        product.property_account_expense or
-                        product.categ_id.property_account_expense_categ)
+                        product.property_account_expense_id or
+                        product.categ_id.property_account_expense_categ_id)
                     taxes = product.supplier_taxes_id
```
pedrobaeza: This is not filtered by company

invoice_fiscal_position_update/models/account_invoice.py
```diff
@@ -1,60 +1,42 @@
...
                 taxes = taxes or account.tax_ids
                 if fp:
                     account = fp.map_account(account)
                     taxes = fp.map_tax(taxes)
```
pedrobaeza: This is not filtered by company

angelmoya: OK, fixed

angelmoya:
Code coverage decreased, but all code of this module is tested, I think it has been decreased from actual commit of branch 9.0, but it's not relevant

pedrobaeza:
Not really. You are not testing everything. It lacks the `else` parts. Look for example https://codecov.io/gh/OCA/account-invoicing/compare/916a44af4fa3c265c76ae54da282153cff44edf8...5d4952f6726fb3acd544776a4daadb066aa43a55#696E766F6963655F66697363616C5F706F736974696F6E5F7570646174652F6D6F64656C732F6163636F756E745F696E766F6963652E7079-47

You have 2 options:
- If it's a weird else condition, you can put `else:  # pragma: no cover` to avoid the coverage check.
- Create a special test for that condition if not.

invoice_fiscal_position_update/README.rst
```diff
@@ -0,0 +1,58 @@
...
+
+No specific configuration needed. This module uses the standard
+configuration of the fiscal positions.
+
+Usage
+=====
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/95/10.0
```
pedrobaeza, line 25: This should be 9.0

sergio-teruel:
cc @Tecnativa

pedrobaeza:
Superseeded by #223

PR #198, opened 2016-10-17
[DEL] Remove odoo_addons directory

adrienpeiffer:


lmignon (approved):
This one should be fast tracked since the changes are just a cleanup of the rep created by setuptools-odoo for odoo9.

pedrobaeza (approved):

PR #199, opened 2016-10-20
[10.0][WIP] add account_invoice_sequential_dates

dcorio:
This is a module developed for the italian localization, but we think it might be useful to others.


dcorio:
cosmic mess...

PR #218, opened 2017-01-16
module account_partner_journal

jweste:
[NEW] Allow to define journal on a supplier. This journal will be used by default on invoices


hbrunn:
is this the same as https://github.com/OCA/purchase-workflow/pull/309 ?

jweste:
@hbrunn : Oh yes, it looks very similar. Sorry, I had not seen your PR :/
But I need a v9 module. Should I work on a v9 version of your code?

Issue #222, opened 2017-01-23
[8.0] Invoice merge validation v8


ljsalvatierra-binovo:
**Warning**
```bash
2017-01-23 10:44:31,959 11296 WARNING odoo openerp.models: account.invoice.read() with unknown field '17'
2017-01-23 10:44:31,959 11296 WARNING odoo openerp.models: account.invoice.read() with unknown field '16'
2017-01-23 10:44:31,960 11296 WARNING odoo openerp.models: account.invoice.read() with unknown field '15'
2017-01-23 10:44:31,960 11296 WARNING odoo openerp.models: account.invoice.read() with unknown field '14'
```

**account_invoice_merge/wizard/invoice_merge.py**
```python
@api.model
    def _dirty_check(self):
...
        invs = inv_obj.read(ids,
                                  ['account_id', 'state', 'type', 'company_id',
                                   'partner_id', 'currency_id', 'journal_id'])
...
```

**openerp/models.py**
```python
# new-style implementation of read()
    @api.v8
    def read(self, fields=None, load='_classic_read'):
```

pedrobaeza:
Solved already.

ljsalvatierra-binovo:
Thank you! Do you have an ETA on when will it be merge?

pedrobaeza:
Follow https://github.com/OCA/account-invoicing/pull/227 and review it for speeding up the process

PR #227, opened 2017-02-10
Fix bug #222 bad use of read in new API

alexis-via:


pedrobaeza:
It should work also with that notation, isn't it?

alexis-via:
Certainly... (by the way, in v9, we use a browse instead of a read).

pedrobaeza:
Travis complains...

alexis-via:
When we merge 2 invoices, why do we bother to re-generate their invoice lines ? Why not just re-attach the existing invoice lines to the new account.invoice ? It would simplify the code quite a lot...

alexis-via:
@pedrobaeza Travis is red on OCA/account-invoicing ; so it's not a regression of this PR.

alexis-via:
I say that for v8 of course (it's green on OCA/account-invoicing/10.0).

alexis-via:
OK, I understand this module also merges invoice lines, that's why they are re-generated... but I complexifies at lot the module :-(

pedrobaeza:
Please solve conflicts

PR #230, opened 2017-02-16
[9.0] New module [purchase_stock_picking_return_invoicing]

JordiBForgeFlow:
Refund Returned Pickings from Purchase Orders
=============================================

This module extends the functionality of purchase orders to better manage
supplier returns and refunds.

In the purchase order you are able to display, for each line:

* Quantity to Bill, and  Quantity to Refund as separate fields. You have the
  option to create a vendor bill or a refund. In the bill or refund the
  correct quantity will be proposed, based on those fields.

* Billed Quantity and Refunded Quantity, as separate fields.

* Received Quantity and Returned Quantity, as separate fields.

Usage
=====

Case 1: When you return to a supplier some products, and you have not yet
received the bill from the supplier

#. Go to *Purchases > Purchase > Purchase Orders > Create*.
#. Choose a supplier and add a product whose *Invoicing Policy* is *On Received
   quantities*, and input some quantity to purchase.
#. Confirm the purchase order.
#. Go to *Shipment > Validate > Apply* so as to receive the quantities ordered.
#. Press the button *Reverse*.
#. In the wizard *Reverse Quantity* Set *Quantity* to the quantity to be
   returned. Press *Return* to complete the wizard.
#. On the return picking press *Validate > Apply*.
#. Go back to the purchase order. You will notice that the field *Returned
   Qty* is now the quantity that was returned. The field *Quantity to
   Bill* will show the quantity received less the quantity returned.
#. Press the button *Invoices* to create the vendor bill.
#. The proposed vendor bill will be proposed for the difference between the
   received and the returned quantity.

Case 2: When you return to a supplier some products, and you have already
received a bill from the supplier.

#. Go to *Purchases > Purchase > Purchase Orders > Create*.
#. Choose a supplier and add a product whose *Invoicing Policy* is *On Received
   quantities*, and input some quantity to purchase.
#. Confirm the purchase order.
#. Go to *Shipment > Validate > Apply* so as to receive the quantities ordered.
#. Press the button *Invoices* to create the vendor bill.
#. The proposed vendor bill will be proposed for the quantity received. The
   *Invoice Status* is now 'Invoiced'
#. Go to the original incoming shipment
#. Press the button *Reverse*.
#. In the wizard *Reverse Quantity* Set *Quantity* to the quantity to be
   returned. Press *Return* to complete the wizard.
#. On the return picking press *Validate > Apply*.
#. Go back to the purchase order. It will have  *Invoice Status* as 'Waiting
   Invoces'. You will notice that the field *Returned Qty* is now the quantity
   that was returned. The field *Quantity to Refund* is now the showing the
   quantity returned that was previously billed.
#. Press the button *Refunds* to create the vendor refund bill (since the
   field *Quantity to Invoice* is negative).
#. The proposed vendor refund bill will be proposed for the quantity that is
   to be refunded.
#. If you back to the purchase order, you will notice that *Invoice Status*
   is now 'Invoiced', even when the quantity ordered does not match with the
   quantity invoiced, because you did return some products.

Remark: As part of the 3-way match proces, if you accept that you will not
claim for a refund for the quantity returned to the supplier, just set the
purchase status as 'Done' at the end of the process, and the quantity to
invoice for the items will be set to 0 (because you have accepted the
discrepancies).


JordiBForgeFlow:
cc @tecnativa @eficent

JordiBForgeFlow:
@lreficent @mreficent please review

JordiBForgeFlow:
For some strange reason Travis fails in OCB, but I cannot reproduce the error, neither in runbot nor in local.

JordiBForgeFlow:
Travis issues an error in the OCB. PR https://github.com/OCA/OCB/pull/590 fixes that.

JordiBForgeFlow:
@pedrobaeza Following up on the discussion we had here https://github.com/OCA/OCB/commit/f6ff9771097ad0be27e63f8f968a6fea4bcdf943#commitcomment-20920814

The problem of your proposal to dynamically select invoice or refund is in purchases the supplier invoice is created manually by the user, either from within the PO using just a link. As he adds the PO to the invoice, then the system proposes the invoice lines 'automagically'.

If you have a PO where in line 1 you still need to invoice quantity 1, and line 2 you need to refund quantity 1, the user should create 2 separate invoices. 1 in_invoice, and 1 in_refund.

If you think about it, it makes total sense. The refunds are normally a result of an issue with a product, and are issued once the vendor has already sent a prior invoice. It makes sense that a separate refund is created for those products that have been returned.

purchase_stock_picking_return_invoicing/models/account_invoice.py
```diff
@@ -0,0 +1,33 @@
...
+            data['account_id'] = invoice_line.with_context(
+                {'journal_id': self.journal_id.id,
+                 'type': 'in_invoice'})._default_account(),
+            account = invoice_line.get_invoice_line_account(
+                'in_invoice', line.product_id,
+                self.purchase_id.fiscal_position_id, self.env.user.company_id)
+            if account:
+                data['account_id'] = account.id
+        msg = 'PR 230. Debug. Data: %s' % data
+        _logger.info(msg)
```
LoisRForgeFlow: Is this just for local debug? it should be removed at some point, shouldn't it?

LoisRForgeFlow (commented):
Tested on runbot and works fine now! I don't now if the issue with travis will involve some changes though.

JordiBForgeFlow:
Can you review @mreficent?

MiquelRForgeFlow (approved):
tested on runbot :+1:

purchase_stock_picking_return_invoicing/README.rst
```diff
@@ -0,0 +1,108 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+=============================================
+Refund Returned Pickings from Purchase Orders
+=============================================
+
+This module extends the functionality of purchase orders to support marking some
+returned pickings as "To refund in Purchase Order",  excluding those quantities
```
pedrobaeza: "To Refund in PO"

pedrobaeza: s/returned pickings/returned picking lines

purchase_stock_picking_return_invoicing/README.rst
```diff
@@ -0,0 +1,108 @@
...
+received the bill from the supplier
+
+#. Go to *Purchases > Purchase > Purchase Orders > Create*.
+#. Choose a supplier and add a product whose *Invoicing Policy* is *On Received
+   quantities*, and input some quantity to purchase.
+#. Confirm the purchase order.
+#. Go to *Shipment > Validate > Apply* so as to receive the quantities ordered.
+#. Press the button *Reverse*.
+#. In the wizard *Reverse Quantity* Set *Quantity* to the quantity to be
+   returned, and check *To Refund*. Press *Return* to complete the wizard.
```
pedrobaeza: s/To Refund/To Refund in PO

purchase_stock_picking_return_invoicing/README.rst
```diff
@@ -0,0 +1,108 @@
...
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/95/9.0
+
+Known issues / Roadmap
+======================
+
+* This addon is a pseudobackport of a functionality that exists natively in
+  v10, plus a fix for https://github.com/odoo/odoo/issues/13974, so this addon
```
pedrobaeza: s/v10/v11

purchase_stock_picking_return_invoicing/models/purchase_order.py
```diff
@@ -0,0 +1,139 @@
...
+            if any(invoice_status[line] == 'to invoice' for line in
+                   invoice_status.keys()):
+                order.invoice_status = 'to invoice'
+            elif all(invoice_status[line] == 'invoiced' for line in
+                     invoice_status.keys()):
+                order.invoice_status = 'invoiced'
+
+    @api.multi
+    def action_view_invoice_refund(self):
+        '''
```
pedrobaeza: Triple double quota for docstring according PEP257

pedrobaeza (changes requested):
Very light code review.

After trying functionally the module, these are the things I see that needs to be improved:

* On "Reverse transfer" dialog, "To refund in PO" column should be hidden for non incoming pickings. The same applies for "To Refund" in the brother module "sale_stock_picking_return_invoicing", but that can be on another module.
* Once you have selected on the dialog if refund or not refund, there's no way of changing this. Can we added somewhere on "Initial demand" page?
* The invoice status doesn't change to "waiting invoices" when returning goods after making the initial invoice.
* There's no visual sign that you need to click on invoices nor refunds for making the operation, but that's also a problem in standard with only invoices.
* The standard "Invoices" smart-button shows both invoices (the normal and the refund one), which is weird, so there are 2 choices:
  1. Change this button to only show normal invoices.
  2. Change the behavior of the normal invoice to change on the fly the invoice type when only having returned quantities.
* What reinforces the choice of the second option is that if you create directly a supplier refund, you are not able to select any purchase order, which indicates that there is some black logic inside the module to get this to work when clicking "Refunds" smart-button.

pedrobaeza:
@sergio-teruel, please continue reviewing the module and solving the OCB issue.

purchase_stock_picking_return_invoicing/__openerp__.py
```diff
@@ -0,0 +1,23 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Eficent Business and IT Consulting Services
+#           <contact@eficent.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Refund Return Pickings in Purchase Orders",
```
sergio-teruel: s/Return/Returned

purchase_stock_picking_return_invoicing/__openerp__.py
```diff
@@ -0,0 +1,23 @@
...
+    "summary": "Add an option to refund return pickings",
```
sergio-teruel: s/return/returned

purchase_stock_picking_return_invoicing/tests/test_refund.py
```diff
@@ -0,0 +1,299 @@
...
+            "name": "Test account type",
+            "type": "payable",
+        })
+        cls.account = cls.env["account.account"].create({
+            "name": "Test account",
+            "code": "TEST_A",
+            "user_type_id": cls.account_type.id,
+            "reconcile": True,
+        })
+        cls.partner = cls.env.ref('base.res_partner_1')
```
sergio-teruel: If you create a new accounts why not create a new partner and a new product instead of use demo data?? in this way we do not need demo data, so this is better for run tests. :smile:

JordiBForgeFlow:
@pedrobaeza @sergio-teruel I removed the field 'to_refund_po' from the return wizard. It does not make sense in the purchasing context to use that indicator. In purchases it is up to the person that validates the supplier invoices to compare the quantities ordered, received, returned with the vendor bill to decide what is acceptable (under the accepted tolerance). This way the process is way more simple and intuitive.

sergio-teruel (changes requested):

purchase_stock_picking_return_invoicing/models/account_invoice.py
```diff
@@ -0,0 +1,33 @@
...
+from openerp import models
+_logger = logging.getLogger(__name__)
+
+
+class AccountInvoice(models.Model):
+    _inherit = "account.invoice"
+
+    def _prepare_invoice_line_from_po_line(self, line):
+        data = super(AccountInvoice,
+                     self)._prepare_invoice_line_from_po_line(line)
```
sergio-teruel: Why not you use
if data:
    if line.product_id.....

by this way you avoid the OCB Error

luismontalba (approved):
Tested on runbot

chienandalu (approved):
- tested functionality

JordiBForgeFlow:
Thanks for your reviews so far. We have introduced an important change. New fields are added to the PO line:
* Returned Qty
* Qty to Refund
* Refunded Qty

Now the process is much more clear.

Thanks to @pedrobaeza for the recommendation.

JordiBForgeFlow:
@chienandalu @luismontalba @sergio-teruel Can you re-test after the latest changes that we introduced?

purchase_stock_picking_return_invoicing/models/account_invoice.py
sergio-teruel: @jbeficent I remove this comment after last changes in origin OCB method.

chienandalu (approved):
The workflow is clearer now indeed. Tested functionality :+1:

luismontalba (approved):
I've tested it on runbot and it works

pedrobaeza:
Merging finally. Thanks all for the efforts made.

pedrobaeza (approved):

ccandito:
very good.
this covers a huge gap in odoo.
when will it be available in 10?
thanks in advance
claudio candito

pedrobaeza:
There are no plans for now for that. You can  also contribute with the migration.

ccandito:
unfortunately I'm not a developer, but I'll test it.
I'd suggest to implement another feature also:
a new field, block payment, to be evaluating manually
in case this is evaluated it is not possible to register the payment.
this field can be evaluated after posting the invoice also.
Regards
claudio candito

PR #233, opened 2017-03-06
[MIG][9.0] account_invoice_merge_purchase module

cubells:
- Updated README
- Updated tests

cc @Tecnativa

### DIFF
--- a/account_invoice_merge_purchase/README.rst
+++ b/account_invoice_merge_purchase/README.rst
@@ -29,6 +29,7 @@ Contributors
 ------------
 
 * Adrien Peiffer <adrien.peiffer@acsone.eu>
+* Luc De Meyer <luc.demeyer@noviat.com>
 
 Maintainer
 ----------
--- a/account_invoice_merge_purchase/README.rst
+++ b/account_invoice_merge_purchase/README.rst
@@ -1,26 +1,33 @@
 .. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
-    :alt: License: AGPL-3
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
 
+==================================
 Merge invoices from purchase order
 ==================================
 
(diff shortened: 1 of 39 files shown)


account_invoice_merge_purchase/README.rst
```diff
@@ -1,26 +1,33 @@
...
+This module provides a compatibility between purchase order workflow and
+account invoice merge module
 
 
 Usage
 =====
 
+Select several vendor bills from the same partner and click on **Merge Partner
+Invoice** in Action button menu.
+
```
rafaelbn, line 18: Hi, I don't see this option:

rafaelbn (changes requested):

chienandalu (commented):
Functionality is ok. However I'd suggest to pop a warning whenever the responsible differ across invoices in the same way it does with other not compatible merging parameters (different partner, paid o open bill...).

pedrobaeza:
@chienandalu That's out of the scope of this module

chienandalu:
@pedrobaeza The problem is that when the bill responsible differs, those bills are not going to merge but the user has no clue about the reason why the merging didn't success. So the user is going to have to review manually every not merged bill and find out.

pedrobaeza:
This module is only a glue module that adds purchase fields to the merge process. What you are mentioning is something for the main module, account_invoice_merge. That's why I say this is out of scope of this PR. Anyway, previous screen before merging informs about the criteria to merge invoices. You should fill an issue with the improvement you are mentioning (which is good, that's out of question), but for being done in the module account_invoice_merge.

chienandalu:
Absolutely right, @pedrobaeza :smile: I missed the point

chienandalu (approved):
functionality :+1:

account_invoice_merge_purchase/README.rst
pedrobaeza: Current runbot contains this.

pedrobaeza (approved):

PR #236, opened 2017-03-16
[FIX] Combining invoices keeps the sale.order.line reference intact

RoelAdriaans:
Because self.env['sale.order'] returns an empty record set, this is evaluated to False.
This in turn never calls the code to restore the link between sale order and invoice.
Because this code was never called in version 9 and above, it never encountered the issue
that the the field invoice_ids is now a computed field without a search.
Searching on that field returned all the sale.orders.
The linked sale_line_ids are now copied like the rest of the fields.

### DIFF
--- a/account_invoice_merge/models/account_invoice.py
+++ b/account_invoice_merge/models/account_invoice.py
@@ -41,7 +41,7 @@ class AccountInvoice(models.Model):
             'product_id', 'account_id', 'account_analytic_id',
             'uom_id'
         ]
-        for field in ['analytics_id']:
+        for field in ['analytics_id', 'sale_line_ids']:
             if field in self.env['account.invoice.line']._fields:
                 fields.append(field)
         return fields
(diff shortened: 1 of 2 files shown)


pedrobaeza (commented):
Why not only removing the line `if so_obj`? The rest of the code is there because you can have installed account, but not sale module.

RoelAdriaans:
Because that code doesn't work :)

The so_obj.search([('invoice_ids', 'in', invoices_info[new_invoice_id])]) does a search on invoice_ids, and that's a computed field, without an search option, so that doesn't work. : https://github.com/odoo/odoo/blob/9.0/addons/sale/sale.py#L135

Fixing that is possible, but ends op being a kludge. 
This code also looks if sale is installed, because if sale is not installed, the field sale_line_ids is not available in self.env['account.invoice.line']._fields.

pedrobaeza (approved):
OK, I see. Thanks for the clarification.

account_invoice_merge/models/account_invoice.py
```diff
@@ -41,7 +41,7 @@ def _get_invoice_line_key_cols(self):
             'product_id', 'account_id', 'account_analytic_id',
             'uom_id'
         ]
-        for field in ['analytics_id']:
+        for field in ['analytics_id', 'sale_line_ids']:
```
MiquelRForgeFlow, line 44: The 'analytics_id' field was added by `account_analytic_plans` base module. But this base module was removed in v9. So this field is not anymore in v9 and then here it should be removed.

MiquelRForgeFlow (changes requested):

RoelAdriaans:
Closing this merge request. 
Does not merge anymore, and no interest in updating or testing it since 9.0 is pretty old.

Issue #239, opened 2017-03-18
[8.0] account_invoice_shipping_address wrong partner


scigghia:
this module in presence of stock_transport_multi_address should transfer the invoice shipping address using right field delivery_address_id from stock.picking

maybe we could add stock_transport_multi_address dependency and use delivery_address_id as proposed

https://github.com/OCA/account-invoicing/blob/8.0/account_invoice_shipping_address/stock.py#L33

pedrobaeza:
Closing as there's a PR for fixing this.

PR #240, opened 2017-03-18
manage correct shipping address invoicing a picking

scigghia:
add stock_transport_multi_address dependency
imp _get_invoice_vals to manage correct shipping address

fix for https://github.com/OCA/account-invoicing/issues/239

### DIFF
--- a/account_invoice_shipping_address/stock.py
+++ b/account_invoice_shipping_address/stock.py
@@ -1,23 +1,9 @@
 # -*- coding: utf-8 -*-
-##############################################################################
-#
-#    Copyright (C) 2014 Agile Business Group sagl (<http://www.agilebg.com>)
-#    Author: Nicola Malcontenti <nicola.malcontenti@agilebg.com>
-#
-#    This program is free software: you can redistribute it and/or modify
-#    it under the terms of the GNU Affero General Public License as published
-#    by the Free Software Foundation, either version 3 of the License, or
-#    (at your option) any later version.
-#
-#    This program is distributed in the hope that it will be useful,
-#    but WITHOUT ANY WARRANTY; without even the implied warranty of
-#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-#    GNU Affero General Public License for more details.
-#
-#    You should have received a copy of the GNU Affero General Public License
-#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
-#
-##############################################################################
+# Copyright (C) 2014 Agile Business Group sagl (<http://www.agilebg.com>)
+# Author: Nicola Malcontenti <nicola.malcontenti@agilebg.com>
+# Copyright 2017 Apulia Software srl - www.apuliasoftware.it
+# Author Andrea Cometa <a.cometa@apuliasoftware.it>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from openerp import models, api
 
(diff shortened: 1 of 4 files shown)


pedrobaeza:
Instead of changing dependencies in a stable version, you should check it dynamically if the field is present (something like `if <field_name> in self.env[<model>]._fields:`)

scigghia:
thank you @pedrobaeza

account_invoice_shipping_address/stock.py
```diff
@@ -19,16 +19,23 @@
...
+from openerp import models
 
 
-class StockPicking(orm.Model):
+class StockPicking(models.Model):
     _inherit = "stock.picking"
 
     def _get_invoice_vals(self, cr, uid, key, inv_type,
-                          journal_id, picking, context=None):
+                          journal_id, move, context=None):
```
pedrobaeza: Why not changing also this to new API with `@api.model`? And was the signature incorrect previously?

account_invoice_shipping_address/stock.py
```diff
@@ -19,16 +19,23 @@
...
         invoice_vals = super(StockPicking, self)._get_invoice_vals(
-            cr, uid, key, inv_type, journal_id, picking, context=context)
-        if picking and picking.partner_id:
-            invoice_vals['address_shipping_id'] = picking.partner_id.id
+            cr, uid, key, inv_type, journal_id, move, context=context)
+        if move and move.partner_id:
+            if 'delivery_address_id' in self.env['stock.picking']._fields:
+                invoice_vals['address_shipping_id'] = (
+                    move.picking_id.delivery_address_id and
```
pedrobaeza: You can use directly `move.picking_id.delivery_address_id.id` without the fear of errors.

account_invoice_shipping_address/stock.py
scigghia: signature was incorrect 
ok for new api

account_invoice_shipping_address/stock.py
scigghia: perfect

account_invoice_shipping_address/stock.py
```diff
@@ -19,16 +19,21 @@
...
-                          journal_id, picking, context=None):
+    @api.model
+    def _get_invoice_vals(self, key, inv_type, journal_id, move):
         invoice_vals = super(StockPicking, self)._get_invoice_vals(
-            cr, uid, key, inv_type, journal_id, picking, context=context)
-        if picking and picking.partner_id:
-            invoice_vals['address_shipping_id'] = picking.partner_id.id
+            key, inv_type, journal_id, move)
+        if move and move.partner_id:
+            if 'delivery_address_id' in self.pool['stock.picking']._fields:
```
pedrobaeza: s/pool/env

pedrobaeza: Put also a comment saying this field is provided by the module x.

account_invoice_shipping_address/stock.py
```diff
@@ -30,7 +30,8 @@ def _get_invoice_vals(self, key, inv_type, journal_id, move):
         invoice_vals = super(StockPicking, self)._get_invoice_vals(
             key, inv_type, journal_id, move)
         if move and move.partner_id:
-            if 'delivery_address_id' in self.pool['stock.picking']._fields:
+            # this field is provided by stock_transport_multi_address module
+            if 'delivery_address_id' in self.pool.env['stock.picking']._fields:
```
pedrobaeza: It's `self.env[...`, not `self.pool.env[...`

pedrobaeza: `s/old/new` means to replace `old` by `new`

PR #241, opened 2017-03-23
[9.0][IMP]sale_stock_picking_return_invoicing. Allow refund after picking has been updated

sergio-teruel:
cc @Tecnativa

### DIFF
--- a/sale_stock_picking_return_invoicing/README.rst
+++ b/sale_stock_picking_return_invoicing/README.rst
@@ -8,7 +8,8 @@ Refund Returned Pickings from Sales Orders
 
 This module extends the functionality of sales orders to support marking some
 returned pickings as "To refund in Sale Order" and to allow you to create
-invoices deducting the products that are returned and refunded.
+invoices deducting the products that are returned and refunded. Also allows to
+modify stock move field to_refund_so after it has been confirmed.
 
 Usage
 =====
@@ -50,6 +51,9 @@ after you created an invoice, you need to:
    Create and View Invoices*.
 #. A refund is created for the quantity you returned before.
 
+For allowing to refund quantities after the picking has been confirmed, you can
+change the value of 'Refund Options' field.
+
 .. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
    :alt: Try me on Runbot
    :target: https://runbot.odoo-community.org/runbot/95/9.0
@@ -81,6 +85,7 @@ Contributors
 ------------
 
 * Jairo Llopis <jairo.llopis@tecnativa.com>
+* Sergio Teruel <sergio.teruel@tecnativa.com>
 
 Maintainer
 ----------
--- a/sale_stock_picking_return_invoicing/models/stock_picking.py
+++ b/sale_stock_picking_return_invoicing/models/stock_picking.py
@@ -24,14 +24,15 @@ class StockPicking(models.Model):
 
     @api.multi
     def _update_stock_moves(self):
-        for pick in self:
-            for move in pick.move_lines:
-                if pick.to_refund_lines == 'keep_line_value':
-                    move.to_refund_so = move.origin_to_refund_so
-                elif pick.to_refund_lines == 'to_refund_so':
-                    move.to_refund_so = True
-                elif pick.to_refund_lines == 'no_refund_so':
-                    move.to_refund_so = False
+        for pick in self.filtered(
+                lambda x: x.to_refund_lines != 'keep_line_value'):
+            pick.move_lines.write({
+                'to_refund_so': (True if pick.to_refund_lines == 'to_refund_so'
+                                 else False)})
+        for move in self.filtered(
+                lambda x: x.to_refund_lines == 'keep_line_value'
+                ).mapped('move_lines'):
+            move.to_refund_so = move.origin_to_refund_so
 
     @api.multi
     def set_delivered_qty(self):
--- a/sale_stock_picking_return_invoicing/models/stock_picking.py
+++ b/sale_stock_picking_return_invoicing/models/stock_picking.py
@@ -1,0 +1,52 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Sergio Teruel <sergio.teruel@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openerp import api, fields, models
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    to_refund_lines = fields.Selection(selection=[
+        ('keep_line_value', 'Keep Line Value'),
+        ('to_refund_so', 'To Refund In Sales Order'),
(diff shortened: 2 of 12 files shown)


chienandalu (approved):
tested on runbot

sale_stock_picking_return_invoicing/README.rst
```diff
@@ -8,7 +8,8 @@ Refund Returned Pickings from Sales Orders
 
 This module extends the functionality of sales orders to support marking some
 returned pickings as "To refund in Sale Order" and to allow you to create
-invoices deducting the products that are returned and refunded.
+invoices deducting the products that are returned and refunded. Also allows
+modify stock move field to_refund_so after it has been confirmed.
```
pedrobaeza: Put `to` before `modify`

sale_stock_picking_return_invoicing/README.rst
```diff
@@ -50,6 +51,9 @@ after you created an invoice, you need to:
    Create and View Invoices*.
 #. A refund is created for the quantity you returned before.
 
+To allow refund quantities after the picking has been confirmed you can change
```
pedrobaeza: s/To allow/For allowing to

pedrobaeza: comma after `confirmed`

sale_stock_picking_return_invoicing/views/stock_picking_view.xml
```diff
@@ -0,0 +1,18 @@
...
+<odoo>
+
+<record id="view_picking_form" model="ir.ui.view">
+    <field name="inherit_id" ref="stock.view_picking_form"/>
+    <field name="model">stock.picking</field>
+    <field name="arch" type="xml">
+        <xpath expr="//field[@name='owner_id']/.." position="after">
+            <field name="location_usage" invisible="1"/>
+            <field name="to_refund_lines"
+                   attrs="{'invisible': [('location_usage', '!=', 'internal')]}"
```
pedrobaeza: This shouldn't use this condition, but be based on the `picking_type_code` field, when it's equal to `outgoing`.

sale_stock_picking_return_invoicing/models/stock_picking.py
```diff
@@ -0,0 +1,56 @@
...
+        string='Refund Options', default='keep_line_value',
+        help="This field allow modify 'to_refund_so' field value in all "
+             "stock moves from this picking after it has been confirmed."
+             "keep_line_value: keep the original value.\n"
+             "to_refund_so: set all stock moves to True value and recompute "
+             "delivered quantities in sale order.\n"
+             "no_refund_so: set all stock moves to False value and recompute "
+             "delivered quantities in sale order.",
+    )
+    location_usage = fields.Selection(
```
pedrobaeza: Don't add this related field, but `picking_type_code`. Explained later.

sale_stock_picking_return_invoicing/models/stock_picking.py
```diff
@@ -0,0 +1,56 @@
...
+        related='location_dest_id.usage',
+        readonly=True,
+    )
+
+    @api.multi
+    def _update_stock_moves(self):
+        for pick in self:
+            for move in pick.move_lines:
```
pedrobaeza: Don't loop here, just make:
```
pick.move_lines.write({'to_refund_so': <value>})
```

sale_stock_picking_return_invoicing/models/stock_picking.py
```diff
@@ -0,0 +1,56 @@
...
+                elif pick.to_refund_lines == 'no_refund_so':
+                    move.to_refund_so = False
+
+    @api.multi
+    def set_delivered_qty(self):
+        so_lines = self.mapped(
+            'move_lines.procurement_id.sale_line_id').filtered(
+            lambda x: x.product_id.invoice_policy in ('order', 'delivery'))
+        for so_line in so_lines:
+            so_line.qty_delivered = so_line._get_delivered_qty()
```
pedrobaeza, line 43: You should only update the lines that change its `to_refund_so` value.

sale_stock_picking_return_invoicing/views/stock_move_view.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+<record id="view_move_picking_tree" model="ir.ui.view">
+    <field name="inherit_id" ref="stock.view_move_picking_tree"/>
+    <field name="model">stock.move</field>
+    <field name="arch" type="xml">
+        <xpath expr="//field[@name='state']" position="after">
+            <field name="to_refund_so" invisible="1"/>
+            <field name="origin_to_refund_so" invisible="1"/>
```
pedrobaeza: This one is not needed in the view

pedrobaeza (changes requested):

sale_stock_picking_return_invoicing/models/stock_picking.py
sergio-teruel: I only write once. And i need maintain the old value stored in move.

pedrobaeza: No, double check please.

pedrobaeza:
@sergio-teruel what happens with the commits? There are a lot of them that doesn't correspond here.

sale_stock_picking_return_invoicing/models/stock_picking.py
sergio-teruel: If I select in pick the option 'keep line values' I need write each move with the field 'origin_to_refund_so', for the other options I am agree because i write the same value for all records.

pedrobaeza: Yeah, but the other 2 can be written in one step.

pedrobaeza:
Check Travis and pending comments.

pedrobaeza:
@chienandalu please finish this

LoisRForgeFlow:
Any new here? I would like to migrate this module to v10.

pedrobaeza:
@lreficent no need to migrate it, as it's included already in core. See https://github.com/OCA/OpenUpgrade/pull/1301.

I will check this PR for getting it merge.

LoisRForgeFlow:
@pedrobaeza oh, great! Thanks you!

sale_stock_picking_return_invoicing/models/stock_picking.py
chienandalu: Is there a flag that can give me the state change at this point?

MiquelRForgeFlow:
@sergio-teruel @chienandalu please, rebase and squash. It should go :green_apple:  now :)

MiquelRForgeFlow:
Any news here?

pedrobaeza:
We are not very interested now in v9, so you can superseed it if needed.

rafaelbn:
I think we should merge this as is, in other case better close it

rafaelbn (approved):
Functionally :+1:

PR #243, opened 2017-03-29
[MIG] Migrated account_invoice_analytic_search v8 to v9

serpentcs-dev1:
@jbeficent Migrated account_invoice_analytic_search to v9


account_invoice_analytic_search/README.rst
```diff
@@ -0,0 +1,43 @@
...
+Invoice Analytic Search
+=======================
+
+Organizations often require to quickly find the invoices associated to a
+project or to an analytic account, searching by it's code, name or
+project/account manager. This module introduces the possibility search
+customer or supplier invoices by analytic account or by project manager.
+
+Installation
+============
```
AaronHForgeFlow: Please remove empty sections

account_invoice_analytic_search/README.rst
```diff
@@ -0,0 +1,43 @@
...
+
+* The module automatically takes care of its dependencies and is ready for use after the installation
+
+Credits
+=======
```
AaronHForgeFlow: Please adjust this to the OCA guidelines:
https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst

Also remove the icon

AaronHForgeFlow (changes requested):
Functional ok

AaronHForgeFlow (changes requested):
I'd replace the image at account_invoice_analytic_search/static/description/icon.png for the OCA one

serpentcs-dev1:
I have added OCA Icon as you suggested @aheficent

account_invoice_analytic_search/models/account_invoice.py
```diff
@@ -0,0 +1,67 @@
...
+        res = {}
+        for invoice in self:
+            res[invoice.id] = []
+            for invoice_line in invoice.invoice_line_ids:
+                if invoice_line.account_analytic_id and\
+                        invoice_line.account_analytic_id.user_id:
+                    res[invoice.id].append(
+                        invoice_line.account_analytic_id.user_id
+                        )
+        return res
```
AaronHForgeFlow: I wonder if it's possible to do something like this:        
for invoice in self:
     invoice.account_analytic_user_ids = invoice.mapped('invoice.invoice_line_ids.account_analytic_id.user_id')

AaronHForgeFlow (commented):
I looks fine, just some comments

serpentcs-dev1:
There was user_id field in Analytic Account which is removed now from [v9](https://github.com/OCA/OpenUpgrade/blob/9.0/addons/analytic/migrations/9.0.1.1/openupgrade_analysis_work.txt#L44
), instead of that partner_id field is introduced. So I have replaced the related field with it.

@aheficent  Is it the proper way?

AaronHForgeFlow:
Hi @darshan-serpent In my opinion this is the proper way. Openupgrade says for the  partner_id in analytic:
[> get partner from user and set it to partner_id field in post-migration](https://github.com/OCA/OpenUpgrade/blob/9.0/addons/analytic/migrations/9.0.1.1/openupgrade_analysis_work.txt#L48)

I understand that it'd contain the partner related to the user

serpentcs-dev1:
Yes, you are right, So can you please review the changes again?

AaronHForgeFlow (changes requested):
I tested as follows:
1. created customer invoice and add an analytic account and a partner_id to the analytic account
2. searched by the analytic_account: no results found
3. searched by the partner_id: no results found

Besides, the field partner_id in the analytic_account is labelled as "Customer" so I changed my mind. The field should be related to the user_id in the project IMHO

What do you think?

PR #245, opened 2017-03-30
account_invoice_fiscal_position_update fix for partner_id change

CGenie:
When creating a Purchase Order, the fiscal position used was not from Contact/supplier
but some default one. Adding an onchange handler fixes this issue.

### DIFF
--- a/account_invoice_fiscal_position_update/models/account_invoice.py
+++ b/account_invoice_fiscal_position_update/models/account_invoice.py
@@ -11,6 +11,12 @@ from openerp.exceptions import ValidationError
 class AccountInvoice(models.Model):
     _inherit = "account.invoice"
 
+    @api.onchange('partner_id', 'company_id')
+    def _onchange_partner_id(self):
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I think this patch is not correct, as you are launching an overwrite of the taxes, not of the fiscal position, which doesn't correspond with your explanations.

CGenie:
I just ran the onchange method for `fiscal_position_id` which isn't triggered initially when creating a Purchase Order.

pedrobaeza:
But this is for invoices, not for purchase orders.

CGenie:
Purchase orders have 'Vendor Bills' which are just account.invoice models.

pedrobaeza:
Then it's not creating the purchase order...

Please explain precisely the flow you want to fix.

alexis-via:
@CGenie as Pedro saif, we need a detailed scenario to be able to understand your problem and be able to reproduce your issue. Otherwise :-1:

PR #249, opened 2017-03-31
account_invoice_validation_workflow: prevent hiding register payment button

sp-jamotion:
Invoice form view refactored to avoid hiding the register payment button
on open invoices.


renzo1974:
I have checked this and it does what it should.

PR #250, opened 2017-04-01
[10.0] [MIG] migrate account_invoice_template to v10

jzahid:
migrate account_invoice_template to v10

### DIFF
--- a/account_invoice_template/__init__.py
+++ b/account_invoice_template/__init__.py
@@ -18,5 +18,5 @@
 #    along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 ##############################################################################
-import account_invoice_template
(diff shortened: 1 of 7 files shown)


account_invoice_template/__init__.py
```diff
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 ##############################################################################
-#
+#    Copyright (C) 2017 Aurium Technologies (<http://www.auriumtechnologies.com>)
```
pedrobaeza, line 3: You shouldn't add your copyright in files that you don't touch (and more if they are `__init__.py` files.

jzahid: sorry. i will remove it.

PR #253, opened 2017-04-03
[9.0] [IMP] account_invoice_supplier_ref_unique: set unique number in refund wizard

ThomasBinsfeld:
This improvement adds a new field on the refund wizard to allow to initialize the unique vendor number of the credit note.

### DIFF
--- a/account_invoice_supplier_ref_unique/wizards/account_invoice_refund.py
+++ b/account_invoice_supplier_ref_unique/wizards/account_invoice_refund.py
@@ -1,0 +1,23 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openerp import api, fields, models
+
+
+class AccountInvoiceRefund(models.TransientModel):
+    _inherit = 'account.invoice.refund'
+
+    supplier_invoice_number = fields.Char(
+        string='Vendor invoice number'
+    )
+
+    @api.multi
+    def compute_refund(self, mode='refund'):
+        res = True
+        for form in self:
+            res = super(
+                AccountInvoiceRefund, form.with_context(
+                    supplier_invoice_number=form.supplier_invoice_number)).\
+                compute_refund(mode=mode)
+        return res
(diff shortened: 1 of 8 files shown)


elicoidal (approved):

account_invoice_supplier_ref_unique/wizards/account_invoice_refund.py
```diff
@@ -0,0 +1,23 @@
...
+    supplier_invoice_number = fields.Char(
+        string='Vendor invoice number'
+    )
+
+    @api.multi
+    def compute_refund(self, mode='refund'):
+        res = True
+        for form in self:
+            res = super(
+                AccountInvoiceRefund, self.with_context(
```
pedrobaeza: s/self/form

account_invoice_supplier_ref_unique/wizards/account_invoice_refund.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2016 ACSONE SA/NV
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+
+<odoo>
+<data>
```
pedrobaeza: this tag is not needed.

account_invoice_supplier_ref_unique/wizards/account_invoice_refund.xml
```diff
@@ -0,0 +1,21 @@
...
+
+    <record model="ir.ui.view" id="account_invoice_refund_form_view">
+        <field name="name">account.invoice.refund.form (in account_invoice_supplier_ref_unique)</field>
```
pedrobaeza: You don't need to put a name to the view if you don't want.

pedrobaeza (changes requested):

account_invoice_supplier_ref_unique/wizards/account_invoice_refund.xml
ThomasBinsfeld: I prefer it with a name :ok_hand:

ThomasBinsfeld:
@pedrobaeza Fixed and squashed, thanks for your review.

account_invoice_supplier_ref_unique/wizards/account_invoice_refund.xml
pedrobaeza: OK, no problem. This is just a matter of taste, although not putting it is one thing less to translate.

pedrobaeza (approved):

zakiuu (approved):

PR #260, opened 2017-04-21
[8.0][ADD] Module account_invoice_so_exception_reset.

NL66278:
Prevent Sale Order from staying in state 'Invoice Exception' when invoice is cancelled and subsequently reset to draft and later on paid after all....


moylop260 (approved):
I prefer returns the same value from orginal inherited method, but LGTM

NL66278:
@moylop260 Thanks for approving. I took notice of your remark, and now return the original result.

PR #238, opened 2017-03-16
[10.0][mig] account_invoice_merge

LoisRForgeFlow:
Task done in this migration:
* Replace openerp tags and import by odoo.
* Update README.
* Adapt header of the files to new template.
* Correct typo error in wizard view.
* Remove the use of workflow in the invoice.
* Squash commit history.

PR #262, opened 2017-04-27
[9.0][FIX] account_invoice_merge: Invoices not related with sale order

carlosdauden:
self.env['sale.order'] is bolean evaluated as False

@Tecnativa

### DIFF
--- a/account_invoice_merge/models/account_invoice.py
+++ b/account_invoice_merge/models/account_invoice.py
@@ -214,7 +214,7 @@ class AccountInvoice(models.Model):
         invoice_line_obj = self.env['account.invoice.line']
         # None if purchase is not installed
         for new_invoice_id in invoices_info:
-            if so_obj != False:
+            if so_obj is not False:
                 todos = old_invoices.mapped(
                     'invoice_line_ids.sale_line_ids.order_id')
                 todos.write({'invoice_ids': [(4, new_invoice_id)]})
--- a/account_invoice_merge/models/account_invoice.py
+++ b/account_invoice_merge/models/account_invoice.py
@@ -1,23 +1,11 @@
 # -*- coding: utf-8 -*-
-##############################################################################
-#
-#    OpenERP, Open Source Management Solution
-#    Copyright (C) 2004-2010 Tiny SPRL (<http://tiny.be>).
-#
-#    This program is free software: you can redistribute it and/or modify
-#    it under the terms of the GNU Affero General Public License as
-#    published by the Free Software Foundation, either version 3 of the
-#    License, or (at your option) any later version.
-#
-#    This program is distributed in the hope that it will be useful,
-#    but WITHOUT ANY WARRANTY; without even the implied warranty of
-#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-#    GNU Affero General Public License for more details.
-#
-#    You should have received a copy of the GNU Affero General Public License
-#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
-#
-##############################################################################
+# Copyright 2004-2010 Tiny SPRL (http://tiny.be).
+# Copyright 2010-2011 Elico Corp.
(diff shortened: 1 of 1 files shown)


account_invoice_merge/models/account_invoice.py
```diff
@@ -214,7 +214,7 @@ def make_key(br, fields):
         invoice_line_obj = self.env['account.invoice.line']
         # None if purchase is not installed
```
pedrobaeza: Why this text?

carlosdauden: Maybe we should ask @pedrobaeza @Cedric-Pigeon 
https://github.com/OCA/account-invoicing/blame/9.0/account_invoice_merge/models/account_invoice.py#L215
:wink:

pedrobaeza: Well, with the blame, I only fix PEP8 issues, but the question is more as curiosity. Have you seen the PR status?

LoisRForgeFlow:
@carlosdauden @pedrobaeza @sergio-teruel We have found this bug:
image
It pops after creating two invoices from the same sale order and trying to merge them.
image

pedrobaeza:
@lreficent, as the later code was never run, there were no issues previously. The problem come in this code:
```
invoice_line_ids = invoice_line_obj.search(
    [('product_id', '=', so_line.product_id.id),
     ('invoice_id', '=', new_invoice_id)])
if invoice_line_ids:
    so_line.write({'invoice_lines': [(6, 0, invoice_line_ids)]})
```
It should be replaced by:
```
invoice_lines = invoice_line_obj.search(
    [('product_id', '=', so_line.product_id.id),
     ('invoice_id', '=', new_invoice_id)])
if invoice_lines:
    so_line.write({'invoice_lines': [(6, 0, invoice_lines.ids)]})
```
It's a bad conversion to new API.

LoisRForgeFlow (changes requested):
https://github.com/OCA/account-invoicing/pull/262/files#diff-e3d50a35a4d448eec1952859dc589751R228 

 `{'invoice_lines': [(6, 0, invoice_line_ids)]})` -> `{'invoice_lines': [(6, 0, invoice_line_ids.ids)]})`

LoisRForgeFlow:
@pedrobaeza yes, exactly. Can you add the fix here and we'll do in v10.

Also, we have found that [this line](https://github.com/OCA/account-invoicing/pull/262/files#diff-e3d50a35a4d448eec1952859dc589751R218) is not searching properly and gets all the sales orders.

pedrobaeza:
@lreficent that can happen if `invoice_ids` is a computed field and there isn't a correct implementation of the search method. Can you check?

LoisRForgeFlow:
@pedrobaeza I don't know if it is the cleanest way but it is fixed now in v10 (4e6e938e580a7d5619e3fcb6d9a116f6af08b771)

account_invoice_merge/models/account_invoice.py
```diff
@@ -214,18 +214,18 @@ def make_key(br, fields):
...
         for new_invoice_id in invoices_info:
-            if so_obj:
-                todos = so_obj.search(
-                    [('invoice_ids', 'in', invoices_info[new_invoice_id])])
+            if so_obj != False:
```
MiquelRForgeFlow: so_obj != False → so_obj is not False to make the travis works

MiquelRForgeFlow (changes requested):

MiquelRForgeFlow:
Can you apply here the last fixes of https://github.com/OCA/account-invoicing/issues/238? 😊

gurneyalex:
#269 fixes the build of the 10.0 branch

pedrobaeza:
@carlosdauden this is waiting your final touch

MiquelRForgeFlow (approved):
:+1:

antoniocanovas (approved):
Ok, functionality ...

LoisRForgeFlow (approved):
Tested on runbot :+1:

PR #265, opened 2017-05-10
[FIX] do_merge now returns a tuple.

NL66278:
do_merge method should have a consistent return value. Now we had a crash in a database where account_invoice_merge_purchase was not installed. This was due to this being a new module, and an existing database. But can also happen in databases were purchase is not needed and installed, so auto_install of account_invoice_merge_purchase will not be done.

### DIFF
--- a/account_invoice_merge/wizard/invoice_merge.py
+++ b/account_invoice_merge/wizard/invoice_merge.py
@@ -83,8 +83,10 @@ class InvoiceMerge(models.TransientModel):
         aw_obj = self.env['ir.actions.act_window']
         ids = self.env.context.get('active_ids', [])
         invoices = inv_obj.browse(ids)
-        allinvoices = invoices.do_merge(keep_references=self.keep_references,
-                                        date_invoice=self.date_invoice)
+        invoices_info, invoice_lines_info = invoices.do_merge(
+            keep_references=self.keep_references,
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

astirpe (approved):

pedrobaeza:
Do you know why the Travis error?

NL66278:
@pedrobaeza I do not quite see where the Travis error comes from. The error only occurs when some modules are excluded from testing. And then it fails on invoice sorting, which I feel has nothing to do with the changes proposed here.

gurneyalex:
the 8.0 build of Travis is read, so unless we fix this all PR on that branch will be red too.

gurneyalex:
is this the same thing as #268 ?

astirpe:
@gurneyalex yes, regarding module `account_invoice_merge_purchase` the fix the same.
This PR include the fix for module `account_invoice_merge` also.

astirpe (changes requested):
@NL66278 after the merge of #268, this PR has some conflicts to be fixed

NL66278:
@astirpe #268 Partially had the same changes that I proposed. I only kept the extra stuff in my branch that was missing in #268

astirpe (approved):

PR #269, opened 2017-05-30
Fix 10.0 branch travis build

gurneyalex:
in account_invoice_refund_link:

* change test class to TransactionCase to avoid a deadlock on the sequence for
  account.move
* add sale as dependency to avoid issues with team_id not being found which
  breaks the tests

### DIFF
--- a/account_invoice_refund_link/tests/test_invoice_refund_link.py
+++ b/account_invoice_refund_link/tests/test_invoice_refund_link.py
@@ -7,21 +7,22 @@ from odoo.tests import common
 from .. import post_init_hook
 
 
-class TestInvoiceRefundLink(common.SavepointCase):
+# using SavepointCase here seems to cause a deadlock
+# -> forcing TransactionCase
+class TestInvoiceRefundLink(common.TransactionCase):
     filter_refund = 'refund'
 
-    @classmethod
-    def setUpClass(cls):
-        super(TestInvoiceRefundLink, cls).setUpClass()
-        cls.partner = cls.env['res.partner'].create({
+    def setUp(self):
+        super(TestInvoiceRefundLink, self).setUp()
+        self.partner = self.env['res.partner'].create({
             'name': 'Test partner',
         })
-        default_line_account = cls.env['account.account'].search([
+        default_line_account = self.env['account.account'].search([
             ('internal_type', '=', 'other'),
             ('deprecated', '=', False),
-            ('company_id', '=', cls.env.user.company_id.id),
+            ('company_id', '=', self.env.user.company_id.id),
         ], limit=1)
-        cls.invoice_lines = [(0, False, {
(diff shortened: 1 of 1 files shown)


gurneyalex:
I'm not too happy with the added dependency, but I have no better idea at the moment. Suggestions are welcome.

tafaRU (approved):
Once travis is green.
Thanks a lot!

account_invoice_refund_link/__manifest__.py
```diff
@@ -18,6 +18,7 @@
     "post_init_hook": "post_init_hook",
     "depends": [
         'account',
+        'sale',
```
adrienpeiffer: Thanks !
Odoo solved this part of the problem https://github.com/odoo/odoo/pull/17273

tafaRU: Good catch!

gurneyalex:
@adrienpeiffer so this means I can remove the dependency and travis will be green tomorrow ?

pedrobaeza:
Why the deadlock is happening? I have switched all my tests to SavepointCase and this can be a problem in other tests as well.

adrienpeiffer:
@gurneyalex Yes :smile: 
About the deadlock, it's quite strange because only the OCB build failed 
https://travis-ci.org/OCA/account-invoicing/builds/234196352

account_invoice_refund_link/tests/test_invoice_refund_link.py
```diff
@@ -7,21 +7,20 @@
 from .. import post_init_hook
 
 
-class TestInvoiceRefundLink(common.SavepointCase):
+class TestInvoiceRefundLink(common.TransactionCase):
```
pedrobaeza: Please drop a comment saying that this is specifically TransactionCase because the deadlock.

gurneyalex:
@pedrobaeza I don't know. I thought it could be a cause, checked and it did fix the issue. 

@adrienpeiffer what is strange is that my branch build correctly (https://travis-ci.org/gurneyalex/account-invoicing/builds/237411606)

gurneyalex:
I just updated the PR with the requested comment, and dropped the added dependency (obviously this will break the PR until tomorrow)

adrienpeiffer:
Travis is red because the last odoo fix isn't on OCB ...

gurneyalex:
yay green!

adrienpeiffer:
@gurneyalex The last build on 10.0 is green. So, I think this one isn't necessary.

https://travis-ci.org/OCA/account-invoicing/builds/234196352

simahawk (approved):

gurneyalex:
@adrienpeiffer ah great then!

@pedrobaeza you can keep using checkpoints in your tests :smile_cat:

PR #273, opened 2017-06-12
[8.0] ADD Module account_invoice_date_period_sync

drl9319:
This module change the period when a date is change in account move.


drl9319:
I close this pr becouse must be in account-financial-tools, this is the new pr: https://github.com/OCA/account-financial-tools/pull/500

PR #274, opened 2017-06-12
[10.0][MIG] account_invoice_line_description

luismontalba:
Account invoice line description
=========================
This module allows to use only the sale description or the purchase description
on the invoice lines depending on the invoice type.

Configuration
-----------------
The user has to belong to group_use_product_description_per_inv_line.
This is possible by selecting the related option in the following menu:
* Invoicing --> Configuration --> Settings

cc @Tecnativa


luismontalba:
Travis fixed

account_invoice_line_description/models/invoice.py
```diff
@@ -0,0 +1,75 @@
...
+from odoo import _, api, models
+
+
+class AccountInvoiceLine(models.Model):
+    _inherit = "account.invoice.line"
+
+    @api.onchange('product_id')
+    def _onchange_product_id(self):
+
+        domain = {}
```
tafaRU: It's better calling super to avoid breaking inheritance chain.

tafaRU (changes requested):

account_invoice_line_description/models/invoice.py
pedrobaeza: Or just use another onchange name: `_onchange_product_id_account_invoice_line_description(...`

tafaRU: In this case I'd prefer calling super as much cleaner.

LoisRForgeFlow (commented):
Tested on runbot :+1: 

Regarding code I agree with @tafaRU comment

luismontalba:
Yes, I agree with you, now the method is calling super

account_invoice_line_description/views/res_config_view.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<odoo>
+    <record id="view_account_config" model="ir.ui.view">
+        <field name="name">account settings</field>
+        <field name="model">account.config.settings</field>
+        <field name="inherit_id" ref="account.view_account_config_settings"/>
+        <field name="arch" type="xml">
+            <xpath
+                expr="//label[@for='module_account_budget']"
```
tafaRU: This field is only visible if user belongs to `account.group_account_user` (aka Accountant) group.
For this reason I see two possibilities:

1. add this info in the README file
2. move `div` after https://github.com/odoo/odoo/blob/10.0/addons/account/views/res_config_view.xml#L104

What do you think?

luismontalba: Sure, I'm going to improve README file

account_invoice_line_description/__manifest__.py
```diff
@@ -1,26 +1,11 @@
...
-#
-##############################################################################
+# Copyright 2015 Agile Business Group sagl
+# (<http://www.agilebg.com>)
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
 {
     'name': "Account invoice line description",
-    'version': '8.0.1.0.0',
+    'version': '10.0.1.0.1',
```
pedrobaeza: 10.0.1.0.0

account_invoice_line_description/__manifest__.py
```diff
@@ -1,26 +1,11 @@
...
     'category': 'Generic Modules/Accounting',
     'author': "Agile Business Group, Odoo Community Association (OCA)",
```
pedrobaeza: Put Tecnativa

account_invoice_line_description/models/invoice.py
```diff
@@ -0,0 +1,33 @@
+# -*- coding: utf-8 -*-
```
pedrobaeza: Please call it `account_invoice.py`

account_invoice_line_description/models/invoice.py
```diff
@@ -0,0 +1,33 @@
...
+        part = self.invoice_id.partner_id
+        type = self.invoice_id.type
+        if self.product_id:
+            if part.lang:
+                product = self.product_id.with_context(lang=part.lang)
+            else:
+                product = self.product_id
+
+            if type in ('in_invoice', 'in_refund'):
+                if product.description_purchase and self.user_has_groups(
```
pedrobaeza: Check as first thing the group and return if not belonging to to avoid all of these lines.

account_invoice_line_description/models/res_config.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
```
pedrobaeza: This should be in `wizards`, not in `models`

account_invoice_line_description/views/res_config_view.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
```
pedrobaeza: Put this inside `wizards`

pedrobaeza (changes requested):

pedrobaeza:
Please add a test for checking the module:

* Set initial data for creating an invoice (partner, product, accounts...)
* Create an invoice of 2 types (purchase and sales).
* Call onchange over the invoices and check result.

account_invoice_line_description/tests/test_account_invoice_line_description.py
```diff
@@ -0,0 +1,102 @@
...
+
+from odoo.tests.common import SavepointCase
+
+
+class TestAccountInvoiceLineDescription(SavepointCase):
+
+    @classmethod
+    def setUpClass(cls):
+        super(TestAccountInvoiceLineDescription, cls).setUpClass()
+
```
pedrobaeza: Why do you waste lines letting these spaces inside a method?

account_invoice_line_description/tests/test_account_invoice_line_description.py
```diff
@@ -0,0 +1,102 @@
...
+        cls.invoice_purchase = cls.env['account.invoice'].create({
+            'partner_id': cls.partner.id,
+            'journal_id': cls.journal_purchase.id,
+            'type': 'in_invoice',
+            'invoice_line_ids': cls.invoice_purchase_vals,
+        })
+
+    def test_onchange_product_id_sale(self):
+        self.invoice_sale.invoice_line_ids.product_id = \
```
pedrobaeza: This can be done on setup

pedrobaeza (changes requested):
You are not sure if the user belongs to the group. Assign it explicitly (`cls.env.user.group_ids |= ...`)

pedrobaeza (approved):
Thanks for the tests. @tafaRU please give your final blessing.

account_invoice_line_description/README.rst
```diff
@@ -1,48 +1,64 @@
...
 ================================
 
-This module allows to use only the product description on the invoice lines.
+This module allows to use only the sale description or the purchase description
+on the invoice lines depending on the invoice type.
 
 Configuration
 =============
 
 The user has to belong to group_use_product_description_per_inv_line.
```
LoisRForgeFlow: I would put the non-technical name here: _Allow using only the product description on th invoice lines_

account_invoice_line_description/README.rst
```diff
@@ -1,48 +1,64 @@
...
-This is possible by selecting the related option in the following menu:
+This is possible having the "Accountant" permission level by selecting the
+related option in the
+following menu:
```
LoisRForgeFlow: this line can be following the above one.

LoisRForgeFlow (approved):
I update my review: 

Some minor comments in the README, tested + code review :+1:

luismontalba:
Yes you're right, README updated

MiquelRForgeFlow (approved):
LGTM :+1: 
Please, squash commits

tafaRU (approved):

account_invoice_line_description/models/account_invoice_line.py
```diff
@@ -0,0 +1,34 @@
...
+    _inherit = "account.invoice.line"
+
+    @api.onchange('product_id')
+    def _onchange_product_id(self):
+        res = super(AccountInvoiceLine, self)._onchange_product_id()
+        if not self.user_has_groups(
+                'account_invoice_line_description.'
+                'group_use_product_description_per_inv_line'):
+            return res
+        else:
```
moylop260: You don't need a else here because the `if` has a `return`

account_invoice_line_description/models/account_invoice_line.py
```diff
@@ -0,0 +1,34 @@
...
+            part = self.invoice_id.partner_id
+            type = self.invoice_id.type
```
moylop260: Redefined built-in `type` variable

pedrobaeza: @luismontalba call it `inv_type`

account_invoice_line_description/models/account_invoice_line.py
```diff
@@ -0,0 +1,34 @@
...
+            if self.product_id:
```
moylop260: Using new api approach you could reduce the cyclomatic complexity here.

```python
product = self.product_id.with_context(lang=self.invoice_id.partner_id)
self.name = (product.description_purchase or self.name if type in ('in_invoice', 'in_refund') else
                      product.description_sale or self.name)
```

luismontalba:
Code refactored. How do you see it now @moylop260 ?

moylop260 (approved):

PR #255, opened 2017-04-12
[8.0][account_invoice_kanban] Kanban view for invoice to manage administrative tasks

sudhir-serpentcs:
Added 'account_invoice_kanban' module for kanban view of invoices like customer invoice and supplier invoice.

@elicoidal

PR #275, opened 2017-06-13
[ADD] New module invoice_margin

legalsylvain:
Trivial module to add margin concept on account.invoice (customer invoices). this Module is highly inspired from "sale_margin" that add margin computation on sale.order. (by Odoo SA) and from "pos_margin" (OCA/pos repository)

account_invoice_form

thanks for your review.


bealdav:
:+1: 

Thanks

invoice_margin/models/account_invoice_line.py
```diff
@@ -0,0 +1,36 @@
...
+    @api.multi
+    @api.depends('product_id', 'quantity', 'price_subtotal')
+    def _compute_multi_margin(self):
+        for line in self:
+            if not line.product_id or\
+                    line.invoice_id.type not in ('out_invoice', 'out_refund'):
+                line.purchase_price = 0
+                line.margin = 0
+            else:
+                line.purchase_price = line.product_id.standard_price
```
moylop260: What about use a related instead?

legalsylvain: A related field will change if the original value changes. does'nt it ?

invoice_margin/models/account_invoice_line.py
```diff
@@ -0,0 +1,36 @@
...
+    purchase_price = fields.Float(
+        'Cost Price', compute='_compute_multi_margin', store=True,
+        multi='multi_margin',
+        digits_compute=dp.get_precision('Product Price'))
+
+    # Compute Section
+    @api.multi
+    @api.depends('product_id', 'quantity', 'price_subtotal')
+    def _compute_multi_margin(self):
+        for line in self:
```
moylop260: Use `for line in self.filtered(lambda l: l.product_id and line.invoice_id.type in ('out_invoice', 'out_refund'))`
if a record.field is not used so a `Null` is stored by default.

invoice_margin/models/account_invoice_line.py
```diff
@@ -0,0 +1,36 @@
...
+        digits_compute=dp.get_precision('Product Price'))
+
+    purchase_price = fields.Float(
+        'Cost Price', compute='_compute_multi_margin', store=True,
+        multi='multi_margin',
+        digits_compute=dp.get_precision('Product Price'))
+
+    # Compute Section
+    @api.multi
+    @api.depends('product_id', 'quantity', 'price_subtotal')
```
moylop260: `invoice_id.type`

invoice_margin/views/view_account_invoice.xml
```diff
@@ -0,0 +1,35 @@
...
+        </field>
+    </record>
+
+    <record id="view_account_invoice_form" model="ir.ui.view">
+        <field name="model">account.invoice</field>
+        <field name="inherit_id" ref="account.invoice_form"/>
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='amount_untaxed']/.." position="after">
+                <group name="margin">
+                    <field name="margin" widget="monetary"/>
```
moylop260, line 26: hidden the field if is not a `out_*` invoice.type

legalsylvain: It is not necessary, because it's not the same view. (in and out). (I tested).

invoice_margin/models/account_invoice_line.py
loxamir: It can't be a related field because when the product cost changes this field have to keep the original amount

invoice_margin/models/account_invoice_line.py
legalsylvain: I'm not sure about your proposal. Use case with your algorithm:
- set a product > margin is computed;
- remove the product > margin is not set to 0 (expected behaviour)

invoice_margin/models/account_invoice_line.py
legalsylvain: thanks. Done.

invoice_margin/models/account_invoice_line.py
moylop260: If you define a related with store=True and readonly=True you will have the same effect like as a computed store=True (but without extra code)

invoice_margin/models/account_invoice_line.py
moylop260: try it you will see the same effect here too

legalsylvain: Ow. Indeed, it works too. i don't understand why ! if product is set and after unset, the algorithm do not pass in the for and so it should not recompute margin. But well ! thanks.

invoice_margin/views/view_account_invoice.xml
```diff
@@ -0,0 +1,35 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+Copyright (C) 2017 - Today: GRAP (http://www.grap.coop)
+@author Sylvain LE GAL (https://twitter.com/legalsylvain)
+License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+-->
+
+<openerp><data>
```
moylop260, line 8: I like this way to easiest migration use just
```diff
- <openerp><data>
+ <odoo>
```
👍

invoice_margin/models/account_invoice_line.py
legalsylvain: Well, 
your solution indeed works, but it doen not allow to overload the function that is the actual purpose. In some context, we could prefer to use a different field as standard_price. don't you think ? (if you want to use value from moves of the picking, or using FIFO / LIFO valuation mechanism.

invoice_margin/models/account_invoice_line.py
legalsylvain: done.

mourad-ehm:
:+1:  code review

legalsylvain:
Hi @bealdav and @mourad-ehm. Thanks for your reviews.

@moylop260, did i answered to all your questions ? 

thanks.

moylop260 (approved):
thanks!

MiquelRForgeFlow:
Could you greenify the PR? 🙂

legalsylvain:
Hi @mreficent. I don't know how. I can not see where is the error. could you help me.

log on runbot : https://runbot.odoo-community.org/runbot/static/build/3295624-275-9bf1ea/logs/job_20_test_all.txt

log on buildout : https://travis-ci.org/OCA/account-invoicing/jobs/242529534

there is just a "Found 1 lines with errors".

if it's the warning, it's not related to my PR. 
```WARNING openerp_test openerp.modules.loading: The model account.invoice.stage has no access rules, consider adding one. E.g. access_account_invoice_stage,access_account_invoice_stage,model_account_invoice_stage,,1,0,0,0```

MiquelRForgeFlow:
Yes, it seems it's that warning. I think the warning is related to https://github.com/OCA/account-invoicing/pull/255.

MiquelRForgeFlow:
@legalsylvain That warning is solved now. Could you try again to greenify the PR?

legalsylvain:
greenification OK. Ok for merging ? 

regards.

MiquelRForgeFlow (approved):
LGTM :+1:

bealdav (approved):

legalsylvain:
hop ! merging.

Issue #277, opened 2017-06-14
[8.0] account_invoice_merge_purchase KeyError when merging


robch2342:
If a purchase order has several invoices attached to it, and you try to merge one (but not all) of those invoices, you will get a KeyError.

2017-06-14 06:52:58,021 2574 ERROR dermanord_11juni openerp.http: Exception during JSON request handling.
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/openerp/http.py", line 544, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/usr/lib/python2.7/dist-packages/openerp/http.py", line 581, in dispatch
    result = self._call_function(**self.params)
  File "/usr/lib/python2.7/dist-packages/openerp/http.py", line 317, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/usr/lib/python2.7/dist-packages/openerp/service/model.py", line 118, in wrapper
    return f(dbname, *args, **kwargs)
  File "/usr/lib/python2.7/dist-packages/openerp/http.py", line 314, in checked_call
    return self.endpoint(*a, **kw)
  File "/usr/lib/python2.7/dist-packages/openerp/http.py", line 810, in __call__
    return self.method(*args, **kw)
  File "/usr/lib/python2.7/dist-packages/openerp/http.py", line 410, in response_wrap
    response = f(*args, **kw)
  File "/usr/lib/python2.7/dist-packages/openerp/addons/web/controllers/main.py", line 948, in call_button
    action = self._call_kw(model, method, args, {})
  File "/usr/lib/python2.7/dist-packages/openerp/addons/web/controllers/main.py", line 936, in _call_kw
    return getattr(request.registry.get(model), method)(request.cr, request.uid, *args, **kwargs)
  File "/usr/lib/python2.7/dist-packages/openerp/api.py", line 268, in wrapper
    return old_api(self, *args, **kwargs)
  File "/usr/lib/python2.7/dist-packages/openerp/api.py", line 399, in old_api
    result = method(recs, *args, **kwargs)
  File "/usr/share/odooext-oca-account-invoicing/account_invoice_merge/wizard/invoice_merge.py", line 87, in merge_invoices
    date_invoice=self.date_invoice)
  File "/usr/lib/python2.7/dist-packages/openerp/api.py", line 266, in wrapper
    return new_api(self, *args, **kwargs)
[...]

PR #278, opened 2017-06-14
[Fix][8.0] account_invoice_merge_purchase KeyError when merging

robch2342:
This is a fix for issue #277


DarioLodeiros (approved):
Checked, I think that fix a significant error

PR #282, opened 2017-06-28
[ADD] account_invoice_line_sequence v10

serpentcs-dev1:
Account Invoice lines with sequence number
==================================

Provides a new sequence field on invoice lines which helps to manage the order of the invoice lines.

Later this sequence can be displayed on the Invoice report.

### DIFF
--- a/account_invoice_line_sequence/README.rst
+++ b/account_invoice_line_sequence/README.rst
@@ -1,0 +1,59 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+==========================================
+Account Invoice lines with sequence number
+==========================================
+
+Provides a new sequence field on invoice lines which helps to manage the order of the invoice lines.
+
+
+Usage
+=====
+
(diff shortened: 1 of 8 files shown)


serpentcs-dev1:
@aheficent Could you please take a look?

AaronHForgeFlow:
It seems that runbot it's not working now :face_with_head_bandage:  This will need to be rebuilt.

moylop260:
FYI I rebuilt and now is green

AaronHForgeFlow (commented):
It works well, the creation and the modification of the Invoices and also when duplicating. Only one think. For For previous Invoices the sequence is 10 for all the lines (See picture). I'm thinking about a pre-init hook to force update old sale orders in the database. Do you agree with me?

image

serpentcs-dev1:
@aheficent Right yes, I'll do the needful.

AaronHForgeFlow (approved):
:+1: Functional + code :ok_hand:

serpentcs-dev1:
Now it works fine, the post-Init resets the sequence of the existing invoice lines. 

Thanks @aheficent  for your review!

screenshot from 2017-06-29 16-06-01

LoisRForgeFlow (approved):
LGTM :+1: 

Thanks for the work!

PR #283, opened 2017-06-28
[10.0][MIG] purchase_stock_picking_return_invoicing

MiquelRForgeFlow:
- The issue https://github.com/odoo/odoo/issues/13974 is still open, so I proceeded with the migration of this module.

- The fix https://github.com/OCA/OCB/pull/598 is also fixed on v10, so the minor hack for the unit test is not needed anymore.

- [x]  Includes cherry-pick 92e667436c3d6fcae850ec4d647211fc3e171847 (https://github.com/OCA/account-invoicing/pull/285).

- [x] Depends of https://github.com/OCA/purchase-workflow/pull/439.


purchase_stock_picking_return_invoicing/README.rst
```diff
@@ -20,76 +20,66 @@ In the purchase order you are able to display, for each line:
...
 =====
 
 Case 1: When you return to a supplier some products, and you have not yet
 received the bill from the supplier
 
 #. Go to *Purchases > Purchase > Purchase Orders > Create*.
-#. Choose a supplier and add a product whose *Invoicing Policy* is *On Received
-   quantities*, and input some quantity to purchase.
+#. Choose a supplier and add a product whose *Invoicing Policy* is
+   *Delivered quantities*, and input some quantity to purchase.
```
LoisRForgeFlow: What was wrong in the orignal sentence was 'Invoicing Policy', actually the field you have to modify is 'Control Purchase Bills' and set it to 'On received quantities'.

purchase_stock_picking_return_invoicing/README.rst
```diff
@@ -20,76 +20,66 @@ In the purchase order you are able to display, for each line:
...
    received and the returned quantity.
 
 Case 2: When you return to a supplier some products, and you have already
 received a bill from the supplier.
 
 #. Go to *Purchases > Purchase > Purchase Orders > Create*.
-#. Choose a supplier and add a product whose *Invoicing Policy* is *On Received
-   quantities*, and input some quantity to purchase.
+#. Choose a supplier and add a product whose *Invoicing Policy* is
+   *Delivered quantities*, and input some quantity to purchase.
```
LoisRForgeFlow: same as above.

purchase_stock_picking_return_invoicing/README.rst
```diff
@@ -20,76 +20,66 @@ In the purchase order you are able to display, for each line:
...
 .. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
    :alt: Try me on Runbot
-   :target: https://runbot.odoo-community.org/runbot/95/9.0
-
-Known issues / Roadmap
-======================
-
-* The functionality of return processing with refunds is being discussed with
-  Odoo here: https://github.com/odoo/odoo/issues/13974, so this addon may not
-  need to be migrated to v10.
```
LoisRForgeFlow: Have you checked this? It is still open, at least do not remove the reference.

purchase_stock_picking_return_invoicing/README.rst
```diff
@@ -20,76 +20,66 @@ In the purchase order you are able to display, for each line:
...
+   :target: https://runbot.odoo-community.org/runbot/95/10.0
 
-* The computation of the quantity invoiced is hacked to overcome an issue in
-  one of the tests of Odoo. See https://github.com/OCA/OCB/pull/598
```
LoisRForgeFlow: Has this fix been done in v10?

LoisRForgeFlow (commented):
I wouldn't describe this as standard migration :sweat_smile: There was some related issues opened in 9 and I think you have to say whether they have been solved or not and which changes they trigger in v10.

purchase_stock_picking_return_invoicing/README.rst
MiquelRForgeFlow: Yes, and the hack is removed as you can see here: https://github.com/OCA/account-invoicing/pull/283/commits/5097f62e495a5f192ee24df3b39ad61e180a31b9#diff-a03198f7ff7f7c9ef2454d443aff2b77L121.

purchase_stock_picking_return_invoicing/README.rst
MiquelRForgeFlow: Ok, I put the reference again. But although it is open, it is pretty dead.

pedrobaeza:
Please check CI status.

MiquelRForgeFlow:
@pedrobaeza CI status checked 😉

pedrobaeza:
@sergio-teruel please review

TimLai125:
Hi @pedrobaeza , May I know if this PR will be merged to the 10.0 branch soon? As one of my customers wishes to use this module in their 10.0 environment. Please let me know if anything I could give a hand on.

pedrobaeza:
@TimLai125 you can review this PR to check if all works correctly. Check https://odoo-community.org/page/review

mreficent:
I will rebase now just in case runbot doesn't have `purchase_open_qty` last fix.

TimLai125:
Hi @mreficent 

I am testing  and reviewing the module.
Before any technical features or behavior discussion, I would like to address an user-side opinion.
The module allows return of a delivery order links back to the purchase order line. Hence when you click the purchase order line in purchase order form view, the following form view will appear,
image

There are two records of "move_ids" however there is not any information about whether it is a delivery order / return which actually brings confusion to user.
I suggest the module also handle the open form view of the purchase order line as well.

castoreando:
I am testing and reviewing the module.

Button for Requests for Quotation appears in in budget it would be better to remove it

MiquelRForgeFlow:
@TimLai125, comment attended.

@castoreando, can you clarify your comment?

pedrobaeza:
@lreficent can you re-review this?

pedrobaeza:
@mreficent it would be good to increase test coverage

purchase_stock_picking_return_invoicing/__manifest__.py
```diff
@@ -0,0 +1,24 @@
...
+    "version": "10.0.1.0.0",
+    "category": "Purchases",
+    "website": "https://github.com/OCA/account-invoicing",
+    "author": "Eficent, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "depends": [
+        "purchase",
+        "purchase_open_qty",
```
pedrobaeza: This shouldn't depend on purchase_open_qty. Why this dependency?

pedrobaeza: I have removed temporarily the dependency for using this PR.

MiquelRForgeFlow:
@pedrobaeza I will try to increase the test coverage

purchase_stock_picking_return_invoicing/__manifest__.py
MiquelRForgeFlow: Without that dependency, the `qty_to_invoice` won't be visible in the view. So, I would like to have that dependency.

pedrobaeza: But that's not a good modularity, because the module makes other thing and includes that feature, but forcing to have that feature for only a visual field is not the correct way. You can do it for example with fields_view_get, or at the end, merge both modules together if the functionality is very tied to this.

pedrobaeza:
@mreficent are you already working on this?

purchase_stock_picking_return_invoicing/__manifest__.py
MiquelRForgeFlow: It's not only a visual field, both modules define the field <code>qty_to_invoice</code> with a compute. We discussed that before in https://github.com/OCA/account-invoicing/pull/285. In v9, both modules were already merged and thus we proceeded with the glue module solution (https://github.com/OCA/account-invoicing/pull/297). But now, this module is not merged yet, so we thought that it's better to add the dependency that migrating and maintaining a glue module.

pedrobaeza:
@mreficent @jbeficent I have cleanup the module, removing the dependency and being IMO very functional. Please check. For code review, the best is to see the final code instead of checking diff.

MiquelRForgeFlow:
@pedrobaeza So, with your changes, we should have still a glue module in v10

pedrobaeza:
I'm discussing right now that with @jbeficent. I don't see clear the dependency of this module for the values qty_to_receive and qty_to_invoice, but I don't know exactly which amount you want to reflect there.

pedrobaeza:
I have added https://github.com/OCA/account-invoicing/pull/283/commits/834034f15c916ec0cc792e69a1ed8c6ae1f6bc34 for avoiding to add lines with quantity = 0

pedrobaeza:
@mreficent @jbeficent what do you think about my changes? Can we move forward?

pedrobaeza:
I'm finally merging this as no feedback and it has been proved to work. The extra field that was added initially has been discarded, because this makes all the existing purchases to not be able to be refunded, and has been replaced by some compute fields. This has the limitation of not being able to receive/return/receive/return and so on. Only first level of return is taken into consideration, but it's not the usual flow. On v11, we will be able to take it into account.

PR #284, opened 2017-06-28
[ADD] Transfer discounts to separate invoice lines

StefanRijnhart:


account_invoice_discount_lines/views/account_invoice.xml
```diff
@@ -0,0 +1,31 @@
...
+    <data>
+        <record id="invoice_supplier_form" model="ir.ui.view">
+            <field name="name">Add applied discount field to invoice line in supplier invoice</field>
+            <field name="model">account.invoice</field>
+            <field name="inherit_id" ref="account.invoice_supplier_form"/>
+            <field name="arch" type="xml">
+                <xpath expr="//field[@name='invoice_line']/tree/field[@name='discount']"
+                       position="attributes">
+                    <attribute name="string">Discount (%)</attribute>
+                    <attribute name="name">discount_display</attribute>
```
legalsylvain, line 12: I'm not sure but do we need here to set a low priority for that view ? This line is changing ``discount`` into ``discount_display``, so if other modules inherit from the view and base the changes on the ``discount`` field, this will break the inheritance mechanism. don't you think ?

legalsylvain (commented):
Hi @StefanRijnhart.

Nice module. Thanks !
i'm asking if it's possible to make it working with the module account_invoice_triple_discount that is quite usefull for supplier invoices. (regular case of a general discount and an extra discount for some lines).

To fix it, i Imagine :
A) call a function "get_discount_value" that could be overloadable in a glue module account_invoice_triple_discount_lines.
B) Lazy test if model contains discount 2 and discount 3 fields. (avoid glue module).

what do you think ?

PR #285, opened 2017-06-30
[9.0][FIX][purchase_stock_picking_return_invoicing] minor fixes

MiquelRForgeFlow:
This PR updates the module a bit
- Update names.
- Update readme.
- Minor things.

### DIFF
--- a/purchase_stock_picking_return_invoicing/README.rst
+++ b/purchase_stock_picking_return_invoicing/README.rst
@@ -2,9 +2,9 @@
    :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
    :alt: License: AGPL-3
 
-=============================================
-Refund Returned Pickings from Purchase Orders
-=============================================
+=======================================
+Purchase Stock Picking Return Invoicing
+=======================================
 
 This module extends the functionality of purchase orders to better manage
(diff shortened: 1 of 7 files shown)


JordiBForgeFlow:
@mreficent Can you attend to the error in Travis?

mreficent:
Yes, the fix was done wrong 😵 

I will try to amend this.

mreficent:
Fixed. It depends now from other PR.

LoisRForgeFlow (approved):
Code review LGTM! Good job

pedrobaeza (changes requested):
You can't change the dependencies of an already merged module. You must create a compatible way to not depend on it or make a glue module

MiquelRForgeFlow:
Ok, @pedrobaeza . I didn't know that. I will proceed now with the glue module approach.

pedrobaeza:
But why can't you make both compatible without glue module?

MiquelRForgeFlow:
Both modules have `qty_to_invoice = fields.Float(compute="_compute_qty_to_invoice"` where this compute method in both are different and both `qty_to_invoice` fields are put in the same view and same place

pedrobaeza (approved):

PR #287, opened 2017-07-03
account_invoice_supplier_ref_unique Update french translation

Martronic-SA:


moylop260:
Use transifex to translate instead, please:
 - https://www.transifex.com/oca/OCA-account-invoicing-10-0/

PR #290, opened 2017-07-06
[10.0][FIX] account_invoice_refund_link: fix init hook by matching invoices and refunds only if in same company

Cedric-Pigeon:
I tried to install this module on a multi-company database and it fails due to the pre-init-hook.
It tried to match records from different companies.

### DIFF
--- a/account_invoice_refund_link/__manifest__.py
+++ b/account_invoice_refund_link/__manifest__.py
@@ -7,7 +7,7 @@
 {
     "name": "Link refund invoice with original",
     "summary": "Link refund invoice with its original invoice",
-    "version": "10.0.1.0.0",
+    "version": "10.0.1.0.1",
     "category": "Accounting & Finance",
     "website": "https://odoo-community.org/",
     "author": "Pexego, "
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):

pedrobaeza:
Please change module version

Cedric-Pigeon:
@pedrobaeza I don't understand

pedrobaeza:
On `__openerp__.py` file, please change the version of the module to one higher.

Cedric-Pigeon:
@pedrobaeza why ? it is just a small fix.

pedrobaeza:
But any change should be reflected on the version (the last number). See https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers

pedrobaeza:
Or someone can say: "stock_picking_invoice_link version 10.0.1.0.0 doesn't install", and another "I have the same version and it installs..."

Cedric-Pigeon:
I don't impact the module features, so from my point of view it is not required in this case

pedrobaeza:
That's why I ask for changing to 10.0.1.0.1, because it's a bug...

Cedric-Pigeon:
done

pedrobaeza:
Thanks! Now the second guy will say (both don't know about Git of course) "Update to 10.0.1.0.1 that installs correctly" :wink:

pedrobaeza:
Same for 9.0: https://github.com/OCA/account-invoicing/pull/291

MiquelRForgeFlow (approved):

Issue #109, opened 2015-10-14
Migration to version 9.0


pedrobaeza:
# Todo

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-9.0
# Modules to migrate
- [ ] account_group_invoice_lines
- [x] account_invoice_check_total - By @ThomasBinsfeld - #162 
- [ ] account_invoice_customer_ref_unique
- [ ] account_invoice_force_number
- [ ] account_invoice_line_description
- [ ] account_invoice_line_sort
- [x] account_invoice_merge - By @BT-hstoeckle - #209 
- [x] account_invoice_merge_payment - By @cubells - #232 
- [x] account_invoice_merge_purchase - By @cubells - #233 
- [ ] account_invoice_partner
- [ ] account_invoice_period_usability
- [x] account_invoice_pricelist - By @sergio-teruel - #211 
- [ ] account_invoice_rounding
- [ ] account_invoice_shipping_address - By @fshahy https://github.com/OCA/account-invoicing/pull/135
- [ ] account_invoice_template
- [ ] account_invoice_uom
- [ ] account_invoice_validation_workflow
- [x] account_invoice_supplier_ref_unique - By @ThomasBinsfeld - #161
- [ ] account_invoice_zero_autopay
- [x] account_payment_term_extension by @jcoux - #141
- [x] invoice_fiscal_position_update > account_invoice_fiscal_position_update - By @sergio-teruel - #223 
- [ ] product_customer_code_invoice
- [X] ~~sale_order_partial_invoice~~ Not needed anymore as possible in core
- [ ] stock_invoice_picking_incoterm
- [ ] stock_picking_invoicing - By @flotho merged in one module in #140
- [ ] stock_picking_invoicing_unified - By @flotho merged in one module in #140

ghost:
@pedrobaeza I will start by migrating account_invoice_shipping_address :)

pedrobaeza:
Cool, I have written it in the summary

flotho:
Hi, 

I'm about to rewrite the stock_pricking_invoice part. As there is no more invoicing defined neither on picking nor on moves everything has to be rewrite

pedrobaeza:
Good idea. You can incorporate https://github.com/OCA/account-invoicing/tree/8.0/stock_picking_invoicing_unified logic to this module.

flotho:
Hi @pedrobaeza 

Do you confirm my analysis, no more invoicing is available on stocks move, pickings etc...

[...]

PR #292, opened 2017-07-12
[9.0][NEW] account_invoice_margin: manage sale margin in invoices

sergio-teruel:
cc @Tecnativa


account_invoice_margin/models/account_invoice.py
```diff
@@ -0,0 +1,42 @@
...
+    )
+    purchase_price = fields.Float(
+        digits=dp.get_precision('Product Price'),
+        string='Cost',
+    )
+
+    @api.multi
+    @api.depends('purchase_price', 'price_subtotal')
+    def _compute_margin(self):
+        for line in self:
```
pedrobaeza: `for line in self.filtered(lambda x: x.invoice_id.type[:2] != 'in'):`

account_invoice_margin/README.rst
```diff
@@ -0,0 +1,63 @@
...
+=============
+
+This module also adds a security group.
+
+Usage
+=====
+
+To use this module, you need to:
+
+#. Go to user and active "Show Invoice Margin" in security options.
```
pedrobaeza: This is configuration, not usage.

sergio-teruel:
@pedrobaeza Changes done

account_invoice_margin/models/account_invoice.py
```diff
@@ -0,0 +1,50 @@
...
+            if line.invoice_id.type == 'out_refund':
+                tmp_margin *= -1
+            line.update({
+                'margin': tmp_margin,
+                'margin_percent': (tmp_margin / line.price_subtotal * 100.0 if
+                                   line.price_subtotal else 0.0),
+            })
+
+    @api.onchange('product_id', 'uom_id')
+    def _onchange_product_id_margin(self):
```
pedrobaeza: Call this onchange with `_onchange` +  `field` +  `module`  = `_onchange_product_id_account_invoice_margin`

account_invoice_margin/models/account_invoice.py
```diff
@@ -0,0 +1,50 @@
...
+    )
+
+    @api.multi
+    @api.depends('purchase_price', 'price_subtotal')
+    def _compute_margin(self):
+        for line in self.filtered(lambda x: x.invoice_id.type[:2] != 'in'):
+            tmp_margin = line.price_subtotal - (
+                line.purchase_price * line.quantity)
+            if line.invoice_id.type == 'out_refund':
+                tmp_margin *= -1
```
pedrobaeza: Better to make `tmp_margin = -tmp_margin`

account_invoice_margin/models/account_invoice.py
```diff
@@ -0,0 +1,50 @@
...
+            line.update({
+                'margin': tmp_margin,
+                'margin_percent': (tmp_margin / line.price_subtotal * 100.0 if
+                                   line.price_subtotal else 0.0),
+            })
+
+    @api.onchange('product_id', 'uom_id')
+    def _onchange_product_id_margin(self):
+        if self.invoice_id.type in ['out_invoice', 'out_refund']:
+            purchase_price = self.product_id.standard_price
```
pedrobaeza: This doesn't take into account if the price is with taxes included. Put it on known issues or change the calculation.

account_invoice_margin/report/__init__.py
```diff
@@ -0,0 +1,3 @@
+# -*- coding: utf-8 -*-
+
+import account_invoice_report
```
pedrobaeza: Always relative imports: `from . import account_invoice_report`

account_invoice_margin/report/account_invoice_report_view.xml
```diff
@@ -0,0 +1,129 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+<data>
+    <record id="view_account_invoice_report_pivot" model="ir.ui.view">
```
pedrobaeza: Why are you creating a totally new invoice analysis view instead of using existing one?

pedrobaeza (changes requested):
I'm missing here a margin summary at invoice level the same as you have it as sales order level.

legalsylvain:
hi. I just discover this module. It seems very similar to invoice_margin v8.0 module, don't you think ? 

https://github.com/OCA/account-invoicing/pull/275/files

regards.

pedrobaeza:
I prefer the new name

legalsylvain:
> I prefer the new name

Indeed, no worry to change the name from 8.0 invoice_margin to 9.0 account_invoice_margin, but could be great to base this module from the code of the existing one, cherry-picking commits, as usual.

pedrobaeza:
There's no commit history to import as there is no start from the other module. If there's a feature missing, we can include it, but there's no sense in adding a fake commit history.

account_invoice_margin/models/account_invoice.py
legalsylvain: I don't think it's necessary : that is the same behaviour as in the Odoo core. (sale_margin module)

legalsylvain:
> There's no commit history to import as there is no start from the other module. If there's a feature missing, we can include it, but there's no sense in adding a fake commit history.

indeed, and this module will have a better behaviour, due to the introduction of the onchange_product_id, allowing changing purchase_price manually. In the other hand, rewriting an existing module from scratch is not a good habit, and is not very fair regarding authorship.

legalsylvain (changes requested):
Hi @sergio-teruel,

thanks for this module.

As the field `account_invoice.purchase_price` is not computed (that is a good choice of implementation), you should initialize the value in a installation script. Otherwise, installing this module in an existing database will render bad values for existing invoices. (100% of margin for all invoices.)

kind regards.

pedrobaeza:
We didn't know about the other module and it's also not included on the migration issue (https://github.com/OCA/account-invoicing/issues/109). Sorry any way for not watching for that.

I'm thinking that maybe this module should be located in https://github.com/OCA/margin-analysis. What do you think?

account_invoice_margin/models/account_invoice.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+class AccountInvoiceLine(models.Model):
+    _inherit = 'account.invoice.line'
+
+    margin = fields.Float(
+        compute='_compute_margin',
+        digits=dp.get_precision('Product Price'),
+        store=True,
+        string='Margin',
```
legalsylvain, line 16: should not be multi='margin' ?

account_invoice_margin/models/account_invoice.py
```diff
@@ -0,0 +1,50 @@
...
+    )
+    margin_percent = fields.Float(
```
legalsylvain: could you add margin_percent field on AccountInvoice model, I think it's an important data for users to know the global margin of an invoices ?

legalsylvain (changes requested):
Hi @pedrobaeza, 

> We didn't know about the other module and it's also not included on the migration issue (#109). Sorry any way for not watching for that.

My bad.

> I'm thinking that maybe this module should be located in https://github.com/OCA/margin-analysis. What do you think?

I have no clear opinion on the matter.

Hi @sergio-teruel again,

I just improved the v8.0 module invoice_margin https://github.com/OCA/account-invoicing/pull/308 to make faster the installation. I think you can copy the whole file hooks.py for this PR.

I made the same improvement for pos_margin (https://github.com/OCA/pos/pull/207), and I reduce installation time from 6 days to 2 hours, so I think this worth the pain !

regards.

account_invoice_margin/models/account_invoice.py
legalsylvain: making margin negative for refund is understandable, but it is not in the spirit of Odoo refund, I think. All the values are positive in a refund, so, to be consistent, margin should be positive too.
Don't you think ? (propose to remove line 35 & 36)

pedrobaeza: The idea for making this is to get proper results on invoice analysis, so there's 2 options:

* Keep it as is.
* Have 2 fields: `margin` and `margin_signed`, both being computed here, but one used on invoice view and the other on the analysis.

account_invoice_margin/models/account_invoice.py
pedrobaeza: Yeah, but it doesn't convert that in the correct behavior.

legalsylvain: What about a field standard_price_vat_excl ? In a new module, and make account invoice margin and pos margin, depending on the new module ?

But are you sure that its possible to have a standard price vat incl ? I never seen any implementation with such settings. Could you explain when it can occur ?

Regards.

pedrobaeza: You can easily see it:

* Define a purchase tax vat included.
* Put it in a product.
* Define cost price (i.e. 100).
* Include that product on a purchase order

You will see that the standard price, it's considered as vat included, so the real one to use for margins is 86.96:

seleccion_009

The other way around can't happen as method `compute_taxes` already excludes taxes for sales price.

legalsylvain: OK ! Thanks. (I never used purchase tax vat included, but if it's a use case, it should be handled.

So what about a standard_price_vat_excl field, as this is what we need here ?

If agree, I can make a PR to do that.

pedrobaeza: I think we don't need an extra module. Just call <purchase_tax> compute taxes method with the cost price and the quantity and use the untaxed amount returned.

sergio-teruel:
Hi @legalsylvain I will review all your comments, I have returned today from my holidays... :smile:

account_invoice_margin/models/account_invoice.py
sergio-teruel: With the new Api if the method have the same name is not mandatory 'multi' argument

sergio-teruel:
Travis fails due to account_invoice_merge module..

pedrobaeza:
@legalsylvain is this OK for you then?

pedrobaeza (approved):

rafaelbn (approved):
:+1: thanks

@legalsylvain what do you think?

legalsylvain:
There is no hook to populate existing invoices. So margin will be bad.
What do you think ? 
I put a link on a sample of script upper in the review. Regards.

rafaelbn:
@sergio-teruel @pedrobaeza do we close this PR?

pedrobaeza:
We are not maintaining this module in 9.0 anymore. 11.0 version is in https://github.com/OCA/margin-analysis/tree/11.0/account_invoice_margin, so closing this. If someone is interested, he/she can take this PR as starting point.

PR #294, opened 2017-07-18
[ADD] account_invoice_triple_discount: New module

chienandalu:
Account Invoice Triple Discount
===============================

This module allows to have three successive discounts on every invoice line.

Usage
=====

Create a new invoice and add discounts in any of the three discount fields
given. They go in order of precedence so discount 2 will be calculated over
discount 1 and discount 3 over the result of discount 2. For example, let's
divide by two on every discount:

Unit price: 600.00 ->

  - Disc. 1 = 50% -> Amount = 300.00
  - Disc. 2 = 50% -> Amount = 150.00
  - Disc. 3 = 50% -> Amount = 75.00

You can also use negative values to charge instead of discount:

Unit price: 600.00 ->

  - Disc. 1 = 50% -> Amount = 300.00
  - Disc. 2 = -5% -> Amount = 315.00

cc @Tecnativa


pedrobaeza (approved):

account_invoice_triple_discount/README.rst
```diff
@@ -6,7 +6,7 @@
 Account Invoice Triple Discount
 ===============================
 
-This module allows to have three successive discounts on every invoice line.
+This module allows to have three consecutive discounts on each invoice line.
```
chienandalu: Well, I'd say _successive_ is the proper term here (https://wordpandit.com/successive-percentage-change/).

pedrobaeza: OK then

luismontalba (approved):
Tested on runbot

Issue #194, opened 2016-10-11
Migration to version 10.0


pedrobaeza:
# Todo

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-10.0
# Modules to migrate
- [x] account_group_invoice_lines - By @alexis-via - #208 
- [x] account_invoice_blocking - By @adrienpeiffer - #197 
- [x] account_invoice_check_total - By @adrienpeiffer - #195 
- [ ] account_invoice_customer_ref_unique - By @aheficent - #257
- [x] account_invoice_force_number - By @tafaRU  - #242
- [x] account_invoice_line_description - By @luismontalba - #274 
- [ ] account_invoice_line_sort
- [x] account_invoice_merge - By @lreficent - #238 
- [x] account_invoice_merge_payment - By @benwillig - #264 
- [x] account_invoice_merge_purchase - By @adrienpeiffer - #251 
- [ ] account_invoice_partner
- [ ] account_invoice_period_usability
- [x] account_invoice_pricelist - By @rvalyi - #342 
- [x] account_invoice_pricelist_sale - By @rvalyi - #343 
- [x] account_invoice_refund_link - By @luismontalba - #254 
- [x] account_invoice_refund_option - ~~By @NeoPolus - #229~~ Not needed in v10
- [x] account_invoice_rounding - By @grindtildeath - #212
- [x] ~~account_invoice_shipping_address~~ - Feature included in core
- [x] account_invoice_supplier_ref_unique - By @adrienpeiffer - #192 
- [x] account_invoice_supplierinfo_update - By @mourad-ehm - #252
- [x] account_invoice_supplierinfo_update_discount - By @legalsylvain - #468
- [ ] account_invoice_template - By @jzahid - #250 
- [ ] account_invoice_uom
- [ ] account_invoice_validation_workflow
- [ ] account_invoice_zero_autopay
- [x] account_payment_term_extension - By @alexis-via - #193 
- [x] invoice_fiscal_position_update > account_invoice_fiscal_position_update - By @alexis-via - #191 
- [ ] product_customer_code_invoice - By @angedras - #267 
- [ ] purchase_batch_invoicing - By @Montxu - #456 
- [ ] purchase_stock_picking_return_invoicing
- [ ] sale_order_partial_invoice
- [x] sale_timesheet_invoice_description - By @pedrobaeza - #300 
- [ ] stock_invoice_picking_incoterm
- [ ] stock_picking_invoicing

[...]

PR #303, opened 2017-08-14
[10.0] [ADD] account_invoice_shipping_address

nikul-serpentcs:
Invoice Shipping Address
========================

This module adds a shipping address field to the invoice, also it fills it in sale and picking invoicing


pedrobaeza:
This module is not needed anymore. See https://github.com/OCA/account-invoicing/issues/194#issuecomment-300507617.

PR #302, opened 2017-08-10
[ADD] Adds account_outstanding_payment

daramousk:
module. Backporting functionality from version 9. Allow an invoice to be paid using already existing customer credit

PR #305, opened 2017-08-18
Removed strange directories following issue #302

elicoidal:


elicoidal:
@pedrobaeza

pedrobaeza (approved):
Fast-tracking

pedrobaeza:
The second part, which is removing resources in Transifex, is also done.

elicoidal:
Thanks @pedrobaeza

PR #308, opened 2017-08-23
[IMP][8.0] invoice_margin module 

legalsylvain:
* ADD : margin_percent fied on account_invoice model;
* IMP : initialize data (cost price on invoice lines), based on purchase price in product_price_history

CC : @ loxamir, @bealdav , @moylop260 , @mreficent


MiquelRForgeFlow (approved):
LGTM :+1: 

Can you explain why you want the margin_percent? And why still working in v8? 😄

invoice_margin/views/view_account_invoice.xml
```diff
@@ -22,8 +23,9 @@ License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
         <field name="inherit_id" ref="account.invoice_form"/>
         <field name="arch" type="xml">
             <xpath expr="//field[@name='amount_untaxed']/.." position="after">
-                <group name="margin">
+                <group name="margin" string="Margin">
```
pedrobaeza, line 26: Why not putting this on the same subtotals block?

legalsylvain:
> Can you explain why you want the margin_percent?

@mreficent, thanks for your review ! 
In an end users PoV, on account_invoice model, margin field is interesting when you sum the margin of many invoices, to know what you won. (by month, by salesman, etc...)
But for a single invoice, if you show margin : 30€, it doesn't talk very good to the user, because 30€ over 200€ of sale is not the same as 30€ over 150€. For that reason, showing 15% is a good information. It allows users to make some discount on a line, checking that the global margin_percent is not under a given value. You can then filter easely in the search view, to check "which invoices are under X % of margin". 

Hope I'm understandable on that point.

> And why still working in v8? :smile:

;-)

invoice_margin/views/view_account_invoice.xml
legalsylvain: because of consistency. sale_margin (of Odoo SA) works like that.
https://github.com/odoo/odoo/blob/10.0/addons/sale_margin/views/sale_margin_view.xml#L10

legalsylvain:
I just refactored the module to improve it like on https://github.com/OCA/pos/pull/207 PR for pos_margin, to take into account remarks from @pedrobaeza and @sebastienbeau about computed field, and how to compute purchase_price for existing invoices.

@mreficent, I think you could take the hook file in your PR. (https://github.com/OCA/account-invoicing/pull/292) (or maybe cherry picking the whole module will be now faster.)

MiquelRForgeFlow:
I think you are confused, that PR is not mine 😅

legalsylvain:
How ! Sorry. CC : @sergio-teruel

regards.

invoice_margin/hooks.py
```diff
@@ -0,0 +1,84 @@
...
+
+def _create_column(cr, table_name, column_name, column_type):
+    req = "ALTER TABLE %s ADD COLUMN %s %s" % (
+        table_name, column_name, column_type)
+    cr.execute(req)
+
+
+def pre_init_hook(cr):
+    _logger.info(
+        "Compute account_invoice_line.purchase_price for existing lines")
```
moylop260, line 19: The creation of a new computed field consider re-compute the existing lines currently.
So this code is not longer required.

moylop260: if you want use sql to faster execution you should add a docstring to justify this method

legalsylvain: > The creation of a new computed field consider re-compute the existing lines currently.
So this code is not longer required.

purchase_price is not a computed field. (like in #292). So it has to be prepopulated by SQL script.
https://github.com/OCA/account-invoicing/pull/308/files#diff-a08017a90fbce7fdbe067b22197b76f9R24

moylop260 (approved):

invoice_margin/hooks.py
legalsylvain: is it now ok ? @moylop260.

legalsylvain: sorry, i did'nt see your approval.

moylop260:
FYI lint errors of the travis result of this PR worked from: https://github.com/OCA/account-invoicing/pull/354

Issue #310, opened 2017-08-31
[RFC][account_invoice_line_discount_fixed_amount]


JordiBForgeFlow:
This is an RFC proposing an approach to handing fixed amount discounts in invoices. 
See https://github.com/OCA/sale-workflow/issues/521 for fixed amount discount in sales order lines.

Account Invoice Line Discount Fixed Amount
==================================
* Introduces a 'Discount (Fixed amount)'
* A user cannot enter both a 'Discount (%)'  and 'Discount (Fixed amount)'  (constraint).
* The total invoice amount is computed according to whatever discount type was applied.
* The total invoice tax amount is computed according to whatever discount type was applied.
* The invoice form shows the 2 discount types.

Configuration
------------------
* Go to 'Invoicing / Settings' and check the box "Use fixed amount discounts'. This will ensure that all employees have access to the new field by adding them to group 'Usability / Use fixed amount discounts'.

In order to implement this, we 'd follow a similar approach as https://github.com/OCA/account-invoicing/blob/10.0/account_invoice_triple_discount/models/account_invoice.py

pedrobaeza:
@sergio-teruel shouldn't we have this for product prices (percentage or fixed price)?

JordiBForgeFlow:
@pedrobaeza What do you mean for product prices? What product is defined as percentage?

pedrobaeza:
For PrestaShop, you can define sales price as a fixed amount or percentage of cost price. The problem with this comes from rounding errors.

JordiBForgeFlow:
Well, how do you define discounts in prestashop? I am dealing with an integration with Drupal Commerce, where the discount comes as a fixed amount. In the magento connector they have a similar issue, but they translate a fixed amount discount to a % discount in Odoo. https://github.com/guewen/connector-magento/blob/2eb2310476e4a7e9035435e424179b15d1cf458e/connector_magento/models/sale_order/importer.py#L639 This, however, can easily translate into discrepancies between the ecommerce order and the odoo one.

pedrobaeza:
I don't remember if it was PrestaShop or other source, but it's something similar. You will have rounding issues any way, but that's why sale_exception exists.

JordiBForgeFlow:
@pedrobaeza Why do you mean that sale_exception occurs? In the end you have rounding errors between Odoo and the ecommerce, so I cannot see where sale_exception would help.

On the other side having the ability to add a fixed discount is often needed.

[...]

PR #311, opened 2017-09-01
[9.0][ADD] account_invoice_fixed_discount

LoisRForgeFlow:
Attend issue https://github.com/OCA/account-invoicing/issues/310

Account Invoice Fixed Discount
======================

This module extends the functionality of Invoicing to allow you to apply fixed
amount discounts at invoice line level.

The module also extends the invoice report to show fixed discount.


JordiBForgeFlow (approved):
Great one!

AaronHForgeFlow (approved):
:+1: Tested on runbot

MiquelRForgeFlow (approved):
LGTM :+1:

account_invoice_fixed_discount/models/account_invoice.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+class AccountInvoice(models.Model):
+    _inherit = "account.invoice"
+
+    @api.multi
+    def get_taxes_values(self):
+        self.ensure_one()
+        vals = {}
+        for line in self.invoice_line_ids.filtered(lambda l: l.discount_fixed):
```
moylop260: filtered('discount_fixed')

moylop260 (changes requested):
Review the beta checks please
https://travis-ci.org/OCA/account-invoicing/jobs/271739620#L743

lreficent:
@moylop260 the method `_compute_price` is api.one in standard odoo: https://github.com/odoo/odoo/blob/9.0/addons/account/models/account_invoice.py#L1103. The other changes has been made.

LoisRForgeFlow:
@moylop260 the method _compute_price is api.one in standard odoo: https://github.com/odoo/odoo/blob/9.0/addons/account/models/account_invoice.py#L1103. The other changes have been made.

account_invoice_fixed_discount/models/account_invoice.py
```diff
@@ -0,0 +1,76 @@
...
+            if line.discount and line.discount_fixed:
+                raise ValidationError(
+                    _("You can only set one type of discount per line."))
+
+    @api.one
+    @api.depends('price_unit', 'discount', 'invoice_line_tax_ids', 'quantity',
+                 'product_id', 'invoice_id.partner_id',
+                 'invoice_id.currency_id', 'invoice_id.company_id',
+                 'discount_fixed')
+    def _compute_price(self):
```
moylop260, line 64: @pedrobaeza @lasley 
What do you think about a inherit of a computed method?

pedrobaeza: Well, we do something similar in https://github.com/OCA/account-invoicing/blob/7a2aa5c6ed53cb0e9d853219b04e7224a93bd2ee/account_invoice_triple_discount/models/account_invoice.py#L49, and these cases inheritance will fail, but there's no other way to do it. You can mention the incompatibility in the README.

LoisRForgeFlow: what do you mean? is it impossible to have installed and working fine both this module and yours?

pedrobaeza: Yeah, as both modifies the price_unit as trick for having the rest without taking into account possible previous modifications. You can't have a fixed price discount and also second and third discount. If you see we can find a way to conciliate both....

LoisRForgeFlow: We could try to add a hook module which we both can inherit into our modules. Do you think that would be possible?

LoisRForgeFlow:
@pedrobaeza @moylop260 I added a warning in the README about the incompatibility with account_invoice_triple_discount. Can we move forward?

moylop260 (approved):

LoisRForgeFlow:
Rebased and squashed.

PR #314, opened 2017-09-21
[10.0][FIX] account_invoice_fiscal_position_update: On invoice creati…

rousseldenis:
…on, tax mapping could be wrong set

Steps to reproduce :

Create a account tax mapping depending on fiscal position
Create a purchase order with products and a partner that have the first fiscal position
From purchase order, go to Vendor Bills, then Create

The generated lines come with product taxes instead of the mapped taxes on fiscal position.

### DIFF
--- a/account_invoice_fiscal_position_update/__manifest__.py
+++ b/account_invoice_fiscal_position_update/__manifest__.py
@@ -6,7 +6,7 @@
 
 {
     'name': 'Invoice Fiscal Position Update',
-    'version': '10.0.1.0.0',
+    'version': '10.0.1.0.1',
     'category': 'Accounting & Finance',
     'license': 'AGPL-3',
     'summary': 'Changing the fiscal position of an invoice will auto-update '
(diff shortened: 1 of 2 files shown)


alexis-via (approved):
I confirm the scenario of the bug and I confirm that this PR fixes it. I just don't understand why we need to do that to make it work...

rousseldenis:
@alexis-via That's because in that case, the entry point of onchanges is 'purchase_id' field (from default context in action) and partner_id is void. But the 'fiscal_position_change' is called and fiscal_position_id is void too.

So, as the onchanges functions are not called twice (in standard Odoo mechanism), we force that second call.

pedrobaeza:
Isn't the method going to be called twice in the rest of the cases?

rousseldenis:
@pedrobaeza Only if fiscal position value is different on partner_id change.

rousseldenis:
@sbidoul @lmignon

pedrobaeza (approved):
But please change module version number

rousseldenis:
@pedrobaeza Ok, I've just thought of that !

rousseldenis:
Done

adrienpeiffer (approved):

PR #324, opened 2017-10-24
[10.0][fix] travis new flake8 errors (E722 do not use bare except)

grindtildeath:


leemannd (approved):
Approved, but is it worth a minor version number increase?

pedrobaeza (approved):
Fast-tracking this

PR #330, opened 2017-11-14
[8.0] [FIX] account_invoice_supplier_ref_unique

chafique-delli:
fix the search domain of the supplier invoice number in multi-company

### DIFF
--- a/account_invoice_supplier_ref_unique/README.rst
+++ b/account_invoice_supplier_ref_unique/README.rst
@@ -22,6 +22,7 @@ Contributors
 * Marc Cassuto <marc.cassuto@savoirfairelinux.com>
 * Mathieu Benoit <mathieu.benoit@savoirfairelinux.com>
 * Alexis de Lattre <alexis.delattre@akretion.com>
+* Chafique Delli <chafique.delli@akretion.com>
 
 Maintainer
 ----------
--- a/account_invoice_supplier_ref_unique/account_invoice.py
+++ b/account_invoice_supplier_ref_unique/account_invoice.py
@@ -44,7 +44,7 @@ class AccountInvoice(models.Model):
                  '=ilike',
                  self.supplier_invoice_number),
                 ('id', '!=', self.id),
-                ('company_id', '=', self.company_id.id)
+                ('state', '!=', 'cancel')
             ])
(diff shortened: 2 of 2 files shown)


legalsylvain:
Hi @chafique-delli ,

I don't understand when you want to have the same reference for the same partner in many companies. Could you define the use case ?

bealdav (approved):

chafique-delli:
Hi @legalsylvain ,
case of an inter-company automatic invoicing (account_invoice_inter_company module):
- in a company A (supplier) have created a customer invoice with a partner company B.When validating the customer invoice, this is created in company B, a supplier invoice with partner = company A and supplier_invoice_number = customer invoice number of the company A.
- If we decide to cancel the customer invoice in company A to put it back in draft and modify the partner, then the supplier invoice in company B will be canceled but always keep the supplier_invoice_number.
- when we try to validate the customer invoice that has been updated, well we can not because of the constraint _check_unique_supplier_invoice_number_insensitive that will see that there is a supplier invoice that has the same supplier_invoice_number.

I do not know if I was clear in my explanation?

legalsylvain:
Ow I see ! 
So in that case, should not be more generic to keep the module as it, and add instead of the code of the current PR, a domain on state != 'cancel' ?

I mean the problem you mention is not a multicompany problem, but a cancel -> recreate problem, doesn't it ?

regards.

chafique-delli:
@legalsylvain ,
Ok thank you for your review.
Indeed, you're right, finally, I will adapt the account_invoice_inter_company module to better manage the supplier_invoice_number when we cancel an invoice.

Regards.

legalsylvain:
Well, in fact I think that this PR should just be modified into 
[('state', '!=', 'cancel')] but this is usefull, seing the use case @bealdav has explained. (and not only on multicompany context, it's a general problem)

legalsylvain (approved):
:+1: . code review, no test. Thanks !

chafique-delli:
@legalsylvain ,
It's good I changed the domain as suggested.

hparfr (approved):

mourad-ehm (approved):

feketemihai (approved):

PR #335, opened 2017-11-27
8.0 port sale order partial invoice

jcarlosmontoya:
8.0 port sale order partial invoice


rubencabrera:
👍 
This is now in production servers.

Check the last commit's author so the CLA bot is also ok.

PR #338, opened 2017-11-28
Remove broken link after module in question was renamed

thomaspaulb:
Hi, our build failed because of this.


pedrobaeza (approved):
Fast-tracking

hbrunn:
@sbidoul (I *think* you're the pip guy in the OCA, if that's wrong, please tell me) should the script that generates the setup folders also clean them up?

sbidoul:
@hbrunn I'll look into it. I created https://github.com/acsone/setuptools-odoo/issues/21 to track this.

PR #342, opened 2017-12-15
[10][MIGR] account_invoice_pricelist

rvalyi:


pedrobaeza:
Why haven't you used the migration guide, but changing 9.0 by 8.0? You need to preserve commit history

rvalyi:
@pedrobaeza my bad. Somebody told me version 8.0 was the most advanced and I admitted it was true without checking if it was ported to 9.0 before... I'll review your 9.0 port and start from here again sorry.

rvalyi:
@pedrobaeza I should say I have been induced in error because I checked too quickly on this addons-table https://github.com/OCA/account-invoicing/tree/10.0#unported-addons where it says last version of 
account_invoice_pricelist was 8.0.1.0.0 and I assumed the table was correct.

is it a known issue this table can miss that kind of information? and also is there a reason we don't we have such a table for v11?

pedrobaeza:
> @pedrobaeza I should say I have been induced in error because I checked too quickly on this addons-table https://github.com/OCA/account-invoicing/tree/10.0#unported-addons where it says last version of
account_invoice_pricelist was 8.0.1.0.0 and I assumed the table was correct.

This is because the 10.0 branch was created before the module was merged in 9.0

> is it a known issue this table can miss that kind of information? and also is there a reason we don't we have such a table for v11?

I removed that tables as part of the 11.0 branch creation for avoiding this kind of situations, as now if the module is not there, it means is not migrated. And even more, there's no possibility of having outdated sources hosted in the repo. That was one of the advantages I enumerated when announcing this change of paradigm if you remember.

account_invoice_pricelist/model/account_invoice.py
```diff
@@ -1,26 +1,8 @@
...
-#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
-#
-##############################################################################
-
-from openerp import models, fields, api
+# -*- coding: utf-8 -*-
+# © 2014-2016 GRAP <http://www.grap.coop>.
+# © 2017 Therp BV <http://therp.nl>.
+# License AGPL-3.0 or later <http://www.gnu.org/licenses/agpl.html>.
+from openerp import api, fields, models
```
bealdav, line 5: s/openerp/odoo

account_invoice_pricelist/model/__init__.py
```diff
@@ -1,3 +1,6 @@
-# -*- encoding: utf-8 -*-
-
+# -*- coding: utf-8 -*-
+# © 2015-2016 GRAP <http://www.grap.coop>.
+# © 2017 Therp BV <http://therp.nl>.
+# License AGPL-3.0 or later <http://www.gnu.org/licenses/agpl.html>.
```
bealdav, line 4: Not blocking for me but here there is no content (just import), then authors and license can be removed (it's just my opinion ;-) )

account_invoice_pricelist/model/account_invoice.py
```diff
@@ -30,29 +12,38 @@ class AccountInvoice(models.Model):
...
-                if partner._model._columns.get(
-                        'property_product_pricelist_purchase', False):
-                    pricelist_id =\
-                        partner.property_product_pricelist_purchase.id
-        res['value']['pricelist_id'] = pricelist_id
-        return res
+                if hasattr(self.partner_id,
+                           'property_product_pricelist_purchase'):
+                    self.pricelist_id =\
+                        self.partner_id.property_product_pricelist_purchase
```
bealdav, line 30: Partner model has attribute property_product_pricelist_purchase which is defined in product module. 
Why a condition here ?

account_invoice_pricelist/view/account_invoice_view.xml
```diff
@@ -0,0 +1,61 @@
...
+        <!-- Model: account.invoice -->
+        <record id="view_account_invoice_customer_form" model="ir.ui.view">
+            <field name="model">account.invoice</field>
+            <field name="inherit_id" ref="account.invoice_form"/>
+            <field name="arch" type="xml">
+                <field name="state" position="before">
+                    <button
+                        type="object"
+                        name="button_update_prices_from_pricelist"
+                        string="Update prices from product and/or pricelist"
```
bealdav, line 19: This button string is too long in the interface (mobile render !? etc)
=> 'Update prices' is sufficient I think

account_invoice_pricelist/model/account_invoice_line.py
```diff
@@ -0,0 +1,39 @@
+# -*- coding: utf-8 -*-
+# © 2015-2016 GRAP <http://www.grap.coop>.
+# © 2017 Therp BV <http://therp.nl>.
+# License AGPL-3.0 or later <http://www.gnu.org/licenses/agpl.html>.
+from openerp import models, api
```
bealdav, line 5: s/openerp/odoo

bealdav (changes requested):

account_invoice_pricelist/model/__init__.py
legalsylvain: :+1:

account_invoice_pricelist/view/account_invoice_view.xml
legalsylvain: Maybe "Apply Pricelist" ?

pedrobaeza: Or "Update prices"?

legalsylvain:
Hi @rvalyi, the res_groups.yml demo file is not useless for me. By this way, it is more easy to test the module. (On runbot, mainly) to avoid functional users to have to set this settings each time.

For that reason, I regurlarly add a demo file, adding admin user to the groups, required for the module. 
In that module, if users is not part of pricelist group, it will not have the possibility to test fully the module. (but maybe some groups changed since V8)

If you're agree, to add it again, maybe add a xml file, as yml is deprecated.

Thanks for porting this module.

legalsylvain (changes requested):
Hi @rvalyi, more deeply review.

1) minor remark, on readme file, I think runbot link is usefull.

2) I think there are a problem in your code. you don't seems to overload the function _compute_price()
This function should be overloaded to allow to compute price, based on the correct pricelist. Don't you think ? otherwise, changing quantity will not recompute the price, based on the correct pricelist.

3) you can add you in creditor section ! This work is not a /s/openerp/odoo work, lot of changes since V8.

rvalyi:
@legalsylvain I'll redo the migration starting from the 9.0 branch again anyway. Thanks for your comments, I'll take them into account.

angelmoya:
Hi @rvalyi, did you redo this migration? Maybe I could do it if you don't plan to do.

Thanks

rvalyi:
Hello guys, just wanted to say, I'm on it today finally

rvalyi:
Here is the new PR superseding this one
https://github.com/OCA/account-invoicing/pull/368

account_invoice_pricelist/__manifest__.py
```diff
@@ -1,41 +1,20 @@
...
-#
-##############################################################################
-
+# -*- coding: utf-8 -*-
+# © 2015-2016 GRAP <http://www.grap.coop>.
+# © 2015-2017 Therp BV <http://therp.nl>.
+# License AGPL-3.0 or later <http://www.gnu.org/licenses/agpl.html>.
 {
     'name': 'Account - Pricelist on Invoices',
     'version': '8.0.1.0.0',
```
NL66278, line 7: Version number should be updated to 10.0 as well

account_invoice_pricelist/view/account_invoice_view.xml
```diff
@@ -0,0 +1,61 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+    © 2015-2016 GRAP <http://www.grap.coop>
+    © 2017 Therp BV <http://therp.nl>.
+    License AGPL-3.0 or later <http://www.gnu.org/licenses/agpl.html>.
+-->
+<openerp>
```
NL66278, line 7: s/openerp/odoo and ```<data>``` is deprecated.

NL66278 (commented):
:+1: LGTM. Some suggestions for style changes and small fixes. See comments.

PR #344, opened 2017-12-20
[IMP] Allow to search invoice based on the no follow up field

benwillig:


account_invoice_blocking/migrations/10.0.1.0.1/post-migration.py
```diff
@@ -0,0 +1,74 @@
...
+    current_inv_id = None
+
+    blocked_by_invoice_ids = {}
+    invoice_ids_to_block = []
+
+    for row in rows:
+        inv_id = row[0]
+
+        if inv_id != current_inv_id and current_inv_id is not None:
+            if all(blocked_by_invoice_ids[current_inv_id]):
```
adrienpeiffer, line 42: This condition will never called for the last invoice block

adrienpeiffer (changes requested):

account_invoice_blocking/migrations/10.0.1.0.1/post-migration.py
benwillig: I added the missing condition at the end of the loop.

adrienpeiffer (approved):
👍

pedrobaeza (approved):

PR #347, opened 2018-01-02
[11.0][MIG] account_invoice_supplier_ref_unique

astirpe:
Porting of `account_invoice_supplier_ref_unique` to V11


pedrobaeza:
Isn't this check already in core?

astirpe:
@pedrobaeza it seems to me that this module is still needed. The 'Vendor invoice number' is an information not present in the account module.

pedrobaeza:
Are you sure? Then, why this only depends on `account`?

astirpe:
@pedrobaeza sorry, I think I'm not understanding your point. This module adds a field on the vendor invoice, so there's no need of extra dependencies. Or maybe I'm missing something?

pedrobaeza:
Ah, sorry, it was me then. The module name has confused me. I thought it was for having an unique supplier invoice number, but on the existing field.

ThijsvOers (approved):
I have tested the 'uniqueness the invoice supplier ref field and it works correct (tested internally and on runbot)

ThijsvOers:
Hi, when the supplier invoice (with status 'open') is for some reason not correct, you cannot change it any more. You can choose to reconcile  the 'wrong' invoice and immediately create a new number. This works fine except for the reference number. Since it must be unique i cannot use it any more.

My proposal would be that in the case of creating a new invoice in the way described the supplier invoice number is copied to the new number, and deleted in the reconciled 'wrong' invoice.

melroy89 (approved):

account_invoice_supplier_ref_unique/models/account_invoice.py
```diff
@@ -0,0 +1,70 @@
+# Copyright 2016 Acsone
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models, fields, api, _
```
congdpt: Imports should be sorted Alphabetically
from odoo import api, fields, models, _

congdpt (approved):

account_invoice_supplier_ref_unique/models/account_invoice.py
```diff
@@ -0,0 +1,70 @@
...
+    @api.constrains('supplier_invoice_number')
+    def _check_unique_supplier_invoice_number_insensitive(self):
+        """
+        Check if an other vendor bill has the same supplier_invoice_number
+        and the same commercial_partner_id than the current instance
+        """
+        if self.supplier_invoice_number and\
+                self.type in ('in_invoice', 'in_refund'):
+            same_supplier_inv_num = self.search([
+                ('commercial_partner_id', '=', self.commercial_partner_id.id),
```
moylop260, line 27: What about `company_id`?

account_invoice_supplier_ref_unique/models/account_invoice.py
```diff
@@ -0,0 +1,70 @@
...
+                ('type', 'in', ('in_invoice', 'in_refund')),
+                ('supplier_invoice_number',
+                 '=ilike',
+                 self.supplier_invoice_number),
+                ('id', '!=', self.id), ])
```
moylop260: `limit=1` in order to avoid use index [0] in the line 38 `same_supplier_inv_num[0].supplier_invoice_number`

account_invoice_supplier_ref_unique/models/account_invoice.py
astirpe: Shouldn't supplier_invoice_number be unique, regardless of the company_id?
In a multicompany environment, if you create two invoices in two different companies with the same vendor and the same supplier_invoice_number, you should still raise the ValidationError.
Did I understand your point correctly? Is this the case you are talking about?

account_invoice_supplier_ref_unique/models/account_invoice.py
astirpe: Yes indeed! Thank you.

astirpe:
@moylop260 are you ok with my answer? Can we merge this one?

PR #353, opened 2018-01-12
[10.0][ADD] account_invoice_fixed_discount

LoisRForgeFlow:
Migration based on https://github.com/OCA/account-invoicing/pull/311.

Account Invoice Fixed Discount
==============================

This module extends the functionality of Invoicing to allow you to apply fixed
amount discounts at invoice line level.

The module also extends the invoice report to show fixed discount.

Installation
============

**Warning**: This module is incompatible with
``account_invoice_triple_discount`` which also belongs to `OCA/account-invoicing
<https://github.com/OCA/account-invoicing>`_.

Configuration
=============

To configure this module, you need to:

#. Go to * Sale > Configuration > Settings*.
#. In the *Discount* option select *Allow discounts on sales order lines*.

Usage
=====

To use this module, you need to:

#. Go to *Invoicing*.
#. Create a Invoice and specify a fixed discount in a line.


account_invoice_fixed_discount/models/account_invoice.py
```diff
@@ -0,0 +1,76 @@
...
+        for line in self:
+            if line.discount and line.discount_fixed:
+                raise ValidationError(
+                    _("You can only set one type of discount per line."))
+
+    @api.one
+    @api.depends('price_unit', 'discount', 'invoice_line_tax_ids', 'quantity',
+                 'product_id', 'invoice_id.partner_id',
+                 'invoice_id.currency_id', 'invoice_id.company_id',
+                 'discount_fixed')
```
feketemihai: @lreficent Missing sone dependencies: https://github.com/odoo/odoo/blob/10.0/addons/account/models/account_invoice.py#L1177

account_invoice_fixed_discount/models/account_invoice.py
```diff
@@ -0,0 +1,76 @@
...
+    def _compute_price(self):
+        prev_price_unit = self.price_unit
+        prev_discount_fixed = self.discount_fixed
+        price_unit = self.price_unit - self.discount_fixed
+        self.update({
```
feketemihai: Isn't it easier to inherit the function and check if you have discount_fix to make other calculation, else super, because as a performance, instead of writing on 2 fields price_subtotal, price_subtotal_signed you're writing on 6 fields..

account_invoice_fixed_discount/models/account_invoice.py
```diff
@@ -0,0 +1,76 @@
...
+
+class AccountInvoiceLine(models.Model):
+    _inherit = "account.invoice.line"
+
+    discount_fixed = fields.Float(
+        string="Discount (Fixed)",
+        digits=dp.get_precision('Product Price'),
+        help="Fixed amount discount.")
+
+    @api.onchange('discount')
```
feketemihai, line 41: Why not leaving both fields, why this restriction, logically you can benefit on both into one promotion, let's say..., it will not breake anything, just when you compute the price to take the discount in calculation

feketemihai (changes requested):

account_invoice_fixed_discount/models/account_invoice.py
LoisRForgeFlow: it's needed to be done this way to respect the odoo calculations. However we could add a quick return when there is no `discount_fixed`.

account_invoice_fixed_discount/models/account_invoice.py
LoisRForgeFlow: To not overcomplicate it and to add the fixed_discount with the least impact/extensions on the standard as it was just what I needed.

account_invoice_fixed_discount/models/account_invoice.py
LoisRForgeFlow: sure, will add them

account_invoice_fixed_discount/models/account_invoice.py
feketemihai: @lreficent Ok, but can you please added in readme roadmap, to know about it..

LoisRForgeFlow: yes, done!

feketemihai (approved):

rafaelbn:
@lreficent Odoo make this in the core. Which is the difference with this module?

LoisRForgeFlow:
@rafaelbn As far as I see Odoo allows discount per SO line based on a percent, this module is to apply a fixed quantity discount to specific SO lines. I cannot find the way to do that in Odoo core. Could you point it to me?

jackjack82 (approved):
The flow is tested and seems good. approved.

pedrobaeza:
Please squash before merging

LoisRForgeFlow:
@pedrobaeza Squash done

PR #360, opened 2018-02-22
[10.0] account_group_invoice_line One fix and code simplification

alexis-via:
We are not supposed to use self.journal_id on an @api.model method ; with this change, it fixes the use of action_invoice_open() on multi-recordset

I removed the inherit of the method inv_line_characteristic_hashcode() which was in fact useless: line_get_convert() is called before inv_line_characteristic_hashcode() and line_get_convert() set the product_id key to False, so there's no point in removing it from the hash (and it fact, it was poping the wrong position !)

### DIFF
--- a/account_group_invoice_line/__manifest__.py
+++ b/account_group_invoice_line/__manifest__.py
@@ -7,7 +7,7 @@
 
 {
     'name': 'Account Group Invoice Lines',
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):
Please change module version number

bguillot (approved):

PR #361, opened 2018-03-02
8.0 backport account invoice triple discount

legalsylvain:
Backport module from 10.0 to 8.0

single commit to review : 2ffab20  

main changes : 
the tax computation is done differently in 8.0 and 10.0 module. So : 
- calling button_reset_taxes in test file.
- removing inheritance from account.invoice and inherit account.invoice.tax instead.

thanks for your review.


StefanRijnhart (approved):
Diff looks clean and the tests are green.
And because that rhymes, the approve button chimes!

legalsylvain:
;-) 
I was suggesting to OCA board some months ago to create rewards for OCA contributors during OCA days.
If it's accepted, for sure, you'll can apply for the the rewards of the most poetic review.

Thanks for your review.

pedrobaeza:
One thing before merging: please squash together "OCA Transbot commits".

legalsylvain:
> One thing before merging: please squash together "OCA Transbot commits".

Unrelated to that PR, but I don't understand why. The 3 commits exists in 10.0 version and have not been squashed. it's a pity if each people cherry-picking commits should resquash. IMO, Should be done upstream, or not.

pedrobaeza:
@legalsylvain it's not unrelated. I don't really understand why some of you bothers so much to do this simple task (it's done in a minute).

The thing is that Transifex dumps each weekend the translations made during that week, but also the changes over already translated terms, creating an artificial bigger diff that doesn't worth to keep. In the original branch, there's no other option, as it's organic in time, but when migrating to newer versions - or this case as a backport -, we should it.

legalsylvain:
>  I don't really understand why some of you bothers so much to do this simple task

Because 
- it takes times
- I don't see any interest
- although it was interesting (Question of perspective), it should be done by a bot

> (it's done in a minute).

Well, in fact, no. cloning my branch, making git manipulation, pushing again, waiting to see if all is ok, travis, etc..

pedrobaeza:
It can't be done by a bot, and if you do it when doing the backporting, no need to build another Travis + runbot. Any way, it's computer time, not yours :wink:

StefanRijnhart:
I agree with @legalsylvain. This commit hygiene is mainly a matter of taste. For large repos with a huge amount of commits it may be noticeable in terms of performance but I don't think this is the case in the OCA.

pedrobaeza:
Everything counts. It's not only this repository, but every OCA repository added to a project, and every module added to a repository. I can assure you that since v11 (both the module cleaning and this translation "hygiene" I'm performing on all PRs - mine and of the others -), my deployments goes dozens of times faster. I don't have problem in doing the same in this PR if you don't want to.

legalsylvain:
Hi @pedrobaeza. Saw you realized commit optimization you desired.
Is it OK to merge this PR ?
regards.

pedrobaeza:
Yes, merging!

PR #362, opened 2018-03-05
[11.0] Add module enabling to reuse supplier invoice references

asaunier:
In some cases, having more than one supplier invoice with the same reference is possible. For instance, in Switzerland, BVR/ESR numbers are used as a type of reference. Generally, this number is unique and used only for one invoice. However, some suppliers do use the same BVR reference for several invoices (usually for recurring invoices).

Related to https://github.com/odoo/odoo/pull/15891

### DIFF
--- a/account_invoice_supplier_ref_reuse/README.rst
+++ b/account_invoice_supplier_ref_reuse/README.rst
@@ -1,0 +1,54 @@
+.. image:: https://img.shields.io/badge/license-AGPL--3-blue.png
+   :target: https://www.gnu.org/licenses/agpl
+   :alt: License: AGPL-3
+
+=================================
+Reuse Supplier Invoice References
+=================================
+
+In some cases, having more than one supplier invoice with the same
+reference is possible. For instance, in Switzerland, BVR/ESR numbers
+are used as a type of reference. Generally, this number is unique and
+used only for one invoice. However, some suppliers do use the same BVR
+reference for several invoices (usually for recurring invoices).
+
+Installation
+============
+
+To install this module, just click the install button.
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/account-invoicing/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smash it by providing detailed and welcomed feedback.
+
+Credits
+=======
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://odoo-community.org/logo.png>`_.
+
+Contributors
+------------
+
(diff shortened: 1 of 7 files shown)


account_invoice_supplier_ref_reuse/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2018 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    'name': 'Reuse Supplier Invoice References',
+    'version': '11.0.1.0.0',
+    'summary': 'Makes it possible to reuse supplier invoice references',
+    'author': 'Camptocamp SA, '
```
guewen: s/Camptocamp SA/Camptocamp/ (we'd like to have to have the same name everywhere for easier lookups in apps.odoo.com)

account_invoice_supplier_ref_reuse/models/account_invoice.py
```diff
@@ -0,0 +1,14 @@
...
+
+from odoo import api, models
+
+
+class AccountInvoice(models.Model):
+    _inherit = "account.invoice"
+
+    @api.multi
+    def _check_duplicate_supplier_reference(self):
+        # Do nothing instead of checking if the reference number already
```
feketemihai, line 12: And when you uninstall the module, would raise errors for each of the invoice in with the same reference, isn't easier to update the reference with reference +"-a number" ex you have on test ABC123 first invoice, second one to be ABC123-1, third ABC123-2 and so on..

asaunier: Hi @feketemihai thanks for the feedback!
Having unique references would indeed be easier but in some cases it's not really an option, as in the swiss example quoted in the description above. For such cases it's likely that people won't uninstall the module ever, since they need it to reuse some references anyway.

asaunier:
I have just squashed the PR, which should now be ready for merging.
@guewen @feketemihai OK for you?

account_invoice_supplier_ref_reuse/README.rst
```diff
@@ -0,0 +1,54 @@
+.. image:: https://img.shields.io/badge/license-AGPL--3-blue.png
+   :target: https://www.gnu.org/licenses/agpl
+   :alt: License: AGPL-3
+
+=================================
+Reuse Supplier Invoice References
+=================================
+
+In some cases, having more than one supplier invoice with the same
+reference is possible. For instance, in Switzerland, BVR/ESR numbers
```
yvaucher, line 10: The english term for BVR/ESR is ISR

yvaucher (approved):
LGTM

One nitpicking in the readme.

feketemihai (approved):

PR #298, opened 2017-07-26
[10.0 ][MIG] account_invoice_analytic_search

nikul-serpentcs:
Invoice Analytic Search
=======================

Organizations often require to quickly find the invoices associated to a project or to an analytic account, searching by it's code, name or project/account manager. This module introduces the possibility search
customer or supplier invoices by analytic account or by project manager.

PR #366, opened 2018-03-19
[10.0][MIG] account_invoice_analytic_search

AaronHForgeFlow:
Supersedes #298


account_invoice_analytic_search/models/account_invoice.py
```diff
@@ -0,0 +1,33 @@
...
+        for invoice in self:
+            invoice.account_analytic_ids =\
+                invoice.mapped('invoice_line_ids.account_analytic_id.id')
+
+    @api.multi
+    def _search_analytic_accounts(self, operator, value):
+        invoice_line_obj = self.env['account.invoice.line']
+        invoice_line_ids = invoice_line_obj.search(
+            [('account_analytic_id', operator, value)])
+        return [('id', 'in', tuple(invoice_line_ids.mapped('invoice_id').ids))]
```
rousseldenis: @aheficent As I said in https://github.com/OCA/account-invoicing/pull/298, this is not sufficent as this can lead to almost three searches (invoice_line.search(), invoice_line_ids.mapped() and return result).

Do just :
```python
def  _search_analytic_accounts(self, operator, value):
    return [('invoice_line_ids.account_analytic_id', operator, value)]
```

rousseldenis (changes requested):

rousseldenis (approved):

AaronHForgeFlow:
tests fixed now

feketemihai (approved):

PR #367, opened 2018-03-29
[FIX] account_invoice_refund_link: Don't choke with standard fields

pedrobaeza:
Since v11, Odoo counts with a field called `refund_invoice_ids` of type one2many, which conflicts with the m2m field defined in this module. Moreover, these fields are not needed anymore.

There is only 2 OCA modules using this module right now, and they don't need significant changes for this adaptation, but I will do it when this get merged.

@Tecnativa

### DIFF
--- a/account_invoice_refund_link/README.rst
+++ b/account_invoice_refund_link/README.rst
@@ -6,19 +6,19 @@
 Link refund invoice with original
 =================================
 
-This module links customer and supplier refunds with the invoice that originate
-them (and invoice lines).
+This module shows the links between refunds and their original invoices in the
+invoice form and also keep track of refund lines and their original invoice
+lines.
 
 Usage
 =====
 
-This module creates in the invoice form a new 'Refund' page.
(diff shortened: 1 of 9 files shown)


pedrobaeza:
Fixing tests. Be ready soon.

pedrobaeza:
Tests fixed. Ready to be reviewed

feketemihai (approved):
tested on runbot and it's ok, not tested the init_hook.

pedrobaeza:
> not tested the init_hook.

I have included a test for the init hook to showcase its use.

feketemihai:
I meant i didn't test locally...to check init_hook...i know it's fine...

StefanRijnhart (approved):

pedrobaeza:
FYI, one of the modules that uses this one is changed in this PR: https://github.com/OCA/vertical-association/pull/57

The other for OCA/l10n-spain is in progress

PR #368, opened 2018-03-30
10 account invoice pricelist from 9

rvalyi:
this is a new attempt of https://github.com/OCA/account-invoicing/pull/342 but properly starting from newer v9 version and adding last fixes/features from v8 atop of it.
cc @pedrobaeza @bealdav @legalsylvain


account_invoice_pricelist/README.rst
```diff
@@ -62,6 +46,8 @@ Contributors
 ------------
 
 * Sylvain LE GAL (https://twitter.com/legalsylvain)
+* Holger Brunn <hbrunn@therp.nl>
+* Sergio Teruel <sergio.teruel@tecnativa.com>
```
pedrobaeza, line 45: Add yourself in the contributors

account_invoice_pricelist/models/account_invoice_line.py
```diff
@@ -0,0 +1,31 @@
+# -*- coding: utf-8 -*-
```
pedrobaeza: Don't split this from main model and put all in the file `account_invoice.py`. See https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#file-naming for explanations

account_invoice_pricelist/models/account_invoice_line.py
```diff
@@ -0,0 +1,31 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openerp import models, api
+
+
+class AccountInvoiceLine(models.Model):
+    _inherit = 'account.invoice.line'
+
+    @api.onchange('product_id')
```
pedrobaeza: You have to recompute also on quantity change, as there are pricelist with escalated prices.

account_invoice_pricelist/views/account_invoice_view.xml
```diff
@@ -0,0 +1,58 @@
...
+<odoo>
+
+    <record id="view_account_invoice_customer_form" model="ir.ui.view">
+        <field name="model">account.invoice</field>
+        <field name="inherit_id" ref="account.invoice_form"/>
+        <field name="arch" type="xml">
+            <field name="user_id" position="after">
+                <group col="4" colspan="4">
+                    <field name="pricelist_id"/>
+                    <button type="object"
```
pedrobaeza, line 11: IMO this button shouldn't be included, but auto-recompute on pricelist change + product/qty change, the same as standard Odoo does in sales orders.

account_invoice_pricelist/views/account_invoice_view.xml
```diff
@@ -0,0 +1,58 @@
...
+                            name="button_update_prices_from_pricelist"
+                            string="Update prices"
+                            attrs="{'invisible': ['|', ('pricelist_id', '=', False), ('state', 'not in', ['draft'])]}"
+                            help="Update price in lines from the pricelist"
+                    />
+                </group>
+            </field>
+            <field name="invoice_line_ids" position="attributes">
+                <attribute name="context">{'partner_id': partner_id, 'price_type': context.get('price_type') or False, 'type': type, 'pricelist_id': pricelist_id}</attribute>
```
pedrobaeza: Is this still needed?

pedrobaeza (changes requested):

account_invoice_pricelist/models/account_invoice_line.py
legalsylvain: We could maybe make depends on invoice_id.pricelist_id too. It make the "update from pricelist" on the invoice obsolete.
* I think it more usable. I think we want to apply the allways pricelist when we change the pricelist
* ~~it's not the same behaviour like in sale module. (odoo core).~~

account_invoice_pricelist/views/account_invoice_view.xml
bealdav: In pratical use cases, you may stick on manual values because of one shot product invoicing. I use it in v8, then I need of this button for v10 too. Thanks

pedrobaeza: I prefer the auto thing and following v8 paradigm.

bealdav: You prefer, ok but do you use this module ?

legalsylvain: > IMO this button shouldn't be included, but auto-recompute on pricelist change + product/qty change, the same as standard Odoo does in sales orders.

Seems it's changed, because in oldest version, the mechanism was not implemented on sale orders.
I think that : 
* both approaches are correct;
* OCA should adapt its module to odoo behaviour. If sale orders now works with onchange on pricelist, we should implement it to improve user experience.

so :+1: for @pedrobaeza proposal.

regards.

account_invoice_pricelist/views/account_invoice_view.xml
pedrobaeza: > You prefer, ok but do you use this module ?

@bealdav Yes, I made the v9 migration ;)

account_invoice_pricelist/views/account_invoice_view.xml
bealdav: ok 2 against 1, you know what to do @rvalyi ;-)

account_invoice_pricelist/views/account_invoice_view.xml
bealdav: ok thanks

pedrobaeza:
@rvalyi please don't flood Travis with constant commits triggering tests. Cancel them or wait a bit for pushing.

pedrobaeza:
(and runbot also)

account_invoice_pricelist/README.rst
rvalyi: done

account_invoice_pricelist/views/account_invoice_view.xml
rvalyi: Hello @pedrobaeza 
I took most of your remarks in consideration except this one. As far as I tested, even in v10, if you change the pricelist in a sale oder it doesn't update the price of the items (there is no such onchange on the pricelist field).

We can very well remove this button and have the same behavior. But IMHO, this is not bad to have the button, I feel rather that the native sale order is missing such button. It reminds me the modules Alexis de Lattre made to update the lines when you change a fiscal position because there were not such update. So isn't better to keep this button? Nobody will force you to use it either...

We do you think?

pedrobaeza: Ah, right, I mixed fiscal position with pricelist. Indeed, prices are not changed on that case. Well, OK then.

bealdav:
from runbot 

WARNING openerp_test odoo.models: @onchange('product_id', 'quantity', 'uom') parameters must be field names

rvalyi:
@bealdav good catch. I fixed the typo. It is ok for you guys then?

bealdav:
@pedrobaeza  since your new comments, you may remove request changes ? Thanks

account_invoice_pricelist/README.rst
```diff
@@ -44,16 +29,15 @@ For further information, please visit:
 Known issues / Roadmap
 ======================
 
-* When changing the pricelist, give the possibility to recompute prices;
+None currently
```
pedrobaeza: You can remove directly the section

account_invoice_pricelist/views/account_invoice_view.xml
```diff
@@ -0,0 +1,52 @@
...
+
+    <record id="view_account_invoice_supplier_form" model="ir.ui.view">
+	      <field name="model">account.invoice</field>
+      	<field name="inherit_id" ref="account.invoice_supplier_form"/>
+      	<field name="arch" type="xml">
+	          <field name="user_id" position="after">
+                <group col="4" colspan="4">
+                    <field name="pricelist_id"/>
+      	            <button type="object"
+			                      name="button_update_prices_from_pricelist"
```
pedrobaeza: Strange indentation

account_invoice_pricelist/models/account_invoice.py
```diff
@@ -0,0 +1,75 @@
...
+
+    @api.onchange('partner_id', 'company_id')
+    def _onchange_partner_id_account_invoice_pricelist(self):
+        result = super(AccountInvoice, self)._onchange_partner_id()
+        if self.partner_id:
+            if self.type in ('out_invoice', 'out_refund')\
+                    and self.partner_id.property_product_pricelist:
+                # Customer Invoices
+                self.pricelist_id = self.partner_id.property_product_pricelist
+            elif self.type in ('in_invoice', 'in_refund')\
```
pedrobaeza: There are no more purchase pricelists, so this should be removed.

pedrobaeza (changes requested):
Please squash your migration commits together

account_invoice_pricelist/README.rst
rvalyi: ok

account_invoice_pricelist/views/account_invoice_view.xml
rvalyi: ok

account_invoice_pricelist/models/account_invoice.py
rvalyi: ok

rvalyi: @pedrobaeza ok to your remarks. But about squashing the commits. Do you think about specific commits? I think I mostly give useful details about what comes from v9, what comes from v8 (because both evolved independently) and also it is not so easy to squash commits because of the merge of the account_invoice_line into the account_invoice file. So can we forget that or may be you give me specific commits I can squash. I'll amend my last commit to take your last remarks into account.

pedrobaeza:
I'm not talking about making a general squashing into 1 commit, but grouped by logical set/author for removing some extra diffs that are not needed (the progress of translations in "OCA Transbot..." commits), or remove meaningless commits like PEP8 adjustments or so. This is a brief proposal (but not exhaustive):

seleccion_009

pedrobaeza:
And the guide on how to do it: https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests

rvalyi:
all right. changes coming...

rvalyi:
@pedrobaeza your changes request has been taken into account. How is it for you now?

rvalyi:
ping @pedrobaeza So is it ok for merging on your side now?

pedrobaeza:
I'm going to squash a bit more the history (specially your latest commits), and merge. Thanks for migrating this.

pedrobaeza (approved):

bealdav (approved):

account_invoice_pricelist/views/account_invoice_view.xml
alexis-via: This view code breaks the invoice form view on v10... the labels of other fields of the right column disappear. I wonder if someone really uses the code in real life... :-(

alexis-via: Going back to PR https://github.com/OCA/account-invoicing/pull/342  At least, that one didn't break views !!!

pedrobaeza: Propose a PR for solving it. Don't make a drama of this, and it's not breaking. It's only UI.

alexis-via: I don't make a drama of this. I just say that this work is not tested.

PR #328, opened 2017-11-13
[11.0][MIG] account_payment_term_extension

nachoserra-domatix:

PR #375, opened 2018-04-16
[MIG] account_payment_term_extension: Migration to 11.0

cubells:
- Update README
- Update spanish translation
- Adding changes of https://github.com/OCA/account-invoicing/pull/328

cc @Tecnativa


account_payment_term_extension/i18n/es.po
```diff
@@ -58,30 +55,30 @@ msgstr "Nombre mostrado"
 #. module: account_payment_term_extension
 #: model:ir.ui.view,arch_db:account_payment_term_extension.view_payment_term_line_form
 msgid "Due Date Computation"
-msgstr ""
+msgstr "álculo de fecha de vencimiento"
```
pedrobaeza: Missing initial C

account_payment_term_extension/i18n/es_ES.po
```diff
@@ -1,29 +1,25 @@
 # Translation of Odoo Server.
```
pedrobaeza: Remove this file

account_payment_term_extension/models/account_payment_term.py
```diff
@@ -1,12 +1,11 @@
-# -*- coding: utf-8 -*-
 # Copyright 2013-2016 Camptocamp SA (Yannick Vaucher)
 # Copyright 2004-2016 Odoo S.A. (www.odoo.com)
 # Copyright 2015-2016 Akretion
 # (Alexis de Lattre <alexis.delattre@akretion.com>)
 # Copyright 2018 Simone Rubino - Agile Business Group
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-
+from functools import reduce
```
pedrobaeza: Why this?

pedrobaeza (changes requested):
Please squash commits by logical set/author

account_payment_term_extension/models/account_payment_term.py
cubells: reduce function is used here:
https://github.com/OCA/account-invoicing/pull/375/files#diff-d0188f250f37daab85f1680d28b455ccR198
and in python3 does not exist. We have two solutions, import from functools or replace reduce function by a for loop.
I choose import option.

pedrobaeza: OK, thanks for the explanations

pedrobaeza (approved):

angelmoya (approved):

PR #376, opened 2018-04-17
[11.0][MIG] account_invoice_transmit_method: Migration to 11.0

jarmokortetjarvi:


mlaitinen (approved):
LGTM

account_invoice_transmit_method/models/partner.py
```diff
@@ -0,0 +1,30 @@
...
+
+from odoo import models, fields, api
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    customer_invoice_transmit_method_id = fields.Many2one(
+        'transmit.method', string='Customer Invoice Transmission Method',
+        company_dependant=True, track_visibility='onchange',
```
astirpe: There's a typo, it should be `company_dependent`. This typo is also present in previous versions of the module.

astirpe (changes requested):

account_invoice_transmit_method/models/partner.py
tafaRU: good catch! I'm taking care of creating PRs to fix that in `8.0` and `10.0`.

astirpe: Ping me when you open the PRs, I will quickly approve them.

astirpe (approved):
Looks good to me.

account_invoice_transmit_method/models/partner.py
tafaRU: https://github.com/OCA/account-invoicing/pull/424 and https://github.com/OCA/account-invoicing/pull/425

jarmokortetjarvi:
pedrobaeza review in https://github.com/OCA/account-invoicing/pull/424#pullrequestreview-156698994
also applies to this PR

astirpe:
@jarmokortetjarvi the comment you are mentioning makes sense in older versions, in case the module is already installed in production. But in this case it's a first release for version 11.0, so I don't see any reason to do that. What is your use case?

jarmokortetjarvi:
> @jarmokortetjarvi the comment you are mentioning makes sense in older versions, in case the module is already installed in production. But in this case it's a first release for version 11.0, so I don't see any reason to do that. What is your use case?

If this PR were to be merged now, and an installation from previous version, say 10.0, would be migrated to 11.0 and updated, the contents of the field would be lost. This wouldn't be a problem if the PR for the 10.0 version was already merged with the necessary migration script.

I'm not sure if this is an actual issue though.

astirpe:
@jarmokortetjarvi I think you're right! Thanks!

rafaelbn (approved):
Tested functionally in runbot :+1: 

- Name is not very popular
- We should take care not having this functionality duplicated in partner-contact in future

rafaelbn:
From: https://github.com/OCA/account-invoicing/pull/376#issuecomment-430213161

> pedrobaeza review in #424 (review)
> also applies to this PR

Ping @pedrobaeza

jarmokortetjarvi:
Would it be fine if I reverted the "Fix typo" -commit to get the migration PR merged?
It could be opened as a new PR. This would mean that 11.0 branch would need the migration script too though.

The migration itself is pretty straight-forward, but the field change is a bit more complicated issue.

pedrobaeza:
Yes, you can remove the type `company_dependant` and put on the roadmap to make the fields company aware.

jarmokortetjarvi:
Removed the commit to fix the company-dependency typo.
It should be a pure 10.0 branch migration now.

pedrobaeza:
Please remove the argument of the field with typo, and add in the roadmap to be converted to company dependent with the migration script.

rafaelbn:
Please @jarmokortetjarvi take in care this PR from @alexis-via https://github.com/OCA/account-invoicing/pull/440

Thanks

jarmokortetjarvi:
> Please @jarmokortetjarvi take in care this PR from @alexis-via #440
> 
> Thanks

While I support that improvement, it hasn't been accepted and merged yet, so I'd prefer not to include the https://github.com/jarmokortetjarvi/account-invoicing/pull/1 PR into this PR. I fear it will delay the merge again.

pedrobaeza:
Hi, @jarmokortetjarvi, on contrary, this migration is the ideal place as the module is not yet in branch, so we can change dependencies.

jarmokortetjarvi:
I'm closing this PR as it seems unlikely it will be merged to `11.0`

If someone needs `11.0` version, a backport from `12.0` version https://github.com/OCA/account-invoicing/commit/5bdf893905b0faabf1ef0623a3d68cb05f97eec4 would be a better idea at this point.

PR #378, opened 2018-05-01
[FIX][#377] possible fix for error in computation of amount_untaxed

### LINKED ISSUE #377: [8.0.][account_invoice_rounding] rounding added twice to amount_untaxed
Module: account_invoice_rounding
Version: 8.0

Steps to reproduce:
1. (install account_invoice_rounding)
2. In Settings -> Accounting set the following:
selection_183
3. Create new customer invoice
4. Add new invoice line and set the unit price to 90.00 and tax 7.7%
5. Save and Validate

The "Rounding" invoice line amount is added twice in the amount_untaxed => 90.04. The strange thing is that this only occurs for specific unit prices, in most cases it seems to work fine.
selection_184
The total amount is computed correctly but the calculation of the amount_untaxed is wrong.

BT-tbaechle:
This seems to fix #377 but I'm sure there is a more elegant solution for this. So another solution (or validation) would be greatly appreciated.
The problem seems to be that the core _compute_amount method already includes the amount of the rounding invoice line while the global_round_line_id on the invoice is not yet set which is why the amount is not removed here: 
https://github.com/OCA/account-invoicing/blob/8.0/account_invoice_rounding/account.py#L168
Before it is added back/again here: 
https://github.com/OCA/account-invoicing/blob/8.0/account_invoice_rounding/account.py#L177

### DIFF
--- a/account_invoice_rounding/account.py
+++ b/account_invoice_rounding/account.py
@@ -172,8 +172,9 @@ class AccountInvoice(models.Model):
                 if 'amount_tax' in swedish_rounding:
                     self.amount_tax = swedish_rounding['amount_tax']
                 elif 'amount_untaxed' in swedish_rounding:
-                    self.amount_untaxed = (
-                        swedish_rounding['amount_untaxed'])
+                    if sum(line.price_subtotal for line in self.invoice_line) != self.amount_untaxed:
+                        self.amount_untaxed = (
+                            swedish_rounding['amount_untaxed'])
 
     @api.multi
     def _get_rounding_invoice_line_id(self):


account_invoice_rounding/account.py
```diff
@@ -174,8 +174,9 @@ def _compute_amount(self):
                 if 'amount_tax' in swedish_rounding:
                     self.amount_tax = swedish_rounding['amount_tax']
                 elif 'amount_untaxed' in swedish_rounding:
-                    self.amount_untaxed = (
-                        swedish_rounding['amount_untaxed'])
+                    if sum(line.price_subtotal for line in self.invoice_line) != self.amount_untaxed:
+                        self.amount_untaxed = (
+                            swedish_rounding['amount_untaxed'])
```
yvaucher, line 177: So basically swedish_rounding['amount_untaxed'] is wrong if there is an adjusting line.

There was an issue solved in version 9.0 that implied two write due to method call cycling (probably not related):
https://github.com/OCA/account-invoicing/commit/be97dfa9248ebcd0f0889e6da999e23a77ce824a#diff-9096b1ac1c24a29588050b4c5a6d3726

The value in swedish_rounding['amount_untaxed'] comes from here:
https://github.com/OCA/account-invoicing/blob/8.0/account_invoice_rounding/account.py#L58

Thus we should probably fix it here. Still this value is set twice and that's a bad idea. I'm doubtful about the first write in `self.amount_untaxed` on line 168. It could probably be moved after the `if swedish_rounding:` in a elif to set it as we probably still want to remove it from untaxed amount.

Could you give a try to backport the tests made in later version of this module to make it stable:

https://github.com/OCA/account-invoicing/blob/10.0/account_invoice_rounding/tests/test_invoice_rounding.py

yvaucher (changes requested):

yvaucher:
Thank you for trying to fix this, as you said there is a better way to solve this, to help in solving it, I think it would really help by replacing the yml tests by the python tests of further versions.

account_invoice_rounding/account.py
BT-sschmid: The test was added by #392. Until the fix is applied, the test will fail.

yvaucher:
@BT-tbaechle thanks, I merged #392  can you rebase this PR?

BT-tbaechle:
@yvaucher thank you. I just did a rebase...

yvaucher:
@BT-tbaechle have you seen my comment about amount_untaxed setted twice? (read line [166]( https://github.com/OCA/account-invoicing/blob/71c58799ca9c4b49b5d3f4c1de93426effbe4304/account_invoice_rounding/account.py#L166) instead of 163)

BT-tbaechle:
@yvaucher yes, I just assumed you wanted to try it yourself. However I'm not sure if I understood your suggestion correctly, here is what I did: 
selection_266
Unfortunately this didn't fix the problem. I would greatly appreciate your help.
Thanks!

PR #384, opened 2018-05-21
[9.0][FIX] purchase_batch_invoicing: No Tax compute in invoice created

igallart:


pedrobaeza:
Please change module version number according https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers

rafaelbn:
Dear @igallart please could you review travis in order to merge this?

carlosdauden (approved):

rafaelbn (approved):
@igallart :+1:

purchase_batch_invoicing/wizards/purchase_batch_invoicing.py
```diff
@@ -92,6 +92,7 @@ def action_batch_invoice(self):
                 invoice.currency_id = po.currency_id
                 invoice.purchase_id = po
                 invoice.purchase_order_change()
+                invoice._onchange_invoice_line_ids()
```
pedrobaeza: Why the onchange? This should be `invoice.compute_taxes` IMO

pedrobaeza (changes requested):

purchase_batch_invoicing/wizards/purchase_batch_invoicing.py
igallart: If you think it's better to change it, that's fine with me. I can not prove it, but with the current solution it works well.

pedrobaeza: Yes, because this might done more things than the expected one, depending also on the extension modules. See the patch we have made for v11: #505

rafaelbn: How to suggest the change in the code? It saw now is possible in Github

rafaelbn:
```suggestion
                invoices.compute_taxes()
```

rafaelbn: This is BETA TEST

igallart: Changed.

pedrobaeza:
Why are you changing here things of account_invoice_merge?

igallart:
> Why are you changing here things of account_invoice_merge?

I have changed it based on the suggestions. I changed the position because I thought it was more correct to place it after the raise

pedrobaeza:
But this PR is for `purchase_batch_invoicing`. This has nothing to do with `account_invoice_merge`, and I don't see any comment about changing that here.

igallart:
You're right Pedro, it's been my fault. This happened because during the pull request I made two corrections in two modules, and github puts it on the same fork. I do not know if I can have two forks from the same repository at the same time. If it is very problematic, I will eliminate the extra one.

pedrobaeza:
You can't have more than one fork, but you are able to create the number of branches you need. I see that you have already fixed it and create another branch, so I think you have understood the concept.

pedrobaeza (approved):

PR #385, opened 2018-05-25
[ADD][11.0] account_invoice_reimbursable

etobella:
Reimbursables management
========================

This module adds the option to create reimbursables.

Reimbursables are charges made to the customer on behalf of a third party.

The customer will receive two invoices, one from the supplier that includes
the reimbursable and one from the third party.

The supplier will pay to the third party its invoice to the customer on behalf
of the customer

Usage
=====

#. Go to 'Accounting/Invoicing > Purchases > Documents > Vendors Bills'
#. Create an invoice for a provider and add the reimbursables on the
   reimbursable page. Reimbursable partner will be required
#. Validate the invoice


pedrobaeza:
Is this the Spanish "suplido"?

etobella:
I am still working on it, because, it only works for supplier invoices. It should be fixed in order to be used on customer invoices and use the elements for the payments.

etobella:
Yes, it is the spanish "suplido"

pedrobaeza:
Great! I will say it for having reviewers in this.

etobella:
I think that is outside the scope of this PR the usage of customer invoices for some reasons:
-  A lot of companies receive invoices with reimbursements (usually when lawyers are involved), but not many make invoices with reimbursements
-  Reimbursements on customer invoices are hard to deal for reconciliation. I think that it needs to be meditated

So, I think the best option is to make hook in order to allow the change of accounts on the second module. What do you think?

Also, I had to add a modification of an account method until odoo/odoo#24915 is applied

etobella:
I finally removed all the unnecessary code and I think it is better to maintain customer invoices out of the current goal of this module.
It is ready for testing now!

JordiBForgeFlow (approved):
:+1:  Technical and functional review.

oca_dependencies.txt
```diff
@@ -0,0 +1 @@
+
```
MiquelRForgeFlow, line 1: why?

etobella: Travis was returning a small issue, not an error, but I tried to solve it.

MiquelRForgeFlow (approved):
Checked on runbot

PR #392, opened 2018-06-15
[#377][ADD][8.0] account_invoice_rounding: Added python unit-tests

BT-sschmid:
Most of the test was backported from v10.
The test covers the problem of issue #377 and will fail until fixed.


yvaucher (approved):
This should raise the concerns of #378

yvaucher:
This fix the tests but make travis fails, I'll merge as I prefer a correct but failing test than a green but wrong test on travis.

PR #393, opened 2018-06-24
[ADD] Add module account_invoice_change_currency

novawish:
Allow users to update and convert to new currency of invoice based on currency exchange rate


pedrobaeza (approved):
This was previously approved in the superseeded PR, so I proceed to merge.

PR #394, opened 2018-06-25
[ADD] account_invoice_pro_forma_sequence

SimoRubi:
Module to track pro-forma invoices with sequences.
Please check, thanks


hurrinico (approved):
👍 just code review

eLBati (approved):
LGTM

tafaRU (changes requested):
Please move https://github.com/OCA/account-invoicing/pull/394/commits/e0425e8f7efad7ca33cc91da49f4644fc3ffbaa5 to a separate PR

SimoRubi:
Done, now this depends on https://github.com/OCA/account-invoicing/pull/401

SimoRubi:
@tafaRU is it ok now? Please see #401

tafaRU (approved):

SimoRubi:
@OCA/accounting-maintainers is this ok to be merged?

SimoRubi:
Thanks

elicoidal (commented):
LGTM but I have a general question. It seems weird to me to have a number different from the final number (I would have tended to use the invoice number in the proforma stage).
At least in China it is not an issue because there is an official invoice with separate number.
Was there a specific legal reason to implement it separately?
Would there be reason (legal or not) to not have it separately?

SimoRubi:
@elicoidal Thanks for the review!
I implemented it in a separate field for a couple of reasons:

- `account.invoice.number` is a related field from `move_id.name` and I do not want mess up with the account move

- In many points (https://github.com/odoo/odoo/blob/10.0/addons/account/models/account_invoice.py#L452 is one) Odoo applies some logic based on the fact if the invoice has a number or not, and I do not want to mess up there either

I do not know about legal reasons but *coding reasons* are enough for me

elicoidal (approved):

elicoidal:
Let's move on: if there is a requirement in the future for same number then a configuration could be added by the requester :wink:

PR #395, opened 2018-06-27
[10.0][IMP] purchase_stock_picking_returm_invoicing: Improve perf for qty_returned field

pedrobaeza:
This field is computed very slow with previous code. I have improved the speed through several techniques like read_group and shared prefetching. It might be less readable, but incredibly faster compared with the previous code.

cc @Tecnativa

### DIFF
--- a/purchase_stock_picking_return_invoicing/models/purchase_order.py
+++ b/purchase_stock_picking_return_invoicing/models/purchase_order.py
@@ -3,6 +3,7 @@
 # Copyright 2018 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
+import collections
 from odoo import api, fields, models
 
 
@@ -127,15 +128,40 @@ class PurchaseOrderLine(models.Model):
         'move_ids.returned_move_ids.state',
     )
     def _compute_qty_returned(self):
+        """Made through read_group for not impacting in performance."""
+        ProductUom = self.env['product.uom']
+        StockMove = self.env['stock.move']
+        groups = StockMove.read_group(
+            [('origin_returned_move_id.purchase_line_id', 'in', self.ids),
+             ('state', '=', 'done'),
+             ('location_id.usage', '!=', 'supplier')],
+            ['origin_returned_move_id', 'product_uom_qty', 'product_uom'],
+            ['origin_returned_move_id', 'product_uom'], lazy=False,
+        )
+        p = self._prefetch
+        # load all records at once on first access
+        origin_move_ids = []
+        uom_ids = set()
+        for g in groups:
+            origin_move_ids.append(g['origin_returned_move_id'][0])
+            uom_ids.add(g['product_uom'][0])
+        StockMove.browse(origin_move_ids, prefetch=p)
+        ProductUom.browse(list(uom_ids), prefetch=p)
+        line_qtys = collections.defaultdict(lambda: 0)
+        for g in groups:
+            uom = ProductUom.browse(g["product_uom"][0], prefetch=p)
+            move = StockMove.browse(
+                g['origin_returned_move_id'][0], prefetch=p,
+            )
+            if uom == move.purchase_line_id.product_uom:
+                qty = g["product_uom_qty"]
+            else:
+                qty = uom._compute_quantity(
+                    g["product_uom_qty"], move.purchase_line_id.product_uom,
+                )
+            line_qtys[move.purchase_line_id.id] += qty
         for line in self:
-            qty = 0.0
-            moves = line.mapped('move_ids.returned_move_ids')
-            for move in moves.filtered(lambda x: x.state == 'done'):
-                if move.location_id.usage != 'supplier':
-                    qty += move.product_uom._compute_quantity(
-                        move.product_uom_qty, line.product_uom,
-                    )
-            line.qty_returned = qty
+            line.qty_returned = line_qtys.get(line.id, 0)
 
     @api.depends('qty_returned')
     def _compute_qty_received(self):
--- a/purchase_stock_picking_return_invoicing/__manifest__.py
+++ b/purchase_stock_picking_return_invoicing/__manifest__.py
@@ -6,7 +6,7 @@
 {
     "name": "Purchase Stock Picking Return Invoicing",
     "summary": "Add an option to refund returned pickings",
-    "version": "10.0.1.0.0",
+    "version": "10.0.1.0.1",
     "category": "Purchases",
     "website": "https://github.com/OCA/account-invoicing",
     "author": "Eficent,"


purchase_stock_picking_return_invoicing/models/purchase_order.py
```diff
@@ -127,15 +127,29 @@ def _compute_qty_refunded(self):
...
-                        move.product_uom_qty, line.product_uom,
-                    )
-            line.qty_returned = qty
+        """Made through read_group for not impacting in performance."""
+        StockMove = self.env['stock.move']
+        origin_moves = StockMove.search([
+            ('origin_returned_move_id.purchase_line_id', 'in', self.ids),
+        ])
+        groups = StockMove.read_group(
+            [('id', 'in', origin_moves.ids),
```
yajo: I think you can put here directly `('origin_returned_move_id.purchase_line_id', 'in', self.ids)` and remove the previous search.

purchase_stock_picking_return_invoicing/models/purchase_order.py
```diff
@@ -127,15 +127,30 @@ def _compute_qty_refunded(self):
...
+             ('location_id.usage', '!=', 'supplier')],
+            ['purchase_line_id', 'product_uom_qty', 'product_uom'],
+            ['purchase_line_id', 'product_uom'], lazy=False,
+        )
+        line_dict = {}
+        p = self._prefetch
+        for g in groups:
+            line_id = g['purchase_line_id'][0]
+            line_dict.setdefault(line_id, 0)
+            uom = ProductUom.browse(g['product_uom'][0]).with_prefetch(p)
```
yajo: I think that if you want to prefetch those UoM, you have to browse all of them first in the same query.

i.e., add this at the top of the method:

```python
self.mapped("product_uom")
```

Then you can use here `ProductUom.browse(g['product_uom'][0]).with_prefetch(self._prefetch)`

purchase_stock_picking_return_invoicing/models/purchase_order.py
```diff
@@ -127,15 +127,30 @@ def _compute_qty_refunded(self):
...
         for line in self:
-            qty = 0.0
-            moves = line.mapped('move_ids.returned_move_ids')
-            for move in moves.filtered(lambda x: x.state == 'done'):
-                if move.location_id.usage != 'supplier':
-                    qty += move.product_uom._compute_quantity(
-                        move.product_uom_qty, line.product_uom,
-                    )
-            line.qty_returned = qty
+            self.qty_returned = line_dict.get(line.id, 0)
```
yajo: Did you mean `line.qty_returned = ...`?

yajo (commented):
Please add some comments in code, it's hard to read... :confused:

purchase_stock_picking_return_invoicing/models/purchase_order.py
```diff
@@ -127,15 +127,29 @@ def _compute_qty_refunded(self):
...
+        groups = self.env['stock.move'].read_group(
+            [('origin_returned_move_id.purchase_line_id', 'in', self.ids),
+             ('state', '=', 'done'),
+             ('location_id.usage', '!=', 'supplier')],
+            ['purchase_line_id', 'product_uom_qty', 'product_uom'],
+            ['purchase_line_id', 'product_uom'], lazy=False,
+        )
+        p = self._prefetch
+        # load all UoM records at once on first access
+        uom_ids = set([g['product_uom'][0] for g in groups])
```
yajo: Remove the brackets, you don't need a list if you have a generator :wink::

```python
uom_ids = set(g['product_uom'][0] for g in groups)
```

yajo (approved):

purchase_stock_picking_return_invoicing/models/purchase_order.py
```diff
@@ -127,15 +127,35 @@ def _compute_qty_refunded(self):
...
+        groups = StockMove.read_group(
+            [('origin_returned_move_id.purchase_line_id', 'in', self.ids),
+             ('state', '=', 'done'),
+             ('location_id.usage', '!=', 'supplier')],
+            ['origin_returned_move_id', 'product_uom_qty', 'product_uom'],
+            ['origin_returned_move_id', 'product_uom'], lazy=False,
+        )
+        p = self._prefetch
+        # load all moves records at once on first access
+        origin_move_ids = set(g['origin_returned_move_id'][0] for g in groups)
```
carlosdauden: Can you check this option?

        p = self._prefetch
        group_dict = defaultdict(list)
        uom_ids = set()
        for g in groups:
            group_dict[g['origin_returned_move_id'][0]].append(
                g['product_uom'][0], g['product_uom_qty']
            )
            uom_ids.add(g['product_uom'][0])
        ProductUom.browse(list(uom_ids)).with_prefetch(p)
        for line in self:
            qty = 0.0
            for gl in group_dict[line.id]:
                if gl[0] == line.product_uom.id:
                    qty += gl[1]
                else:
                    uom = ProductUom.browse(gl[0]).with_prefetch(p)
                    qty += line.product_uom
                    x = uom._compute_quantity(gl[1], line.product_uom)
            line.qty_returned = qty

purchase_stock_picking_return_invoicing/models/purchase_order.py
```diff
@@ -127,15 +127,35 @@ def _compute_qty_refunded(self):
...
+        StockMove.browse(origin_move_ids).with_prefetch(p)
+        # load all UoM records at once on first access
+        uom_ids = set(g['product_uom'][0] for g in groups)
+        ProductUom.browse(list(uom_ids)).with_prefetch(p)
```
carlosdauden: You can use prefetch as paremeter:
`ProductUom.browse(list(uom_ids), prefetch=p)`
https://github.com/odoo/odoo/blob/020906659e55bd41a07ab0e4e79da903af7b50c1/odoo/models.py#L4324

In the below browses, too.

purchase_stock_picking_return_invoicing/models/purchase_order.py
```diff
@@ -127,15 +127,35 @@ def _compute_qty_refunded(self):
...
+        line_qtys = {}
+        for g in groups:
+            uom = ProductUom.browse(g["product_uom"][0]).with_prefetch(p)
+            move = StockMove.browse(
+                g['origin_returned_move_id'][0]
+            ).with_prefetch(p)
+            line_qtys.setdefault(move.purchase_line_id.id, 0)
+            line_qtys[move.purchase_line_id.id] += uom._compute_quantity(
```
carlosdauden: Add condition `uom == move.purchase_line_id.product_uom` to set g['product_uom_qty'] directly instead of call _compute_quantity function

carlosdauden (changes requested):

purchase_stock_picking_return_invoicing/models/purchase_order.py
pedrobaeza: The same is evaluated on the `_compute_quantity` method, so no gain adding that lines

carlosdauden: I don't see it:
https://github.com/odoo/odoo/blob/3112cd2dbe7459739f00eb775d5510f0aefcb3c2/addons/product/models/product_uom.py#L91-L106

carlosdauden (approved):

purchase_stock_picking_return_invoicing/models/purchase_order.py
```diff
@@ -127,15 +127,35 @@ def _compute_qty_refunded(self):
         'move_ids.returned_move_ids.state',
     )
     def _compute_qty_returned(self):
+        """Made through read_group for not impacting in performance."""
+        ProductUom = self.env['product.uom']
+        StockMove = self.env['stock.move']
```
moylop260, line 133: Use [snake_case for variables names](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#variable-names) and [CamelCase for class names](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#odoo-python-classes).

purchase_stock_picking_return_invoicing/models/purchase_order.py
```diff
@@ -127,15 +127,35 @@ def _compute_qty_refunded(self):
...
+        groups = StockMove.read_group(
+            [('origin_returned_move_id.purchase_line_id', 'in', self.ids),
+             ('state', '=', 'done'),
+             ('location_id.usage', '!=', 'supplier')],
+            ['origin_returned_move_id', 'product_uom_qty', 'product_uom'],
+            ['origin_returned_move_id', 'product_uom'], lazy=False,
+        )
+        p = self._prefetch
```
moylop260, line 141: `p = self._prefetch` is using the same pointer variable.
if you use `id(p)` and `id(self._prefetch)` you will the same id result.

What about uses the same `self` without extra `p` variable and without extra `with_prefetch(p)`?

purchase_stock_picking_return_invoicing/models/purchase_order.py
```diff
@@ -127,15 +127,35 @@ def _compute_qty_refunded(self):
...
+        # load all moves records at once on first access
+        origin_move_ids = set(g['origin_returned_move_id'][0] for g in groups)
+        StockMove.browse(origin_move_ids).with_prefetch(p)
+        # load all UoM records at once on first access
+        uom_ids = set(g['product_uom'][0] for g in groups)
```
moylop260: Avoid iteration 2 times for the same `groups` variable in order to get the `origin_move_ids` and `uom_ids` at same time.

purchase_stock_picking_return_invoicing/models/purchase_order.py
moylop260: Don't compute_quantity if it is the same uom

purchase_stock_picking_return_invoicing/models/purchase_order.py
```diff
@@ -127,15 +127,35 @@ def _compute_qty_refunded(self):
...
+        ProductUom.browse(list(uom_ids)).with_prefetch(p)
+        line_qtys = {}
+        for g in groups:
+            uom = ProductUom.browse(g["product_uom"][0]).with_prefetch(p)
+            move = StockMove.browse(
+                g['origin_returned_move_id'][0]
+            ).with_prefetch(p)
+            line_qtys.setdefault(move.purchase_line_id.id, 0)
```
moylop260: Use a `collections.defaultdict` instead.

purchase_stock_picking_return_invoicing/models/purchase_order.py
```diff
@@ -127,15 +127,35 @@ def _compute_qty_refunded(self):
...
         for line in self:
-            qty = 0.0
-            moves = line.mapped('move_ids.returned_move_ids')
-            for move in moves.filtered(lambda x: x.state == 'done'):
-                if move.location_id.usage != 'supplier':
-                    qty += move.product_uom._compute_quantity(
-                        move.product_uom_qty, line.product_uom,
-                    )
-            line.qty_returned = qty
+            line.qty_returned = line_qtys.get(line.id, 0)
```
moylop260, line 164: You can avoid use this extra loop if you use in the main loop:
`move.purchase_line_id['qty_returned']  += value`

since that is sharing the same object

Review the following example:
 - https://github.com/odoo/odoo/pull/25280/commits/0f237f129e369b0edbab83209ce6bfd58436060d

purchase_stock_picking_return_invoicing/models/purchase_order.py
yajo: He can use `self._prefetch` directly without `p`, but he needs to add `.with_prefetch()` after a browse call if he wants to reduce SQL queries, because `.browse()` automatically sets a new prefetch object if it doesn't get another one explicitly.

purchase_stock_picking_return_invoicing/models/purchase_order.py
yajo: TIL. It's cool!

```py
>>> d=collections.defaultdict(lambda: 0)
>>> d[1]+=5
>>> d[1]
5
```

purchase_stock_picking_return_invoicing/models/purchase_order.py
pedrobaeza: This is a special case that has been discussed a lot of times. This represents something similar as a class in Odoo context, and it's used by Odoo itself this way.

purchase_stock_picking_return_invoicing/models/purchase_order.py
pedrobaeza: I'm not sure that's a good example, as dependent values are going to be recomputed each time you assign the variable. Another option can be to enclose everything on a `with self.env.norecompute():` clause, but current option is not bad at all.

pedrobaeza:
Suggestions made or answered.

MiquelRForgeFlow (approved):
If this doesn't break anything, then it's fine for me.

purchase_stock_picking_return_invoicing/models/purchase_order.py
moylop260: If you like a new special case you should added to our guidelines.
For now, our guidelines of the OCA (not odoo) is use CamelCase for class name and snake_case for variables, uppercase for global variable.
If odoo uses other standard that is the odoo standard.
If you want join the odoo and oca standard and avoid discuss too many times you need added to the guidelines document.

purchase_stock_picking_return_invoicing/models/purchase_order.py
moylop260: if you uses “move.purchase_line_id[“qty_returned”] += qty” this is not re-comouted
It is just updating the cache value and you avoid an extra loop-for unneeded.

It just requires at start a self.update(qty_returned: 0) for the case where one record is not found.

pedrobaeza:
> If this doesn't break anything, then it's fine for me.

Tests are there for checking this :wink:

purchase_stock_picking_return_invoicing/models/purchase_order.py
pedrobaeza: It's not a special case, as this is not the variable mentioned in the guidelines, and the linter is not forcing us to do the way you want. Open if you want the discussion for forbidding it and then I won't use it if approved.

purchase_stock_picking_return_invoicing/models/purchase_order.py
pedrobaeza: I'm not sure about that and it's difficult for me to check right now, but hey, that's not so important. That loop is very cheap in performance terms, so I prefer to keep it. Remember the sentence about performance:
```
Programmers waste enormous amounts of time thinking about, or worrying about, the 
speed of noncritical parts of their programs, and these attempts at efficiency
actually have a strong negative impact when debugging and maintenance are considered.

We should forget about small efficiencies, say about 97% of the time: premature optimization
is the root of all evil. Yet we should not pass up our opportunities in that critical 3%.
```

purchase_stock_picking_return_invoicing/models/purchase_order.py
moylop260: I got it.
He is using a new browse for each iteration of the loop and he needs to re-use the same prefetch with all ids.

Thanks @Yajo to clarify.

moylop260 (approved):
Just small comments but not blocking.

pedrobaeza:
@moylop260 although we don't agree on that small things that can be discussed at other level, they are not blocking for the merging, so I proceed for continuing my flow with the migration to v11 (you can review it as well on #396 :wink: ).

Thanks for your review (and the others' ones as well).

purchase_stock_picking_return_invoicing/models/purchase_order.py
moylop260: Your first comment:
> This is a special case...

Your second comment:
> It's not a special case...

I think that you are contradicting the guidelines and your self comments.

But I don't have problem if you want skip our guideline [variable-names](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#variable-names) for now. If in the future you decide added it to avoid discussed it again and again it is good for me too.

I just approved this one.
Thanks for the great work here.

pedrobaeza: Well, I haven't used the proper words, but they were following your comments. Any way, what I meant to say basically is: "this one is out of the guidelines". Call is "special case" or not. Remember that I was answering your sentence:

> If you like a new special case you should added to our guidelines.

So the proper answer would be "this is not the kind of special case inside the rule, as this can't be considered a normal variable (and thus is a special variable case)". Better?

I don't want to skip things, simply that they are not covered, but you know me, we can discuss what you want for generalizing this rule, but I don't want to get blocked this PR for that reason, and this want blocks me a huge migration to the the performance question.

PR #401, opened 2018-07-03
[FIX] Do not overwrite account's view when adding kanban

SimoRubi:
Add kanban as first view without overwriting account's views.

### DIFF
--- a/account_invoice_kanban/__manifest__.py
+++ b/account_invoice_kanban/__manifest__.py
@@ -4,7 +4,7 @@
 {
     'name': 'Account Invoice Kanban',
     'summary': 'Add Kanban view in Invoice to follow administrative tasks',
-    'version': '10.0.1.0.0',
+    'version': '10.0.1.1.0',
     'website': 'https://www.elico-corp.com/',
     'author': 'Elico Corp, Odoo Community Association (OCA)',
     'depends': [
--- a/account_invoice_kanban/views/account_invoice_view.xml
+++ b/account_invoice_kanban/views/account_invoice_view.xml
@@ -74,15 +74,15 @@
         <field name="view_mode">kanban,tree,form,calendar,pivot,graph</field>
     </record>
 
-    <record id="account.action_invoice_tree1_view1" model="ir.actions.act_window.view">
-        <field eval="1" name="sequence"/>
+    <record id="action_invoice_tree1_view0" model="ir.actions.act_window.view">
+        <field eval="0" name="sequence"/>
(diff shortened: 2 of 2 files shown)


hurrinico (approved):
👍

eLBati:
> Do not overwrite account's view when adding kanban

Why not? Wasn't this done on purpose?
cc @elicoidal

tafaRU (approved):

SimoRubi:
@eLBati The issue here is that, before the PR, the view `account.action_invoice_tree1_view1` is replaced with a view of type kanban while originally (i.e in `account` module) it is of type tree.
Then, when the tree view is opened, it is the tree view of supplier invoices instead of the correct one (the tree view declared in `account`).

With this PR, I am adding a new view of type `kanban` and placing it before (`sequence = 0`) the tree view, instead of editing the original tree view declared in the menu's action.

I know it is complicated but let me try to clarify the situation:
In module `account`, the menu's action declared two views:
- tree `account.invoice_tree`
- form

Before the PR, in this module
- ~tree~ kanban Then the tree view opened was determined by the system and it might be no more `account.invoice_tree`
- form


After the PR:
- kanban 
- tree `account.invoice_tree`
- form

Same applies for supplier views/actions

pedrobaeza:
I think the purpose is to not have that kanban view normally visible by default, but having it for mobile view with `web_responsive` / `web_enterprise` to show it

elicoidal (approved):
Change looks OK to me (I do not know the reasons of the original design but your proposal looks sane to me).
Version number should be 10.0.1.1.0 (you need to update the module)

pedrobaeza:
But as said, I prefer to not have kanban on desktop views.

SimoRubi:
Right @elicoidal I updated the version number.

@pedrobaeza I've read the README of the module again, I'm sorry but it just enjoys using kanban views for viewing invoices and says nothing about mobile/desktop.
Anyway, I'm not changing the behavior of this module, I'm just fixing what I explained earlier and I think it was a bug, you can check in any runbot of 10.0 for this repository with the following steps:
1. Go to invoicing > Customer invoices
2. Open the tree view
3. In debug, check that the tree view is `account.invoice_supplier_tree` (in standard odoo it is `account.invoice_tree`)

SimoRubi:
squashed commits

SimoRubi:
Thanks

elicoidal:
@pedrobaeza please confirm we can move on.

pedrobaeza:
OK, let's go with this.

PR #405, opened 2018-07-15
[FIX] do not use sale pricelist for purchase invoices

legalsylvain:
This fix applies the same behaviour in account.invoice.line as in account.invoice file : 
The pricelist will be selected from property_product_pricelist for 'out_invoice' / 'out_refund'
The pricelist will be selected from property_product_pricelist_purchase for 'in_invoice' / 'in_refund'

thanks for your review.

### DIFF
--- a/account_invoice_pricelist/model/account_invoice_line.py
+++ b/account_invoice_pricelist/model/account_invoice_line.py
@@ -25,15 +25,18 @@ class AccountInvoiceLine(models.Model):
         context = self._context
         partner = self.env['res.partner'].browse(partner_id)
         pricelist_id = context.get('pricelist_id', False)
-        if type in ('out_invoice', 'out_refund'):
-            # Customer Invoices
-            pricelist_id = partner.property_product_pricelist.id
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):

AaronHForgeFlow (approved):
:+1: It makes sense to me

legalsylvain:
Hi @pedrobaeza and @aheficent.

Thanks for the review ! 
I just added two other commits that fixes two bugs in the same module : 

* the context is not taken into account. if an invoice has a pricelist different that the partner one, it will be ignored, and the partner pricelist will be applied.
* domain is not set, so it is possible to select a purchase pricelist on a sale invoice. (and sale pricelist no purchase invoice)

Could you update your review ? 

Thanks !

pedrobaeza (approved):

pedrobaeza:
Travis errors unrelated.

legalsylvain:
thanks @pedrobaeza for your quick review.

PR #406, opened 2018-07-16
[WIP][FIX][#379] remove rounding line if no longer needed

### LINKED ISSUE #379: [8.0.][account_invoice_rounding] rounding invoice line not removed after tax calucaltion
Module: account_invoice_rounding
Version: 8.0

Steps to reproduce:
1. (install account_invoice_rounding)
2. In Settings -> Accounting set the following:
selection_183
3. Create new invoice
3. Add new invoice line and set the unit price to 47.77 and tax 7.7%
4. Save 

A "Rounding" invoice line with a unit price of -0.02 is created before the tax is calculated (which seems to be correct although it might make more sense to create the rounding line only after the tax is calculated).
selection_185

When the tax is calculated, the rounding is no longer necessary and thus is not considered in the amount_untaxed but the rounding invoice line still exists which can cause a lot of confusion.
selection_186

BT-tbaechle:
Attempt to fix #379

### DIFF
--- a/account_invoice_rounding/account.py
+++ b/account_invoice_rounding/account.py
@@ -132,7 +132,7 @@ class AccountInvoice(models.Model):
         if float_compare(rounded_total, invoice.amount_total,
                          precision_digits=prec) == 0:
             if invoice.global_round_line_id:
-                invoice.global_round_line_id.unlink()
+                invoice.write({'invoice_line': [[2, invoice.global_round_line_id.id, False]]})
             return {}
 
         # To avoid recursivity as we need to write on invoice or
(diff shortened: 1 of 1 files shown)


BT-tbaechle:
@yvaucher in my local installation with the current Odoo core version this works just fine. However on the runbot I get an error, I suspect that the runbot is built using an older commit (4045e4fe6 see https://runbot.odoo-community.org/runbot/static/build/3332824-406-e35b05/logs/job_10_test_base.txt). If I switch back to this commit as well I get the same error. Is there a way to make sure the latest Odoo core commit is used for the runbot build?

yvaucher:
@BT-tbaechle you can probably trigger a new build by doing a rebase of this PR

PR #409, opened 2018-07-31
[FIX][8.0] account_invoice_pricelist

legalsylvain:
Before the PR
when creating a invoice from the sale.order -> stock.picking workflow the pricelist of the sale order is not recovered.

this come from a call to self in a api.model function that return no sale. 

CC : @quentinDupont

### DIFF
--- a/account_invoice_pricelist_sale_stock/models/stock_picking.py
+++ b/account_invoice_pricelist_sale_stock/models/stock_picking.py
@@ -13,8 +13,9 @@ class StockPicking(models.Model):
         inv_vals = super(StockPicking, self)._get_invoice_vals(
             key, inv_type, journal_id, move
         )
-        if self.sale_id and self.sale_id.pricelist_id:
+        sale = move.picking_id.sale_id
+        if sale and sale.pricelist_id:
             inv_vals.update({
-                'pricelist_id': self.sale_id.pricelist_id.id,
+                'pricelist_id': sale.pricelist_id.id,
             })
         return inv_vals


account_invoice_pricelist_sale_stock/models/stock_picking.py
```diff
@@ -13,8 +13,9 @@ def _get_invoice_vals(self, key, inv_type, journal_id, move):
         inv_vals = super(StockPicking, self)._get_invoice_vals(
             key, inv_type, journal_id, move
         )
-        if self.sale_id and self.sale_id.pricelist_id:
+        sale = move.picking_id.sale_id
+        if sale and sale.pricelist_id:
```
pedrobaeza, line 17: You can put directly `if sale.pricelist_id:`

pedrobaeza (approved):

account_invoice_pricelist_sale_stock/models/stock_picking.py
legalsylvain: no because sale is not defined always.

legalsylvain:
note : I'm working on a second commit to set the pricelist of the partner, if there is no sale defined.

legalsylvain:
Wow. it really more complicated. The whole module is bad designed. and doesn't handle purchase pricelist workflow.
the module should be named account_invoice_pricelist_stock_account in fact.

closing this PR and I'll will create a new one.

account_invoice_pricelist_sale_stock/models/stock_picking.py
pedrobaeza: Although not set, new API returns an empty recordset and then you can query the same `sale.pricelist_id`

legalsylvain: Ow, thanks for the Tips !
I did'nt know.

regards.

legalsylvain:
superseded by https://github.com/OCA/account-invoicing/pull/410.

PR #411, opened 2018-08-01
[ADD] New module: account_invoice_tax_note

acsonefho:
**New:**
- module: account_invoice_tax_note
- Add unit test

**Description:**
In some situations, a mention must be displayed on invoices when a specific tax is used.
This module lets you define such a mention on Tax Groups.
This mechanism complements the invoice note coming from the fiscal position.
These tax group notes are translatable.


account_invoice_tax_note/README.rst
```diff
@@ -0,0 +1,56 @@
...
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit https://odoo-community.org.
```
sbidoul: Can you please use the new readme structure?

account_invoice_tax_note/__manifest__.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2018 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    'name': "Account invoice tax note",
+    'version': "11.0.1.0.0",
+    "author": "ACSONE SA/NV,Odoo Community Association (OCA)",
+    "summary": "In some situations, a mention must be displayed on invoices "
+               "when a specific tax is used.\nThis module lets you define such"
+               " a mention on Tax Groups.\nThis mechanism complements the "
+               "invoice note coming from the fiscal position.",
```
sbidoul: This is a bit long for the summary. I'd say "Print tax notes on customer invoices"

account_invoice_tax_note/models/account_invoice.py
```diff
@@ -0,0 +1,27 @@
...
+        company = invoices.mapped("company_id")
+        if len(company) > 1:
+            # If we have more than 1 company, it's because a user with
+            # full access (ex: admin) see invoices mixed between companies
+            message = _("To print invoices, they must belong to the same "
+                        "company!")
+            raise exceptions.UserError(message)
+        if company.id != self.env.user.company_id.id:
+            # Update the context only if it's necessary
+            invoices = invoices.with_context(force_company=company.id)
```
sbidoul: these company checks may be a bit overkill? Why not doing a `ensure_one()`, since this method is always used in the context of a single invoice.

account_invoice_tax_note/reports/report_invoice_document.xml
```diff
@@ -0,0 +1,16 @@
...
+<!-- Copyright 2018 ACSONE SA/NV
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+<odoo>
+    <template id="report_invoice_document_inherit" inherit_id="account.report_invoice_document">
+        <p t-if="o.fiscal_position_id.note" position="after">
+            <t t-set="account_tax_groups" t-value="o._get_account_tax_groups()"/>
+            <t t-if="account_tax_groups">
+                <t t-foreach="account_tax_groups" t-as="account_tax_group">
+                    <p t-field="account_tax_group.report_note"/>
+                </t>
```
sbidoul: skip empty notes?

sbidoul (changes requested):
Thanks! A few comments.

account_invoice_tax_note/reports/report_invoice_document.xml
acsonefho: Fixed (into _get_account_tax_groups() function of account.invoice)

account_invoice_tax_note/models/account_invoice.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class AccountInvoice(models.Model):
+    _inherit = 'account.invoice'
+
+    @api.multi
+    def _get_account_tax_groups(self):
+        """
+        Get every account.tax.group used on current recordset where
+        report_note is filled.
```
sbidoul: I suggest renaming `_get_account_tax_groups_with_notes()` (and removing the docstring comment which become redundant with the method name).

Because in the template you still read a call to _get_account_tax_groups and you don't immediately understand there is a filtering behind.

account_invoice_tax_note/models/account_invoice.py
```diff
@@ -0,0 +1,26 @@
...
+        Force the company into context if invoices are not in the user company
+        :return: account.tax.group recordset
+        """
+        invoices = self
+        company = invoices.mapped("company_id")
+        company.ensure_one()
+        if company.id != self.env.user.company_id.id:
+            # Update the context only if it's necessary
+            invoices = invoices.with_context(force_company=company.id)
```
sbidoul: I suggest `self.ensure_one()` and not testing for the company at all (or maybe there is a use case for using this method on a recordset that I don't understand).

pedrobaeza:
Are you sure this should be attached to tax groups instead of taxes? I'm feeling that having this as group level makes this more complicated (with company_dependent fields and all the possible problems with sudo operations), and reduce the possibilities.

sbidoul:
@pedrobaeza I don't think anything needs to be company dependent here. I see @acsonefho removed it in the last commit.

I proposed to put it on tax group because it's what Odoo uses in the tax summary of the standard invoice template, and to avoid repeating the same note on several taxes of the same group. That seems to be the simplest for the use case we have mind. 

Is it ok for you if we evolve it later if/when we discover uses cases for putting the note at the tax level?

sbidoul (approved):
Thanks!

pedrobaeza:
Well, I was thinking on several possible notes depending on specific taxes of the same group. For example, we have on Spain a tax for intra-community VAT (that charges and deduces at the same time the same tax percentage - this is a trick for presenting the VAT declaration - which has to be done this way), and it's in the same group as the regular taxes (21% VAT, 10% VAT or 4% VAT). The intra-community one can have a text explaining the application of it (we have one customer that forces us to put this text "hard-coded" by conditions), but not the rest, so if we want to use this module, we must change `l10n_es` to accommodate it to different taxes groups.

Any way, if it's out of your budget/use case, I understand that you can't perform it.

sbidoul:
For intra community we have the same, but the note for that comes from the fiscal position.

account_invoice_tax_note/models/account_invoice.py
```diff
@@ -0,0 +1,15 @@
...
+
+class AccountInvoice(models.Model):
+    _inherit = 'account.invoice'
+
+    @api.multi
+    def _get_account_tax_groups_with_notes(self):
+        self.ensure_one()
+        tax_groups = self.mapped("tax_line_ids.tax_id.tax_group_id")
+        tax_groups = tax_groups.filtered(
+            lambda g: g.report_note).with_prefetch(self._prefetch)
```
sbidoul: last question: why this with_prefetch?

account_invoice_tax_note/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2018 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    'name': "Account invoice tax note",
+    'version': "11.0.1.0.0",
+    "author": "ACSONE SA/NV,Odoo Community Association (OCA)",
+    "summary": "Print tax notes on customer invoices",
+    'website': "https://odoo-community.org",
```
MeetKD: Website should be `https://github.com/OCA/account-invoicing`

sbidoul:
@MeetKD good catch, thanks! fixed, rebased and squashed.

MeetKD (approved):

etobella (approved):
Great proposal!! 👍

JordiBForgeFlow (approved):
Great module! :+1:

sbidoul:
Sweet. Thanks all!

PR #415, opened 2018-08-28
[11.0][NEW] stock_picking_return_refund_option: New module for set refund options in picking directly

sergio-teruel:
cc @Tecnativa
This module is extracted from this one:
https://github.com/OCA/account-invoicing/tree/9.0/sale_stock_picking_return_invoicing
I need help to migration scripts... :smiley:


stock_picking_return_refund_option/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+    'version': '11.0.1.0.0',
+    'development_status': 'Production/Stable',
+    'category': 'Sales',
+    'website': 'https://github.com/OCA/account-invoicing',
+    'author': 'Tecnativa, Odoo Community Association (OCA)',
+    'license': 'AGPL-3',
+    'application': False,
+    'installable': True,
+    'depends': [
+        'sale_stock',
```
pedrobaeza: It should depend on `stock_account`, as the field is now defined in that module and for allowing to use this option also on purchases.

stock_picking_return_refund_option/models/stock_move.py
```diff
@@ -0,0 +1,9 @@
+# Copyright 2018 Tecnativa - Sergio Teruel
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import fields, models
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    origin_to_refund_so = fields.Boolean("Origin To Refund in SO")
```
pedrobaeza: Call it simply `origin_to_refund`

stock_picking_return_refund_option/models/stock_picking.py
```diff
@@ -0,0 +1,50 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import api, fields, models
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    to_refund_lines = fields.Selection(selection=[
+        ('keep_line_value', 'Keep Line Value'),
+        ('to_refund_so', 'To Refund In Sales Order'),
```
pedrobaeza: Remove all the sales references

stock_picking_return_refund_option/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,2 @@
+This module extends the functionality of sales orders to support modify stock
```
pedrobaeza: Change text for talking about picking returns in general

stock_picking_return_refund_option/views/stock_picking_view.xml
```diff
@@ -0,0 +1,19 @@
...
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<odoo>
+
+<record id="view_picking_form" model="ir.ui.view">
+    <field name="inherit_id" ref="stock.view_picking_form"/>
+    <field name="model">stock.picking</field>
+    <field name="arch" type="xml">
+        <xpath expr="//field[@name='owner_id']" position="after">
+            <field name="to_refund_lines"
+                   attrs="{'invisible': [('picking_type_code', '!=', 'incoming')]}"
```
pedrobaeza: Instead of this condition, you should put one that hides it when this is not a return.

pedrobaeza (changes requested):

stock_picking_return_refund_option/__manifest__.py
sergio-teruel: The method _get_delivered_qty declared on sale_stock is used..

stock_picking_return_refund_option/README.rst
```diff
@@ -0,0 +1,102 @@
...
+
+**Table of contents**
+
+.. contents::
+   :local:
+
+Usage
+=====
+
+To use this module, when some customer returns some refundable products to you
```
pedrobaeza, line 45: Include INSTALL.rst file specifying that this module requires the additional installation of `sale` or `purchase` for enabling the features it contains.

You have to add also instructions here for the purchase part.

stock_picking_return_refund_option/hooks.py
```diff
@@ -0,0 +1,21 @@
...
+def pre_init_hook(cr):
+    if not tools.config.options.get('without_demo', False):
+        env = api.Environment(cr, SUPERUSER_ID, {})
+        IrModule = env['ir.module.module']
+        purchase = IrModule.search([
+            ('name', '=', 'purchase'),
+            ('state', '!=', 'installed'),
+        ])
+        if purchase:
+            purchase.button_immediate_install()
```
pedrobaeza: You can't do this, but write `state = 'to install'` and let second pass to act.

stock_picking_return_refund_option/models/stock_picking.py
```diff
@@ -0,0 +1,69 @@
...
+            if any(x.origin_returned_move_id for x in picking.move_lines):
+                picking.is_return = True
+
+    @api.multi
+    def _update_stock_moves(self):
+        for pick in self.filtered(
+                lambda x: x.to_refund_lines != 'keep_line_value'):
+            pick.move_lines.write({
+                'to_refund':
+                    True if pick.to_refund_lines == 'to_refund' else False,
```
pedrobaeza: Put directly:

```python
'to_refund': pick.to_refund_lines == 'to_refund',
```

pedrobaeza (changes requested):

sergio-teruel:
Changes done..

stock_picking_return_refund_option/__manifest__.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2018 Tecnativa - Sergio Teruel
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    'name': 'Stock Picking Return Refund Option',
+    'summary': 'Update the refuund options in pickings',
```
pedrobaeza: s/refuund/refund

stock_picking_return_refund_option/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+    'version': '11.0.1.0.0',
+    'development_status': 'Production/Stable',
```
pedrobaeza: Beta

stock_picking_return_refund_option/__init__.py
```diff
@@ -0,0 +1,5 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from .hooks import _install_module_hook
```
pedrobaeza: It's advisable to name the method the same way as the hook is using (`pre_init_hook`)

pedrobaeza (approved):

pedrobaeza:
@sergio-teruel can you finish this one?

@rafaelbn please review

stock_picking_return_refund_option/tests/test_stock_picking_return_refund_options.py
```diff
@@ -0,0 +1,107 @@
+# Copyright 2018 Tecnativa - Sergio Teruel
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo.tests import SavepointCase
+from odoo import fields
+
+
+class TestSaleOrderLineInput(SavepointCase):
+    at_install = False
```
lmignon: @sergio-teruel con you remove `at_install` and `post_install` plz.

stock_picking_return_refund_option/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+    'license': 'AGPL-3',
+    'application': False,
+    'installable': True,
+    'depends': [
+        'stock_account',
+    ],
+    'data': [
+        'views/stock_picking_view.xml',
+    ],
+    'pre_init_hook': '_install_module_hook',
```
lmignon: No maintainer? https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/oca_module_lifecycle_maintainer_role.rst

stock_picking_return_refund_option/hooks.py
```diff
@@ -0,0 +1,12 @@
+# Copyright 2018 Tecnativa - Sergio Teruel
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import api, SUPERUSER_ID, tools
+
+
+def _install_module_hook(cr):
+    if not tools.config.options.get('without_demo', False):
+        env = api.Environment(cr, SUPERUSER_ID, {})
+        env['ir.module.module'].search([
```
lmignon, line 9: Why do you install manually these addons? If these addons must be installed, these should be declared into the manifest under the `depends`key. So it's clear for the user by reading the manifest that this module will install others modules.

stock_picking_return_refund_option/models/stock_picking.py
```diff
@@ -0,0 +1,77 @@
...
+            move.to_refund = move.origin_to_refund
+
+    @api.multi
+    def set_delivered_qty(self):
+        """
+        Check if exists sale_line_id field in stock.move model that has been
+        added by sale_stock module, this module has not dependency of this,
+        Update sale order line qty_delivered for allow do a refund invoice
+        """
+        if hasattr(self.env['stock.move'], 'sale_line_id') and self.sale_id:
```
lmignon, line 65: In the init hook you install sale_stock. Here you write code checking if the module is installed. If your code work in a different way if the module `sale_stock` is installed and the module doens't depend of `sale_stock` you must define a glue addon with the specific code.

stock_picking_return_refund_option/models/stock_picking.py
```diff
@@ -0,0 +1,77 @@
...
+                so_line.qty_delivered = so_line._get_delivered_qty()
+
+    @api.multi
+    def set_received_qty(self):
+        """
+        Check if exists purchase_line_id field in stock.move model that has
+        been added by purchase module, this module has not dependency of this,
+        Update purchase order line qty_received for allow do a refund invoice.
+        """
+        if (hasattr(self.env['stock.move'], 'purchase_line_id') and
```
lmignon, line 79: Same as https://github.com/OCA/account-invoicing/pull/415/files#diff-1b085086fe8c6305c2cf2e9ea414a25fR48

lmignon (changes requested):

stock_picking_return_refund_option/hooks.py
pedrobaeza: This is a trick for not depending on both modules and not creating 2 glue modules, but testing everything.

stock_picking_return_refund_option/models/stock_picking.py
pedrobaeza: Sometimes you need only the sales part, sometimes only the purchase part, and sometimes both. With this you avoid 2 ugly glue modules and perform the same without problems.

stock_picking_return_refund_option/hooks.py
lmignon: You are wrong. The final result is that these modules are isntalled.

pedrobaeza: Only in demo mode, not in production

stock_picking_return_refund_option/models/stock_picking.py
lmignon: @pedrobaeza It's not a good practice at all. We must ensure a proper separation between modules and this kind of code is not sustainable.

stock_picking_return_refund_option/hooks.py
lmignon: I strongly disagree with this kind of practice. Tests are  always run in a wrong context.

stock_picking_return_refund_option/models/stock_picking.py
pedrobaeza: I don't agree. I prefer this option. This technique avoids a lot of module overhead and it's sustainable through the trick in testing (we are testing all the code). In upper versions, if for example a field name changes, we will detect it the same as having 3 modules.

We have spent time assuring that all code is tested and thus the `_install_module_hook` trick. You can't dismiss this solution if it's only a matter of choice. Technically, it has no problem. Functionally neither.

stock_picking_return_refund_option/hooks.py
pedrobaeza: Why a wrong context?

stock_picking_return_refund_option/models/stock_picking.py
lmignon: Conceptually it has.

stock_picking_return_refund_option/hooks.py
lmignon: Either the module depends on these two other modules or it does not depend on them. If it does not depend on it, we must be able to test them without them being installed. With this trick it will never be possible to test this module without both dependencies and thus ensure that it can work without them.

pedrobaeza: There's no code to test if the modules are not installed, so it's not a problem. It's only a new field, that we put together for not needing a base module with the field plus 2 extra modules with the logic.

stock_picking_return_refund_option/models/stock_picking.py
```diff
@@ -0,0 +1,77 @@
+# Copyright 2018 Tecnativa - Sergio Teruel
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import api, fields, models
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    to_refund_lines = fields.Selection(selection=[
```
pedrobaeza, line 9: @sergio-teruel I'm thinking that this can be a field with `compute` and `inverse` with only the 2 possible values, so that you also have a visibility of the refund option. Right now, when you perform a return, you don't know which is the selected value.

sergio-teruel:
Changes done

stock_picking_return_refund_option/models/stock_picking.py
sergio-teruel: I'm not sure... this record is picking and to_refund field is in sotck.move so you can have a picking with stock_move to refund and others to not refund, in this case, What value would we assign to to_refund_lines field in picking?.
I think that we need three values.. Also we can do visible to refund at stock_move level.
What do you think?

sergio-teruel:
@pedrobaeza Changes done.

rafaelbn (approved):
:+1:

rafaelbn:
Please @sergio-teruel  could make las push here in order to merge?

>  codecov/patch — 88.23% of diff hit (target 94.14%) 

:thinking:

stock_picking_return_refund_option/models/stock_picking.py
pedrobaeza: We talked about this directly, but probably don't remember it. The result of the discussion was that the field must be `compute` and `inverse` with that 2 values + no value, that means that there are a mix, but when you set it one time, there's no sense nor possibility of going again to the "mix option".

sergio-teruel:
@rafaelbn 100% of coverage

rafaelbn (approved):
Nice :green_heart: :100:

rafaelbn:
I know sorry!  Do you know how to make that @OCA-git-bot re-review and old PR?

rafaelbn:
Hello @sergio-teruel @pedrobaeza could you please answer @lmignon in order to agree and merge this one? Thanks a lot!

pedrobaeza:
It's already answered, @rafaelbn

stock_picking_return_refund_option/i18n/es.po
```diff
@@ -0,0 +1,86 @@
...
+#: selection:stock.picking,to_refund_lines:0
+msgid "To Refund"
+msgstr "Para abonar"
+
+#. module: stock_picking_return_refund_option
+#: model:ir.model,name:stock_picking_return_refund_option.model_stock_picking
+msgid "Transfer"
+msgstr "Transferir"
+
+#~ msgid "Keep Line Value"
```
pedrobaeza: Why these lines?

stock_picking_return_refund_option/models/stock_move.py
```diff
@@ -0,0 +1,9 @@
+# Copyright 2018 Tecnativa - Sergio Teruel
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import fields, models
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    origin_to_refund = fields.Boolean("Origin To Refund")
```
pedrobaeza: This field shouldn't be needed with new approach of compute/inverse

pedrobaeza (changes requested):

sergio-teruel:
Changes done and rebase

pedrobaeza (approved):

PR #417, opened 2018-08-29
[FIX] Reimbursables

etobella:
The accounting lines where added on `invoice_line_move_line_get`, but that was a problem with some modules. They should be added on `compute_invoice_totals`.

### DIFF
--- a/account_invoice_reimbursable/__manifest__.py
+++ b/account_invoice_reimbursable/__manifest__.py
@@ -3,7 +3,7 @@
 
 {
     'name': 'Reimbursables management',
-    'version': '11.0.1.0.0',
+    'version': '11.0.1.0.1',
     'summary': 'Create the option to add reimbursables on invoices',
     'author': "Creu Blanca, Odoo Community Association (OCA)",
     "category": "Accounting & Finance",
--- a/account_invoice_reimbursable/models/account_invoice.py
+++ b/account_invoice_reimbursable/models/account_invoice.py
@@ -96,25 +96,41 @@ class AccountInvoice(models.Model):
         self.reimbursable_ids._onchange_partner_id()
         return res
 
-    @api.model
-    def invoice_line_move_line_get(self):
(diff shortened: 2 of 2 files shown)


pedrobaeza:
I think it should be more in `finalize_invoice_move_lines`, which the docstring says is for this purpose.

etobella:
Well, it's true that it should be added on `finalize_invoice_move_lines`, but we must edit `compute_invoice_totals` in order to recompute the total.

pedrobaeza:
Uhm, I see this a bit cumbersome... Maybe the initial approach was not the best. I didn't review the module then, but in my mind, reimbursable should be invoice lines with a product marked some way and an extra partner column for indicating to whom reimburse. That way, you don't have to change too much logic in invoice ORM.

etobella:
I tried to use them as invoice lines, but at the end, I refused the idea. They have a complete different behaviour, because they need a partner and no taxes should be assigned. In fact, each reimbursable should be assigned to a different invoice and the total of the original invoice should not change.
The invoice total should be the amount of the invoice of the vendor, but the executable amount is the total amount that you will pay to the vendor. If reimbursables are included as lines, the total amount will not be the real total amount.
I checked the information with spanish facturae format in order to be sure which was the better approach.

pedrobaeza:
OK, I suppose I haven't dug enough in this, so continue as is. I have to use this in a near future, so I'll tell my feedback once reviewed.

JordiBForgeFlow (approved):
:+1: Technical and functional review.

JordiBForgeFlow:
@etobella Can you squash your commits? IMHO this one is ready to merge!

etobella:
@jbeficent It is rebased now

JordiBForgeFlow:
@pedrobaeza Do you agree with this PR?

AaronHForgeFlow (approved):
Functional tests on runbot. :+1:  Thanks.

PR #420, opened 2018-09-05
Migrated module account_invoice_merge from branch 10.0 to branch 11.0

dsmerghetto:


pedrobaeza:
There has been another PR in the same time for the same module: #421 

Any way, you should preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-11.0

I'm closing both until one of them is correct in this.

PR #421, opened 2018-09-05
IMP:	- Migrated module account_invoice_merge to branch 11.0

OmniaGit:
Migrated module account_invoice_merge to branch 11.0


pedrobaeza:
There has been another PR in the same time for the same module: #420 

Any way, you should preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-11.0

I'm closing both until one of them is correct in this.

PR #425, opened 2018-09-19
[8.0][FIX] account_invoice_transmit_method: fix typo

tafaRU:
Replace company_dependant by company_dependent


astirpe (approved):
Please bump the version

tafaRU:
@astirpe, versione bumped and rebased!

pedrobaeza (changes requested):
You can't do this without:

* Using another field name, as this operation removes immediately the columns and makes you lose all data (no pre-migration renaming option)
* Creating a migration script that moves data from column to ir.property.

PR #426, opened 2018-09-19
[FIX] account_invoice_anglo_saxon_no_cogs_deferral wrong version in __manifest__.py 

emagdalenaC2i:


AaronHForgeFlow (approved):
I think we should fast-track this

PR #435, opened 2018-10-02
[11.0][IMP] account_invoice_force_number: forward port of https://github.com/OCA/account-invoicing/pull/350

tafaRU:


bodedra (approved):
LGTM :+1:

account_invoice_force_number/views/account_invoice_view.xml
```diff
@@ -1,6 +1,21 @@
-<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0"?>
```
pedrobaeza: Don't remove encoding. This is needed for XML (but not for .py).

pedrobaeza (changes requested):

account_invoice_force_number/views/account_invoice_view.xml
tafaRU: ok, thanks! Done in https://github.com/OCA/account-invoicing/pull/435/commits/b0bafe08770b8dab03aabf913d20d5e46c6505da

pedrobaeza (approved):

PR #439, opened 2018-10-18
 [12.0] account_invoice_change_currency migrate to v12.0

hugho-ad:


account_invoice_change_currency/README.rst
```diff
@@ -0,0 +1,54 @@
...
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://odoo-community.org/logo.png>`_.
+
+Contributors
+------------
+
+* Hugo Adan <hugo@vauxoo.com>
```
moylop260: And the other ones?

account_invoice_change_currency/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,5 @@
+Configuration
```
moylop260, line 1: What about merge all files in the README file instead?

Or Are there changes in the guidelines that Im missing?
cc @pedrobaeza @sbidoul

account_invoice_change_currency/README.rst
hugho-ad: done

moylop260: I think that you need use the other rst files since that this module is using splitted files

account_invoice_change_currency/readme/CONFIGURE.rst
hugho-ad: @pedrobaeza @sbejaoui 

we need your input here.

Regards!

moylop260: Hi @hugho-ad 
I just find the thread about:
 - https://github.com/OCA/maintainer-tools/pull/339

You need to use the current template:
 - https://github.com/OCA/maintainer-tools/tree/master/template/module/readme

moylop260 (approved):

moylop260:
I just tested it functionality from runbot and it is working like expected:
 - invoice

@hugho-ad For another PR you should move the button from header to next to currency field with a good icon. What do you think?

For now, I will merge it since that it is a migrated module working fine.

PR #440, opened 2018-10-22
[10.0] account_invoice_transmit_method: use base_view_inheritance_extension

alexis-via:


rafaelbn (approved):

pedrobaeza:
The problem I see with this is that you break current installations if they don't have that module installed. For this stable version, I'm afraid you need to overwrite `fields_view_get` method add inject values dynamically. You can see an example in https://github.com/OCA/sale-workflow/pull/685/files#diff-aeeeb49cc9f7ca9243fa4dd88acfd11fR28

bguillot (approved):

pedrobaeza (changes requested):
See my previous comment

mlaitinen:
Shouldn't this PR also add `server-tools` to oca-dependencies.txt?

EDIT: of course, if the change is done by overriding `fields_view_get`, then `server-tools` is not needed. Please ignore my comment.

alexis-via:
@mlaitinen I added server-tools in oca_dependencies.

alexis-via:
Adding a dependancy in a module in a stable branch is something that has already been done on several OCA modules, including major modules used by a lot of users, for example account_financial_report_qweb, cf this PR https://github.com/OCA/account-financial-reporting/pull/485/files

pedrobaeza:
If it has been previously done doesn't mean that is a good technique. And more having a reasonable alternative.

I'm personally not affected, as I have migrated now all customers to upper version, but this is not for us "advanced users" that know how to deal with the problem, but OK, you have merged it, so we'll see how this affects people.

PR #441, opened 2018-10-23
Missed @api.multi

FullBringEc:
TypeError: assign_outstanding_credit() takes exactly 2 arguments (5 given)


account_outstanding_payment/models/account_invoice.py
```diff
@@ -7,8 +7,9 @@
 from openerp.tools.translate import _
 import json
 from openerp.tools.float_utils import float_is_zero
+import logging
 
-
+logging.basicConfig(level=logging.DEBUG)
```
moylop260: This change is out-of-scope of the description of the PR

FullBringEc: Yes, sorry I was testing and forgot to delete that line.

PR #444, opened 2018-10-24
[REF] Move change currency button to a better place inside the invoice form

hugho-ad:
fix https://github.com/OCA/account-invoicing/issues/442
fix https://github.com/OCA/account-invoicing/issues/443

PR #446, opened 2018-10-30
[REF] Move change currency button to a better place inside the invoice form

hugho-ad:


moylop260:
Merging since that is a backport of an approved feature from:
 - https://github.com/OCA/account-invoicing/pull/444

PR #454, opened 2018-11-25
12.0 mig sale_timesheet_invoice_description

ernestotejeda:
Cc @Tecnativa


pedrobaeza:
Please comment when Travis is green.

ernestotejeda:
Travis in green

sale_timesheet_invoice_description/models/sale.py
```diff
@@ -29,7 +29,12 @@ class SaleOrderLine(models.Model):
     def _prepare_invoice_line_details(self, line, desc_rule):
         details = []
         if desc_rule[0] == '1':
-            details.append(line.date)
+            # If project_timesheet_time_control module is installed,
+            # date field is hidden and the new date_time field is used instead
+            if hasattr(line, 'date_time'): # pragma: no cover
```
pedrobaeza: Although the field is hidden, `date` field is still synchronized, so I prefer to keep the previous behavior showing only date. Customer doesn't need this kind of detail.

sale_timesheet_invoice_description/models/sale.py
```diff
@@ -47,9 +52,14 @@ def _prepare_invoice_line(self, qty):
...
-        for line in self.env['account.analytic.line'].search(
-                domain, order='date, id'):
+            last_date = fields.Datetime.to_string(last_invoice.create_date)
+            domain.append(('create_date', '>', last_date))
+
+        # If project_timesheet_time_control module is installed,
+        # date field is hidden and the new date_time field is used instead
+        line_obj = self.env['account.analytic.line']
+        order = hasattr(line_obj, 'date_time') and 'date_time' or 'date'
+        for line in line_obj.search(domain, order=order + ', id'):
```
pedrobaeza: Why don't rely in default order set by the model instead (and more now that we are going to set it properly)?

sale_timesheet_invoice_description/models/sale.py
ernestotejeda: Yes, I agree, thanks

sale_timesheet_invoice_description/models/sale.py
ernestotejeda: I think you're right, thank you

pedrobaeza (approved):

rafaelbn (approved):
Tested in runbot :+1:

PR #458, opened 2018-11-30
[FIX] account_invoice_change_currency: the custom_rate field has a bad position in supplier invoices.

### LINKED ISSUE #457: [12.0][account_invoice_change_currency] The custom_rate field has a bad position in supplier invoices.
**Supplier Invoice:**

image

hugho-ad:
fix https://github.com/OCA/account-invoicing/issues/457

image

### DIFF
--- a/account_invoice_change_currency/views/account_invoice.xml
+++ b/account_invoice_change_currency/views/account_invoice.xml
@@ -22,11 +22,18 @@
         <field name="model">account.invoice</field>
         <field name="inherit_id" ref="account.invoice_supplier_form"/>
         <field name="arch" type="xml">
-            <xpath expr="//form/header" position="inside">
-                - <field name="custom_rate" placeholder="Custom Rate" class="oe_inline" attrs="{'invisible': [('custom_rate', '=', 0)], 'readonly': [('state', '!=', 'draft')]}"/>
-                <button aria-label="Opened" type="object" name="action_account_change_currency" class="btn-link" states="draft" groups="account.group_account_manager">
-                    <i title="Update Currency: Will convert the amount lines to the new currency set" class="fa fa-fw fa-refresh"/>
-                </button>
+            <xpath expr="//field[@name='currency_id']" position="attributes">
+                <attribute name="invisible">1</attribute>
+            </xpath>
+            <xpath expr="//field[@name='currency_id']" position="after">
+                <label for="currency_id" groups="base.group_multi_currency"/>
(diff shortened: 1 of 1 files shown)


account_invoice_change_currency/views/account_invoice.xml
```diff
@@ -22,11 +22,15 @@
         <field name="model">account.invoice</field>
         <field name="inherit_id" ref="account.invoice_supplier_form"/>
         <field name="arch" type="xml">
-            <xpath expr="//form/header" position="inside">
-                - <field name="custom_rate" placeholder="Custom Rate" class="oe_inline" attrs="{'invisible': [('custom_rate', '=', 0)], 'readonly': [('state', '!=', 'draft')]}"/>
-                <button aria-label="Opened" type="object" name="action_account_change_currency" class="btn-link" states="draft" groups="account.group_account_manager">
-                    <i title="Update Currency: Will convert the amount lines to the new currency set" class="fa fa-fw fa-refresh"/>
-                </button>
+            <xpath expr="//field[@name='currency_id']" position="replace">
```
moylop260: Is there a way without replace?

hugho-ad: No, We need this as is it

moylop260: The problem is that if there is another side inheriting this view and this fields then will be overwritten

What about a `position=after`?

hugho-ad: No, because field currency must be inside a div like:
https://github.com/odoo/odoo/blob/12.0/addons/account/views/account_invoice_view.xml#L456-L460

pedrobaeza: Don't replace, but hide elements instead

hugho-ad: what do you mean by hiding it?

Instead, I better put it in a new line after currency field

pedrobaeza: Using `position='attributes'` and `invisible=1`

hugho-ad: Got it, Changes done

moylop260:
@yaniaular Could you check it, please?

yaniaular:
@moylop260 Can you merge this please?

yaniaular (approved):

moylop260 (approved):

PR #419, opened 2018-09-04
[11.0][IMP] account_invoice_check_total: improve ux

astirpe:
One of our customer asked us the following changes, in order to improve the user experience of the module:

- [x]  Improve the error message, displaying the total amounts and their difference
- [x] Display the check total amount difference after the invoice total amount

This PR includes https://github.com/OCA/account-invoicing/pull/418

PR #459, opened 2018-12-03
[12.0][IMP] account_invoice_check_total: improve ux

astirpe:
This is the porting of PR #419 to V12.


yung-wang (approved):
Nice and works fine in V12

pedrobaeza (approved):

PR #462, opened 2018-12-10
[ADD] account_invoice_analytic_line

hbrunn:
This module is a rewrite of the deprecated v8 module hr_timesheet_invoice.

It allows you to invoice analytic lines derived from time recorded on project tasks to the project's customer without a detour via sales orders. The invoice analytic lines will be attached to the invoice document when printed.


account_invoice_analytic_line/readme/USAGE.rst
```diff
@@ -0,0 +1,3 @@
+To use this module, you need to:
+
+#. go to `Accounting / Invoice analytic lines`, select the lines you want to invoive, and click `More / Invoice lines`
```
thomaspaulb: invoice

thomaspaulb (changes requested):

account_invoice_analytic_line/readme/USAGE.rst
thomaspaulb: it's only a typo, but it's still there, so unresolving this

ntodorova:
@hbrunn is there a runbot instance for this branch? I would like to review and test it.

Regards
Nikolina

account_invoice_analytic_line/models/account_analytic_line.py
```diff
@@ -0,0 +1,112 @@
...
+            'invoice_id': invoice.id,
+            'discount': first.invoice_discount_id.discount,
+            'quantity': quantity,
+            'price_unit': product.with_context(
+                pricelist=invoice.partner_id.property_product_pricelist.id,
+                partner=invoice.partner_id.id,
+                quantity=quantity,
+            ).price,
+        }
+        invoice_line_vals = self.env['account.invoice.line'].play_onchanges(
```
ntodorova: Is the play_onchanges function coming from this repo: https://github.com/akretion/ak-odoo-incubator
Is this mentioned somewhere?

account_invoice_analytic_line/models/account_analytic_line.py
```diff
@@ -0,0 +1,112 @@
...
+# Copyright 2018 Therp BV <https://therp.nl>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError
+
+
+class AccountAnalyticLine(models.Model):
+    _inherit = 'account.analytic.line'
+
+    invoice_discount_id = fields.Many2one(
```
ntodorova, line 11: Do we really need a separate object for discount? On the invoice this is only a float field and I don`t see any special  functionality or so implemented in the  account.invoice.analytic.line.discount module. But if we have an object one needs to create a Discount record each time one use different number.

account_invoice_analytic_line/models/account_analytic_line.py
```diff
@@ -0,0 +1,112 @@
...
+        }
+        invoice_line_vals = self.env['account.invoice.line'].play_onchanges(
+            invoice_line_vals,
+            ['product_id', 'price_unit', 'discount', 'quantity'],
+        )
+        invoice_line_vals['name'] = first.project_id.name
+        invoice_line = self.env['account.invoice.line'].create(
+            invoice_line_vals,
+        )
+        self.write({'invoice_line_id': invoice_line.id})
```
ntodorova, line 98: Why do we need this write?

ntodorova (commented):
Code looks good, but I was not able to test because of missing instance in runbot

hbrunn:
@ntodorova please read https://odoo-community.org/page/review, this will explain how to recreate the runbot instance

account_invoice_analytic_line/models/account_analytic_line.py
hbrunn: the module depends on https://github.com/OCA/server-tools/tree/10.0/onchange_helper

ntodorova:
@hbrunn, I have read it - it is written "On the next page, click on the blue square (if there is none, leave a comment on Github) and log into Odoo with the following credentials:". I don`t see the icon, I have force rebuild, but also nothing happens

account_invoice_analytic_line/models/account_analytic_line.py
hbrunn: for our use case, we only have fixed discounts, and we want to control which discounts employees can select

hbrunn:
you need to wait a few minutes, this should be done 10 or so

account_invoice_analytic_line/models/account_analytic_line.py
hbrunn: note that `self` contains the analytic lines we've just created the invoice line for. So this takes care that we have a link between the analytic lines and the invoice line created. This is used to determine which analytic lines are not yet invoiced, for example in https://github.com/OCA/account-invoicing/pull/462/files/ad8bc705b3d7e279a2bdbf75d26dc2641d856d71#diff-c1268f2774a7fdc6bcc96f5e1bcedee9R21

account_invoice_analytic_line/models/account_analytic_line.py
ntodorova: Wouldn't be nice then to have an oca_dependencies.txt file like for example this one https://github.com/OCA/connector/blob/12.0/oca_dependencies.txt?

hbrunn: this repo depends on account-financial-tools and this again depends on server-tools: https://github.com/OCA/account-financial-tools/blob/10.0/oca_dependencies.txt,

ntodorova:
I am testing the module and when I create an Invoice, I don't see the discount on the line: 
Screenshot from 2019-03-18 11-37-41

As you see the discount is applied (Line one has 50% discount), but it is not mentioned anywhere. Is that how it should be?

hbrunn:
you don't have the `Show discount on invoice lines` (or similar name) group activated

ntodorova:
Yes, now all is fine. So I have review and test. You have answered all my questions. From my side the module is OK

ntodorova (approved):

pedrobaeza:
I would prefer the name `account_analytic_line_invoicing` for a better description of the aim of the module. Apart from that, the module seems OK.

AaronHForgeFlow:
Is the play onchanges working? When I create the invoice from the analytic lines the account_id is not set so the configuration error raises. When creating the invoice with the same product manually the account is filled accordingly.

hbrunn:
@aheficent this works as it should for us in production for a while. Can you reproduce the problem on runbot?

AaronHForgeFlow:
@hbrunn I tested on runbot indeed:
analytic

I set this product
image

and the accounts seems set in the product.

Are there anything else I should have configured before?

hbrunn:
ah, you need to set a customer on the project(s) you want to invoice. A better error message about that would be nice indeed

AaronHForgeFlow (approved):
:+1: Tested. Thanks @hbrunn

pedrobaeza:
@hbrunn what do you think about the module name?

hbrunn:
I take you suggestion as the opinion of a single OCA member, given we have two approvals and you're a little bit late to the party ;-)

pedrobaeza:
OK, no problem

CasVissers-360ERP:
Is there an OCA module which contains similar functionality to this but for later versions? (v12>)  I am looking for a way to create invoices from analytic lines, not only timesheet lines but also analytic lines for consumable and stockable products. If not so is it something which could be proposed?

PR #463, opened 2018-12-11
[ADD] module to set version on partner from account_invoice validation

bguillot:
Hi

This module depends on the module of this PR : OCA/partner-contact#669

It sets a version on addresses of the account invoice when the account invoice is validated.


yvaucher (approved):
LGTM

pedrobaeza:
I don't get the use case for needing this, or something is not well designed in the base if this is needed.

PR #467, opened 2018-12-18
[11.0][add][account_invoice_alternate_payer]

JordiBForgeFlow:
This module allows to enter an alternative payer in the customer invoice or vendor bill.

This allows, that we pay or expect to get paid by another partner other than the main partner.

I plan to introduce also a module for compatibility with SEPA modules, due to https://github.com/OCA/bank-payment/blob/12.0/account_payment_partner/models/account_invoice.py#L22


account_invoice_alternate_payer/views/account_invoice_views.xml
```diff
@@ -0,0 +1,28 @@
...
+                <field name="partner_id" position="after">
+                    <field name="alternate_payer_id" string="Alternate Payee"/>
+                </field>
+            </field>
+        </record>
+
+        <record id="invoice_form" model="ir.ui.view">
+            <field name="name">account.invoice.form</field>
+            <field name="model">account.invoice</field>
+            <field name="inherit_id" ref="account.invoice_supplier_form"/>
```
etobella: I think that may be invoice_form

pedrobaeza:
Is this a "suplido"?

account_invoice_alternate_payer/models/account_invoice.py
```diff
@@ -0,0 +1,39 @@
...
+    _inherit = 'account.invoice'
+
+    alternate_payer_id = fields.Many2one(
+        'res.partner', string='Alternate Payer',
+        help="If set, this partner will be the that we expect to pay or to "
+             "be paid by. If not set, the payor is by default the "
+             "commercial")
+
+    @api.multi
+    def finalize_invoice_move_lines(self, move_lines):
```
etobella: That is probably a bad choice. Now, the system is no longer usable with reimbursables. It is not probable to have a reimburasable invoice with an alternate payor, but it is good to know.

It think it could be better to use: https://github.com/odoo/odoo/blob/11.0/addons/account/models/account_invoice.py#L1189
Here, the partner is assigned to the move line. Also, you should not change the partner of all the lines, but only the line used for payment (type='dest'). Isn't it?

etobella (changes requested):
Just a few comments. I may be wrong.

etobella:
This should be used when we invoice to a partner but another one will pay the invoice.

JordiBForgeFlow:
@pedrobaeza It is surprising that Odoo has not thought that it is common that an invoice can be paid by another partner.

pedrobaeza:
But this is very weird, and AFAIK this is not legal to be done: if the invoice is issued to one entity, another can't pay it.

etobella:
It is totally legal. If it is not legal, then "suplidos" is also ilegal. I will try to show it in an example (using "suplidos"):
Three companies: A, B and C. A makes an invoice for C but it is payed by B, then B makes an invoice to C and includes the invoice he payed on the name of C as a "suplido".
This module is trying to solve the first invoice (from A to C payed by B) and suplidos tries to solve the second invoice.

pedrobaeza:
That's why I asked first if this is "suplidos". It's not the same to pay a specific concept in the name of other person than to pay the full invoice.

etobella:
In suplidos, usually, they pay a full invoice. At least on the cases I have found at work.

pedrobaeza:
OK, but I have always seen this as only a part, as part of a whole operation where you are authorized to receive that part that was covered by other company, and pay to them later.

etobella:
You are right when we are talking about the invoice that is made from the intermediary to the final customer. However, suplidos are invoices (usually full) for the final customers payed by the intermediary to the initial supplier and charged through the invoice of the intermediary to the final customer. Suplidos, do not have taxes and are not included on the total of the invoice, but it is an amount that has to be charged to the customer, as the intermediary has payed it.

This module is not trying to solve this issue. It tries to solve the problem that the initial supplier has an invoice payed by an intermediary. The supplier does not know how the final invoice and the suplido will be made.

Suplido's problem should be solved on `account_reimbursable`.

pedrobaeza:
OK, thanks for the extra explanations.

account_invoice_alternate_payer/models/account_invoice.py
```diff
@@ -0,0 +1,66 @@
...
+        'res.partner', string='Alternate Payer',
+        readonly=True, states={'draft': [('readonly', False)]},
+        help="If set, this partner will be the that we expect to pay or to "
+             "be paid by. If not set, the payor is by default the "
+             "commercial")
+
+    @api.model
+    def line_get_convert(self, line, part):
+        res = super(AccountInvoice, self).line_get_convert(line, part)
+        if line.get('type', False) and line['type'] == 'dest':
```
etobella: line.get('type', False) == 'dest should be enough

etobella (approved):
Just a minor comment

AaronHForgeFlow (commented):
Created a customer invoice for partner A. Assign Partner B as Payee. The receivable account in the invoice does not change for the one in the partner B. Am I doing anything wrong?

account_invoice_alternate_payer/tests/test_account_invoice.py
```diff
@@ -0,0 +1,257 @@
...
+               }
+        register_payments = self.register_payments_model.with_context(
+            ctx).create({
+                'payment_date': fields.Date.today(),
+                'journal_id': self.bank_journal_usd.id,
+                'payment_method_id': self.env.ref(
+                    'account.account_payment_method_manual_out').id,
+                })
+        register_payments.create_payments()
+        self.assertEqual(self.customer_invoice.state, 'paid')
```
AaronHForgeFlow: self.assertEqual(self.vendor_invoice.state, 'paid')

AaronHForgeFlow (approved):
Please dismiss my previous comment. I was selecting as a customer a partner that is a company. In that case the payor is the customer even if we choose an alternate payor. The same for vendor bills.

If we choose an individual as partner in the invoice, then the system will take the alternate payor/payee as it is intended in this module.

:+1: LGTM

etobella:
@jbeficent Can you recheck travis?

jarroyomorales (approved):
Travis was red because of another module, I restarted the jobs and its green now :green_apple:. Can this be merged? @jbeficent

PR #252, opened 2017-04-03
10.0 mig account invoice supplierinfo update

mourad-ehm:

PR #468, opened 2018-12-18
[PORT][10.0] mig account invoice supplierinfo update discount

legalsylvain:
* port the module account_invoice_supplierinfo_update_discount (mainly adapt code to new supplierinfo model, removing partnerinfo)
* New OCA convention. (new readme file, new keys in manifest)
* add tests and demo data.

* Fix in account_invoice_supplierinfo_update  to restore inheritance mechanism that has been lost in #252

Commit to review the migration : https://github.com/OCA/account-invoicing/pull/468/commits/6a5823569655897c64a52b55c6b55e3d81d8e198

CC : @mourad-ehm, @bealdav, @StefanRijnhart 

Thanks for your review !


mourad-ehm (approved):
Code review. Not tested

account_invoice_supplierinfo_update_discount/wizard/wizard_update_invoice_supplierinfo.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<openerp><data>
+
+    <record id="view_wizard_update_invoice_supplierinfo_form" model="ir.ui.view">
+        <field name="model">wizard.update.invoice.supplierinfo</field>
+        <field name="inherit_id" ref="account_invoice_supplierinfo_update.view_wizard_update_invoice_supplierinfo_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='line_ids']/tree/field[@name='price_variation']" position="after">
+                <field name="current_discount" attrs="{'invisible': [('current_discount', '=', 0)]}"
+                    groups="sale.group_discount_per_so_line"/>
```
StefanRijnhart: No group is applied to the related discount field on the purchase order line: https://github.com/OCA/purchase-workflow/blob/10.0/purchase_discount/views/purchase_discount_view.xml#L31. I don't like the dependency on sale. How would you feel about removing that group restriction here?

legalsylvain: Hi @StefanRijnhart. Happy new year ! 

Indeed, you're pointing a change in Odoo that I didn't see.

In the first version of the module, I set a ``groups="sale.group_discount_per_so_line"``
as odoo does it in odoo / 8.0 / account (invoice form)
https://github.com/odoo/odoo/blob/8.0/addons/account/account_invoice_view.xml#L212

remark : the issue description in the module said other things :
[see Issue description](https://github.com/OCA/account-invoicing/tree/8.0/account_invoice_supplierinfo_update_discount#known-issues--roadmap)

But in the 10.0 version, Odoo displays the discount field with ``groups="base.group_no_one"``
https://github.com/odoo/odoo/blob/10.0/addons/account/views/account_invoice_view.xml#L275

So : 
1) OK to remove dependency to sale, and obsolete "group" value.

2) adding a ``groups="base.group_no_one"`` ?
- To be consistent with account module we should add it
- But in the other hand, there is no group in purchase module (as you said)


What do you think ?

StefanRijnhart: Hi @legalsylvain thanks for your exhaustive writeup of the issue! I'm thinking that discount field should always be visible in this module, because otherwise it makes no sense to install the module at all.

legalsylvain:
Travis is red, and I'll not work any more on this PR. Closing.

PR #470, opened 2018-12-24
[FIX] Fixed performance problem when searchn tracking values by subty…

### LINKED ISSUE #469: [PERF] account_invoice_change_currency: Slow query from domain
The following line of code:
 - https://github.com/OCA/account-invoicing/blob/7cb1413180db0b97d406c966097ac400884f872b/account_invoice_change_currency/models/account_change_currency.py#L115

Run the following sql query:
 - `SELECT mail_message.id FROM mail_message WHERE (mail_message.subtype_id != 58) ORDER BY mail_message.id DESC;`

This is so slow (2s) because the result is too big if you have too many records.
After that will process the result of all ids fetched for a second query, then it is so slow.

@hugho-ad 
Could you help us, please?

hugho-ad:
…pe id, instead search them by messages ids then filter them

fix https://github.com/OCA/account-invoicing/issues/469

### DIFF
--- a/account_invoice_change_currency/models/account_change_currency.py
+++ b/account_invoice_change_currency/models/account_change_currency.py
@@ -107,17 +107,25 @@ class AccountInvoice(models.Model):
         subtype_id = self.env.ref(
             'account_invoice_change_currency.mt_currency_update')
         subtype_create_id = self.env.ref('account.mt_invoice_created')
-        domain = [
-            ('mail_message_id', 'in', self.message_ids.ids),
-            ('field', '=', 'currency_id'),
-        ]
+        query = """
+SELECT mtv.old_value_integer, mtv.new_value_integer, mm.subtype_id
+FROM mail_tracking_value as mtv INNER JOIN mail_message AS mm
+ON mtv.mail_message_id = mm.id """
         if skip_update_currency:
-            domain += [('mail_message_id.subtype_id', '!=', subtype_id.id)]
-        last_value = self.env['mail.tracking.value'].sudo().search(
-            domain, limit=1, order='write_date desc, id desc')
-        value = last_value.old_value_integer
-        if last_value.mail_message_id.subtype_id == subtype_create_id:
-            value = last_value.new_value_integer
+            query += " AND mm.subtype_id != %s "
+            params = (subtype_id.id, tuple(self.message_ids.ids))
+        else:
+            params = (tuple(self.message_ids.ids), )
+        query += """WHERE mtv.field = 'currency_id' AND
+mtv.mail_message_id IN %s
+ORDER BY mtv.write_date DESC, mtv.id DESC LIMIT 1"""
+        self.env.cr.execute(query, params)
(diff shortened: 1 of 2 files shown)


moylop260:
Is not this change slower too?
Is possible using a query with join and limit 1?

hugho-ad:
@moylop260 

could you review please, I tried to use sql queries, but not sure.

Regards!

hugho-ad:
@moylop260 

Ready to review/ merge

account_invoice_change_currency/tests/test_account_invoice_change_currency.py
```diff
@@ -237,12 +237,12 @@ def test_custom_rate_zero_update_currency(self):
         expected_value = before_amount * rate / old_rate
         # TODO: Check float comparation, 26179.05 vs 26179.07656
         self.assertEqual(float_compare(
-            inv.amount_total, expected_value, self.precision), -1,
+            inv.amount_total, expected_value, 1), 0,
```
moylop260, line 240: Why a improvement of performance changed the result here?

hugho-ad: It was failing because of the decimals, I did not dig dipper

moylop260: Maybe is not a decimals issue since that you are not changing numbers in the diff.

hugho-ad: Here I'm getting following
image

account_invoice_change_currency/models/account_change_currency.py
```diff
@@ -107,17 +107,25 @@ def get_last_currency_id(self, skip_update_currency=False):
...
-            value = last_value.new_value_integer
+            query += " AND mm.subtype_id != %s "
+            params = (subtype_id.id, tuple(self.message_ids.ids))
+        else:
+            params = (tuple(self.message_ids.ids), )
+        query += """WHERE mtv.field = 'currency_id' AND
+mtv.mail_message_id IN %s
+ORDER BY mtv.write_date DESC, mtv.id DESC LIMIT 1"""
+        self.env.cr.execute(query, params)
+        res = self.env.cr.dictfetchall()
```
moylop260: self.env.cr.dictfetchone()

account_invoice_change_currency/models/account_change_currency.py
```diff
@@ -107,17 +107,25 @@ def get_last_currency_id(self, skip_update_currency=False):
...
+        value = False
+        if res:
+            value = res[0].get('old_value_integer')
```
moylop260: `value = res['old_value_integer']`
Using fetchone is not needed use `res[0]`

And it is not needed use `res.get('field')` since that the query will be return the name of fields ever.

hugho-ad:
@moylop260  Done

moylop260 (approved):

PR #473, opened 2019-01-03
[11.0][FIX] purchase_stock_picking_return_invoicing: Change buggy compute code

pedrobaeza:
We don't need to add dependencies to the method `_get_invoiced`, as the qty_received field is touched when returning a picking with the option `to_refund` marked

We need also to touch the computation method, as we were not using float_compare method nor taking into account products with purchase invoicing policy as "ordered quantities".

cc @Tecnativa


pedrobaeza:
The error happens in other module.

chienandalu (approved):
Looks good :+1:

carlosdauden (approved):

cubells (approved):
:+1: 

LGTM

Code review and test

PR #478, opened 2019-01-14
[10.0][FIX][account_invoice_triple_discount] Fix the discount rounding problem

robyf70:
After playing with this module I've found some rounding problem compared against odoo standard calculation, so using only one discount. For example using value such as:

quantity = 4
price_unit = 28.5
discount = 35

on standard Odoo this will lead to:
total_amount = 74.10

while instead triple discount wrongly calculate
total_amount = 74.12

so avoid such wrong rounding problems I've avoided to do call the super() and instead replace the method by the Odoo standard including the 2 additional discounts

### DIFF
--- a/account_invoice_triple_discount/__manifest__.py
+++ b/account_invoice_triple_discount/__manifest__.py
@@ -3,7 +3,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 {
     'name': 'Account Invoice Triple Discount',
-    'version': '10.0.1.0.0',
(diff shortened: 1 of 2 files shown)


StefanRijnhart:
@robyf70 thanks! Does https://github.com/OCA/account-invoicing/pull/500 solve this problem as well?

robyf70:
I haven't checked, but #500 looks a bit "tricky" as solution

StefanRijnhart:
@robyf70 it would still be interesting to see if either solution fixes both problems so that we can have a discussion over which is the least tricky one!

PR #479, opened 2019-01-15
[10.0] invoice_supplierinfo_update reworked naming on wizard buttons

PierrickBrun:
The buttons names where confusing, this commit makes it more
understandable.
It is only cosmetic change

### DIFF
--- a/account_invoice_supplierinfo_update/__manifest__.py
+++ b/account_invoice_supplierinfo_update/__manifest__.py
@@ -9,7 +9,7 @@
     'summary': 'In the supplier invoice, automatically updates all products '
                'whose unit price on the line is different from '
                'the supplier price',
-    'version': '10.0.1.0.0',
+    'version': '10.0.1.0.1',
     'category': 'Accounting & Finance',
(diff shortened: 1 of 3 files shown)


PierrickBrun:
@chafique-delli  @mourad-ehm @legalsylvain 
What do you think about it ? I think I will propagate the changes in 8.0 when it will be merged

PierrickBrun:
@chafique-delli  @mourad-ehm @legalsylvain 
What do you think about it ? I think I will propagate the changes in 8.0 when it will be merged

youring:
I think you were misunderstanding the button function. According to code,
**update_supplierinfo**: create/update supplier info _and_ mark the invoice as checked;
**update_supplierinfo_validate**: do above and validate the invoice.
But on the contrary, i think this module should not do stuff like validating invoice. Your modification on button names still make sense if change the wizard method a little bit.

legalsylvain:
I see the new string as more confusing. 
Update without validate is clear ans reuse core terms. Update temporarily has less sense for me.

Issue #483, opened 2019-01-19
10.0 - account_invoice_change_currency: Test error


moylop260:
` AssertionError: 2004.64 != 2004.67 : Total amount of invoice does not equal to expected value!!!

https://travis-ci.org/OCA/account-invoicing/jobs/481472498#L1707

@hugho-ad 
Could you help us with this issue, please?

hugho-ad:
@novawish 

could you help us with this issue?

Best regards

novawish:
@hugho-ad I’m really sorry for not awaring of this until now because I did not enable Notifications by mention in my github settings.
It seems that the issue was fixed by https://github.com/OCA/account-invoicing/pull/487, is it ok for now?
If yes can we close this issue?

Thanks and best regards.

PR #412, opened 2018-08-06
[FIX] Store invoice currency on the supplierinfo

StefanRijnhart:
Also: minor refactoring to allow override of update values on existing supplierinfos

PR #485, opened 2019-01-22
[FIX] Convert prices to purchase UOM

StefanRijnhart:
Convert the invoice' prices to the UOM that is valid for the supplierinfo's
to which they are written. This UOM is the product's purchase order UOM.

### DIFF
--- a/account_invoice_supplierinfo_update/__manifest__.py
+++ b/account_invoice_supplierinfo_update/__manifest__.py
@@ -9,9 +9,9 @@
     'summary': 'In the supplier invoice, automatically updates all products '
                'whose unit price on the line is different from '
                'the supplier price',
-    'version': '10.0.1.0.0',
+    'version': '10.0.1.0.1',
     'category': 'Accounting & Finance',
-    'website': 'http://akretion.com',
+    'website': 'https://github.com/OCA/account-invoicing',
     'author':
         'Akretion,'
         'GRAP,'
--- a/account_invoice_supplierinfo_update/README.rst
+++ b/account_invoice_supplierinfo_update/README.rst
@@ -51,7 +51,6 @@ Known Issues / Roadmap
 
 * This module does not manage correctly difference if
 
-    * invoice line UoM are not the same as Supplierinfo UoM
     * invoice line taxes are not the same as products taxes. (If one is
       marked as tax included in the price and the other is marked as
       tax excluded in the price)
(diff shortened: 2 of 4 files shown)


legalsylvain (approved):
Hi @StefanRijnhart 
Thanks for this fix.

- I allowed myself to update the readme file on your branch, removing obsolete "know issues" part.

- ~Just a question : why did you added the two lines in the test file ?~ (OK, I got it.)

        self.currency = self.env.ref('base.GBP')
        self.journal.write({'currency_id': self.currency.id})

- Travis is red for unrelated reason IMHO

LGTM. :+1:

pedrobaeza (commented):
Why not extracting computation which is repeated in several places (2 at least) to a method (like `_get_supplierinfo_unit_price`) for avoiding a code desynchronization between them?

pedrobaeza:
Travis error might be related to this PR changing unit prices in invoices...

legalsylvain:
Hi @pedrobaeza, thanks for your review.

> Travis error might be related to this PR changing unit prices in invoices...

I don't understand that point. Travis is red on 10.0 (HEAD), with the same error than for that branch. 

Could be more precise ? 

regards !

pedrobaeza:
OK, if main branch is red then it's not related, but we should fix it.

legalsylvain:
> OK, if main branch is red then it's not related, but we should fix it.

Pending work : https://github.com/OCA/account-invoicing/issues/483

account_invoice_supplierinfo_update/__manifest__.py
```diff
@@ -9,7 +9,7 @@
     'summary': 'In the supplier invoice, automatically updates all products '
                'whose unit price on the line is different from '
                'the supplier price',
-    'version': '10.0.1.0.0',
+    'version': '10.0.1.0.1',
     'category': 'Accounting & Finance',
     'website': 'http://akretion.com',
```
Nikul-Chaudhary: @StefanRijnhart URL should be `https://github.com/OCA/account-invoicing`

Nikul-Chaudhary (commented):
Minor change

account_invoice_supplierinfo_update/__manifest__.py
StefanRijnhart: Agreed, and changed.

StefanRijnhart:
@pedrobaeza thanks, refactored into a separate method.

@legalsylvain thanks, I had not even noticed it was listed as a known issue!

pedrobaeza (approved):

StefanRijnhart:
@legalsylvain Oh, and those currency lines in the test are a remnant of an other PR of mine (https://github.com/OCA/account-invoicing/pull/412). I removed them from this one.

StefanRijnhart:
Tests are fixed here: https://github.com/OCA/account-invoicing/pull/487

StefanRijnhart:
rebased to fix CI

PR #494, opened 2019-01-27
[12.0] [MIG] account_invoice_refund_link

Nikul-Chaudhary:
Link refund invoice with original
=================================

This module shows the links between refunds and their original invoices in the invoice form and also keep track of refund lines and their original invoice lines.


account_invoice_refund_link/views/account_invoice_view.xml
```diff
@@ -5,82 +5,82 @@
...
-                   attrs="{'invisible': [('refund_invoice_id', '=', False)]}"
-            />
-            <field name="refund_reason"
-                   readonly="1"
-                   attrs="{'invisible': [('refund_reason', '=', False)]}"
-            />
-        </xpath>
-    </field>
-</record>
+    <record id="view_customer_account_invoice_add_refunds_details_form" model="ir.ui.view">
```
pedrobaeza: Please don't change this only for esthetic reasons, as it creates a great diff with no need.

pedrobaeza (commented):
Squash together adjacent "OCA Transbot..." commits.

account_invoice_refund_link/views/account_invoice_view.xml
Nikul-Chaudhary: > Please don't change this only for esthetic reasons, as it creates a great diff with no need.

@pedrobaeza Yes, But it's look ugly, so I changed.

Nikul-Chaudhary:
@pedrobaeza Squashed

account_invoice_refund_link/__manifest__.py
```diff
@@ -6,9 +6,9 @@
 {
     "name": "Link refund invoice with original",
     "summary": "Link refund invoice with its original invoice",
-    "version": "11.0.2.0.0",
+    "version": "12.0.2.0.0",
```
pedrobaeza: 12.0.1.0.0

account_invoice_refund_link/models/account_invoice.py
```diff
@@ -9,7 +9,7 @@
 class AccountInvoice(models.Model):
     _inherit = 'account.invoice'
 
-    refund_reason = fields.Text(string="Refund reason")
+    refund_reason = fields.Text("Refund reason")
```
pedrobaeza: Don't remove keyword attribute. They are always more flexible and it's already there. Why increasing the diff?

account_invoice_refund_link/views/account_invoice_view.xml
pedrobaeza: Please don't

pedrobaeza (changes requested):

Nikul-Chaudhary:
@pedrobaeza Done

pedrobaeza (approved):

misern2 (approved):

PR #497, opened 2019-01-28
[11.0][purchase_stock_picking_return_invoicing] - Add compatibility with purchase_force_invoiced of OCA/purchase-workflow

JordiBForgeFlow:
This PR intends to force the purchase invoice status to = 'Nothing to bill' when the status has been forced in module purchase_force_invoiced. See https://github.com/OCA/purchase-workflow/pull/642

### DIFF
--- a/purchase_stock_picking_return_invoicing/models/purchase_order.py
+++ b/purchase_stock_picking_return_invoicing/models/purchase_order.py
@@ -26,6 +26,12 @@ class PurchaseOrder(models.Model):
             'Product Unit of Measure'
         )
         for order in self.filtered(lambda x: x.state in ('purchase', 'done')):
+            # the field 'force_invoiced' is defined in the
+            # 'purchase_force_invoiced' module of OCA/purchase-workflow.
+            if getattr(order, 'force_invoiced', None) and \
+                    order.force_invoiced and \
+                    order.invoice_status == 'invoiced':
+                continue
             if any(
                 float_compare(
                     line.qty_invoiced, line.product_qty
--- a/purchase_stock_picking_return_invoicing/__manifest__.py
+++ b/purchase_stock_picking_return_invoicing/__manifest__.py
@@ -6,7 +6,7 @@
 {
     "name": "Purchase Stock Picking Return Invoicing",
     "summary": "Add an option to refund returned pickings",
-    "version": "11.0.1.0.0",
+    "version": "11.0.1.0.1",
     "category": "Purchases",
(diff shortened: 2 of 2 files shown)


purchase_stock_picking_return_invoicing/models/purchase_order.py
```diff
@@ -26,6 +26,11 @@ def _get_invoiced(self):
             'Product Unit of Measure'
         )
         for order in self.filtered(lambda x: x.state in ('purchase', 'done')):
+            # the field 'force_invoiced' is defined in the
+            # 'purchase_force_invoiced' module of OCA/purchase-workflow.
+            if hasattr(order, 'force_invoiced') and order.force_invoiced and \
```
moylop260: What about using `getattr` instead?
`if getattr(order, 'force_invoiced', None) and order.invoice_status == 'invoiced':`

moylop260 (approved):
LGTM
Just small comment non-blocking

pedrobaeza:
I don't like the module `purchase_force_invoiced` as it's cheating the system, and this leads to errors with other modules like this one, or worst, create unaware inconsistencies that later gets shown (for example on migrations). You should teach you users to do things right and alternative flows instead of this.

For me, :-1: on the whole module (but I'm afraid you have already merged), and :-1: for merging this depending code here.

HviorForgeFlow (approved):

MiquelRForgeFlow (approved):
LGTM

pedrobaeza:
@jbeficent I didn't agree to put this.

JordiBForgeFlow:
@pedrobaeza I understand, but we need this feature, we made sure that the code is not going to cause others a problem, and it contains approvals, both from our team and from @moylop260.

If we don't implement this, we have a problem.

pedrobaeza:
I'm not saying you shouldn't implement this feature. I'm saying not to do it in this module for having slave code that is activated through that "black magic".

JordiBForgeFlow:
@pedrobaeza can we at least add a hook then? We'll use a separate module to implement our feature.

pedrobaeza:
But you already have the hook.. it's `_get_invoiced` itself. Overwrite and only call super with recordsets without that flag.

pedrobaeza:
If not enough, I'm OK to add the hook.

JordiBForgeFlow:
@pedrobaeza as per our discussion, I have created https://github.com/OCA/account-invoicing/pull/557. I'm closing this PR.

PR #498, opened 2019-01-29
[FIX] self_invoice: report header

etobella:
The report was not being correctly parsed, as you can see on the header.
imagen


etobella:
It should look like:
imagen

pedrobaeza (commented):
Why not using `external_layout` as the rest of documents?

etobella:
Because the company on the top should be the company that makes the invoice, your supplier.

pedrobaeza (approved):
OK, got it

etobella:
@jarroyomorales @aheficent

jarroyomorales (approved):

PR #499, opened 2019-01-30
[FIX] add _change_pricelist to update currency

mourad-ehm:


account_invoice_pricelist/models/account_invoice.py
```diff
@@ -16,11 +16,20 @@ class AccountInvoice(models.Model):
     @api.onchange('partner_id', 'company_id')
     def _onchange_partner_id_account_invoice_pricelist(self):
         result = super(AccountInvoice, self)._onchange_partner_id()
-        if self.partner_id and self.type in ('out_invoice', 'out_refund')\
-                and self.partner_id.property_product_pricelist:
-            self.pricelist_id = self.partner_id.property_product_pricelist
+        for invoice in self:
```
pedrobaeza: onchange are always executed for one record, so don't change this.

account_invoice_pricelist/models/account_invoice.py
```diff
@@ -16,11 +16,20 @@ class AccountInvoice(models.Model):
...
+                    ('out_invoice', 'out_refund') and\
+                    invoice.partner_id.property_product_pricelist:
+                invoice.pricelist_id =\
+                    invoice.partner_id.property_product_pricelist
+                invoice.currency_id = invoice.pricelist_id.currency_id
         return result
 
+    @api.onchange('pricelist_id')
+    def _change_pricelist(self):
+        for invoice in self:
```
pedrobaeza: Only for one record.

pedrobaeza (changes requested):

account_invoice_pricelist/models/account_invoice.py
mourad-ehm: In some case, we call this method by script and we can get incorrect result.

pedrobaeza: Then the error is in your script. You should call the onchange record per record, not all at once.

mourad-ehm: I do not think it's a good thing. Because the method can be called in the code or in the user interface as well. in this way if improvements are made in the orm one can benifit them without modifying the code. In some cases, oca can be a leader and lead odoo afterwards. That's my opinion, but in the meantime I'm going to make the changes.

pedrobaeza: Yeah, but meanwhile Odoo doesn't change this approach, it's an effort in vain to go against current, as inheriting onchange methods from upstream will have to be single-record aware, so you will end up with a mix of both, possible errors not knowing all the inheritance chain, etc.

Note that at the beginning, I started to do the same thing, but at the end I realized what I'm now telling you.

mourad-ehm: OK

mourad-ehm:
Hi @pedrobaeza could you merge this please ?

pedrobaeza (approved):
@mourad-ehm I'm not anymore OCA's community manager. Any way, this requires 2 approvals. I'm giving mine as personal one.

PR #500, opened 2019-01-30
[10.0][FIX] account_invoice_triple_discount: Fix rounding

SimoRubi:
The fix is pretty much the same applied in https://github.com/OCA/sale-workflow/pull/768.
This is to fix the behavior described in the unit test:
Use case of an invoice line having:

-  price_unit = 15.2,
-  qty = 131.04
-  discount = 50
-  discount2 = 0
-  discount3 = 3

The result should be the same as applying only one discount of 51.5

### DIFF
--- a/account_invoice_triple_discount/__manifest__.py
+++ b/account_invoice_triple_discount/__manifest__.py
@@ -3,7 +3,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 {
     'name': 'Account Invoice Triple Discount',
-    'version': '10.0.1.0.0',
+    'version': '10.0.1.0.1',
     'category': 'Accounting & Finance',
     'author': 'Tecnativa, '
               'Odoo Community Association (OCA)',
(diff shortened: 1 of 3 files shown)


StefanRijnhart (approved):
Technical review. Curious to know what the opinion is of the original author @chienandalu on the functional side.

StefanRijnhart:
@SimoRubi  after looking at https://github.com/OCA/sale-workflow/pull/768, I think I may have misunderstood the problem. Is it indeed the case here as well that the problem is that with global rounding enabled, discount2 and discount3 are not actually applied?

SimoRubi:
Hi @StefanRijnhart thanks for your review!
Here the issue is different (it is the one described in description and in the automated test) but the same fix applied in https://github.com/OCA/sale-workflow/pull/768 seems to fix it.

pedrobaeza (commented):
I don't think `invalidate_cache` call serves for anything

SimoRubi:
Thanks @pedrobaeza for the review, you are right, I also retried it in runbot after https://github.com/OCA/account-invoicing/pull/500/commits/556db9667fe43bc1659a979d36992eaa5d916a40.

stevech091 (approved):
functional ok

lmignon:
@SimoRubi @pedrobaeza @StefanRijnhart This merge seems to break oca/sale-worklow.... https://travis-ci.org/OCA/sale-workflow/builds/501135521

pedrobaeza:
Curious, it was green before... @SimoRubi can you please check?

PR #504, opened 2019-02-15
[ADD] purchase_invoice_no_zero_lines: avoid zero quantity lines in ac…

jonerikceberio:
…count invoice lines from purchase
To answer OCA/OCB#814
While coding I found nearly the same in account-invoicing/purchase_stock_picking_return_invoicing


pedrobaeza:
I talk about `purchase_invoicing` instead of `purchase_invoice`, but I don't want to be picky...

purchase_invoicing_no_zero_line/models/account_invoice.py
```diff
@@ -0,0 +1,25 @@
...
+
+    @api.onchange('purchase_id')
+    def purchase_order_change(self):
+        """
+        after the creation of the lines, delete the zero qty lines
+        if the journal is marked as such
+        """
+        res = super(AccountInvoice, self).purchase_order_change()
+        if self.journal_id and self.journal_id.avoid_zero_lines:
+            self.invoice_line_ids -= self.invoice_line_ids.filtered(
```
pedrobaeza, line 20: Maybe for being more exact, we should filter only the lines of the added purchase order, isn't it?

purchase_invoicing_no_zero_line/models/account_journal.py
```diff
@@ -0,0 +1,14 @@
...
+
+from odoo import api, fields, models, _
+
+
+class AccountJournal(models.Model):
+    _inherit = "account.journal"
+
+    avoid_zero_lines = fields.Boolean(
+        string="Avoid zero lines?",
+        help="If this option is checked, the purchase invoices in this journal will have no lines with zero quantity"
```
pedrobaeza: Please restrict to PEP8 80 cols

pedrobaeza:
No need to copy SVG file, as it's unmodified.

jonerikceberio:
Uups, also adding spanish translation

jonerikceberio:
Travis hiccup, is there a way to retry?

> $ curl -sSf -o python-3.5.tar.bz2 ${archive_url}
curl: (56) SSL read: error:00000000:lib(0):func(0):reason(0), errno 104
Unable to download 3.5 archive. The archive may not exist. Please consider a different version.

pedrobaeza:
Yeah, sometimes happens. Done.

jonerikceberio:
@pedrobaeza , what is with codecov?

pedrobaeza:
It's a metric for seeing test coverage. As your module doesn't have tests, then it's lowering the repository total coverage, and thus the red cross. They are not mandatory, but recommended.

purchase_invoicing_no_zero_line/__init__.py
```diff
@@ -0,0 +1,4 @@
+# -*- coding: utf-8 -*-
```
pedrobaeza: This is not needed in any Python file since v11 (due to Python 3)

pedrobaeza (approved):

enriquemartin (approved):

PR #140, opened 2016-03-19
[V9]Stock Picking Invoicing

flotho:
Since Odoo v9 doesn't support anymore invoicing neither on move nor on picking, this migration has the purpose to summarize some modules :
- https://github.com/OCA/account-invoicing/tree/9.0/stock_picking_invoicing
- https://github.com/OCA/stock-logistics-workflow/tree/9.0/stock_picking_invoice_link
- https://github.com/OCA/account-invoicing/tree/8.0/stock_picking_invoicing_unified
- And the odoo core one : https://github.com/odoo/odoo/blob/8.0/addons/stock_account/stock.py

PR #320, opened 2017-10-13
account_invoice_from_picking initial commit

rvalyi:
This module allows to create invoice lines from picking orders.

A usage of this module is it allows you to create partial invoices for partial receipts/deliveries in case you forgot to create them after each receipt/delivery. Care has been taken for not creating discrepancies with the normal invoices created from the sale orders or pulling lines from the purchase order.

This module might also be required in some federal countries such as Brazil were you may produce invoice-like documents (a bit like European intrastat but with accounting entries too) when you simply move goods from one federal state to another without any sale or purchase order.

**This a work in progress** as there is still a bug: you can still pull lines from pickings related to paid sale orders. And tests are still missing.

PR #509, opened 2019-02-28
[10.0][MIG] stock_picking_invoicing

acsonefho:
**Migration of module stock_picking_invoicing**

**Functional:**
This module is used to create invoices from pickings

**I edit a lot this module to improve some points**
- Readibility (about the code)
- Make it more extendable (in case of inheritance etc)
- Add more case to group new invoices (by partner, products etc)
- Add unit test about the group
- Fix some bug with the journal to use
- Remove useless things (company_id etc)

**Technically**
- [x] Forward port existing MR #140 (v9.0)
- [x] Generate new readme
- [x] Migration of code (old API etc)
- [x] Make unit test pass
- [x] Migration of xml files (remove openerp, data etc)
- [x] Refactor a little bit code (remove duplicates etc)


stock_picking_invoicing/README.rst
```diff
@@ -1,42 +1,93 @@
...
+
 Contributors
-------------
+~~~~~~~~~~~~
 
 * Lorenzo Battistini <lorenzo.battistini@agilebg.com>
 * Leonardo Pistone <leonardo.pistone@agilebg.com>
 * Daniel Sadamo <daniel.sadamo@kmee.com.br>
 * Alex Comba <alex.comba@agilebg.com>
+* Florent THOMAS <florent.thomas@mind-and-go.com>
```
Cedric-Pigeon, line 77: @acsonefho You can add yourself...

stock_picking_invoicing/__init__.py
```diff
@@ -1,21 +1,37 @@
...
 
-from . import stock
+
+def pre_init_hook(cr):
+    """
+    Remove views who exists into previous version but not into this one
+    :param cr: database cursor
+    :return:
+    """
+    _logger.info("Remove previous views who doesn't exist anymore "
```
Cedric-Pigeon: which doesn't exist...

Cedric-Pigeon (commented):
Small comments

mileo (approved):
LGTM, tested with only stock and invoice modules installed.

We must think about how the integration with the sale/purchase module will be made.

mileo:
There is duplicate PR with a nice feature in https://github.com/OCA/account-invoicing/pull/320 that can be used to import a picking on the invoice model.

AaronHForgeFlow (commented):
Functional tests good so far. Invoicing quantities updated properly in the SO or PO.

mileo:
@nikul-serpentcs please see this PR, is similar of https://github.com/OCA/account-invoicing/pull/320

@aheficent your comment seems to approve the pr, if it is true so please use the github feature to it.

AaronHForgeFlow:
@mileo  I did not approve because this is modifying the data model and I am not sure if this may lead to discrepancies between the sales-invoices and picking-invoices. I will not block this, but I prefer to go for another approach. 

Regards.

pedrobaeza:
I don't see either this approach is the correct one. To invoice orders and the "dashboard" in there is the best change in several Odoo versions.

rvalyi:
been very busy these days to deliver something to a customer, but will make my best to review this PR. Thanks you all for your review.

stock_picking_invoicing/tests/test_picking_invoicing.py
```diff
@@ -30,41 +14,511 @@ def setUp(self):
         self.invoice_wizard = self.env['stock.invoice.onshipping']
         self.invoice_model = self.env['account.invoice']
         self.partner_model = self.env['res.partner']
+        self.partner = self.env.ref('base.res_partner_2')
+        self.partner2 = self.env.ref('base.res_partner_address_4')
+        self.partner2 = self.env.ref('base.res_partner_address_4')
```
nikul-serpentcs: @acsonefho  duplicate? 18 & 19 lines

TheCloneMaster:
In Odoo 10, if products are set to invoice on "delivered qty", and then you return products from picking, you will be able to "create invoice" for the refunded products, and it will create a refund invoice...
No need to add any new modules...

rvalyi (approved):
I approve the PR (I'll remove the test duplicate line if we go for the merge).

@pedrobaeza @TheCloneMaster yes in general invoicing from the sale orders (base on delivered quantities) or selecting purchase orders is way more consistent than it used to be in Odoo. But it may not fit all the use cases.

First this module allows you to create partial invoices for partial receipts/deliveries in case you forgot to create them right after each receipt/delivery. So say you make a big sale order, then send a partial A, then a partial B and then a partial C. After you delivered A, you made an invoice for A, perfect. Then you delivered B but you forgot to create the invoice. Then you delivered C. If you use use the core Odoo sale order wizard, you cannot make separate invoices for B and C anymore. They will all come together. This can be a problem, your customer may refuse your invoice... With this module you can make separate invoices for B and C. This is a real case @florian-dacosta faced with one of our customer in France for instance.

Another very important use case can happen when you move goods betweem two states of a federated country even if it still belongs to your company (so there is simply no sale or purchase order).
Generally you should legally bookmark theses goods moves a bit like oca/intrastat-extrastat.
But Brazil at least, the legal document you should produce is very similar to an invoice: it should make special accounting move with tax amounts all like a real invoice except ther is no payment reconciliation to be made.
This module is absolutely required for our localization, all the 3 PSCs of the project agree on that. Now this would be stupid we move it to our repo only if there are a few other non Brazilian usages, albeit not very obvious.


Finally I found the following limits in this module:
- if you create a supplier invoice from a receipt: the price lines will be 0. It could have taken the standard_price instead like it does in the vendor invoice. Now this is not shocking either as the purchase order put 0 too unless you have a supplier price defined.
- if you invoice a delivery or a receipt, you will be able to double invoice it from the sale order or the purchase order respectively. This is because sale.order,.line and purchase.order.line have am invoice relation that is not set by this module. This is not shocking either as this module doesn't depend on sale nor purchase. It could be done in glue module or also with just a few lines testing if sale or purchase modules are installed. So I propose to adopt the module this way on v10. Then I can port it on v12.

pedrobaeza:
@rvalyi, indeed there are use cases where invoicing specific pickings is needed, but the problem in my opinion is the flow itself: invoicing from pickings provokes you the need of controlling the proper synchronization of all the involved objects, have exception cases (like pickings without sale), or losing features (like money in advance).

I think a better approach would be a module `sale_invoicing_picking_filter` that still controls the flow from the sales orders, but allow you to set pickings to invoice, controlling quantities and lines through inheritance.

What do you think about this approach?

rvalyi:
@pedrobaeza we agree about the drawbacks when you invoice from pickings using this module and I agree that in some cases a sale order  filter to select specific pikings to invoice would do the job too.

However, there is also really the case where you will move goods between 2 states (possibly the 2 federal states of the same country) and where you may need to register this with a special kind of invoice (much like an intrastat declaration). Or for instance if you move products for testing purpose between 2 companies. So it means invoice from picking WITHOUT ANY SALE NOR PURCHASE ORDER!

At least this is an accounting requirement here in Brazil with the "remessa", see here:
https://www.valor.srv.br/matTecs/matTecsIndex.php?idMatTec=63

So I guess other countries may need this module and there is nothing specific to Brazil however it is required here. So how do we do? If you want we can put some kind of best practise warning in the README, but I feel it is better here instead of burried inside our very own localization. The best proof it is not so specific for Brazil  and synergies may exist is that this module has been created in Europe and still we need it...

pedrobaeza:
But why don't you implement the approach I have told (the module `sale_invoicing_picking_filter`)?

rvalyi:
@pedrobaeza because ir our case there will often be NO such sale order document nor purchase order! It's really just a stock picking that will generate an accounting document that is so much similar to an invoice in every aspect that it's just better to say it's a special kind of invoice. 

So for the invoice we don't have the choice (it'a a kind of proforma in the sense there will be nothing to pay, however there will still be some account moves with all the tax computation dance. But on the contrary, creating a fake kind of sale order or purchase order onnly to generate these "remessa" invoice would also force us to tweak all sale/purchase features like reports and would generate unecessary operations with strange access rights (like stock user having to use sales menu) that would just make no sense for the ERP users...

rvalyi:
@acsonefho @Cedric-Pigeon any take on this one? As I explained the module is mandatory for us in Brazil, all the 3 projects PSC agree there is no alternative on this. Now the question is if we merge this great module here or move it to our OCA/l10n-brazil repo and make it invisible to potential users in other countries...

pedrobaeza:
IMO you should always rely on orders (sales/purchase), but go your own way.

rvalyi:
@pedrobaeza the "remessa" case I gave with the link makes just no sense with an order. For instance you just move products in your own company from a warehouse from say the Rio de Janeiro federal state to a a warehouse in the Sao Paulo federal state and you should create such an electronic "remessa" invoice with the proper accounting move (yet nothing to reconcile) including all the tax account move lines.
That makes just no sense to create sale order or purchase orders that will be shown among the other orders for this. Potentially you may even need to do this without even the sale or purchase module installed, requiring these modules would somewhat break the Odoo modularity...

This is really because the federal gov simply wants to track product moves to deal with the tax compensation between the different federal states, a bit like the European intrastat system but really normalized as part of the Brazilian accounting here. And I guess other federation countries may need such system as they probably have such tracking system to avoid tax evasion or fiscal competition between their different tax juridictions. Instead in Europe the intrastat documents obviously aren't part of the national accounting as every European country has different national accounting norms...

Now the real question is if we merge the module here or move it to our localization repo...

pedrobaeza:
We do the same in Spain when sending goods to Canarian Islands, but we create a sales orders for that, and issuing the invoice from it, and there's no problem in having the orders there. You can filter out them by partner, sale order type...

rvalyi:
@pedrobaeza interresting, well that's worth analysing... Could you point to the relevant parts of your localization for that? Specially how you filter out the sale orders later...

In our case, national ERPs also usually offer possibility to import the XML of the electronic invoice of such incoming pickings to generate the appropriate invoice and we would have to get it working with such case. Also, in the case the company would not usually use the sale module, would it be legit to require installing the sale module just to handle these special cases of invoicing?

pedrobaeza:
We don't need nothing extra. We are just using `sale_order_type` module for specifying the warehouse, and a saved filter excluding that sales order type is enough (in the case I have).

For the electronic invoicing you mention, you import the `account.invoice` model, so nothing to do with sales orders nor pickings...

rvalyi:
> We don't need nothing extra. We are just using sale_order_type module for specifying the warehouse, and a saved filter excluding that sales order type is enough (in the case I have).

thanks for the explanation.

> For the electronic invoicing you mention, you import the account.invoice model, so nothing to do with sales orders nor pickings...

Well more or less because here the national ERP's typically have a feature to generate the incoming picking with the exact product quantities based on this incoming electronic invoice (this avoids quite a lot of manual operations even if some corrections can be made).

So well eventually we construct the picking back again from the invoice. But if we don't use this stock_picking_invoicing module, are we sure such generated incoming picking will have a proper status and be properly related to the invoice and even invoice lines? Or will we go untill back-generating even a fake sale order and a related picking all related with the invoice we just imported? These are the important questions if we opt to always use a sale order for such product moves...

pedrobaeza:
For a proper link, just use `stock_picking_invoice_link` module, but don't require all the invoicing stuff IMO. Use the same digital format for creating invoice, picking, and the link between both, but don't recreate invoicing mechanism in that case.

Cedric-Pigeon:
Hi All,

sorry for the late reaction. I completely share @rvalyi's opinion. We also need this feature in some particular cases where no sale orders are generated. So please leave this addon in this repo.

Cedric-Pigeon (approved):

stock_picking_invoicing/__init__.py
```diff
@@ -1,21 +1,37 @@
...
+    module = 'stock_picking_invoicing' AND
+    model = 'ir.ui.view'"""
+    # Remove the view
+    query_view = """
+    DELETE FROM ir_ui_view
+    WHERE id IN (%(main_query)s);"""
+    # Remove the XML ID of the view
+    query_data = """
+    DELETE FROM ir_model_data
+    WHERE id IN (%(main_query)s);"""
```
gabrielcardoso21: In this case wouldn't it be more correct to delete the ir_model_data register wich the res_id is the one selected by "main_query"?

gabrielcardoso21 (changes requested):

rousseldenis:
I agree with this module behaviour.

In some flows, the entry point is the stock move.

The typical usage is an intercompany stock request (using https://github.com/OCA/stock-logistics-warehouse/tree/11.0/stock_request for example).

rousseldenis (approved):
Code review

rvalyi:
@rousseldenis all right, so may be you can just fix the duplicate line mentioned at https://github.com/OCA/account-invoicing/pull/509#pullrequestreview-237038397 ?

rousseldenis:
> @rousseldenis all right, so may be you can just fix the duplicate line mentioned at [#509 (review)](https://github.com/OCA/account-invoicing/pull/509#pullrequestreview-237038397) ?

@acsonefho

rvalyi:
@rousseldenis @acsonefho up

stock_picking_invoicing/__init__.py
acsonefho: Oh yes! I make a mistake! Thank you for your review!

stock_picking_invoicing/models/stock_invoice_state_mixin.py
```diff
@@ -0,0 +1,63 @@
+# -*- coding: utf-8 -*-
+# Copyright (C) 2019-Today: Odoo Community Association (OCA)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import api, fields, models
+
+
+class StockInvoiceStateMixin(models.AbstractModel):
```
rousseldenis, line 7: IMHO, this one should be externalized in a generic module ... this code would never be reused as this...

renatonlima:
Hi Guys,

Travis there is one error on test https://travis-ci.org/OCA/account-invoicing/jobs/550298624#L1794 in module account_invoice_change_currency it is necessary rounding the sum of invoice lines https://github.com/acsone/account-invoicing/blob/10.0-stock_picking_invoicing/account_invoice_change_currency/tests/test_account_invoice_change_currency.py#L85

rousseldenis:
> Hi Guys,
> 
> Travis there is one error on test https://travis-ci.org/OCA/account-invoicing/jobs/550298624#L1794 in module account_invoice_change_currency it is necessary rounding the sum of invoice lines https://github.com/acsone/account-invoicing/blob/10.0-stock_picking_invoicing/account_invoice_change_currency/tests/test_account_invoice_change_currency.py#L85

@acsonefho You take care of this ?

stock_picking_invoicing/models/stock_invoice_state_mixin.py
rvalyi: @rousseldenis I am not shocked by the current design. I didn't understand what you mean, could you elaborate?

rousseldenis: @rvalyi What I wanted to say is that this part of code could be reused in any other model that wants to implement invoice states without necessary pulling this module.

rvalyi: Ok. Well may be some improvement to be made later? Or in v12? I'm a bit worried it delays the merge again with speculations about possible usage...

rousseldenis: Ok let's merge it after fix is done

renatonlima (approved):

renatonlima:
ping @pedrobaeza

Cedric-Pigeon (approved):

Cedric-Pigeon:
@acsonefho Could you resolve conflict please?

Cedric-Pigeon (approved):

PR #511, opened 2019-03-13
[FIX] account_invoice_change_currency: Fixed get rate at first currency convertion.

hugho-ad:
Before this commit, was returned the same currency that the invoice on
the onchange context now is returned the currency recorded at the
messages with tracking values.

FIX https://github.com/OCA/account-invoicing/issues/512

### DIFF
--- a/account_invoice_change_currency/models/account_change_currency.py
+++ b/account_invoice_change_currency/models/account_change_currency.py
@@ -131,6 +131,7 @@ ORDER BY mtv.write_date DESC, mtv.id DESC LIMIT 1"""
     @api.multi
     def get_last_rate(self):
         self.ensure_one()
+        subtype_create_id = self.env.ref('account.mt_invoice_created')
         last_values = self.env['mail.tracking.value'].sudo().search([
             ('mail_message_id', 'in', self.message_ids.ids),
             ('field', 'in', ['rate', 'currency_id']),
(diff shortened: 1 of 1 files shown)


moylop260 (approved):

hugho-ad:
@pedrobaeza 

could you review / merge please?

Regards!

pedrobaeza:
I don't like too much the method of extracting previous value from tracking values. That can fail in thousands of unexpected ways.

hugho-ad:
@pedrobaeza 

If you delete manually those messages is the only way that I see that could fail

pedrobaeza:
Or if there is a manual message added... As said, it's very loose method.

hugho-ad:
> Or if there is a manual message added... As said, it's very loose method.

For this, is not possible, we are creating own message subtypes to handle this, How could you replicate?
Can you show us step by step?

Regards!

pedrobaeza:
This is a very tricky method. As said, I don't like it and it's going to have side effects sooner or later, so I don't approve it. If others consider it OK, then go ahead with it.

hugho-ad:
@novawish 

Could you help us please with your review and feedback, please

Regards!

hugho-ad:
@novawish 
Could you help us please with your review and feedback, please

Regards!

luistorresm (approved):
@hugho-ad LGTM, works fine.

luistorresm:
Hello @pedrobaeza 

Do you have other option to fix this case?

Regards

pedrobaeza:
I don't have any right now, but proceed with it if you think so.

moylop260:
Merging since that it is a fix, a full refactoring is required using another PR:

/ocabot merge

PR #386, opened 2018-06-10
[IMP] account_invoice_merge: refactor invoice merging

mpanarin:
Refactor of this module. It used deprecated methods, was obscure and hard to extend. 
Now it is much cleaner and readable, switching to SavepointCase tests helped a lot as well (from ~30 seconds to ~4)

PR #514, opened 2019-03-15
Revert "[IMP] account_invoice_merge: refactor invoice merging"

elicoidal:
Reverts OCA/account-invoicing#386


elicoidal:
@mreficent @intero-chz

elicoidal:
@mpanarin

elicoidal:
Travis is green again: Fast tracking

mpanarin:
Okay, although I don't really understand the reason for the revert. Because of this refactor modules started failing or what? As during the development the branch was green. Maybe instead of reverting this, fixing the failing modules was worth the shot?

intero-chz:
@mpanarin i think this fixes should be done with this PR here. As i already wrote only the _attachment module seems a bit more complicated. The other two modules should be fixed in no time :-)

PR #517, opened 2019-03-28
[10.0][FIX] account invoice triple discount: onchange cache management

SimoRubi:
Fix https://github.com/OCA/sale-workflow/issues/817

### DIFF
--- a/account_invoice_triple_discount/models/account_invoice.py
+++ b/account_invoice_triple_discount/models/account_invoice.py
@@ -54,7 +54,9 @@ class AccountInvoiceLine(models.Model):
         Resetting discount2 and discount3 to 0.0 avoids issues if
         this method is called multiple times."""
         prev_values = dict()
-        self.invalidate_cache(['discount', 'discount2', 'discount3'], self.ids)
+        if not self.env.in_onchange:
+            self.invalidate_cache(
+                ['discount', 'discount2', 'discount3'], self.ids)
         for line in self:
             prev_values[line] = dict(
                 discount=line.discount,
--- a/account_invoice_triple_discount/models/account_invoice.py
+++ b/account_invoice_triple_discount/models/account_invoice.py
@@ -54,6 +54,7 @@ class AccountInvoiceLine(models.Model):
         Resetting discount2 and discount3 to 0.0 avoids issues if
         this method is called multiple times."""
         prev_values = dict()
+        self.invalidate_cache(['discount', 'discount2', 'discount3'], self.ids)
         for line in self:
             prev_values[line] = dict(
                 discount=line.discount,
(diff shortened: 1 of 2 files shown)


account_invoice_triple_discount/models/account_invoice.py
```diff
@@ -54,6 +54,9 @@ def triple_discount_preprocess(self):
         Resetting discount2 and discount3 to 0.0 avoids issues if
         this method is called multiple times."""
         prev_values = dict()
+        if not self.env.in_onchange:
```
lmignon, line 57: @SimoRubi Can you elaborate why this trick is required?

SimoRubi: @lmignon I don't know exactly why this works, I expected all the changes of #500 to work without any cache management.

The really wrong thing in my opinion is that the automated tests do not work without the `invalidate_cache` trick.
While debugging the tests issue (that note, 1 in 5 times were not failing!) I noticed the following (extracted from this same file):
```
    @api.multi
    @api.depends('discount2', 'discount3')
    def _compute_price(self):
        prev_values = self.triple_discount_preprocess()
        # Here discount is 50 and discount3 is 0, but inside the following method discount is 0 again.
        # More in particular, the record's discounts values change after a _browse call in the model's __iter__ method
        super(AccountInvoiceLine, self)._compute_price()
        self.triple_discount_postprocess(prev_values)
```
Moreover, this happened only for the 2nd of the lines created in `test_04`..
So I concluded that something was really messed up with the cache management

So I had to add the `invalidate_cache` line, but that somehow breaks the interface behavior as explained in https://github.com/OCA/account-invoicing/pull/517/commits/4411c63c9c2b807f8c17b8354e1ce85085a50f2a..

lmignon: @SimoRubi IMO it's not the right fix. If something is wrong with the cache we must fix it into Odoo. This error could  also reveal a problem into the way the 2 addons are designed.

SimoRubi: @lmignon I agree this is not the right fix, I would simply call this a workaround.
Anyway I probably won't have time to investigate further into the Odoo cache management and find the right fix (for these modules, or for the Odoo core).

lmignon:
@SimoRubi Can you check #525. It fix the same issue in an other way and more details on what's wrong...

SimoRubi:
@lmignon I have done my review in #525, please take a look over there to see the latest updates on these issues/PRs management

SimoRubi:
@OCA/accounting-maintainers any news about this?
I understand from https://github.com/OCA/sale-workflow/issues/817#issuecomment-537047405 that this might be closed in the next future

pedrobaeza:
Closing in favor of #525

PR #520, opened 2019-04-17
[FIX][8.0] account_invoice_pricelist fail with fiscal position Incl -> Excl

legalsylvain:
This PR fixes incompatibility between account_invoice_pricelist and fiscal_position, in the case when fiscal position map taxes from TAX incl to Tax Excl. (B2B fiscal position in a B2C context).

**Scenario**
- Product : price 12€ vat Incl, (VAT 20%). (So 10€ Vat Excl)
- Pricelist : discount of 10%. (sale price * 0.9)
- Fiscal position that map Tax 20% Incl --> Tax 20% Excl

**Current Result**
- Saling the product, the result is 10.8 € Vat Excl. (compute : 12€ - 10%) that is wrong

**Expected Result**
- saling the product, the result should be 9€ Vat Excl.
    - Pricelist process : 12,00€ Incl --> 10,80 € Incl  ( x 0.9)
    - Fiscal Position Process : 10,80 € Incl --> 9€ Excl ( / (1 + 0,2))

**Origin**
In the core, the function `product_id_change` call a function `_fix_tax_included_price` that fix the unit price with such fiscal position. 
This module overwrite product_id_change, and doesn't call this function. _fix_tax_included_price

### DIFF
--- a/account_invoice_pricelist/model/account_invoice_line.py
+++ b/account_invoice_pricelist/model/account_invoice_line.py
@@ -46,6 +46,8 @@ class AccountInvoiceLine(models.Model):
         )
         pricelist_model = self.env['product.pricelist']
         product_model = self.env['product.product']
+        AccountTax = self.env['account.tax']
+        AccountAccount = self.env['account.account']
         product = product_model.browse(product)
         pricelist = pricelist_model.browse(pricelist_id)
         values = res['value']
@@ -57,6 +59,8 @@ class AccountInvoiceLine(models.Model):
         else:
             pricedict = pricelist.price_get(product.id, qty, partner_id)
         price_unit = pricedict[pricelist_id]
+
+        # MultiCurrency context
         if currency_id:
             company = self.env['res.company'].browse(company_id)
             currency = self.env['res.currency'].browse(currency_id)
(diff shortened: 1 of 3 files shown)


legalsylvain:
CC :  Author : @NL66278. Reviewers : @pedrobaeza, @hbrunn, @lfreeke

Thanks for your review !

account_invoice_pricelist/model/account_invoice_line.py
```diff
@@ -65,6 +69,18 @@ def product_id_change(
                     price_unit, currency, round=True
                 )
                 price_unit = currency.round(price_unit)
+
+        # Fiscal position with Tax Excl / Incl context
+        if fposition_id:
+            fp_tax_ids = res['value']['invoice_line_tax_id']
+            account = AccountAccount.browse(res['value']['account_id'])
+            if type in ('out_invoice', 'out_refund'):
```
nikul-serpentcs, line 77: @legalsylvain here No need to` else`
```
taxes = product.supplier_taxes_id or account.tax_ids
if type in ('out_invoice', 'out_refund'):
    taxes = product.taxes_id or account.tax_ids
```

legalsylvain: Yes, you're right. But it's not my code, it's a copy of odoo code, so, I'll prefer to keep the same algorigthm, to make the code more easy to compare. 

Are you ok ? 
https://github.com/odoo/odoo/blob/8.0/addons/account/account_invoice.py#L1375

nikul-serpentcs: @legalsylvain  keep it

PR #522, opened 2019-04-29
[ADD] Added module account_auto_reconcile_rounding_differences

wpichler:


account_auto_reconcile_rounding_differences/__init__.py
```diff
@@ -0,0 +1,4 @@
+# -*- coding: utf-8 -*-
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
```
CarmenMiranda: Here you should put the License like it's said in the manteiner-tools "init" template (https://github.com/OCA/maintainer-tools/blob/master/template/module/__init__.py#L2)

CarmenMiranda:
Hi @wpichler, I think that you need put a description with an explanation of what this PR do or why it's need it.
I notice that it's missing the License in the manifest file, you should put one like it's said in the OCA template (https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py#L3)

Also in the models and the view the licenses and author are missing.  Examples:
- Models: (https://github.com/OCA/maintainer-tools/blob/master/template/module/models/model_name.py#L2)
- Views: (https://github.com/OCA/maintainer-tools/blob/master/template/module/views/model_name_view.xml#L2).

Besides I note that there is not a README file in this module, you should add one, to explain what this Module do, it's configuration, etc. For example, the README in the module "account_group_invoice_line"(https://github.com/OCA/account-invoicing/blob/10.0/account_group_invoice_line/README.rst).

Another thing, for future references/changes/improvements you should add a description to the methods.

account_auto_reconcile_rounding_differences/models/account_payment.py
```diff
@@ -0,0 +1,17 @@
+# -*- coding: utf-8 -*-
+
+from odoo import models, fields, api, _
+
+
+class account_payment(models.Model):
```
CarmenMiranda: Here you should follow the UpperCaseCamelCase convention to name the class.
```python
class AccountPayment(models.Model):
```

pedrobaeza:
Please don't use plurals on module names.

PR #525, opened 2019-05-10
[10.0][FIX] account_invoice_triple_discount: Fix dependency issue

acsonefho:
**Bug**
In some case, amounts are not corrects (subtotal etc).
It's because each iteration (of api.multi) could be done in different cache.
As we're updating cache in this case, it could be possible that another cache is used, so without edited values. So the amount is not the expected one.
To fix that, we just have to use api.one (like the original method in Odoo) to stay into the same cache.

(we need this fix to avoid issues into sale-workflow/sale_triple_discount)

### DIFF
--- a/account_invoice_triple_discount/models/account_invoice.py
+++ b/account_invoice_triple_discount/models/account_invoice.py
@@ -32,8 +32,15 @@ class AccountInvoiceLine(models.Model):
         default=0.0,
     )
 
-    @api.multi
-    @api.depends('discount2', 'discount3')
+    # Each iteration of api.multi is done into a different cache!
+    # As we're using cache to define a value into discount field, we have
+    # to stay into the same cache. So we can not use api.multi in this case.
+    @api.one
+    @api.depends(
+        'price_unit', 'discount', 'invoice_line_tax_ids', 'quantity',
+        'product_id', 'invoice_id.partner_id',
+        'invoice_id.currency_id', 'invoice_id.company_id',
+        'invoice_id.date_invoice', 'invoice_id.date', 'discount2', 'discount3')
     def _compute_price(self):
         prev_values = self.triple_discount_preprocess()
         super(AccountInvoiceLine, self)._compute_price()
@@ -60,7 +67,7 @@ class AccountInvoiceLine(models.Model):
                 discount2=line.discount2,
                 discount3=line.discount3,
             )
-            line.update({
+            line._cache.update({
                 'discount': line._get_triple_discount(),
                 'discount2': 0.0,
                 'discount3': 0.0
--- a/account_invoice_triple_discount/models/account_invoice.py
+++ b/account_invoice_triple_discount/models/account_invoice.py
@@ -5,6 +5,7 @@
 
 from odoo import api, fields, models
 from odoo.addons import decimal_precision as dp
+from odoo.tools import float_compare
 
 
 class AccountInvoice(models.Model):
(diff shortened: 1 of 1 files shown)


lmignon:
@acsonefho A little bit tricky to deal with the cache... it feels to me like a design problem into the addons or an issue into Odoo. #517 try to solve the same issue in an other way (the one with the cache not the triggers)

account_invoice_triple_discount/models/account_invoice.py
```diff
@@ -60,7 +67,7 @@ def triple_discount_preprocess(self):
                 discount2=line.discount2,
                 discount3=line.discount3,
             )
-            line.update({
+            line._cache.update({
```
lmignon, line 81: @acsonefho  Why do you deal with the cache? Is it required. Behind the scene it's what the `update`method do

account_invoice_triple_discount/models/account_invoice.py
```diff
@@ -32,8 +32,15 @@ class AccountInvoiceLine(models.Model):
         default=0.0,
     )
 
-    @api.multi
-    @api.depends('discount2', 'discount3')
+    # Each iteration of api.multi is done into a different cache!
+    # As we're using cache to define a value into discount field, we have
+    # to stay into the same cache. So we can not use api.multi in this case.
+    @api.one
```
SimoRubi, line 39: Deprecated, see https://github.com/odoo/odoo/blob/feef39aee083294f8c8d19aecef0434af04c48d0/odoo/api.py#L359

account_invoice_triple_discount/models/account_invoice.py
```diff
@@ -32,8 +32,15 @@ class AccountInvoiceLine(models.Model):
         default=0.0,
     )
 
-    @api.multi
-    @api.depends('discount2', 'discount3')
+    # Each iteration of api.multi is done into a different cache!
```
SimoRubi, line 36: I don't think this is the expected behavior, so looks like it is a bug of odoo core.
You might open an issue in there addressing the wrong behavior or create a PR if you have actually found the misbehaving code.

SimoRubi (changes requested):
Thanks for investigating this issue!
I already tried a different workaround in https://github.com/OCA/account-invoicing/pull/517

account_invoice_triple_discount/models/account_invoice.py
acsonefho: @lmignon I have to deal with the cache because (as the module is currently designed), every discount fields are computed into the standard discount field (into cache). Then, after calling the standard behaviour (`super`) the custom `discount` fields are re-updated with filled value.
But as there is `api.multi` and `api.one`, Odoo change the cache in some case.
I agree with your previous comment: maybe it's a design problem or an issue into Odoo core (so in this case, my fix is only a workaround).

account_invoice_triple_discount/models/account_invoice.py
acsonefho: @SimoRubi Thank you for your comment but I explain why I was forced to use `@api.one` (instead of `multi`) on the line just before.

grindtildeath (approved):
Looks ok :+1:

i-vyshnevska (approved):

rafaelbn:
Hi @SimoRubi @lmignon do you approve technically this PR?

SimoRubi:
> Hi @SimoRubi @lmignon do you approve technically this PR?

@rafaelbn as I explained in the review https://github.com/OCA/account-invoicing/pull/525#pullrequestreview-241586146 I'm actually torn from this because it re-introduces a deprecated decorator.
I'd like to know why it does not work and at least point it out (with an issue) to odoo core so it can hopefully fix it.
I can't do that because I didn't investigated enough and don't know the details, but maybe @acsonefho knows it better and could open an issue in odoo/odoo

rafaelbn:
Please @acsonefho @SimoRubi could we you clarify this in the Code Sprint?

eLBati (approved):
https://github.com/OCA/sale-workflow/issues/817#issuecomment-537047405

pedrobaeza:
@lmignon do you agree with this solution?

rousseldenis:
@lmignon

gurneyalex:
Hello, we use this PR in production, and we experienced a bug when, in some rare random cases, the cache would be reset in the middle of the call to super()._compute_price(). It maybe because of some custom code for our customer which changes the dependencies. Still, it's not nice, because it shows that the code in this module is fragile (and we also had issues with sale_triple_discount for the same customer). 

This patch https://github.com/acsone/account-invoicing/pull/2 works around the issue for us by detecting the failure and re running the method. 

On the long term, I think that the approach of these modules should be changed by adding a discount1 field and making the standard discount field a computed field combining discount1 discount2 and discount3, and stopping to mess with the cache. 


cc @guewen @yvaucher

pedrobaeza:
Let's merge finally this patch as the most satisfactory one:

/ocabot merge patch

PR #529, opened 2019-05-21
[FIX] account_invoice_pricelist call _fix_tax_included_price

legalsylvain:
if the module account_invoice_pricelist is installed, the call to _fix_tax_included_price is not realized. So the price_unit can be bad;

This patch fix this problem.


bealdav (approved):

legalsylvain:
Hi @bealdav . thanks for your review. could you review also this related simple PR : 
https://github.com/OCA/account-invoicing/pull/528

regards.

legalsylvain:
Travis is red because this PR use ``type`` that is mandatory because it is the name of the argument of the function. Merging trivial patch.

PR #533, opened 2019-05-28
[ADD] New module account_analytic_billing_plan

oihane:


fclementic2c:
title sounds interesting... What is the purpose exactly ?

anajuaristi:
Hi @fclementic2c 
We have got a couple of customers who just make an agreement with the customer to invoice a project depending on finished milestone tasks. 
They don't make recurrent invoices, nor invoice the sale order, nor invoice real hours but... supose this case of use:
I will invoice 20% of the project at the begining
10% more when installation is done estimated data 31 May
20% more when development1 is finished estimated data first June
20% more when devolopment2 is finished estimated data 15 June
30% more when training is finished estimated 30June
so you can have an invoicing plan on your projects, where team members just mark the done tasks and accountants can see what is ready to invoice or not, depending on tasks status and stimated datas of the lines.
Pleas ask me if you need more information

account_analytic_billing_plan/__manifest__.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2018 Alfredo de la Fuente - AvanzOSC
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+{
+    "name": "Account Analytic Billing Plan",
+    "version": "12.0.1.0.0",
```
nikul-serpentcs, line 6: @oihane If added new Module then the version should be `"12.0.0.0.0"`.
Or Add `Summary` key

nikul-serpentcs (commented):
Nitpicking

account_analytic_billing_plan/__manifest__.py
pedrobaeza: That's not true... Version must be always 12.0.1.0.0 for new version, and summary is not mandatory.

Daniel-CA (approved):
LGTM

PR #544, opened 2019-06-25
[10.0][ADD] back porting account_invoice_tax_required

labaggio:


pedrobaeza:
This is already present in https://github.com/OCA/account-financial-tools/tree/10.0/account_invoice_tax_required. It was moved to the other repository because it fits better.

labaggio:
> This is already present in https://github.com/OCA/account-financial-tools/tree/10.0/account_invoice_tax_required. It was moved to the other repository because it fits better.

thank you very much, I had not found it

pedrobaeza:
I use OCA "apps store" for knowing these things:

https://odoo-community.org/shop/product/tax-required-in-invoice-323?search=account_invoice_tax_required+

PR #546, opened 2019-06-26
12 migr inv pricelist sale

bealdav:


account_invoice_pricelist_sale/__manifest__.py
```diff
@@ -8,10 +8,10 @@
     'license': 'AGPL-3',
     'website': 'http://therp.nl',
```
nikul-serpentcs: URL should be `https://github.com/OCA/account-invoicing`

account_invoice_pricelist_sale/__manifest__.py
```diff
@@ -8,10 +8,10 @@
...
     'category': 'Sales',
```
nikul-serpentcs: 'category': 'Accounting & Finance',

nikul-serpentcs (commented):
nitpicking

PR #547, opened 2019-06-26
[12.0][ADD] base_ebill_payment_contract

TDu:
The goal of this module is to be a common ground for ebilling addons.
Alone this module has no purpose. It needs to be installed with a module
that implements a specific ebilling system.
This depending module has to implement his own 'transmit.method'.


pedrobaeza:
Why here and not in OCA/edi? About the name, base modules should prefix (not suffix) `base` in the name. Also ebilling is a very diffuse name...

guewen:
> Why here and not in OCA/edi

Can be discussed... this is mainly to represent payment contracts for e-billing services. It's a thin layer on top of `account_invoice_transmit_method` which is in `account-invoicing`. This module does not do any EDI by itself, even if it is indeed meant as a basis for sort of EDI (as `account_invoice_transmit_method`).

> Also ebilling is a very diffuse name...

What about `base_ebill_payment_contract`?

pedrobaeza:
Uhm, I'm not fully understanding the aim of the module then and the weird contract model. We have also on OCA/contract modules for setting agreements. I abstain then from giving more opinions.

TDu:
@pedrobaeza I did not know about 'base' should prefix a module name. I fixed it following proposed name from @guewen

base_ebill_payment_contract/models/ebill_payment_contract.py
```diff
@@ -0,0 +1,56 @@
...
+    _description = "Ebill Payment Contract"
+    _rec_name = "name"
+
+    transmit_method_id = fields.Many2one(
+        comodel_name="transmit.method",
+        string="Transmit method",
+        ondelete="restrict",
+    )
+    partner_id = fields.Many2one(comodel_name="res.partner", required=True)
+    name = fields.Char(related="partner_id.name")
```
sebalix, line 23: Should it be readonly?

base_ebill_payment_contract/models/ebill_payment_contract.py
```diff
@@ -0,0 +1,56 @@
...
+    date_start = fields.Date(
+        string="Start date",
+        required=True,
+        default=lambda self: fields.Date.today(),
```
sebalix:
```suggestion
        default=fields.Date.today,
```

base_ebill_payment_contract/readme/CONFIGURE.rst
```diff
@@ -0,0 +1 @@
+
```
sebalix: This file can be removed I guess (no content)?

base_ebill_payment_contract/readme/INSTALL.rst
```diff
@@ -0,0 +1 @@
+
```
sebalix: Same

base_ebill_payment_contract/readme/USAGE.rst
```diff
@@ -0,0 +1,3 @@
+To use this module, you need to:
+
+#. Install a specific e-billing module.
```
sebalix: An module example is welcome, but I suppose the module is not ready yet :)

base_ebill_payment_contract/tests/test_base_ebill_payment_contract.py
```diff
@@ -0,0 +1,56 @@
+# Copyright 2019 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from datetime import date, timedelta
+from odoo.tests.common import Form, SingleTransactionCase
```
sebalix, line 6: Nitpick: one line space between standard and third party imports.

base_ebill_payment_contract/views/ebill_payment_contract.xml
```diff
@@ -0,0 +1,64 @@
...
+
+<odoo>
+
+  <record model="ir.ui.view" id="ebill_payment_contract_form_view">
+    <field name="name">ebill.payment.contract.form</field>
+    <field name="model">ebill.payment.contract</field>
+    <field name="arch" type="xml">
+      <form>
+        <header>
+        </header>
```
sebalix: Is it a placeholder for future buttons? If so a comment is welcome to not think it's a dead tag.

base_ebill_payment_contract/views/ebill_payment_contract.xml
```diff
@@ -0,0 +1,64 @@
...
+      </form>
+    </field>
+  </record>
+
+  <record model="ir.ui.view" id="ebill_payment_contract_search_view">
+    <field name="name">ebill.payment.contract.search</field>
+    <field name="model">ebill.payment.contract</field>
+    <field name="arch" type="xml">
+      <search>
+      </search>
```
sebalix, line 33: Same?

base_ebill_payment_contract/models/ebill_payment_contract.py
i-vyshnevska: in v12 related readonly by default

TDu: Indeed tested, and it does not seem to be specificity mentioned in the official doc.

base_ebill_payment_contract/models/ebill_payment_contract.py
TDu: Thanks fixed

base_ebill_payment_contract/readme/USAGE.rst
TDu: Nope, so I left a comment to inform

base_ebill_payment_contract/tests/test_base_ebill_payment_contract.py
TDu: Done

base_ebill_payment_contract/views/ebill_payment_contract.xml
TDu: I just added some search fields and filter :+1:

sebalix:
@TDu Travis is red: `base_ebill_payment_contract/models/ebill_payment_contract.py:18:80: E501 line too long (94 > 79 characters)`

sebalix (approved):

base_ebill_payment_contract/models/ebill_payment_contract.py
```diff
@@ -0,0 +1,60 @@
...
+    date_end = fields.Date(string="End date")
+    state = fields.Selection(
+        selection=[("draft", "Draft"), ("open", "Open"), ("cancel", "Cancel")],
+        required=True,
+        default='draft',
+    )
+    is_valid = fields.Boolean(compute="_compute_is_valid", store=True)
+
+    @api.onchange('state')
+    def _compute_state_changed(self):
```
grindtildeath:
```suggestion
    def _onchange_state(self):
```

base_ebill_payment_contract/models/ebill_payment_contract.py
```diff
@@ -0,0 +1,60 @@
...
+        """Change the end date if contract is canceled."""
+        if self.state == 'cancel' and self.date_end > fields.Date.today():
+            self.date_end = fields.Date.today()
+
+    @api.multi
+    @api.depends("state", "date_start", "date_end")
```
grindtildeath, line 44: I'm just wondering here, if `is_valid` was computed at record creation, and after a while `date_end` is yesterday, the field won't be recomputed since there is no trigger, is it right ?

grindtildeath (commented):
Small things, otherwise looks good.

base_ebill_payment_contract/models/ebill_payment_contract.py
TDu: This is a good point. Maybe `is_valid` should not be a field then but simply a function ?

grindtildeath: If you remove `store=True` it will be recomputed anyway so there wouldn't be any issue. But if it must be stored, I don't know what's the best practice to recompute the field everyday.

TDu:
I am closing this as the customer moved to the next Odoo version.
For version 13.0, I am moving it to the edi repository as suggested. https://github.com/OCA/edi/pull/207
Thanks for all the reviews, they have been attended.

PR #550, opened 2019-07-01
FIX acc_invoice_pricelist: tax include in multicompany

bealdav:
This bug is only present in multicompany context.

Odoo only call _fix_tax_included_price() once

https://github.com/odoo/odoo/blob/12.0/addons/account/models/account.py#L1181

Concerned versions: 10.0 11.0 12.0

cc @mourad-ehm @florian-dacosta @sebastienbeau

### DIFF
--- a/account_invoice_pricelist/models/account_invoice.py
+++ b/account_invoice_pricelist/models/account_invoice.py
@@ -57,8 +57,10 @@ class AccountInvoiceLine(models.Model):
             fiscal_position=(
                 self.invoice_id.partner_id.property_account_position_id.id)
         )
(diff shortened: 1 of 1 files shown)


mourad-ehm (approved):

PierrickBrun (approved):

legalsylvain:
@OCA-git-bot is wonderfull. ;-)

bealdav:
yes but ocabot is good for unemployment increase.

Now Pedro has a little free time.

Don't make ocabot too clever, otherwise ...

;-)

PR #555, opened 2019-07-02
[12.0][ADD] account_menu_invoice_refund

kittiu:
By Odoo standard, Invoices and Refunds are in different menu, and can't be mixed upon payment.

This module add 2 new menus,

1. Customers > Invoices / Credit Notes
2. Vendors > Bills / Refunds

Additionally it allow register net payment by selecting both invoice and refund.

**Note:**
This is accomplished by simply remove the condition that disallow mixing invoice and refund in account.payment's default_get()


pedrobaeza:
Good idea

etobella:
ping @jarroyomorales

jarroyomorales:
LGTM!

account_menu_invoice_refund/hooks.py
```diff
@@ -0,0 +1,75 @@
...
+
+MAP_INVOICE_TYPE_PARTNER_TYPE = {
+    'out_invoice': 'customer',
+    'out_refund': 'customer',
+    'in_invoice': 'supplier',
+    'in_refund': 'supplier',
+}
+
+
+def post_load_hook():
```
AaronHForgeFlow, line 16: It works so this seems correct, but why not putting it on the account_abstract_payment class?

kittiu: You mean, why not inherit account.abstract.payment?
I want to remove the checking, I need use hook. I don't know how to do it any other way though.

kittiu:
Ready for merge

account_menu_invoice_refund/hooks.py
AaronHForgeFlow: :+1:

AaronHForgeFlow (approved):
:+1: Functional tests

jarroyomorales (approved):

PR #556, opened 2019-07-03
[10.0] [ADD] account_invoice_split_refund

grindtildeath:
This module adds a new "Advance Payment Method" to "Sales Advance Payment Invoice"
wizard in order to split invoices and refunds, and not have any line with a negative
quantity on the invoice or the refund.

### DIFF
--- a/account_invoice_split_refund/tests/test_invoice_split_refunds.py
+++ b/account_invoice_split_refund/tests/test_invoice_split_refunds.py
@@ -34,17 +34,19 @@ class TestInvoiceSplitRefunds(TestSale):
         })
         self.sale_order.action_confirm()
 
-        delivery_picking = self.sale_order.picking_ids
-        delivery_picking.force_assign()
-        delivery_picking.pack_operation_product_ids.write({'qty_done': 5})
-        delivery_picking.do_new_transfer()
+        self.delivery_picking = self.sale_order.picking_ids
+        self.delivery_picking.force_assign()
+        self.delivery_picking.pack_operation_product_ids.write({'qty_done': 5})
+        self.delivery_picking.do_new_transfer()
 
+    def test_account_invoice_split_refunds(self):
         invoice_id = self.sale_order.action_invoice_create()
         self.invoice = self.env['account.invoice'].browse(invoice_id)
         self.invoice.action_invoice_open()
 
         return_wiz = self.env['stock.return.picking'].with_context(
-            active_ids=delivery_picking.ids, active_id=delivery_picking.id
+            active_ids=self.delivery_picking.ids,
+            active_id=self.delivery_picking.id
         ).create({})
         return_wiz.product_return_moves.write({
             'quantity': 2,
@@ -55,8 +57,6 @@ class TestInvoiceSplitRefunds(TestSale):
         return_picking.force_assign()
         return_picking.pack_operation_product_ids.write({'qty_done': 2})
         return_picking.do_new_transfer()
-
-    def test_account_invoice_split_refunds(self):
 
         self.assertEqual(self.sale_order_line.qty_delivered, 3)
         self.assertEqual(self.sale_order_line.qty_invoiced, 5)
@@ -96,3 +96,10 @@ class TestInvoiceSplitRefunds(TestSale):
         )
 
         self.assertEqual(new_invoice.invoice_line_ids.quantity, 6)
+
+    def test_invoice_without_refunds(self):
+        create_invoice_wiz = self.env['sale.advance.payment.inv'].with_context(
+            active_ids=self.sale_order.ids, active_id=self.sale_order.id
+        ).create({})
+        create_invoice_wiz.create_invoices()
+        self.assertEqual(len(self.sale_order.invoice_ids), 1)
--- a/account_invoice_split_refund/tests/test_invoice_split_refunds.py
+++ b/account_invoice_split_refund/tests/test_invoice_split_refunds.py
@@ -39,7 +39,7 @@ class TestInvoiceSplitRefunds(TestSale):
         self.delivery_picking.pack_operation_product_ids.write({'qty_done': 5})
         self.delivery_picking.do_new_transfer()
 
-    def test_account_invoice_split_refunds(self):
+    def test_account_invoice_split_refund_and_invoice(self):
         invoice_id = self.sale_order.action_invoice_create()
         self.invoice = self.env['account.invoice'].browse(invoice_id)
         self.invoice.action_invoice_open()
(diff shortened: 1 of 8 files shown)


pedrobaeza:
Please don't use plurals on the module name.

account_invoice_split_refund/tests/test_invoice_split_refunds.py
```diff
@@ -0,0 +1,102 @@
...
+        self.sale_order_line = self.env['sale.order.line'].create({
+            'order_id': self.sale_order.id,
+            'name': self.delivery_product.name,
+            'product_id': self.delivery_product.id,
+            'product_uom_qty': 5,
+            'product_uom': self.delivery_product.uom_id.id,
+            'price_unit': self.delivery_product.list_price
+        })
+        self.sale_order.action_confirm()
+        # import pdb; pdb.set_trace()
```
i-vyshnevska:
```suggestion
```

account_invoice_split_refund/tests/test_invoice_split_refunds.py
```diff
@@ -0,0 +1,102 @@
...
+        delivery_picking = self.sale_order.picking_ids
+        delivery_picking.force_assign()
+        delivery_picking.pack_operation_product_ids.write({'qty_done': 5})
+        delivery_picking.do_new_transfer()
+        # wiz = self.env[wiz_act['res_model']].browse(wiz_act['res_id'])
+        # wiz.process()
```
i-vyshnevska: cleanup here too

account_invoice_split_refund/tests/test_invoice_split_refunds.py
```diff
@@ -0,0 +1,102 @@
...
+        self.assertEqual(
+            new_refund.invoice_line_ids.product_id, self.delivery_product
+        )
+        # As we invoiced 5 and returned 2, we must refund 2
+        self.assertEqual(new_refund.invoice_line_ids.quantity, 2)
+        self.assertEqual(
+            new_invoice.invoice_line_ids.product_id, service_product
+        )
+
+        self.assertEqual(new_refund.invoice_line_ids.quantity, 5)
```
i-vyshnevska: line 96 and 101 ?

account_invoice_split_refund/wizard/sale_make_invoice_advance.py
```diff
@@ -0,0 +1,35 @@
...
+    @api.multi
+    def create_invoices(self):
+        if self.advance_payment_method != 'all_split':
+            return super(SaleAdvancePaymentInv, self).create_invoices()
+        sale_orders = self.env['sale.order'].browse(
+            self._context.get('active_ids', []))
+        # Create all the invoices first
+        sale_orders.action_invoice_create()
+        # Create any refund afterwards
+        sale_orders.action_invoice_create(final=True)
```
i-vyshnevska: if this line fails it will affect result of 29?

grindtildeath: If it fails, the transaction is rolled back and that's perfectly fine.

i-vyshnevska: if first call successful, second fails, result of first rolled back?
 I am thinking that we twice calls same function and it raise Usererror

TDu (approved):

grindtildeath:
Travis error unrelated and fixed in https://github.com/OCA/account-invoicing/pull/561

i-vyshnevska (approved):

account_invoice_split_refund/wizard/sale_make_invoice_advance.py
jbaudoux: It won't work if there is no invoice or no refund to create. I think it will work only if there is an invoice and refund to create. You need to add test condition or try/catch

jbaudoux (changes requested):
Does not work as expected. You cannot anymore make the invoicing if there is not an invoice and refund to create.

account_invoice_split_refund/wizard/sale_make_invoice_advance.py
grindtildeath: Fixed in 7bcf35d

grindtildeath:
@i-vyshnevska Yes, just fixed it afterwards with another commit :wink:

grindtildeath:
@jbaudoux Last change should fix any issue and I added more tests with the three use cases, i.e:
 - Only invoices to generate
 - Only refunds to generate
 - Invoices and refunds to generate

account_invoice_split_refund/wizard/sale_make_invoice_advance.py
```diff
@@ -0,0 +1,40 @@
...
+        """Force splitting refunds from invoices as default method"""
+        return 'all_split'
+
+    @api.multi
+    def create_invoices(self):
+        if self.advance_payment_method != 'all_split':
+            return super(SaleAdvancePaymentInv, self).create_invoices()
+        sale_orders = self.env['sale.order'].browse(
+            self._context.get('active_ids', []))
+        qties_to_invoice = sale_orders.mapped('order_line.qty_to_invoice')
```
i-vyshnevska:
```suggestion
        qties_to_invoice = sum(sale_orders.mapped('order_line.qty_to_invoice'))
```

account_invoice_split_refund/wizard/sale_make_invoice_advance.py
```diff
@@ -0,0 +1,40 @@
...
+        to_invoice = bool(filter(lambda qty: qty > 0, qties_to_invoice))
+        to_refund = bool(filter(lambda qty: qty < 0, qties_to_invoice))
+        # Create all the invoices
+        if to_invoice:
+            sale_orders.action_invoice_create(final=False)
+        # Create all the refunds
+        if to_refund:
+            sale_orders.action_invoice_create(final=True)
```
i-vyshnevska, line 36: should we filter orders  and perform
     sale_orders.action_invoice_create(final=True)
only for them,
can get here again UserError with no lines?

account_invoice_split_refund/tests/test_invoice_split_refunds.py
```diff
@@ -0,0 +1,132 @@
...
+            'to_refund_so': True
+        })
+        return_id = return_wiz.create_returns()['res_id']
+        return_picking = self.env['stock.picking'].browse(return_id)
+        return_picking.force_assign()
+        return_picking.pack_operation_product_ids.write({'qty_done': 2})
+        return_picking.do_new_transfer()
+        self.assertEqual(self.sale_order_line.qty_to_invoice, -2)
+        create_invoice_wiz = self.env['sale.advance.payment.inv'].with_context(
+            active_ids=self.sale_order.ids, active_id=self.sale_order.id
```
i-vyshnevska, line 127: add other sale order without refund

account_invoice_split_refund/wizard/sale_make_invoice_advance.py
grindtildeath: There's no use in filtering here since that's what is done by https://github.com/odoo/odoo/blob/12.0/addons/sale/models/sale.py#L556 and the conditions used here match the conditions there.
So no we can't get a UserError anymore after last changes

account_invoice_split_refund/tests/test_invoice_split_refunds.py
grindtildeath: There's no need to add another sale order as what's important is the `qty_to_invoice` on lines and we have the three possibilities now covered in the test...

i-vyshnevska (approved):

account_invoice_split_refund/wizard/sale_make_invoice_advance.py
i-vyshnevska: thanks for explanation

PR #571, opened 2019-07-25
[IMP] account_invoice_rounding: remove overriding methods of account.tax

tafaRU:
As their base methods were removed since 9.0

See https://github.com/OCA/OpenUpgrade/blob/9.0/addons/account/migrations/9.0.1.1/openupgrade_analysis_work.txt#L460-L461 for further info


hurrinico (approved):
LGTM

dreispt (approved):

PR #573, opened 2019-08-05
[FIX] bad display on invoices form

legalsylvain:
Trivial patch. Fix xml. (currently, the button "Update price" create a bug in the right part of the form.

Before : 

image

With this PR : 

image

Thanks for your review.

### DIFF
--- a/account_invoice_pricelist/views/account_invoice_view.xml
+++ b/account_invoice_pricelist/views/account_invoice_view.xml
@@ -5,14 +5,16 @@
         <field name="model">account.invoice</field>
         <field name="inherit_id" ref="account.invoice_form"/>
         <field name="arch" type="xml">
-            <field name="user_id" position="after">
-                <field name="pricelist_id"/>
+            <button name="action_invoice_open" position="before">
                 <button type="object"
                         name="button_update_prices_from_pricelist"
                         string="Update prices"
                         attrs="{'invisible': ['|', ('pricelist_id', '=', False), ('state', 'not in', ['draft'])]}"
                         help="Update price in lines from the pricelist"
                 />
+            </button>
+            <field name="user_id" position="after">
+                <field name="pricelist_id"/>
             </field>
         </field>
     </record>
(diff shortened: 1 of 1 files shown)


nikul-serpentcs (approved):
Code Review LGTM :+1: 

@pedrobaeza  Fast Tracking

pedrobaeza (commented):
I still prefer the button in the context of the pricelist instead on the header. For doing that without problems in the layout, you must do the same that I have done in:

https://github.com/OCA/delivery-carrier/pull/208/files#diff-6842358945a6663501dd1b676fc37254R8-R20

legalsylvain:
> I still prefer the button in the context of the pricelist instead on the header. For doing that without problems in the layout, you must do the same that I have done in: https://github.com/OCA/delivery-carrier/pull/208/files#diff-6842358945a6663501dd1b676fc37254R8-R20

Hi, @pedrobaeza. yes it's a preference. But it the meantime :

1) I see a **move in Odoo recent versions**. a lot of buttons in **form views** are moving in the head part.
For exemple : 
- **V8** / product / update quantity on hand : 
image
- **V12** / product : update quantity on hand : 
image

2) The "button" that are staying **in form view are mainly links**. 

image


3) **we win place in little screen**. with web_responsive all the button are in the "Quick Action" part : 

image

I think we should follow Odoo SA changes, to have an homogenous software. Don't you think ? 

Well in the meantime, if it's a blocking point for you, feel free to change the code in the current PR to fit with your preference. Both solutions are technically correct, and the main objective of this PR is to fix a bug.

hparfr (approved):
I like it too in the header bar.

remi-filament (approved):

pedrobaeza:
For me the main problem is the context. If you provide the button on the header, you are not associating the action with the fact that changing pricelist will require that update, and people can wrongly think that they need to push that button each time they add a line or similar.

Any way, if all prefer that placement, I won't block.

PR #581, opened 2019-08-20
[UPD] Travis.yml

StefanRijnhart:
https://twitter.com/PedroMBaeza/status/1162448940953952256

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,38 +1,69 @@
 language: python
-
-python:
-  - "2.7"
-
 sudo: false
 cache: pip
 
+python:
+  # For branches > 10.0, put `- "3.5"`
+  - "2.7.13"
+
 addons:
+  # By default postgresql-9.1 is installed but there is issue related:
+  #  https://github.com/OCA/maintainer-quality-tools/issues/432#issuecomment-281580935
+  # Better use higher PostgreSQL version
+  postgresql: "9.6"
   apt:
+#    sources:
+#    Search your sources alias here:
+#      https://github.com/travis-ci/apt-source-whitelist/blob/master/ubuntu.json
     packages:
       - expect-dev  # provides unbuffer utility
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,38 +1,69 @@
...
   global:
-  - VERSION="10.0" TESTS="0" LINT_CHECK="0"
+  - VERSION="10.0" TESTS="0" LINT_CHECK="0" MAKEPOT="0"
+  # Set this variable to some version existing as linux-generic build on
+  # https://github.com/wkhtmltopdf/wkhtmltopdf/releases
+  # if you need to install wkhtmltopdf
+  # - WKHTMLTOPDF_VERSION="0.12.5"
+  # Set the above to install a `wkhtmltopdf` version that is not the one provided
+  # by the `pov-wkhtmltopdf` repo.
+  - PHANTOMJS_VERSION="latest"
```
pedrobaeza: No need for this

.travis.yml
```diff
@@ -1,38 +1,69 @@
...
+  # The above line controls the PhantomJS version that is used for JS testing.
+  #   It is not necessary to include this value unless you are altering the default.
+  #   Use `OS` to skip the PhantomJS upgrade & use the system version instead.
+  - WEBSITE_REPO="1"
```
pedrobaeza: and this

pedrobaeza (changes requested):
Isn't better to leave out all the comments?

.travis.yml
StefanRijnhart: Removed website stuff, but I thought keeping the comments will make it easier to determine what has changed when yet another new version is committed.

pedrobaeza: Well, for instructions, the template is there, and I don't see this as a diff comparison, as you usually need to fine-tune specific repos with things, and thus you might remove needed things thinking that the template has changed.

PR #582, opened 2019-08-20
[UPD] Travis.yml

StefanRijnhart:


pedrobaeza (approved):
I still think comments should be stripped out, but this should work.

rafaelbn (approved):
:green_apple:

PR #588, opened 2019-08-29
[11.0][ADD] sale_order_invoicing_queued: Enqueue sales order invoicing

pedrobaeza:
This module allows to enqueue in several jobs the sales orders invoicing process to be executed in paralell on background, that's it's normally done serially and on foreground.

Jobs are split following the same criteria as standard Odoo: grouping by order invoicing address and order currency.

Installation
============

This module depends on *queue_job* module that is hosted on https://github.com/OCA/queue.

Configuration
=============

If you want to see queued jobs, you need "Job Queue / Job Queue Manager" permission in your user.

Usage
=====

* Go to *Sales > Invoicing > Orders to Invoice*.
* Select sales orders whose invoicing you want to enqueue.
* Click on *Action > Invoice Order*.
* Click on "Enqueue Invoicing" button.
* On that moment, jobs will be enqueued on foreground, but the control will be returned to you after that.
* Having the "Job Queue Manager" permissions, you can go to the sales order, and see the tab "Invoicing Jobs". There a list with all the jobs related to that sales order can be found.

Known issues / Roadmap
======================

* This module doesn't take into account any possible modification on grouping criteria, like all the modules that uses `sale_order_action_invoice_create_hook`.

cc @Tecnativa


rafaelbn:
Please @pedrobaeza notify me for testing this in production with customer, we need lots of sale order to test this correctly. Thanks

sale_order_invoicing_queued/models/queue_job.py
```diff
@@ -0,0 +1,17 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+
+from odoo import models
+
+
+class QueueJob(models.Model):
+    _inherit = 'queue.job'
+
+    def cancel_sudo(self):
```
yajo: It seems to me like this should go directly in the queue addon, isn't it? This doesn't seem to be specific for invoices...

sale_order_invoicing_queued/models/queue_job.py
```diff
@@ -0,0 +1,17 @@
...
+        self.sudo().filtered(
+            lambda x: x.state in ['pending', 'enqueued']
+        ).unlink()
+
+    def requeue_sudo(self):
```
yajo: Same here.

sale_order_invoicing_queued/models/sale_order.py
```diff
@@ -0,0 +1,32 @@
+# Copyright 2019 Tecnativa - Pedro M. Baeza
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+try:
+    from odoo.addons.queue_job.job import job
+except ImportError:
```
yajo, line 8: You can add a TODO note to remove this try-except in v12.

sale_order_invoicing_queued/models/sale_order.py
```diff
@@ -0,0 +1,32 @@
...
+    import logging
+    import functools
+
+    logging.getLogger().debug("Can't `import queue_job`.")
+
+    def empty_decorator_factory(*argv, **kwargs):
```
yajo:
```suggestion
    def job(*argv, **kwargs):
```

sale_order_invoicing_queued/models/sale_order.py
```diff
@@ -0,0 +1,32 @@
...
+        return functools.partial
+    job = empty_decorator_factory
```
yajo:
```suggestion
```

sale_order_invoicing_queued/models/sale_order.py
```diff
@@ -0,0 +1,32 @@
...
+
+from odoo import fields, models
+
+try:
+    from odoo.addons.queue_job.job import job
+except ImportError:
+    import logging
+    import functools
+
+    logging.getLogger().debug("Can't `import queue_job`.")
```
yajo:
```suggestion
    logging.getLogger(__name__).warning("Can't `import queue_job`.")
```

sale_order_invoicing_queued/models/sale_order.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def empty_decorator_factory(*argv, **kwargs):
+        return functools.partial
+    job = empty_decorator_factory
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    invoicing_job_ids = fields.Many2many(
```
yajo, line 21: :thinking: By using a m2m, do you mean that there could be several jobs generating the same invoice for the same SO?

sale_order_invoicing_queued/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,6 @@
+This module allows to enqueue in several jobs the sales orders invoicing
+process to be executed in paralell on background, that's it's normally done
```
yajo:
```suggestion
process to be executed in paralell on background, that is usually done
```

sale_order_invoicing_queued/readme/ROADMAP.rst
```diff
@@ -0,0 +1,3 @@
+* This module doesn't take into account any possible modification on grouping
+  criteria, like all the modules that uses
+  `sale_order_action_invoice_create_hook`.
```
yajo:
```suggestion
  ``sale_order_action_invoice_create_hook``.
```

sale_order_invoicing_queued/readme/ROADMAP.rst
```diff
@@ -0,0 +1,3 @@
+* This module doesn't take into account any possible modification on grouping
+  criteria, like all the modules that uses
```
yajo:
```suggestion
  criteria, like all the modules that use
```

sale_order_invoicing_queued/readme/USAGE.rst
```diff
@@ -0,0 +1,9 @@
+#. Go to *Sales > Invoicing > Orders to Invoice*.
+#. Select sales orders whose invoicing you want to enqueue.
+#. Click on *Action > Invoice Order*.
+#. Click on "Enqueue Invoicing" button.
+#. On that moment, jobs will be enqueued on foreground, but the control will
```
yajo, line 5:
```suggestion
#. On that moment, jobs will be enqueued on the background, but the control will
```

sale_order_invoicing_queued/views/queue_job_views.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright 2019 Tecnativa - Pedro M. Baeza
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+<odoo>
+    <record id="view_queue_job_sale_order_invoicing" model="ir.ui.view">
+        <field name="model">queue.job</field>
+        <field name="priority" eval="9999"/>
```
yajo, line 7: I'd remove this if possible.

```suggestion
```

pedrobaeza:
For testing you only need one sales order, but this is thought to be powerful on large scale.

sale_order_invoicing_queued/wizards/sale_advance_payment_inv.py
```diff
@@ -0,0 +1,35 @@
...
+
+class SaleAdvancePaymentInv(models.TransientModel):
+    _inherit = "sale.advance.payment.inv"
+
+    def enqueue_invoices(self):
+        queue_obj = self.env['queue.job']
+        order_obj = self.env['sale.order']
+        context = self.env.context
+        final = (self.advance_payment_method == 'all')
+        if (self.advance_payment_method not in ('delivered', 'all')):
```
yajo:
```suggestion
        if (self.advance_payment_method not in {'delivered', 'all'}):
```

yajo (changes requested):

sale_order_invoicing_queued/models/queue_job.py
pedrobaeza: No, upstream already has `unlink` functionality. This is only a bridge for being executed through a button.

sale_order_invoicing_queued/models/sale_order.py
pedrobaeza: Same answer.

sale_order_invoicing_queued/models/sale_order.py
pedrobaeza: This is working and the standard solution when this is used as decorator, so I prefer to not touch it

sale_order_invoicing_queued/models/sale_order.py
pedrobaeza: Not sure this change will work

sale_order_invoicing_queued/models/sale_order.py
pedrobaeza: It mustn't be warning: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#imports

sale_order_invoicing_queued/models/sale_order.py
pedrobaeza: Not the same invoice, but several invoices. One sales order > several invoices, thus several jobs.

sale_order_invoicing_queued/readme/DESCRIPTION.rst
pedrobaeza: OK about `'s`, but I have chosen normally because the etymological difference about what is "the rule": https://www.google.com/search?client=ubuntu&channel=fs&q=normally+vs+usually&ie=utf-8&oe=utf-8

sale_order_invoicing_queued/readme/USAGE.rst
pedrobaeza: No, they are enqueued on foreground, and executed on background.

sale_order_invoicing_queued/views/queue_job_views.xml
pedrobaeza: Not possible. This view mustn't be used by default. It's on purpose.

pedrobaeza:
Comments attended or answered

sale_order_invoicing_queued/models/queue_job.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+from odoo import models
+
+
+class QueueJob(models.Model):
+    _inherit = 'queue.job'
+
+    def cancel_sudo(self):
+        self.sudo().filtered(
```
yajo: I think you have to use ACLs here that gives unlink access to `queue.job` records linked to a sale order if the user has read access to the sale order.

Otherwise, using sudo and unlink in a public method is a very dangerous operation. You can abuse the XMLRPC to do things you shouldn't be able to do.

sale_order_invoicing_queued/models/queue_job.py
```diff
@@ -0,0 +1,17 @@
...
+            lambda x: x.state in ['pending', 'enqueued']
+        ).unlink()
+
+    def requeue_sudo(self):
+        self.sudo().requeue()
```
yajo: For the same reason, use ACLs here too, instead of sudo.

yajo (changes requested):
Oops, security hole... :speak_no_evil:

sale_order_invoicing_queued/models/sale_order.py
yajo: Same as what? :confused:

sale_order_invoicing_queued/models/sale_order.py
yajo: Yes, it will :wink: (if you do https://github.com/OCA/account-invoicing/pull/588#discussion_r319387683 too).

sale_order_invoicing_queued/models/sale_order.py
yajo:
```suggestion
    logging.getLogger(__name__).debug("Can't `import queue_job`.")
```

sale_order_invoicing_queued/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,6 @@
+This module allows to enqueue in several jobs the sales orders invoicing
+process to be executed in paralell on background, that it's normally done
```
yajo:
```suggestion
process to be executed in paralell on background, which is normally done
```

sale_order_invoicing_queued/views/queue_job_views.xml
yajo: Oh, that's not very obvious... A comment would help then. I didn't know it works like that.

Also do you really need 9999, or is it enough to use something more than the default (16 if I remember correctly)?

yajo (commented):
---

sale_order_invoicing_queued/views/queue_job_views.xml
pedrobaeza: Yes, it is. It's the normal priority field use. I take that number as a very high one for being sure the view is not used as default.

sale_order_invoicing_queued/models/queue_job.py
pedrobaeza: OK, good point, but no need to rework ACL. It's already good suited. Just removing the sudo.

sale_order_invoicing_queued/models/queue_job.py
pedrobaeza: I remove the method at all, as no sudo is required.

pedrobaeza:
@Yajo changes done. Twin PR to review #589

sale_order_invoicing_queued/views/queue_job_views.xml
```diff
@@ -0,0 +1,29 @@
...
+                <field name="func_string"/>
+                <field name="date_done"/>
+                <field name="state"/>
+                <button type="object"
+                        name="requeue"
+                        string="Requeue"
+                        class="oe_highlight"
+                        attrs="{'invisible': [('state', '!=', 'failed')]}"/>
+                 <button type="object"
+                         name="cancel"
```
yajo, line 20:
```suggestion
                         name="unlink"
```

Since this button is invisible unless state is pending or enqueued, actually you're doing the same and this way you can remove the extra method from the model. :blush:

sale_order_invoicing_queued/models/queue_job.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+from odoo import models
+
+
+class QueueJob(models.Model):
+    _inherit = 'queue.job'
+
+    def cancel(self):
+        self.filtered(lambda x: x.state in ['pending', 'enqueued']).unlink()
```
yajo, line 12: IMHO this method can be removed. See comment on view.

sale_order_invoicing_queued/models/sale_order.py
```diff
@@ -0,0 +1,31 @@
+# Copyright 2019 Tecnativa - Pedro M. Baeza
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+try:  # The try can be removed in v12 as not needed
```
yajo, line 6:
```suggestion
try:  # TODO The try can be removed in v12 as not needed
```

yajo (changes requested):
Cool, then I see now a little more improvement. Almost there. :wink:

pedrobaeza:
All already commented in the twin PR. Please check.

yajo (approved):

sale_order_invoicing_queued/models/sale_order.py
```diff
@@ -0,0 +1,31 @@
...
+    def job(*argv, **kwargs):
+        return functools.partial
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    invoicing_job_ids = fields.Many2many(
+        comodel_name='queue.job',
+        column1='invoice_id',
```
sergio-teruel: ¿invoice_id or sale_order_id?

pedrobaeza: Yeah, copy/paste error. Changed.

carlosdauden (approved):

PR #589, opened 2019-08-29
[11.0][ADD] account_invoice_validation_queued: Enqueue account invoice validation

pedrobaeza:
This module allows to enqueue in several jobs the account validation process to be executed in paralell on background, that's it's normally done serially and on foreground.

Each invoice creates a job for performing the validation, but you will only be able to enqueue invoices for the same date.

Installation
============

This module depends on *queue_job* module that is hosted on https://github.com/OCA/queue.

Configuration
=============

If you want to see queued jobs, you need "Job Queue / Job Queue Manager" permission in your user.

Usage
=====

* Go to *Invoicing > Sales > Documents > Customer Invoices* or  *Invoicing > Purchases > Documents > Vendor Bills*.
* Mark at least one check on the left part of one draft invoice line in the list view.
* Click on *Action > Confirm Draft Invoice*.
* On the dialog popup that appears, click on "Enqueue Validation".
* If any of the selected invoices are not in draft state, you will receive an error.
* If any of the invoices is already enqueued, there will be a message saying so and avoiding to perform the process.
* Once enqueued, and having the "Job Queue Manager" permission, you can go to the invoice, and see the tab "Validation Jobs". A list with all the jobs related to that invoice can be found there.

Known issues / Roadmap
======================

* Allow to enqueue for validating invoices of different date.
* There's a chance that if you perform several enqueues of invoice validations from different dates, the order of the validated invoices, and thus, the number given for it, will result disordered.

cc @Tecnativa


account_invoice_validation_queued/models/account_invoice.py
```diff
@@ -0,0 +1,31 @@
+# Copyright 2019 Tecnativa - Pedro M. Baeza
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+try:  # the try/except statements can be removed on v12
```
yajo, line 6:
```suggestion
try:  # TODO the try/except statements can be removed on v12
```

account_invoice_validation_queued/models/queue_job.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+from odoo import models
+
+
+class QueueJob(models.Model):
+    _inherit = 'queue.job'
+
+    def cancel(self):
+        self.filtered(lambda x: x.state in ['pending', 'enqueued']).unlink()
```
yajo, line 12: This can possibly be removed. See comment in view.

account_invoice_validation_queued/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,5 @@
+Jobs are enqueued in the channel ``root.account_invoice_validation_queued``,
+so you must adjust your Odoo configuration according this.
```
yajo: A link to docs on how to configure odoo would be helpful here.

account_invoice_validation_queued/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,6 @@
+This module allows to enqueue in several jobs the account validation process
+to be executed in paralell on background, that's it's normally done
```
yajo:
```suggestion
to be executed in paralell on background, which is normally done
```

account_invoice_validation_queued/views/queue_job_views.xml
```diff
@@ -0,0 +1,28 @@
...
+                <field name="date_created"/>
+                <field name="date_done"/>
+                <field name="state"/>
+                <button type="object"
+                        name="requeue"
+                        string="Requeue"
+                        class="oe_highlight"
+                        attrs="{'invisible': [('state', '!=', 'failed')]}"/>
+                 <button type="object"
+                         name="cancel"
```
yajo, line 19:
```suggestion
                         name="unlink"
```

Since the button is invisible in other states, this makes the model method unnecessary.

account_invoice_validation_queued/wizards/account_invoice_confirm.py
```diff
@@ -0,0 +1,35 @@
...
+                raise exceptions.UserError(_(
+                    "You can't enqueue invoices with different dates."
+                ))
+        for invoice in invoices:
+            if invoice.validation_job_ids.filtered(
+                lambda x: x.state in {'pending', 'enqueued', 'started'}
+            ):
+                raise exceptions.UserError(_(
+                    "There's already an enqueued job for validating the "
+                    "invoice #%s. Please wait until it's finished or remove "
```
yajo, line 30: Not sure if the invoice ID is very helpful for a end user. Maybe the invoice number is better, don't you think?

account_invoice_validation_queued/wizards/account_invoice_confirm_views.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2019 Tecnativa - Pedro M. Baeza
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<odoo>
+    <record id="account_invoice_confirm_view" model="ir.ui.view">
+        <field name="model">account.invoice.confirm</field>
+        <field name="inherit_id" ref="account.account_invoice_confirm_view"/>
```
yajo, line 7: Don't you need a group in this view?

yajo (changes requested):

account_invoice_validation_queued/models/account_invoice.py
pedrobaeza: This is not a TODO, as for v11 this is totally done. Even in v12, this is something harmless, so I'm not going to mark it that way.

account_invoice_validation_queued/models/queue_job.py
pedrobaeza: Yes, but being a public method that can be called from any place, I prefer to filter.

account_invoice_validation_queued/views/queue_job_views.xml
pedrobaeza: Already answer above

account_invoice_validation_queued/wizards/account_invoice_confirm_views.xml
pedrobaeza: Group? This is only adding a button in the header...

account_invoice_validation_queued/wizards/account_invoice_confirm.py
pedrobaeza: They are in draft, so no number yet given. Better this than anything :wink:

account_invoice_validation_queued/wizards/account_invoice_confirm_views.xml
yajo: Yes, but don't you need to have queue permissions to enqueue invoices?

pedrobaeza: The enqueue operation is done with `sudo`, as the permission is not needed for that. If you are an accountant, you want to have invoices validated on parallel without needing to have job permissions.

pedrobaeza:
Changes done

account_invoice_validation_queued/README.rst
```diff
@@ -0,0 +1,122 @@
...
+   the invoice, and see the tab "Validation Jobs". A list with all the jobs
+   related to that invoice can be found there.
+
+Known issues / Roadmap
+======================
+
+* Allow to enqueue for validating invoices of different date.
+* There's a chance that if you perform several enqueues of invoice validations
+  from different dates, the order of the validated invoices, and thus, the
+  number given for it, will result disordered.
```
sergio-teruel, line 81: I think that you must include as know issue that no gap sequence implementation not allowed due to database locks

pedrobaeza: Great point! I have included in CONFIGURE section, as it's something to be configured for the module.

carlosdauden (approved):

sergio-teruel (approved):

account_invoice_validation_queued/models/queue_job.py
```diff
@@ -0,0 +1,12 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+
+from odoo import models
+
+
+class QueueJob(models.Model):
+    _inherit = 'queue.job'
+
+    def cancel(self):
```
lmignon, line 11: This code should never be there... Moreover it's a duplicate of  https://github.com/OCA/account-invoicing/commit/6ac79c298cb07f61f6f1888ef26eca702e4aa284#diff-01026d97e6f71aadeb837a145ddac1caR11

account_invoice_validation_queued/wizards/account_invoice_confirm.py
```diff
@@ -0,0 +1,35 @@
...
+                    "it from the selection."
+                ) % invoice.id)
+            new_delay = invoice.sudo().with_delay().action_invoice_open_job()
```
lmignon, line 33: we lost the traceability of the user at the origin of the invoice confirmation.

account_invoice_validation_queued/views/queue_job_views.xml
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright 2019 Tecnativa - Pedro M. Baeza
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+<odoo>
+    <record id="view_queue_job_account_invoice_validation" model="ir.ui.view">
```
lmignon, line 5: This view is a duplicate of https://github.com/OCA/account-invoicing/commit/6ac79c298cb07f61f6f1888ef26eca702e4aa284#diff-9478f8f14832f66efde975e04566b2d4R4

account_invoice_validation_queued/wizards/account_invoice_confirm.py
```diff
@@ -0,0 +1,35 @@
...
+            return self.invoice_confirm()
+        # Check if all invoices are for the same date
+        inv_date = invoices[:1].date_invoice
+        for invoice in invoices:
+            if invoice.date_invoice != inv_date:
+                raise exceptions.UserError(_(
+                    "You can't enqueue invoices with different dates."
+                ))
+        for invoice in invoices:
+            if invoice.validation_job_ids.filtered(
```
lmignon, line 25: @pedrobaeza Queue job allows you to define a job_identity key to avoid the creation of duplicates. No need of specific code here... https://github.com/OCA/queue/commit/2e86357819315e82bba12a39149ed82614f67d8b

lmignon (commented):
I'm too late but I think that you should take care of my comments...

account_invoice_validation_queued/models/queue_job.py
pedrobaeza: Yes, because you can install each module and need the same feature, but there's no problem in having it "duplicated", and this is serving an specific UI purpose (to be able to cancel the job), so no better way to do it that I know of.

account_invoice_validation_queued/wizards/account_invoice_confirm.py
pedrobaeza: Can I put in the user to execute the work somehow?

account_invoice_validation_queued/views/queue_job_views.xml
pedrobaeza: It's not the same. That view is primary and it's intended to show the specific fields for `queue.job` that are interesting to be shown.

account_invoice_validation_queued/wizards/account_invoice_confirm.py
pedrobaeza: But this is not unique... You can validate the invoice several times (you cancel it and validate it again).

account_invoice_validation_queued/views/queue_job_views.xml
lmignon: Does this mean that if a module has properly extended the tree view for queue_job, its changes will never appear again once this module is installed?

account_invoice_validation_queued/wizards/account_invoice_confirm.py
lmignon: image

account_invoice_validation_queued/views/queue_job_views.xml
pedrobaeza: Not really. This view is only intended for showing the jobs attached to an invoice, with the relevant fields in it for an "accountant". List in queue section will be the same as standard, with the same added fields.

account_invoice_validation_queued/wizards/account_invoice_confirm.py
pedrobaeza: OK, good to know. I will change this in a new PR when we clarify all the comments.

guewen: Also, jobs must be idempotent: the first thing they should do is verify the state and return if it doesn't match its expectation (e.g. return if the state is already 'open').

By removing this custom code, you will be able to remove the override of `action_cancel()` that you added on `queue.job`. I agree with @lmignon that it shouldn't be there. You shouldn't change the behavior of the `queue.job` model here.

pedrobaeza: The logic for having that method is that `unlink` is already public, so this is not a backdoor, but it's a safe method for not needing to worry about which jobs can be removed in this context, but if it's so problematic, I'll remove it and call `unlink` directly. That's not a big problem at all.

@guewen what about changing the user that will execute the job. Is it possible from the `with_delay` call?

pedrobaeza: About the idempotent you mentioned, @guewen, I have double checked again, and the method that is call from the job is already handling the cases where the state is different from `open`, and that's why I prefer to leave them that way, so that the job fails in that case with the exception raised from upstream.

guewen: >   what about changing the user that will execute the job. Is it possible from the with_delay call?

The job is executed as the user delaying the job. However, the user does not need access rights on `queue.job` because `with_delay()` creates the record using `sudo()`.



>  I have double checked again, and the method that is call from the job is already handling the cases where the state is different from `open`, and that's why I prefer to leave them that way, so that the job fails in that case with the exception raised from upstream.

If the state is already good, the job has nothing to do: it should be `done`.

pedrobaeza: > The job is executed as the user delaying the job. However, the user does not need access rights on queue.job because with_delay() creates the record using sudo().

Great then, changing the code for removing `sudo`.

> If the state is already good, the job has nothing to do: it should be done.

Well, Odoo thinks the contrary on this, as it raises the exception if you are trying to validate again one validated invoice. What do you think about this, @lmignon ?

guewen: > Well, Odoo thinks the contrary on this, as it raises the exception if you are trying to validate again one validated invoice

My point is not about odoo's code but about jobs. Jobs are not user-oriented, they are background, low-level representation of "a method running a background" (think celery). Users should not have to deal with them and have to care about them (beside a few advanced users tracking issues). A good functioning system using jobs should have *zero* jobs failing (in practice it's inevitable to have some but that should be the goal). The thing is you modified the UI to show the job records to users, that's why you have different expectations, but IMO that's a bad idea. The job should only validate the invoice, if it's already validated then it should return and be done; if there is an error during validation, the job should log a message in the invoice's thread and be done. But "normal" users should not have to deal with failed jobs.

pedrobaeza: We have different vision about this and what the user should know and interact with, but in this case, as the scope is very limited, I'm adding the condition about the state in https://github.com/OCA/account-invoicing/compare/fc5af5be55fcb243fa7ef1500be236415c264959..85444ce0e5d09adcf335059db1f9a78013f50eb4 for #593 for not having the job failing.

PR #593, opened 2019-09-09
[IMP] *_queued: Remove `cancel` method + use same user

pedrobaeza:
Thanks to @guewen and @lmignon for their comments.

I'm missing an argument for the method `with_delay` for indicating the user who is going to execute the job (if different from current one).

I can't use `identity_key` in sales order invoicing as you can instruct a sales order to be invoiced in different groupings, having different identity key.

cc @Tecnativa


lmignon (approved):
Thank you @pedrobaeza for the changes. This code looks better. (code review only)

pedrobaeza:
Fixed tests + add comments from @guewen from the other place (no need to make all the sudo stuff).

@guewen please continue commenting here any extra thing.

account_invoice_validation_queued/models/account_invoice.py
```diff
@@ -26,6 +26,9 @@ class AccountInvoice(models.Model):
         copy=False,
     )
 
+    def _get_validation_queued_identity_key(self, job):
+        return self.id
```
guewen: The identity key should at least include the model and method names so 2 kind of jobs won't conflict, you can take example on https://github.com/OCA/queue/blob/fdcde048348694df37ce2f63686467cf762aeda5/queue_job/job.py#L102-L144 (or directly use it, I think you want the `identity_exact` here)

pedrobaeza: Thanks for the tip. I have used `identity_exact`.

account_invoice_validation_queued/views/queue_job_views.xml
```diff
@@ -16,7 +16,7 @@
                         class="oe_highlight"
                         attrs="{'invisible': [('state', '!=', 'failed')]}"/>
                  <button type="object"
-                         name="cancel"
+                         name="unlink"
```
yajo, line 19: You can apply this if you want the same functionality as before, with the `cancel()` method.

```suggestion
                         name="unlink"
                         attrs="{'invisible': [('state', 'not in', ('pending', 'enqueued'))]}"
```

sale_order_invoicing_queued/views/queue_job_views.xml
```diff
@@ -17,7 +17,7 @@
                         class="oe_highlight"
                         attrs="{'invisible': [('state', '!=', 'failed')]}"/>
                  <button type="object"
-                         name="cancel"
+                         name="unlink"
```
yajo, line 20: Here too.

```suggestion
                         name="unlink"
                         attrs="{'invisible': [('state', 'not in', ('pending', 'enqueued'))]}"
```

yajo (approved):
Indeed it looks much better. :blush: 

A couple of suggestions.

guewen (approved):
Great, thanks!

account_invoice_validation_queued/views/queue_job_views.xml
pedrobaeza: It's already put 3 lines below

sale_order_invoicing_queued/views/queue_job_views.xml
pedrobaeza: It's already put 3 lines below

pedrobaeza:
OK, proceeding to merge then. Thanks all for your suggestions.

PR #598, opened 2019-09-23
[12.0][ADD] account_invoice_repair_link

AdriaGForgeFlow:
Adds module account_invoice_repair_link.

This module adds a link between invoices and repair orders. Invoices are generated from
the repair orders. With this module you can navigate back to the repair orders that
generated a specific invoice.

This module is useful if you want to add information from the repair in the invoice report.

cc ~ @Eficent


account_invoice_repair_link/migrations/12.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2019 Eficent Business and IT Consulting Services S.L.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+
+def migrate(env, version):
```
JordiBForgeFlow: This should go in an init hook, not a migration, otherwise it will never be triggered

JordiBForgeFlow: Well, I don't think that you need this anyway. Because the invoice id is already in the repair order. The one2many does not store data in the model where it is used, but in the related model

account_invoice_repair_link/models/repair.py
```diff
@@ -0,0 +1,16 @@
...
+from odoo import api, models
+
+
+class Repair(models.Model):
+    _inherit = 'repair.order'
+
+    @api.multi
+    def action_invoice_create(self, group=False):
+        Invoice = self.env['account.invoice']
+        res = super().action_invoice_create(group)
```
JordiBForgeFlow: Same as above. You only need to store data in the field invoice_id of the repair.order.

account_invoice_repair_link/migrations/12.0.1.0.0/post-migration.py
AdriaGForgeFlow: You're right, fixed!

account_invoice_repair_link/models/account_invoice.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2019 Eficent Business and IT Consulting Services S.L.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo import models, fields, _
```
quannguyen1560: Import order
https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#imports

quannguyen1560 (commented):
please have a look about my comment.

JordiBForgeFlow (approved):
:+1: functional and technical review.

account_invoice_repair_link/views/account_invoice_views.xml
```diff
@@ -0,0 +1,25 @@
...
+                            type="object"
+                            attrs="{'invisible': [('repair_ids', '=', [])]}"
+                            groups="stock.group_stock_user">
+                    </button>
+                </xpath>
+            </field>
+        </record>
+
+    </data>
+</odoo>
```
HviorForgeFlow, line 25: Nit: Add a blank line at the end

HviorForgeFlow (approved):
Functional review
LGTM :+1:

AaronHForgeFlow (approved):
Please squash :+1:

AdriaGForgeFlow:
> Please squash

Done :smile:

PR #611, opened 2019-10-11
[11.0][MIG] product_supplierinfo_for_customer_invoice

MiquelRForgeFlow:
Standard migration.


product_supplierinfo_for_customer_invoice/models/account_invoice.py
```diff
@@ -0,0 +1,25 @@
...
+
+class AccountInvoiceLine(models.Model):
+    _inherit = 'account.invoice.line'
+
+    @api.multi
+    def _get_product_customer_code(self):
+        product_supplierinfo_obj = self.env['product.supplierinfo']
+        for line in self.filtered(lambda il: il.product_id.supplier_ids):
+            product = line.product_id
+            code_id = product_supplierinfo_obj.search([
```
AaronHForgeFlow, line 16: Does not look good to me. I created a customer invoice and the customer code in the invoice was totally incorrect.

It should filter by the customer, and also the product.supplierinfo must be type customer

AaronHForgeFlow (changes requested):

product_supplierinfo_for_customer_invoice/models/account_invoice.py
```diff
@@ -0,0 +1,27 @@
...
+                ('type', '=', 'customer'),
```
AaronHForgeFlow: `supplierinfo_type`

AaronHForgeFlow (changes requested):

AaronHForgeFlow (approved):
LGTM now

merge_small

nikul-serpentcs (approved):
Perfecto :+1:

PR #624, opened 2019-10-31
[IMP][11.0] purchase_stock_picking_return_invoicing: ease multi refund

chienandalu:
- Show "Add purchase order" field in supplier refunds to allow refunding
multiple orders.
- Improve "Refunds" stat button to fit 12.0 improvements like:
  - Enter draft mode as soon as the stat button is clicked.
  - [FIX] When no refund was made yet, the whole customers refunds would
appear.

cc @Tecnativa TT18623

### DIFF
--- a/purchase_stock_picking_return_invoicing/views/account_invoice_view.xml
+++ b/purchase_stock_picking_return_invoicing/views/account_invoice_view.xml
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+    <!-- Add purchase orders to refund -->
+    <record id="view_invoice_supplier_purchase_form" model="ir.ui.view">
+        <field name="model">account.invoice</field>
+        <field name="inherit_id" ref="purchase.view_invoice_supplier_purchase_form"/>
+        <field name="arch" type="xml">
+            <field name="reference" position="after">
+                <field name="purchase_id" attrs="{'readonly': [('state','not in',['draft'])], 'invisible': ['|', ('state', '=', 'purchase'), ('type', '=', 'in_invoice')]}" class="oe_edit_only" options="{'no_create': True}" context="{'show_total_amount': True}"/>
+            </field>
+        </field>
+    </record>
+
+</odoo>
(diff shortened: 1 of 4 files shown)


purchase_stock_picking_return_invoicing/views/account_invoice_view.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+    <!-- Add purchase orders to refund -->
+    <record id="view_invoice_supplier_purchase_form" model="ir.ui.view">
+        <field name="model">account.invoice</field>
+        <field name="inherit_id" ref="purchase.view_invoice_supplier_purchase_form"/>
+        <field name="arch" type="xml">
+            <field name="reference" position="after">
+                <field name="purchase_id" attrs="{'readonly': [('state','not in',['draft'])], 'invisible': ['|', ('state', '=', 'purchase'), ('type', '=', 'in_invoice')]}" class="oe_edit_only" options="{'no_create': True}" context="{'show_total_amount': True}"/>
```
pedrobaeza, line 10: Not sure about this, as the problem is that the amount should appear reversed and there's no sense on showing all the purchases here.

chienandalu: There's a constraint if the total value of the invoice is negative, so you can't add positive balance to a refund as you can't add negative balance to a normal invoice. And there's no limitation on wich orders are available or not. I just chose to show up the same field in the case of refunds to ease the creation of refunds in the same way the normal invoices do. It's on the customer's side to choose what orders are the right ones.

pedrobaeza: OK, fine

pedrobaeza (approved):

sergio-teruel:
Tested in runbot and ok!! Great @chienandalu

sergio-teruel (approved):

pedrobaeza:
Please put it on 12.0 as well.

PR #625, opened 2019-10-31
[12.0][MIG] purchase_batch_invoicing: Migration to 12.0

ernestotejeda:
Cc @Tecnativa TT20176


purchase_batch_invoicing/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 {
     "name": "Purchase Batch Invoicing",
     "summary": "Make invoices for all ready purchase orders",
-    "version": "11.0.1.0.1",
+    "version": "12.0.1.0.1",
```
pedrobaeza: 12.0.1.0.0

purchase_batch_invoicing/tests/test_invoicing.py
```diff
@@ -66,7 +67,7 @@ def setUpClass(cls):
                     "product_qty": 1,
                     "price_unit": 100,
                     "product_uom": cls.uom1.id,
-                    "date_planned": "2016-05-12",
+                    "date_planned": date(2016, 5, 12),
```
pedrobaeza: AFAIK, you don't need to change this

purchase_batch_invoicing/views/purchase_order_view.xml
```diff
@@ -24,8 +24,8 @@
         <field name="model">purchase.order</field>
         <field name="inherit_id" ref="purchase.purchase_order_tree"/>
         <field name="arch" type="xml">
-            <xpath expr="//field[@name='state']" position="after">
-                <field name="invoice_status"/>
+            <xpath expr="//field[@name='invoice_status']" position="attributes">
+                <attribute name="attrs" ></attribute>
```
pedrobaeza: `<attribute name="attrs"/>`

pedrobaeza (approved):

purchase_batch_invoicing/wizards/purchase_batch_invoicing.py
```diff
@@ -0,0 +1,129 @@
...
+        """Generate invoices for all selected purchase orders.
+
+        :return dict:
+            Window action to see the generated invoices.
+        """
+        invoices = self.env["account.invoice"]
+        for pogroup in self.grouped_purchase_orders():
+            invoice = invoices.create({
+                "partner_id": pogroup.mapped("partner_id").id,
+                "type": "in_invoice",
```
chienandalu: It'd be nice to have a method returning these vals so it can be overloaded in another module and ovoid incompatibilities, like `purchase_stock_picking_return_invoicing`

purchase_batch_invoicing/tests/test_invoicing.py
ernestotejeda: I was wondering if it is necessary, however I have changed it because I think that if in the future the Odoo date/datetime format changes, that value will not have to be changed. What do you think ?

pedrobaeza: No, this doesn't change any way. The python datetime object is introduced in this version, but you are always able to pass the previous string format and it's converted when needed.

ernestotejeda:
Changes done

purchase_batch_invoicing/wizards/purchase_batch_invoicing.py
```diff
@@ -60,6 +61,9 @@ def _purchase_order_domain(self, ids=None):
             domain[1] = ("id", "in", pos.ids)
         return domain
 
+    def _prepare_batch_invoice_vals(self, partner):
```
chienandalu, line 64: You can add a docstring documenting the purpose of it:

"""Allow to override the invoice defaults by a third module. i.e.: set invoice type to in_refund"""

chienandalu (approved):

PR #629, opened 2019-11-06
[FEAT] Choose the purchase order on the vendor bill

aisopuro:
Sometimes, due to e.g. integration, users may be given pre-made refunds
instead of being able to generate the refund from the purchase order. In
these cases it is useful to be able to select the purchase order directly
onto the bill, similarly to how it can be selected onto vendor bills.


pedrobaeza:
Please see #624 instead

PR #633, opened 2019-11-14
[REF] account_invoice_change_currency: Fixed get rate at first currency convertion

luistorresm:
Before this commit, was returned the same currency that the invoice on
the onchange context, now is returned the currency recorded at the
messages with tracking values.

Cherry-pick from https://github.com/OCA/account-invoicing/pull/511/


moylop260:
Merging since that it is just a migration using cherry-pick

PR #642, opened 2019-12-04
[MIG][12.0] stock_picking_return_refund_option: Migrationto v12.0

sergio-teruel:
cc @Tecnativa TT20141
@Tardo @chienandalu Can yopu review?


pedrobaeza (approved):

stock_picking_return_refund_option/models/stock_picking.py
```diff
@@ -0,0 +1,83 @@
...
+            Each line to original value selected in return wizard by user.
+        """
+        for picking in self:
+            picking._update_stock_moves()
+            picking.set_delivered_qty()
+            picking.set_received_qty()
+
+    def _compute_is_return(self):
+        for picking in self:
+            if any(x.origin_returned_move_id for x in picking.move_lines):
```
Tardo, line 48:
```suggestion
            if any(picking.move_lines.mapped('origin_returned_move_id')):
```

stock_picking_return_refund_option/readme/USAGE.rst
```diff
@@ -0,0 +1,48 @@
...
+#. Go to *Sales > Sales Orders > Create*.
+#. Choose a customer and add a product whose *Invoicing Policy* is *Delivered
+   quantities*, and input some quantity to sell.
+#. Confirm the sale.
+#. Go to *Delivery > Validate > Apply*.
+#. Return to the sale order.
+#. Press *Create Invoice > Invoiceable lines > Create and View Invoices*.
+#. The created invoice's amount is the same you sold.
+#. Return to the sale order.
+#. Go to *Delivery > Reverse*.
```
Tardo:
```suggestion
#. Go to *Delivery > Return*.
```

Tardo (approved):
:+1: Tested on runbot... Only non-blocking suggestions.

chienandalu (approved):

stock_picking_return_refund_option/models/stock_picking.py
sergio-teruel: @Tardo I have done this for performance, with mapped method all stock move lines will be readed, with for method if the first record has field origin_returned_move_id to true, the other records will not be read...

sergio-teruel:
Changes Done!!

stock_picking_return_refund_option/models/stock_picking.py
pedrobaeza: When taking that performance design decisions, it's better to put a comment for avoiding future doubts about it.

sergio-teruel: Ok.. yes!!

PR #644, opened 2019-12-06
[11.0] [IMP] account_invoice_fiscal_position_update

feketemihai:
Add mapping also for the invoice account (receivable or payable)
Take invoice line account from normal method (done also in v12.0) to fix anglo saxon accounts mapping.

### DIFF
--- a/account_invoice_fiscal_position_update/models/account_invoice.py
+++ b/account_invoice_fiscal_position_update/models/account_invoice.py
@@ -29,26 +29,20 @@ class AccountInvoice(models.Model):
         inv_type = self.type
         for line in self.invoice_line_ids:
             if line.product_id:
+                account = line.get_invoice_line_account(
+                    inv_type, line.product_id, fp, self.company_id
+                )
                 product = line.with_context(force_company=self.company_id.id).\
                     product_id
                 if inv_type in ('out_invoice', 'out_refund'):
-                    account = (
-                        product.property_account_income_id or
-                        product.categ_id.property_account_income_categ_id)
-                    # M2M fields don't have an option 'company_dependent=True'
-                    # so we need per-company post-filtering
                     taxes = product.taxes_id.filtered(
                         lambda tax: tax.company_id == self.company_id)
                 else:
-                    account = (
-                        product.property_account_expense_id or
-                        product.categ_id.property_account_expense_categ_id)
                     taxes = product.supplier_taxes_id.filtered(
                         lambda tax: tax.company_id == self.company_id)
                 taxes = taxes or account.tax_ids.filtered(
                     lambda tax: tax.company_id == self.company_id)
                 if fp:
-                    account = fp.map_account(account)
                     taxes = fp.map_tax(taxes)
 
                 line.invoice_line_tax_ids = [(6, 0, taxes.ids)]


legalsylvain (changes requested):
Hi @feketemihai. Thanks for this PR.
I think that you introduce a mapping that is not present in Odoo core, but maybe I'm wrong.

could you mention the mapping you try to reproduce ? 

thanks.

feketemihai:
@legalsylvain Yes, we have some situations where we have to overwrite the invoice line account (even if it's coming from product expense/income or stock accounts) with a different account defined on certain configuration (since we use some verticals for agriculture/forestry), but this will not change the "default" Odoo behaviour is working, it will just improve the code in this cases, where somebody overwrites the account. And in default Odoo behaviour the account is reset in the mapping to the one get from the product.

legalsylvain:
Sorry if I'm not clear. You are introducing mapping for invoice account. 
My question is, is it the default feature of Odoo ? And if yes, where ? 
If not, I think that you should not introduce that feature in this module.

regards.

feketemihai:
@legalsylvain Yes, in Odoo default code, the invoice account is not mapped everywhere in the code, only invoice line accounts and taxes, even it should be, because the payable and receivable could change also.

legalsylvain:
So, my position is that introducing that feature is not the good place here, and should be designed in another module.

what is your PoV @OCA/accounting-maintainers 
regards.

pedrobaeza:
This should replicate what Odoo does in core. If that mapping is in core, OK that changing the fiscal position triggers the same changes, but if not, this is not the module.

feketemihai:
@legalsylvain @pedrobaeza Can i leave at least the invoice line overwritten account mapping?

pedrobaeza:
But isn't that what was before?

account_invoice_fiscal_position_update/models/account_invoice.py
```diff
@@ -27,28 +27,30 @@ def fiscal_position_change(self):
...
+        if fp:
+            self.account_id = fp.map_account(self.account_id)
         for line in self.invoice_line_ids:
             if line.product_id:
+                account = line.get_invoice_line_account(
+                    inv_type, line.product_id, fp, self.company_id
+                )
+                # Use mapping of line account in case it was overwritten
+                # the product account
+                line_account = fp.map_account(line.account_id)
```
feketemihai: @pedrobaeza I mean this one, in case we overwrite the invoice line account with a different account (not the one taken from product/category), the normal mapping is reverting it, mapping the account from product/category.

pedrobaeza: No, that would be incorrect. Imagine you have product with account A. Previous fiscal position maps A to B, so B is the one before switching to another fiscal position. Imagine that one maps A to C, so as is, the final account is C, which is correct. With this code, it will try to map B, so leaving B, and that's incorrect.

feketemihai: Revert it. Thanks

feketemihai:
@pedrobaeza I created some time ago in issue about the anglo saxon accounting mapping, this will fix at least that one, since previous implementation took income/expense account. And this already in v12.0. https://github.com/OCA/account-invoicing/commit/be7e0a573b399e0a718c124b48f7429ec1fec2f0

pedrobaeza (approved):

legalsylvain (approved):
make sense.
thanks.

PR #645, opened 2019-12-12
[10.0][ADD] account_invoice_positive_total

acsonefho:
Module used to check if the total of OUT invoice/refund is positive. If it's not the case, an exception is raised. It doesn't check IN invoice/refund


account_invoice_positive_total/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+# -*- coding: utf-8 -*-
+# Copyright 2019 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    'name': "Account invoice positive total",
+    'summary': "Invoices/Refunds can be validated if the total is positive.",
```
Cedric-Pigeon: only if...

Cedric-Pigeon (changes requested):

account_invoice_positive_total/tests/test_account_invoice.py
```diff
@@ -0,0 +1,100 @@
+# -*- coding: utf-8 -*-
+# Copyright 2019 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo.tests.common import TransactionCase
+from odoo import exceptions
+
+
+class TestAccountInvoice(TransactionCase):
```
baimont: Why use TransactionCase and not SavepointCase?
Wouldn't you have faster tests with SavepointCase?

baimont (approved):

PR #647, opened 2019-12-16
[13.0][MIG] account_invoice_force_number

kongrattapong:
Ready to review.


kittiu:
@kongrattapong 
I checked from the v12 of this module, it simply open the "move_name" field (not create the new field).
I think it is not right to migrate by adding a new field "move_name".

Things are quite different between v12 and v13. There is no longer model "account.invoice", may be we should not migrate this module at all.
WDYT? @pedrobaeza

pedrobaeza:
I have a similar case in Spain with `l10n_es_account_invoice_sequence`, that isolates the invoice number from the move number, but I haven't analyzed yet how to implement it. I can tell you back when done.

patel1008b:
@kongrattapong @pedrobaeza @kittiu

kittiu:
Any further comment on this ? @bhaveshselarka

patel1008b:
Hi @kittiu, might be I missed information, often required this feature for import, I am using currently in v12.  I need to have this feature in v13 in one of the recent projects. What do you think about this PR?
will it merge very soon?

Thanks.

kittiu (approved):
@bhaveshselarka @pedrobaeza 
IMHO, I think the use case for import is also justified to have this new field. WDYT?
Code and Functional reviewed :+1:

kittiu:
We have been using also, in case data migration.
1 more approve and we can merge.

ps-tubtim:
@kongrattapong Please update pre-commit before merge.

ps-tubtim (approved):

PR #650, opened 2019-12-17
[13.0][MIG] account_invoice_fixed_discount

kongrattapong:
Ready to review


AdriaGForgeFlow:
Is this still WIP? @kongrattapong

kongrattapong:
> Is this still WIP? @kongrattapong

@AdriaGForgeFlow I forgot this PR sorry, I will make this pr greenify.

AdriaGForgeFlow (changes requested):
Tested locally. When you add a fixed discount the tax amount is not correctly updated:

image


Tax amount equals $48 = 15% of $320 but it should be $45 = 15% of $300 right?

kongrattapong:
> Tested locally. When you add a fixed discount the tax amount is not correctly updated:
> 
> image
> 
> Tax amount equals $48 = 15% of $320 but it should be $45 = 15% of $300 right?

Thank you for review, Fixed it

Selection_039

account_invoice_fixed_discount/models/account_move.py
```diff
@@ -5,6 +5,23 @@
...
+    _inherit = "account.move"
+
+    def _recompute_tax_lines(self, recompute_tax_base_amount=False):
+        vals = {}
+        for line in self.invoice_line_ids.filtered("discount_fixed"):
+            vals[line] = {"price_unit": line.price_unit}
+            price_unit = line.price_unit - line.discount_fixed
+            line.update({"price_unit": price_unit})
+        res = super(AccountMove, self)._recompute_tax_lines(
+            recompute_tax_base_amount=False
```
AdriaGForgeFlow: Super function does not have this parameter as argument, thus the error. You need to remove it from here and maybe from the function declaration as well.

AdriaGForgeFlow (changes requested):
After the changes I got this error every time I try to add or edit a line.

image

Find below a comment where the error occurs:

account_invoice_fixed_discount/models/account_move.py
kongrattapong: I Think you need to fetch code from V13 because it's need.
Selection_040

AdriaGForgeFlow: You're right, I didn't have my Odoo code up to date. My bad.

AdriaGForgeFlow (approved):
Functional review :+1:

account_invoice_fixed_discount/models/account_move.py
kongrattapong: It's fine, Thank for your review. XD

AaronHForgeFlow (approved):
Tested on local LGTM :+1:

PR #653, opened 2020-01-15
FIX:	- Triple discount recursion

OmniaGit:
Triple discount recursion


kittiu:
@OmniaGit thanks for your contribution.
I am not sure how to test it. Would you guide me the step to test recursion please?

OmniaGit:
Dear @kittiu,
the problem came from an elentronic invoice import.
the compere 
`if 0 != float_compare(self.price_subtotal, expected, precision_digits=prec):`
fail each time and go into recursion.
It was few weeks ago so we are not able to give you the value for witch the function fail

regards,
Matteo

kittiu:
@OmniaGit I really unsure how to test it, I am a bit worry. Any hint?
But in fact, the existing computation is quite complex.
I did change the logic in v13 like this.
https://github.com/OCA/account-invoicing/pull/661/files#diff-0f72ae7ef2a02142acdb4018e8bc8246
What do you think?

PR #654, opened 2020-01-16
[11.0][FIX] sale_order_invoicing_grouping_criteria: An user without invoicin…

eantones:
…g permissions cannot access to the partners

Accessing to partner form or tree views leads to an error if your user does not belong to any of the invoicing groups.

@pedrobaeza 

Thanks!

### DIFF
--- a/sale_order_invoicing_grouping_criteria/readme/CONTRIBUTORS.rst
+++ b/sale_order_invoicing_grouping_criteria/readme/CONTRIBUTORS.rst
@@ -1,3 +1,7 @@
 * `Tecnativa <https://www.tecnativa.com>`__:
 
   * Pedro M. Baeza
+
+* `NuoBiT Solutions, S.L. <https://www.nuobit.com>`__:
+
+  * Eric Antones <eantones@nuobit.com>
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Why changing the inheritance? Just put the required group in the view.

eantones:
Yes, I thought that solution too but I'd rather to leave that job to the child view and inherit that behavior too.

In my humble opinion my approach is better but if you don't accept that I can do what you said.

Please, let me know.

Thank you very much

pedrobaeza:
Well, my criteria for defining the inheritance is that the used selector to which view it depends. The security involved may change according other criteria. What about if the groups of the other view is changed in a custom module? You should control that in your own module.

eantones:
Yes but the current module affects the invoicing workflow so the natural view to inherit is the one related to the invoice module not the one related to the base module.

pedrobaeza:
I don't agree and I have told you the general criteria that I consider the decision factor for putting the inheritance, but if you want to keep that approach, please put the group also in this view for the other side effect I commented.

eantones:
Ok, no problem, here's the other approach. Thank you

pedrobaeza (approved):
OK, thanks for pleasing me :)

eantones:
jajajajajajja no, please thank you ;-)

pedrobaeza:
Sorry for not remembering this one when merging #665.

PR #658, opened 2020-01-22
[11.0][ADD]account_invoice_tier_validation

AaronHForgeFlow:
Invoice validation using tier definitions.

cc @LoisRForgeFlow


account_invoice_tier_validation/models/account_invoice.py
```diff
@@ -0,0 +1,17 @@
...
+class AccountInvoice(models.Model):
+    _name = "account.invoice"
+    _inherit = ["account.invoice", "tier.validation"]
+    _state_from = ["draft"]
+    _state_to = ["open", "paid"]
+
+    @api.model
+    def _get_under_validation_exceptions(self):
+        res = super(AccountInvoice, self)._get_under_validation_exceptions()
+        res.append("route_id")
```
LoisRForgeFlow, line 16: this is probably not needed, right?

account_invoice_tier_validation/data/account_invoice_tier_definition.xml
```diff
@@ -0,0 +1,12 @@
...
+<!-- Copyright 2020 ForgeFlow S.L.
+     License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl). -->
+<odoo noupdate="1">
+    <record id="account_invoice_default_tier_definition" model="tier.definition">
+        <field name="model_id" ref="model_account_invoice"/>
+        <field name="definition_type">domain</field>
+        <field name="definition_domain"/>
+        <field name="review_type">group</field>
+        <field name="reviewer_group_id" ref='account.group_account_manager'/>
+    </record>
```
LoisRForgeFlow, line 11: I'm not sure about this, maybe it should not change default behavior by default, this is probably more suitable as demo data, what do you think?

LoisRForgeFlow (commented):
Why there is no runbot? :cry:

account_invoice_tier_validation/data/account_invoice_tier_definition.xml
AaronHForgeFlow: yes, agree

AaronHForgeFlow:
Will open a new one, looking for runbot

account_invoice_tier_validation/models/account_invoice.py
AaronHForgeFlow: nope

PR #660, opened 2020-01-25
[12.0] [MIG] account group invoice line

saxomanu:


kittiu:
@saxomanu thank you for your contribution. I have some initial comments to this (also some question).

First small point, but I would think the name of this PR should be
`[12.0][MIG] account_group_invoice_line`, this should help for readability.

* I see this jump from 10.0 -> 12.0. Would you consider doing for 11.0 first and then 11.0 -> 12.0?
* Please consider use readme folder, then you don't need the README file (bot will generate it). The complete sample https://github.com/OCA/mis-builder/tree/10.0/mis_builder/readme
* It is good time to add some tests.

Now the questions (I am wondering myself, guide me if you know the answer :)
* Why would one want to remove the product from move line?
* Even so, this module is not really the merge of lines. Just to set null to product_id...

FFernandez-PlanetaTIC (approved):

flotho (approved):
LGTM

flotho:
@kittiu , actually accoun.invoice.line allow to have no product in it. Some accountant prefer to deal only with accounts.
Additionnaly, this method is called in action_move_create and will help to group it

PR #661, opened 2020-01-28
[13.0][MIG] account_invoice_triple_discount

kittiu:
Standard migration

Test error from other modules, awaiting for following merge,
- [X] https://github.com/OCA/account-invoicing/pull/663


kittiu:
@bealdav @pedrobaeza @Nikul-Chaudhary 

Since v13 is a big change in account invoice, so I have to change the calc logic quite a bit. But I tested it have the same result. I think it is ok, but you might want to look?

Due to v13 also, I do have difficulty writing test. I have the error here,
https://travis-ci.org/OCA/account-invoicing/jobs/643164653#L1019

Because when we change the discount, on screen onchange do not trigger write immediately, and so it works fine.
But on test script, when I assign the new discount, it fire the write() method immediatelly and so the error occur as you see.

Are there anyway I to imitate onchange in test script properly? (I try with Form, it also not work)

ps-tubtim:
@kittiu I think I found some bug.

Process:
1. Create Customer Invoice
2. Add a Invoice line by
      - Product = Product 1
      - Quantity = 1
      - Price = 100
      - Discount 1 (%) = 10
      - Discount 2 (%) = 10
3. System Compute "Subtotal" = 81
4. Click Save button
5. Subtotal = 81 as before but Discount 1, Discount 2 = 0 

discount

Selection_018

kittiu:
@ps-tubtim thanks for reporting. This is interesting, I found resolution here,
https://www.odoo.com/forum/help-1/question/adding-field-to-invoice-line-ids-in-account-move-odoo-13-165052
Basically, we need to add the discount field in journal entry tab also, even it is not used. Otherwise, on create line, it won't saved.

kittiu:
Travis failed coz of black. So I did the rebase.

account_invoice_triple_discount/models/account_move.py
```diff
@@ -0,0 +1,20 @@
...
+
+class AccountMoveLine(models.Model):
+    _inherit = "account.move.line"
+
+    discount1 = fields.Float("Discount 1 (%)", digits="Discount")
+    discount2 = fields.Float("Discount 2 (%)", digits="Discount")
+    discount3 = fields.Float("Discount 3 (%)", digits="Discount")
+
+    @api.onchange("discount1", "discount2", "discount3")
+    def _onchange_triple_discount(self):
```
legalsylvain: should not be a compute function ? otherwise, a self.env["account.move.line"].create({"discount1": 10, "discount2": 20}) will not raise the compute of discount. Don't you think ?

benwillig:
Could you fix travis please? Then we can see the results of the unit tests.

EDIT: tests fails when started them locally

benwillig:
Hi @kittiu, any news about this?

kittiu:
> Hi @kittiu, any news about this?

Sorry, I am back now. :) Will continue.

account_invoice_triple_discount/models/account_move.py
kittiu: Thank you, this is a good comment.
Fixed.

kittiu:
> > Hi @kittiu, any news about this?
> 
> Sorry, I am back now. :) Will continue.

Done.

kittiu:
@benwillig @legalsylvain @ps-tubtim ready for review. Thanks!

benwillig (approved):
functionnal tests and code review, LGTM

legalsylvain (changes requested):
Not sure but introducing discount 1 will break all the native feature when invoices are created from sale /purchase order. Dont you think ?

benwillig (changes requested):
> Not sure but introducing discount 1 will break all the native feature when invoices are created from sale /purchase order. Dont you think ?

@legalsylvain might be right as we can see [here](https://github.com/odoo/odoo/blob/13.0/addons/sale/models/sale.py#L1415). Don't we simply need to set an inverse function on discount to store the value in discount1? Not sure if we will get an infinite loop here.

legalsylvain:
what about the previous design ? without ``discount_1`` field ?

kittiu:
@legalsylvain
I did try that but quite troublesome due to the new dynamic journal entry of new account.move. I did tried but remember it is very difficult.

If you still insist, I will try again.

kittiu:
> Not sure but introducing discount 1 will break all the native feature when invoices are created from sale /purchase order. Dont you think ?

Another though, what if when the original discount has value, and others don't. I will distribute it into discount_1. In other words, I will ensure that, the discount will always equal to discount_1 * discount_2 * discount_3

The new account.move in v13, is much more complex than v12, that is one reason. Another reasons is the code of v12 by itself is much more complex without having discount_1.
https://github.com/OCA/account-invoicing/blob/12.0/account_invoice_triple_discount/models/account_invoice.py

I still think, this PR is also the improvement. WDYT?

benwillig:
The problem here is changing ```discount``` as a compute field will break other odoo processes which populate that field on creation (or modify it afterwards). such as the sale one as @legalsylvain pointed it out. Even if there is a discount on the sale order, it won't be propagated to the invoice because ```_compute_triple_discount``` will be triggered and will override that value. With the current PR, we would need to add other modules to fix it and this is not how it should be done.

kittiu:
@benwillig OK.
@legalsylvain, WDYT?
I may open another PR and comply with all opinion here.

kittiu:
@benwillig @legalsylvain I must confess.
I tried again, and I found that without discount_1 + the complexity of v13 account.move, I can't really proceed. It is too complex for me.
I have 2 options now.
1) withdraw myself from this PR (I may close it).
2) keep the discount_1, ensure passing discount alone is ok for non-triple discount cases, and for sales/purchase triple discount make sure that related addons comply with this, i.e.,
https://github.com/ecosoft-odoo/sale-workflow/blob/13.0-mig-sale_triple_discount/sale_triple_discount/models/sale_order_line.py#L90

I understand this PR also need migration script from v12 to v13, not very elegant PR also.

benwillig:
I will make further investigations on my side and make a PR if possible.

kittiu:
As most people agree to move to this approach https://github.com/OCA/account-invoicing/pull/733, so I would like to close this for now.

@ps-tubtim please help fix https://github.com/OCA/sale-workflow/pull/1138 for this #733 (when it is ready)

PR #665, opened 2020-02-01
[FIX] sale_order_invoicing_grouping_criteria: Error accessing partners

omar7r:
Hi,

Small fix that change the inheritance of partner's view because the record that use to position the new data in the view is from other inheritance limited to group "account.group_account_invoice" and if the user has not this group, it fails accessing partner's view.

Regards.

### DIFF
--- a/sale_order_invoicing_grouping_criteria/views/res_partner_views.xml
+++ b/sale_order_invoicing_grouping_criteria/views/res_partner_views.xml
@@ -5,7 +5,7 @@
 
     <record id="view_partner_form" model="ir.ui.view">
         <field name="model">res.partner</field>
-        <field name="inherit_id" ref="base.view_partner_form"/>
+        <field name="inherit_id" ref="account.view_partner_property_form"/>
         <field name="arch" type="xml">
             <xpath expr="//group[@name='acc_sale']" position="inside">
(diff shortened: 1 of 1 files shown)


sale_order_invoicing_grouping_criteria/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 
 {
     'name': 'Sales order invoicing grouping criteria',
-    'version': '11.0.1.0.0',
+    'version': '11.0.1.1.0',
```
pedrobaeza: Please don't change the module version anymore. The bot does it and avoids conflicts.

omar7r:
Done @pedrobaeza

pedrobaeza (approved):
I discussed with other contributor about adding security group to this view instead of changing inheritance in a similar case, but let's not delay this more.

kittiu (approved):

pedrobaeza:
@omar7r please squash both commits into one and I'll merge. Also FYI, `[REM]` prefix in the commit is used when you remove a module from the repo, not for that.

omar7r:
no, ahí se añaden los no sujetos, yo hablo de añadir el iva exento repercutido

omar7r:
squash done @pedrobaeza

eantones:
Thanks @pedrobaeza

PR #669, opened 2020-02-11
[FIX] uom onchange of account invoice line in account_invoice_pricelist

### LINKED ISSUE #592: [12.0] Bug with account_invoice_pricelist
With module account_invoice_pricelist installed.

Create an invoice with a specific pricelist.

Create a new invoice line with an article having a different price on the form view than in the pricelist.

You'll see you don't get the price from the pricelist but the price from the product view 'standard price or lst_price' in function of the invoice type).

The reason is that the on change on UOM is done after the onchange of the product and the price is recomputed in Odoo standard code with this price and not the pricelist price. If you change again the product to select the same one on the already created invoice line. The price is correctly set because UOM is not redefined.

ticket_oca

Thank you,

Regards,

Julien

layadelcarmen:
fixes #592

### DIFF
--- a/account_invoice_pricelist/models/account_invoice.py
+++ b/account_invoice_pricelist/models/account_invoice.py
@@ -65,10 +65,9 @@ class AccountInvoiceLine(models.Model):
 
     @api.onchange('uom_id')
     def _onchange_uom_id(self):
+        super(AccountInvoiceLine, self)._onchange_uom_id()
         if self.invoice_id.pricelist_id:
             self._onchange_product_id_account_invoice_pricelist()
-        else:
-            super(AccountInvoiceLine, self)._onchange_uom_id()
 
     @api.multi
     def update_from_pricelist(self):
--- a/account_invoice_pricelist/models/account_invoice.py
+++ b/account_invoice_pricelist/models/account_invoice.py
@@ -44,7 +44,7 @@ class AccountInvoice(models.Model):
 class AccountInvoiceLine(models.Model):
     _inherit = 'account.invoice.line'
 
-    @api.onchange('product_id', 'quantity', 'uom_id')
+    @api.onchange('product_id', 'quantity')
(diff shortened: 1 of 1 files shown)


account_invoice_pricelist/models/account_invoice.py
```diff
@@ -63,6 +63,13 @@ def _onchange_product_id_account_invoice_pricelist(self):
             product.price, product.taxes_id, self.invoice_line_tax_ids,
             self.company_id)
 
+    @api.onchange('uom_id')
+    def _onchange_uom_id(self):
+        if self.invoice_id.pricelist_id:
```
legalsylvain, line 69: Dont you have to test also self.invoice_id.partner_id ? Like line 49 ?

account_invoice_pricelist/models/account_invoice.py
```diff
@@ -63,6 +63,13 @@ def _onchange_product_id_account_invoice_pricelist(self):
...
+            self._onchange_product_id_account_invoice_pricelist()
+        else:
```
legalsylvain: Is it normal not to call super always ?

legalsylvain (commented):
Some questions inline.

account_invoice_pricelist/models/account_invoice.py
layadelcarmen: That might not be needed because create invoice line checks that partner exists, and in the later is checked if partner exists to run the same previous onchange.

account_invoice_pricelist/models/account_invoice.py
layadelcarmen: I added a new commit. I have tested with pricelist and normal price and seams to be working.

josepmy:
@layadelcarmen @legalsylvain the fix seem ok isn't?

legalsylvain (approved):

marcelsavegnago (approved):

PR #674, opened 2020-02-14
[ADD] 13.0 account_move_tier_validation

pegonzalezspesol:
Installation
============

This module depends on base_tier_validation. You can find it at [OCA/server-ux](https://github.com/OCA/server-ux)

Configuration
=============

To configure this module, you need to:

#. Go to *Settings > Technical > Tier Validations > Tier Definition*.
#. Create as many tiers as you want for Account Move model.

Usage
=====

To use this module, you need to:

1. Create an Account Invoice triggering at least one "Tier Definition".
2. Click on Request Validation button.
3. Under the tab Reviews have a look to pending reviews and their statuses.
4. Once all reviews are validated click on Post.

Additional features:

* You can filter the Account Invoices requesting your review through the filter Needs my Review.
* User with rights to confirm the Account Invoice (validate all tiers that would be generated) can directly do the operation, this is, there is no need for her/him to request a validation.


kittiu:
@OCA/community-maintainers seem that runbot and codecov is not active?

account_move_tier_validation/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+# Copyright <2020> PESOL <info@pesol.es>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+{
+    "name": "Account Move Tier Validation",
+    "summary": "Extends the functionality of Account Moves to "
+    "support a tier validation process.",
+    "version": "13.0.1.0.0",
+    "category": "Accounts",
+    "website": "https://github.com/OCA/purchase-workflow",
```
kittiu: This should be /account-invoicing

account_move_tier_validation/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+    "contributors": ["Odoo Peru"],
```
kittiu: Many not need this. Contributors goes to readme/CONTRIBUTORS.rst already.
But you can add yourself as maintainer instead :)
```
"maintainers": ["<your git account>"],
```

kittiu:
@pegonzalezspesol 
I have tested when create Invoice (account.move) which is ok.
My guess is this module created with the intention to tier validate invoice, right?
But this will also apply normal journal entry too (which is not normal).

I understand that, user can test with type != 'entry' in the domain. But are there any good way warn user of this?

What about when select model = "account.move" then give a nice warning for user to beware of this.

@LoisRForgeFlow what do think about this issue?

account_move_tier_validation/__manifest__.py
pegonzalezspesol: fixed in last commit

account_move_tier_validation/__manifest__.py
pegonzalezspesol: removed contributors in last commit

pegonzalezspesol:
> @pegonzalezspesol
> I have tested when create Invoice (account.move) which is ok.
> My guess is this module created with the intention to tier validate invoice, right?
> But this will also apply normal journal entry too (which is not normal).
> 
> I understand that, user can test with type != 'entry' in the domain. But are there any good way warn user of this?
> 
> What about when select model = "account.move" then give a nice warning for user to beware of this.
> 
> @LoisRForgeFlow what do think about this issue?

Thanks @kittiu for your revision and functional test!

About the issue you point out.. I suggest that user defines in the Tier Validation in which type of move is needed the validation. (Customer Invoice, Supplier Invoice, .. refunds)
image

kittiu:
@pegonzalezspesol yes, and would it be nice to have remarks in README?
WDYT?

pegonzalezspesol:
> @pegonzalezspesol yes, and would it be nice to have remarks in README?
> WDYT?

Absolutely. The more info the better.. let's wait for some other suggestions / revisions and I will include it.

dreispt (changes requested):
<s>I'm sorry, you need to bring in the module history from the 12.0 branch, and then apply any change needed in a final commit.
This guide will help you: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-13.0</s>
Please ignore

dreispt (approved):

kittiu (approved):
I am ok, only left the readme. :)

AaronHForgeFlow (approved):
:+1: code + functional review

PR #675, opened 2020-02-14
[FIX] account_invoice_transmit_method: Transmit method does not propagated to invoice when user invoices from saler orders

sergio-teruel:
cc @Tecnativa TT22168
When you do a invoice from sale order the method _prepare_invoice invokes an invoice_onchage only with partner_id key, so we can not presuppose that partner has not a n invoice transmit method...
https://github.com/odoo/odoo/blob/4e90f7bbc5e32998942ec6d0542ffa6cb5ed474a/addons/sale/models/sale.py#L472-L474
With this patch, the create method will do a query to partner to get transmit method.


pedrobaeza (approved):
Maybe you should check context for getting type in that onchange?

carlosdauden (approved):

sergio-teruel:
I think that is not necessary

PR #676, opened 2020-02-17
[FIX] purchase_batch_invoicing: fix visibility of invoice_status field

ernestotejeda:
Cc @Tecnativa TT22179

### DIFF
--- a/purchase_batch_invoicing/views/purchase_order_view.xml
+++ b/purchase_batch_invoicing/views/purchase_order_view.xml
@@ -25,7 +25,7 @@
         <field name="inherit_id" ref="purchase.purchase_order_tree"/>
         <field name="arch" type="xml">
             <xpath expr="//field[@name='invoice_status']" position="attributes">
-                <attribute name="attrs"/>
+                <attribute name="invisible">0</attribute>
             </xpath>
         </field>
     </record>


pedrobaeza:
I remember that it has also about column visibility in the context

ernestotejeda:
Yes, but that was in v11 https://github.com/odoo/odoo/blob/11.0/addons/purchase/views/purchase_views.xml#L401
In v12 it has just invisible="1" https://github.com/odoo/odoo/blob/b405da2d3fbdf2dfc881877fd4663f0a40aa6bd6/addons/purchase/views/purchase_views.xml#L375

purchase_batch_invoicing/views/purchase_order_view.xml
```diff
@@ -25,7 +25,7 @@
         <field name="inherit_id" ref="purchase.purchase_order_tree"/>
         <field name="arch" type="xml">
             <xpath expr="//field[@name='invoice_status']" position="attributes">
-                <attribute name="attrs"/>
+                <attribute name="invisible"/>
```
pedrobaeza: Is this working? Shouldn't be:

```suggestion
                <attribute name="invisible">0</attribute>
```
?

ernestotejeda: I tested it locally and it worked, but to be sure let do it as you stated, ;-)

pedrobaeza: Well, if it was working no need of changing it. I didn't know for sure. That's why I asked. Any way, we both know for the next time.

pedrobaeza (approved):

pedrobaeza:
Fast-tracking this as obvious.

PR #678, opened 2020-02-25
[MIG][13.0] stock_picking_return_refund_option: Migration to v13.0

sergio-teruel:
cc @Tecnativa
TT21691


Tardo:
Please, check travis errors.

sergio-teruel:
In the pre_init_hook all mandatory modules was checked to be install, but in v13.0 these modules not install their dependencies.... 
¿Can yoy help me?

Tardo:
Check Travis

sergio-teruel:
@Tardo please, read my comments...
This module not needs from sales and purchase modules...
But for testing I have tried install sales and purchase... but I have not succeeded..

sergio-teruel:
ping @Tardo @carlosdauden @pedrobaeza

rousseldenis:
> In the pre_init_hook all mandatory modules was checked to be install, but in v13.0 these modules not install their dependencies....
> ¿Can yoy help me?

If I understand, you want to install those modules when installing this one... Bad idea. I think you don't want to do glue modules but that's the only way to do it right.

sergio-teruel:
@rousseldenis Yes I know!! but only need it for tests.

rousseldenis:
> @rousseldenis Yes I know!! but only need it for tests.

Nope, you don't.

You need to remove that part : https://github.com/OCA/account-invoicing/pull/678/files#diff-a32444d8cce3c719a91a6f7bcbb2998b2811ca8cc965a03c1502a157bc31fae5R76

and put it in a purchase module.

You need to test only related picking things.

stock_picking_return_refund_option/hooks.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2018 Tecnativa - Sergio Teruel
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import SUPERUSER_ID, api, tools
+
+
+def pre_init_hook(cr):
```
rousseldenis: Please don't do that.

As already said, you must respect dependencies.

If your code depends on purchase_stock and sale_stock, put them in depends key in manifest. IMHO, splitting in more specialized modules is the best option.

rousseldenis: https://github.com/Tecnativa/account-invoicing/pull/3

This is not the best, but improve module stability

rousseldenis:
@sergio-teruel All :green_apple: !

pedrobaeza:
Not really comfortable adding such heavy dependencies, although our customers using that module have both installed. Please squash some administrative commits and rebasing we will get the merge commit out, that will cause problems retrieving commit history for later versions.

rousseldenis:
> Not really comfortable adding such heavy dependencies, although our customers using that module have both installed. Please squash some administrative commits and rebasing we will get the merge commit out, that will cause problems retrieving commit history for later versions.

That was the easiest solution. The better is to split logic into separated modules but you cannot have code that depends on modules that are not in dependency tree...

pedrobaeza:
Well, such code was conditional and is inferred by introspection if I don't remember bad. This technique is used in other places, but let's not use more time on this and if we ever have the case where not both should be installed, then we come back to the debate.

pedrobaeza (approved):

PR #679, opened 2020-03-01
[DUMMY] account_invoice_check_total: Translate term with missing arg

luisg123v:
This is to test the case where the number of arguments (%s) of the
original string doesn't match the translated string.

This is related to https://github.com/OCA/pylint-odoo/issues/274

### DIFF
--- a/account_invoice_check_total/i18n/es.po
+++ b/account_invoice_check_total/i18n/es.po
@@ -55,7 +55,7 @@ msgid ""
 "There is a difference of %s"
 msgstr ""
 "Por favor, verifique el total de la factura.\n"
-"                        El total introducido (%s) no "
+"                        El total introducido (s) no "
 "corresponde                        con el total calculado (%s)!\n"
 "Hay una diferencia de %s"
 


AaronHForgeFlow:
Is it happening only in es.po file?

luisg123v:
@AaronHForgeFlow no, it happens in any file. This is just an demonstration.

AaronHForgeFlow:
Is is this line? https://github.com/OCA/account-invoicing/blob/11.0/account_invoice_check_total/models/account_invoice.py#L50
It seems correct. The error must be in the .pot file, isn't it?

luisg123v:
@AaronHForgeFlow the module is correct.

This is just a dummy PR to demonstrate that CI doesn't fail when the number of arguments doesn't match original string.

This is related to https://github.com/OCA/pylint-odoo/issues/274, it's on the description.

AaronHForgeFlow:
:sweat_smile: understood. Sorry.

luisg123v:
Solved on https://github.com/OCA/pylint-odoo/pull/290, so closing.

PR #680, opened 2020-03-02
[12.0][IMP] sale_order_invoicing_queued: Make it compatible with grouping criteria

pedrobaeza:
We compatibilize this module with `sale_order_invoicing_grouping_criteria` one checking explicitly if there's a method with the name that the other module declares, and if so, calling it for getting the grouping criteria.

This is the only valid approach for now, as we don't have any hook for getting that criteria.

cc @Tecnativa TT21478

### DIFF
--- a/sale_order_invoicing_queued/wizards/sale_advance_payment_inv.py
+++ b/sale_order_invoicing_queued/wizards/sale_advance_payment_inv.py
@@ -1,4 +1,4 @@
-# Copyright 2019 Tecnativa - Pedro M. Baeza
+# Copyright 2019-2020 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from odoo import _, exceptions, models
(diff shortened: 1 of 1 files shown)


sale_order_invoicing_queued/wizards/sale_advance_payment_inv.py
```diff
@@ -20,6 +20,10 @@ def enqueue_invoices(self):
         grouped_orders = defaultdict(lambda: order_obj.browse())
         for order in sale_orders:
             group_key = (order.partner_invoice_id.id, order.currency_id.id)
+            # If we have `sale_order_invoicing_grouping_criteria` module
+            # installed, we take that grouping criteria
+            if getattr(order, '_get_sale_invoicing_group_key'):
```
sergio-teruel:
```suggestion
            if hasattr(order, '_get_sale_invoicing_group_key'):
```

sergio-teruel (changes requested):

sale_order_invoicing_queued/wizards/sale_advance_payment_inv.py
pedrobaeza: Well, technically both works, as `getattr` return the method if exists and resolves as boolean `True`, but changed for being more meaningful.

carlosdauden (approved):

sergio-teruel (approved):

PR #681, opened 2020-03-02
[13.0][FIX] sale_order_invoicing_grouping_criteria: Return recordset instead of list

sergio-teruel:
cc @Tecnativa


pedrobaeza (approved):
Fast-tracking this patch

pedrobaeza:
Ouch, pre-commit fails. Why don't you install locally for being run always (`pre-commit install -f` in your repository)?

sergio-teruel:
Ohh Yes, I already have it installed, but in some cases I forget execute it locally....

PR #570, opened 2019-07-23
[12.0] account_invoice_transmit_method: use base_view_inheritance_extension

alexis-via:
Same PR as https://github.com/OCA/account-invoicing/pull/440/files

Seems the module base_view_inheritance_extension hasn't been ported to v12 yet ; I use the module for v11 which works without change for v12.

PR #682, opened 2020-03-02
[13.0] MIG account_invoice_transmit_method

alexis-via:
This PR depends on the PR to migrate base_view_inheritance_extension to v13 https://github.com/OCA/server-tools/pull/1777
The start point of this v12 PR to improve account_invoice_transmit_method https://github.com/OCA/account-invoicing/pull/570/files


account_invoice_transmit_method/views/account_move.xml
```diff
@@ -0,0 +1,34 @@
...
+
+
+<record id="view_move_form" model="ir.ui.view">
+    <field name="name">Add Transmit Method on Invoice form view</field>
+    <field name="model">account.move</field>
+    <field name="inherit_id" ref="account.view_move_form"/>
+    <field name="arch" type="xml">
+        <field name="journal_id" position="after">
+            <field name="transmit_method_id" widget="selection"
+                domain="[('customer_ok', '=', True)]"
```
TDu: While testing I noticed a problem with this domain, it is never set properly. And will always be set to the next one below `domain="[('supplier_ok', '=', True)]`. So a transmit_method declared has customer only will not be displayed.
It seems that Odoo does not handle properly the same field declared twice on the same view with different domains. I didn't dig into the source code to know why.

sebalix: Indeed Odoo doesn't support that, fields are unique in a view, and Odoo will consider only the last one.
If the goal is to have a different domain depending of the invoice type, we could overload the `fields_view_get` method?

TDu: This field is defined twice so it is placed in a different position on customer vs vendor invoice.
This was not the case in previous version.
But is it really necessary ? Maybe the idea behind it is that it is more important to see the transmit method for customer invoices because it will require an action from the user to transmit the invoice ?

alexis-via: I fixed the domain ; it is now set via @api.onchange('type').
I kept the placing at 2 different locations because, on customer invoice, the field is quite important so I want to place it at the top ; on vendor bills, it's not important, so I put it in the last tab.

TDu: @alexis-via Thank you for the fix. I actually did not know you could do this in an `onchange`.
Could you take care of the linting errors in the xml files, I think it is the only thing that prevents this PR to be merged

TDu:
@alexis-via To help on merging this one I took the liberty to open #808 
And hopefully help other module depending to this one

alexis-via:
The other PR has been merged, so let's cancel this one.

PR #701, opened 2020-04-14
[12.0][FIX] account_invoice_refund_link: Fix tests

ernestotejeda:
TT18546


pedrobaeza:
Why is this needed? Please explain the problem you faced.

ernestotejeda:
I got this error during tests performance:
https://github.com/odoo/odoo/blob/12.0/addons/account/models/account_invoice.py#L842
This PR could fix that.

pedrobaeza:
That's not possible. The possible problem is that the scaffolding doesn't contain `l10n_generic_coa` in the included addons.

pedrobaeza:
This is finally not needed.

PR #707, opened 2020-04-25
[12.0][ENH] account_menu_invoice_refund, remove payment netting feature

### LINKED ISSUE #705: [12.0] account_menu_invoice_refund: Shouldn't include feature for payments
The module should include only the menu part, and not the feature of allowing to mix both invoice types in payments, due to the following reasons:

- I want the first thing (the menu), but not the second, and I'm forced to used both.
- The module name and initial scope don't include the second feature.
- The second feature doesn't belong here (it's a payment feature).
- The monkey-patch technically is conflicting and I prefer to avoid it in my set of modules.

@kittiu can you please check my reasoning?

kittiu:


account_menu_invoice_refund/readme/DESCRIPTION.rst
```diff
@@ -1,13 +1,11 @@
 By Odoo standard, Invoices and Refunds are in different menu,
-and can't be mixed upon payment.
 
-This module add 2 new menus,
+This module simply add 2 new menus,
```
pedrobaeza:
```suggestion
This module simply adds 2 new menus,
```

pedrobaeza (approved):

pedrobaeza:
OK to go on with this?

kittiu:
Yes, you can merge?

PR #710, opened 2020-05-08
11.0

alain-binovo:
[FIX] Fixed error on different currency than default one.

### DIFF
--- a/account_invoice_reimbursable/models/account_invoice.py
+++ b/account_invoice_reimbursable/models/account_invoice.py
@@ -118,9 +118,8 @@ class AccountInvoice(models.Model):
                 self)
             currency = self.currency_id.with_context(
                 date=date)
-            if not (self.get('currency_id') and self.get('amount_currency')):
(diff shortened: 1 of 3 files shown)


kittiu:
> [FIX] Fixed error on different currency than default one.

Thank you for your contribution. By reading the code, I am not sure what you try to fix. Can you give the step to error? or explain more here.

Also, according to the guide line, I think you can name your branch like,
* 11.0-fix-account_invoice_reimbursable-do_xxx
And the issue name could be
* [11.0][FIX] account_invoice_reimbursable, do xxx

alain-binovo:
Hi, the error happens when you try to validate an invoice from a different currency from the default one and looking for the 12.0 and 13.0 versions there is an if that is not necessary that was the one which raises the error.

kittiu:
It seem right that, the "if" is not necessary, I am ok with the removal of that.
Please consider the above comments for the naming convention of this PR. :+1:

kittiu (approved):

PR #713, opened 2020-05-14
[12.0][FIX] Copy purchase data from picking to invoice, module stock_picking_invoicing.

mbcosta:
Copy purchase data from picking to invoice.

### DIFF
--- a/stock_picking_invoicing/wizards/stock_invoice_onshipping.py
+++ b/stock_picking_invoicing/wizards/stock_invoice_onshipping.py
@@ -345,6 +345,7 @@ class StockInvoiceOnshipping(models.TransientModel):
             'journal_id': journal.id,
             'picking_ids': [(4, p.id, False) for p in pickings],
         })
+        # HACK: using hasattr for not depending on Purchase
         if hasattr(picking, 'purchase_id'):
             values['purchase_id'] = picking.purchase_id.id
         invoice, values = self._simulate_invoice_onchange(values)
@@ -456,8 +457,10 @@ class StockInvoiceOnshipping(models.TransientModel):
             'move_line_ids': move_line_ids,
             'invoice_id': invoice.id,
         })
+        # HACK: using hasattr for not depending on Purchase
         if hasattr(move, 'purchase_line_id'):
             values['purchase_line_id'] = move.purchase_line_id.id
+        # HACK: using hasattr for not depending on Sale
         if hasattr(move, 'sale_line_id'):
             if move.sale_line_id:
                 values['sale_line_ids'] = [(6, 0, move.sale_line_id.id)]
--- a/stock_picking_invoicing/wizards/stock_invoice_onshipping.py
+++ b/stock_picking_invoicing/wizards/stock_invoice_onshipping.py
@@ -345,6 +345,8 @@ class StockInvoiceOnshipping(models.TransientModel):
             'journal_id': journal.id,
             'picking_ids': [(4, p.id, False) for p in pickings],
         })
+        if hasattr(picking, 'purchase_id'):
+            values['purchase_id'] = picking.purchase_id.id
         invoice, values = self._simulate_invoice_onchange(values)
         return invoice, values
 
@@ -454,6 +456,8 @@ class StockInvoiceOnshipping(models.TransientModel):
             'move_line_ids': move_line_ids,
             'invoice_id': invoice.id,
         })
+        if hasattr(move, 'purchase_line_id'):
+            values['purchase_line_id'] = move.purchase_line_id.id
         values = self._simulate_invoice_line_onchange(values)
         return values
 
--- a/stock_picking_invoicing/wizards/stock_invoice_onshipping.py
+++ b/stock_picking_invoicing/wizards/stock_invoice_onshipping.py
@@ -458,6 +458,9 @@ class StockInvoiceOnshipping(models.TransientModel):
         })
         if hasattr(move, 'purchase_line_id'):
             values['purchase_line_id'] = move.purchase_line_id.id
+        if hasattr(move, 'sale_line_id'):
+            if move.sale_line_id:
+                values['sale_line_ids'] = [(6, 0, move.sale_line_id.id)]
         values = self._simulate_invoice_line_onchange(values)
         return values
 


rvalyi (changes requested):
there is no such purchase_id field with the stock_picking_invoicing and its dependencies. Indeed, it doesn't depend on purchase nor purchase_stock. You can check for yourself: no purchase_id in https://github.com/odoo/odoo/blob/12.0/addons/stock/models/stock_picking.py
On the contrary this field is brought by the purchase_stock module:
https://github.com/odoo/odoo/blob/12.0/addons/purchase_stock/models/stock.py#L11

So what you should do instead is override the _build_invoice_values_from_pickings method and adjust the invoice in the localization module (writting after create is bad but given this is a per picking manual operation this is still okay).

Or if you feel it can be done quickly in a clean way, extract some _prepare_invoice_from_picking that would contain much of the def _build_invoice_values_from_pickings(self, pickings) code, be called by _build_invoice_values_from_pickings but just return values to _build_invoice_values_from_pickings which would in turn call _simulate_invoice_onchange to create the invoice and would be overridable before the invoice is created.

rvalyi:
or alternatively you could test hasattr(picking, 'purchase_id') and update the values dictionary only if purchase_id is found. This is often done to avoid the multiplication of glue modules and would be acceptable for me.

stock_picking_invoicing/wizards/stock_invoice_onshipping.py
```diff
@@ -338,13 +338,18 @@ def _build_invoice_values_from_pickings(self, pickings):
             'partner_id': partner_id,
             'account_id': account_id,
             'payment_term_id': payment_term,
+            'purchase_id': picking.purchase_id.id or False,
```
rvalyi: but of course if you go for the if hasattr(picking, 'purchase_id'), then you should remove this line with picking.purchase_id.id or you will have an error...

stock_picking_invoicing/wizards/stock_invoice_onshipping.py
```diff
@@ -452,8 +457,13 @@ def _get_invoice_line_values(self, moves, invoice_values, invoice):
             'price_unit': price,
             'invoice_line_tax_ids': [(6, 0, taxes.ids)],
             'move_line_ids': move_line_ids,
+            'purchase_line_id': move.purchase_line_id.id or False,
```
rvalyi: same comment as above

rvalyi (changes requested):

mbcosta:
Hi @rvalyi , thank you for review. I made the last sugestion "test hasattr(picking, 'purchase_id')" . 

"Indeed, it doesn't depend on purchase nor purchase_stock."  it's true, if should be better make it in the localization, there is no problem for me.

stock_picking_invoicing/wizards/stock_invoice_onshipping.py
```diff
@@ -454,6 +458,10 @@ def _get_invoice_line_values(self, moves, invoice_values, invoice):
             'move_line_ids': move_line_ids,
             'invoice_id': invoice.id,
         })
+        if hasattr(move, 'purchase_line_id'):
+            values.update({
+                'purchase_line_id': move.purchase_line_id.id or False,
```
rvalyi: can be made simpler as the above suggestion

rvalyi (changes requested):

stock_picking_invoicing/wizards/stock_invoice_onshipping.py
```diff
@@ -345,6 +345,10 @@ def _build_invoice_values_from_pickings(self, pickings):
             'journal_id': journal.id,
             'picking_ids': [(4, p.id, False) for p in pickings],
         })
+        if hasattr(picking, 'purchase_id'):
+            values.update({
+                'purchase_id': picking.purchase_id.id or False,
```
rvalyi: I think this could be written simply as:
```
if hasattr(picking, 'purchase_id'):
    values['purchase_id'] = picking.purchase_id
```
No??

rvalyi (changes requested):

mbcosta:
Hi @rvalyi @mileo 
Talking with @renatonlima about the implemantation of stock_picking_invoicing in brazilian localization we noticed there is a functionality related to this module, in version 8.0, that maybe should be reimplemented Invoice Policy

Purchase  https://github.com/OCA/OCB/blob/8.0/addons/purchase/res_config.py#L30 
Sale https://github.com/OCA/OCB/blob/8.0/addons/sale_stock/res_config.py#L39 

The initial idea would be create two new modules purchase_stock_picking_invoicing and sale_stock_picking_invocing where we include fields related purchase/sale during the Invoice creation based on Picking and this re implementation. 

What do you think ?

rvalyi:
Hello Magno, In general, I think adding a "glue module" just because of 1 or 2 fields is a bad idea because it adds a lot of boilerplate code to maintain and remember to install just for 2 or 3 useful lines of code. Then you also tend to create a multiplication of the modules that would depend on these modules...

Instead I think testing for the existence of a field/column like you just did to check if a module is actually installed and if the field should be propagated in some _prepare method is a popular way to fix this kind of situation in the OCA ecosystem. Moreover, sale and purchase are really core Odoo modules generating pickings so we can anticipate their presence in this module without bloating the codebase with extra glue modules.

As for the settings you just mentioned, I would go for adding them in OCA/l10n-brazil/l10n_br_sale_stock and OCA/l10n-brazil/l10n_br_purchase_stock instead of in this repo. The reason is we have more control in OCA/l10n-brazil to merge things faster and this is really where we need them until proven otherwise.

mbcosta:
I understand your point @rvalyi , we could end up with more README data than code in the modules and maybe the Invoice Policy just matter to brazilian localization.

josepmy:
Hello @mbcosta, I'm very interested to test it, it's ready?

mbcosta:
Hi @josepmy , you can test is ready, we are implement this module in brazilian localization.

josepmy:
Hi, ok. Let me try.

marcelsavegnago:
@mbcosta esta PR ready for review? I saw that there are some blocks from @rvalyi that I believe were treated.

rvalyi (approved):
@marcelsavegnago you were right, it's all correct now.
@renatonlima are we not using this already in our projects?

renatonlima (approved):

renatonlima:
@rvalyi 

This PR is very important because if you create one invoice based on sale or purchase picking it needs related to sale/purchase line to update some fields like quantity invoiced and others.

@mbcosta KUDOS for your PR!

PR #619, opened 2019-10-21
13.0 mig account payment term extension

emagdalenaC2i:

PR #714, opened 2020-05-19
[13.0][MIG] account_payment_term_extension: Migration to 13.0

carlos-domatix:
Standard migration of account_payment_term_extension and make the day_of_the_month field invisible because it is not used.


HaraldPanten:
Hi @carlos-domatix I'm afraid, this PR is open yet. https://github.com/OCA/account-invoicing/pull/619

Maybe you can talk to @emagdalenaC2i and agree to keep just one.

carlos-domatix:
Sorry, I had only seen this PR that is closed  #617

emagdalenaC2i:
No, problem. Go ahead #617 is closed too

HaraldPanten (approved):
Hi @carlos-domatix 

Functional review 👍

emagdalenaC2i (approved):
LGTM

HaraldPanten:
Hi @carlos-domatix @emagdalenaC2i 

Few weeks ago I realised that "Odoo base" payment terms options have changed a little bit.

If you remember, in other versions we had the possibility to select "day(s) after the end of the invoice month" in the Due computation; in the terms lines.

On the other hand, this option has been removed and we have only:

- days after the invoice date
- of the following month
- of the current month

Do you think that this option should be added here as an [IMP]? is any other way to set "day(s) after the end of the invoice month" in V13?

@pedrobaeza what do you think?

THX

pedrobaeza:
I don't see that option very useful sincerely, but if anyone consider so, he/she can add that option as IMP.

What I consider is that this module should be moved to OCA/account-payment, as it has been too much time incorrectly in this repository: payment terms are not only something of invoicing, and less now that there are no more invoices.

HaraldPanten:
> I don't see that option very useful sincerely, but if anyone consider so, he/she can add that option as IMP.
> 
> What I consider is that this module should be moved to OCA/account-payment, as it has been too much time incorrectly in this repository: payment terms are not only something of invoicing, and less now that there are no more invoices.

In account_payment repo would make sense, right.

@carlos-domatix , @emagdalenaC2i what do you think? Do you mind opening a new PR in the other repo?

On the other hand, for the "day(s) after the end of the invoice month" option, we'll open an [IMP] PR unless you decide to include this functionality in the [MIG] to V13.

Is that OK for you? THX!

carlos-domatix:
Moved to account-payment: https://github.com/OCA/account-payment/pull/331

PR #718, opened 2020-05-27
[11.0][FIX]account_invoice_fiscal_position_update: Keep taxes from pu…

javierjcf:
…rchase order line when create invoice from smart button

### DIFF
--- a/account_invoice_fiscal_position_update/models/account_invoice.py
+++ b/account_invoice_fiscal_position_update/models/account_invoice.py
@@ -40,6 +40,12 @@ class AccountInvoice(models.Model):
                 else:
                     taxes = product.supplier_taxes_id.filtered(
                         lambda tax: tax.company_id == self.company_id)
+                    # If you change from purchase line the taxes, maintain it.
+                    # HACK: using hasattr for not depending on purchase
+                    if (hasattr(line.invoice_id, 'purchase_id') and
+                            line.invoice_id.purchase_id and
+                            line.purchase_line_id.taxes_id):
+                        taxes = line.purchase_line_id.taxes_id
                 taxes = taxes or account.tax_ids.filtered(
                     lambda tax: tax.company_id == self.company_id)
                 if fp:


pedrobaeza:
I think you have both cases: you want to keep the taxes or you screwed up the fiscal position in the purchase order and want to use a new one, so I don't think this patch is valid (or not less than the previous one).

Roodin:
But if you want to force the change of taxes in a purchase line  i don think that Odoo must change these taxes silently when you create the invoice, ¿no?

pedrobaeza:
But that should be done on the invoice creation, not on the fiscal position change.

javierjcf:
Maybe the problem is to get into position change method when creating a invoice from a purchase. The tax is overwrited by this method in the end because an automatic onchange trigger.

If we can avoid to get into this onchange i think this method should contemplate that the invoice line is created from a purchase line with a different tax.

pedrobaeza:
Which line creating the invoice from the purchase order is triggering that onchange?

javierjcf:
When i was debugging, i found that this method is called twice. One of them
i think is because of:
 /odoo/odoo/models.py(5068)_onchange_eval()

I don't know how odoo internally manages automatic onchange trigger.

The second time is because of:

account-invoicing/account_invoice_fiscal_position_update/models/account_invoice.py(19)_onchange_partner_id()
self.fiscal_position_change().

I think onchanges are triggered because you are creating the invoice from
an on_chage method in first instance.

account_invoice_fiscal_position_update/models/account_invoice.py
```diff
@@ -40,6 +40,10 @@ def fiscal_position_change(self):
                 else:
                     taxes = product.supplier_taxes_id.filtered(
                         lambda tax: tax.company_id == self.company_id)
+                    # If you change from purchase line the taxes, maintain it.
+                    if line.purchase_line_id and \
```
pedrobaeza:
```suggestion
                    if line.invoice_id.purchase_id and line.purchase_line_id and \
```

pedrobaeza (changes requested):
Thanks for the extra information. Indeed the onchanges are triggered in chain by Odoo itself, but with this suggestion I think we cover both cases:

- Adding the purchase order in the invoice, keeping purchase lines taxes
- Changing the fiscal position on the invoice latter.

Please try to confirm. It would be great to have a test to cover all of this, but maybe it's too much to ask.

javierjcf:
I will try to do the tests soon

pedrobaeza (approved):
Thanks! Approving although no tests

pedrobaeza (changes requested):
Oh, wait, there's a basic problem: this module doesn't depend on purchase!! You have to add a new one as glue module.

javierjcf:
You're right!
I add purchase to manifest and rebased all the PR commits

pedrobaeza:
Yeah, but that is not valid, as you can't force all the users of this module to install purchase.

You need to create a new `account_invoice_fiscal_position_update_purchase` with this specific code (well, and adaptation with a hook in this one) for being correct. Other "hack" is to test the presence of the fields (`if hasattr(line.invoice_id, 'purchase_id') and ....`) and don't require the dependency and the extra module.

javierjcf:
I chose the second option. Simpler and not requires to remember you need a
special extra module for this fiscal position update if you are using the
purchase module

account_invoice_fiscal_position_update/models/account_invoice.py
```diff
@@ -40,6 +40,12 @@ def fiscal_position_change(self):
                 else:
                     taxes = product.supplier_taxes_id.filtered(
                         lambda tax: tax.company_id == self.company_id)
+                    # If you change from purchase line the taxes, maintain it.
+                    if hasattr(line.invoice_id, 'purchase_id') and \
+                            line.invoice_id.purchase_id and \
+                            line.purchase_line_id and \
+                            line.purchase_line_id.taxes_id:
+                        taxes = line.purchase_line_id.taxes_id
```
pedrobaeza: Clarify hack + remove extra useless condition
```suggestion
                    # HACK: using hasattr for not depending on purchase
                    if (hasattr(line.invoice_id, 'purchase_id') and
                            line.invoice_id.purchase_id and
                            line.purchase_line_id.taxes_id):
                        taxes = line.purchase_line_id.taxes_id
```

pedrobaeza (approved):

PR #720, opened 2020-05-28
[12.0][FIX] account_invoice_fiscal_position_update: Fix display warning

igallart:
When we have sections on an invoice, if we change the fiscal position,
on the invoice, an error appears when trying to show a warning.

TypeError: not all arguments converted during string formatting


igallart:
Anyone interested in merging?

PR #722, opened 2020-06-01
[12.0][ADD] stock_picking_invoicing_purchase

sergiocorato:
Add glue module to get data from purchase when installed.


kittiu:
Following is what I see since installing stock_picking_invoicing alone, with or without this stock_picking_invoicing_purchase. Not sure it is intentional?

When click on Picking's Invoices link, I expect it to show only the invoice related to the picking.

Selection_140
Selection_141

kittiu:
From user perspective,
I think this module should be auto_installed when `stock_picking_invoicing` and `purchase` is installed (and not purchase_discount).
But I see that auto_install=True and it is also depends on `purchase_discount` too, which make it not auto install as I didn't have purchase_discount yet.

Should it have a glue module for `purchase_discount` and `stock_picking_invoicing_purchase` too?

Or if we don't want to have more module, any other technique to auto install without `purchase_discount` and let the `stock_picking_invoicing_purchase` do the install what it depends on (purchase_discount).

I don't have the best solution also. What is others opinion?

kittiu:
This module will (not yet) working with triple discount. Should we mark it in manifest?
https://github.com/OCA/purchase-workflow/blob/12.0/purchase_triple_discount/models/purchase_order.py#L45

josepmy:
> Following is what I see since installing stock_picking_invoicing alone, with or without this stock_picking_invoicing_purchase. Not sure it is intentional?
> 
> When click on Picking's Invoices link, I expect it to show only the invoice related to the picking.
> 
> Selection_140
> Selection_141

I agree with you, the user is fuzzed and must show only the invoiced document related to the picking.
Also, I see that the purchase order don't show the invoices icon, this is confusing, and let the user reinvoice again. 
One purchase could have many pickings and many invoices related to one purchase.

sergiocorato:
@kittiu @josepmy this module do not touch those parts, so you can open an issue related.
I think I'll remove the auto_install option, to give user the ability to install when needed.
The dependency on purchase discount is to do not have another module, at the end I need triple discount so I'll propose this one #737 but it's simple to make another module without discount if someone need it.

josepmy:
Hi @sergiocorato looks is fine, isn't?

sergiocorato:
I ended using `purchase_batch_invoicing` as it permits group invoicing, so close this one.

PR #731, opened 2020-06-10
[MIG] account_invoice_pricelist Migration to  13

yassine-teoshore:


pedrobaeza:
Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-13.0.

If the jump is between several versions, you have to modify the source branch in the main command to accommodate it to this circumstance.

ValentinVinagre:
Hi @yassine-teoshore,
some errors:
1.- If I select the "Show public price & discount to the customer" option in the "Discount Policy" field, the discount doesn't appear in the invoice line.
2.- The new price is only calculated when the invoice is saved, but it should change before saving. Same as a Sale Orders.

P.D: preserve commit history

Thanks.

manuelregidor:
@pedrobaeza , as the person developing this migration has not repplied, we have finished it and opened a new PR (https://github.com/OCA/account-invoicing/pull/757). Could you please close this one? Thank you.

PR #737, opened 2020-06-19
[12.0][IMP] add module to get data from purchase with triple discount when invoicing

sergiocorato:


kittiu:
@sergiocorato thanks for this module.
I tested your module related to triple discount without problem. At first I wonder why you have to calculate price again.

Than, I notice that the module `stock_picking_invoicing` seem to not work with purchase order well.
* PO has no tax, but Invoice have tax
* PO set some price, invoice set another product price.

Not a bug by itself, but no glue module? WDYT?

sergiocorato:
I ended using `purchase_batch_invoicing` as it permits group invoicing, so close this one.

PR #738, opened 2020-06-22
[10.0] account_invoice_transmit_method: Remove arg with wrong syntax

alexis-via:
In the initial dev of the module, I made a typo in the definition of the
fields customer_invoice_transmit_method_id and supplier_invoice_transmit_method_id
on res.partner: I wrote "company_dependant=True" instead of
"company_dependent=True".
As a result, these 2 fields were not company-specific.
So we have 2 options:
1) Fix the typo and write migration scripts
2) Keep these 2 fields with company_dependent=False, as nobody as
complained so far about the fact that the value of this parameter didn't
vary from one company to another.

This PR implement option 2.

### DIFF
--- a/account_invoice_transmit_method/models/partner.py
+++ b/account_invoice_transmit_method/models/partner.py
@@ -11,14 +11,14 @@ class ResPartner(models.Model):
 
     customer_invoice_transmit_method_id = fields.Many2one(
         'transmit.method', string='Customer Invoice Transmission Method',
-        company_dependant=True, track_visibility='onchange',
+        track_visibility='onchange',
         domain=[('customer_ok', '=', True)], ondelete='restrict')
     customer_invoice_transmit_method_code = fields.Char(
         related='customer_invoice_transmit_method_id.code',
         string='Customer Invoice Transmission Method Code', readonly=True)
     supplier_invoice_transmit_method_id = fields.Many2one(
         'transmit.method', string='Vendor Invoice Reception Method',
-        company_dependant=True, track_visibility='onchange',
+        track_visibility='onchange',
         domain=[('supplier_ok', '=', True)], ondelete='restrict')
     supplier_invoice_transmit_method_code = fields.Char(
         related='supplier_invoice_transmit_method_id.code',


alexis-via:
I'm ready to implement option 1 if you guys think that option 1 is best.

bunhin:
Not using the module yet, but vote for option 1

etobella (approved):
Usually when a customer wants to receive invoices, they use a single method, so I vote for not making it company_dependent. However, on my experience I found that customer might want to receive invoices in two different methods. For example:

* Integrated invoice with a WS (Spanish FACe) 
* Invoice by mail for a manual validation with all the attached information

Might be interesting to unify the code between both projects. WDYT?

https://github.com/OCA/l10n-spain/blob/12.0/l10n_es_facturae/models/account_invoice_integration.py

The idea is to create a new model where logs are attached.

etobella:
Maybe it is interesting to make this change on 13.0 as existent modules are not easily migrated.

alexis-via:
@etobella I don't think you can compare this module to what you do in l10n_es_facturae. I understand you need more evolved stuff for the automatic transmission to your national tax administration. But the module account_invoice_transmit_method is a small and simple module that is designed to be used by everybody.

etobella:
I know, I am not expecting to integrate facturae here, otherwise I would like to create a extended base for everyone from this module.
Let me explain, your module specifies a integration method, when many integration methods could be performed by an invoice, also, you just created a method, but you are not following its states.

For example, let's imagine three different integration methods:
* Mail: no follow-up or maybe printing of the report
* Email: follow-up if the email has been sent (to whom, issues, encrypted?)
* Chorus: follow-up states and so on

Also, transmit methods are not exclusive, a customer might ask for invoices on chorus but want to receive all the documentation by mail.

In Spain we have facturae but at least two platforms to send them that are completely different. Creating a simplified system allows to monitor them in an integrated way.

etobella:
Maybe we could do this approach:

Modify the relation of transmit method from many2one to many2many on this module and create a new module with the integration if the transmit method allows integration.

alexis-via:
Multiple transmission method seems to me a very corner case (in fact, when I send an invoice on Chorus, I also send all the attachments). So I'm very much against converting the transmit method field on invoice from M2O to M2M.
I agree that the status of the invoice could be improved. Currently, we just have a 'sent' boolean field. This is fine for several scenarios, but electronic invoicing sometimes allows for more accurate status. These statuses often depend on the method itself. I would be OK to add such a field for v14.

etobella:
Well, on our case, Spain does not allow a lot of attachments and sometimes we send a lot of information on postal mail (a box of paper each month) to some customers. So they receive information in two methods.

Also, that might be the case if some of the information is confidential. You might need to send some information using the official channel (Chours, FACe...) and the confidential data using a different channel

pedrobaeza:
I agree that multiple transmit methods is a need. I also miss an integration with current >= v12 "Send" core option.

alexis-via:
@pedrobaeza What do you mean exactly by "current >= v12 Send core option" ?

pedrobaeza:
@alexis-via I mean this:

Selección_025
Selección_026

etobella:
I made a branch on 12.0 that makes the change and tries to maintain the current fields in order to maintain the current behavior.
https://github.com/tegin/account-invoicing/tree/12.0-multi_transmit

etobella:
With the change, it will be necessary to refactor substitution rule. As it only expects one transmit method...

alexis-via:
OK. So we could take into account the transmit method of the invoice to set the default value for "is_email" and "snailmail_is_letter" (defined in snailmail_account) of the wizard "account.invoice.send". Currently, the default value of these 2 fields come from 2 fields on res.company ("invoice_is_email" and "invoice_is_snailmail"). So I guess you just want to inherit the button name="action_invoice_sent" on the form view of the invoice to add the right context keys.

pedrobaeza:
My mental concept of this is to do for each invoice what is put in transmit method. How? Linking in the transmit method the "send" action.

Example: you create a transmit method called "By mail", and link it to "Send mail" core action. Another transmit method called "Printed", and the same you link to "Print" core action.

When you click on "Send" on that set of invoices, the ones with "By mail" are sent by mail and the one "Printed" are printed. That can be achieved adding an extra option in the send dialog called "By transmit method" or similar.

What do you think?

BTW, this is a bit off-topic of this PR. Sorry for that.

etobella:
I will make the PR on 12.0 with the improvements and the idea of @pedrobaeza and we can continue the main topic of this PR. Is it right for you? @alexis-via

bodi000 (approved):

alexis-via:
Let's go with option 1, I agree that it is a better option.
I implemented the required migration scripts in v13 in this PR https://github.com/OCA/account-invoicing/pull/822
v14 will be OK from start.

PR #743, opened 2020-06-25
[IMP] account_invoice_transmit_method: Make it many2many, integrate with odoo default send

etobella:
This PR contains two commits. The first one, changes the behavior from a Many2One field to a Many2Many.
The second one integrates transmit_method with the default send method from odoo.

### DIFF
--- a/account_invoice_transmit_method/models/transmit_method.py
+++ b/account_invoice_transmit_method/models/transmit_method.py
@@ -3,6 +3,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from odoo import models, fields
+from odoo.tests.common import Form
 
 
 class TransmitMethod(models.Model):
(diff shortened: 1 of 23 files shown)


etobella:
@alexis-via This is what we were talking on #738 
@pedrobaeza

account_invoice_transmit_method/models/transmit_method.py
```diff
@@ -18,9 +19,48 @@ class TransmitMethod(models.Model):
...
+
+    def _transmit_send_invoice(self, invoice):
+        template = self.mail_template_id or self.env.ref(
+            'account.email_template_edi_invoice', False)
+        lang = self.env.context.get('lang')
+        if template and template.lang:
+            lang = template._render_template(
+                template.lang, 'account.invoice', self.id)
+        self = self.with_context(lang=lang)
+        with Form(self.env['mail.compose.message'].with_context(
```
pedrobaeza: My idea here is to use the code upstream for doing this (and avoid duplicating code - DRY principle), so can you change this to create a new send transient record with corresponding active_ids that send the mail?

etobella: @pedrobaeza I don't understand what do you want exactly :thinking: 
Do you want to open the mail.compose.message in order to change it?
I though that the main idea was to send the mail directly with intervention of the user. Also, If there are more than one transmit method, this could be a problem

pedrobaeza: I don't want to do it interactively, just reuse the existing transient model that do the dirty stuff for you (like render_template).

etobella: ok!!! I understand now, thanks :wink:

etobella: I think this is what you expected :wink:

alexis-via:
I'm still very much against converting the fields from M2O to M2M. I've read your implementation, but my opinion is that it will complexify things for all users while only a small fraction of them need to have several transmit method for the same partner/invoice. An alternative would be to add the M2M fields in another module, so that the added complexity will only be supported by users who install the additional module ; users with simple needs (the vast majority) would just use the base module.
I'll be available to discuss that during the OCA dev days 2020.

pedrobaeza:
@alexis-via it's more usual than what you think, specially due to fiscal requirements (you need to send the physical invoice here in Spain always), so restricting to only that and avoiding to have in advance the digital PDF is not something desirable for a base module.

etobella:
@alexis-via I understand your concern :thinking: 
I can make a proposal that will not change the behavior and we can change it using a `res.group`. Using the inverse we can simplify the problem. What do you think?

etobella:
@alexis-via I think the new proposal will work fine for you :smile:

PR #744, opened 2020-06-30
[13.0][ADD] account_invoice_section_sale_order

TDu:
Description from the readme :

When invoicing multiple sale orders at the same time, sale orders may be grouped
by customer into a single invoice. Unfortunately when this happens, it is hard
to know which invoice line belongs to which sale order.

This module helps by grouping invoicing lines into sections with the name of the
targeted sale order.
This is only done when an invoice targets multiple sale order.

Also include a commit to fix the `account_invoice_supplier_ref_unique` broken tests

### DIFF
--- a/account_invoice_section_sale_order/models/sale_order.py
+++ b/account_invoice_section_sale_order/models/sale_order.py
@@ -1,0 +1,54 @@
+# Copyright 2020 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo import models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    def _create_invoices(self, grouped=False, final=False):
+        """Add sections by sale order in the invoice line.
+
+        Order the invoicing lines by sale order and add lines section with
+        the sale order name.
+        Only do this for invoices targetting multiple sale order
+        """
+        invoice_ids = super()._create_invoices(grouped=grouped, final=final)
+        for invoice in invoice_ids:
+            if len(invoice.line_ids.mapped("sale_line_ids.order_id.id")) == 1:
+                continue
+            so = None
+            sequence = 10
+            section_lines = []
+            lines = invoice.line_ids.sorted(
+                key=lambda r: r.sale_line_ids.order_id.id
+            ).filtered(lambda r: not r.exclude_from_invoice_tab)
+            for line in lines:
+                if line.sale_line_ids.order_id and so != line.sale_line_ids.order_id:
+                    so = line.sale_line_ids.order_id
+                    section_lines.append(
+                        (
+                            0,
+                            0,
+                            {
+                                "name": so._get_saleorder_section_name(),
+                                "display_type": "line_section",
+                                "sequence": sequence,
+                            },
+                        )
+                    )
+                    sequence += 10
+                line.sequence = sequence
+                sequence += 10
+            invoice.line_ids = section_lines
+
+        return invoice_ids
+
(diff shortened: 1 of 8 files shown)


jgrandguillaume (approved):
LGTM, thx for this awesome cool feature !

account_invoice_section_sale_order/models/sale_order.py
```diff
@@ -0,0 +1,46 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo import models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    def _create_invoices(self, grouped=False, final=False):
+        """Add sections by sale order in the invoice line.
```
i-vyshnevska, line 11: you cannot use docstring here as it will override docstring from the original method, you can use the comment line

account_invoice_section_sale_order/models/sale_order.py
```diff
@@ -0,0 +1,46 @@
...
+                if line.sale_line_ids.order_id and so != line.sale_line_ids.order_id:
+                    so = line.sale_line_ids.order_id
+                    section_lines.append(
+                        (
+                            0,
+                            0,
+                            {
+                                "name": so.name,
+                                "display_type": "line_section",
+                                "sequence": sequence,
```
i-vyshnevska, line 37: can you use line.sequence +1 instead

account_invoice_section_sale_order/models/sale_order.py
```diff
@@ -0,0 +1,46 @@
...
+                continue
+            so = None
+            sequence = 10
+            section_lines = []
+            lines = invoice.line_ids.sorted(
+                key=lambda r: r.sale_line_ids.order_id.id
+            ).filtered(lambda r: not r.exclude_from_invoice_tab)
+            for line in lines:
+                if line.sale_line_ids.order_id and so != line.sale_line_ids.order_id:
+                    so = line.sale_line_ids.order_id
```
i-vyshnevska, line 29:
```suggestion
                so = line.sale_line_ids.order_id
                if line.sale_line_ids.order_id and so != line.sale_line_ids.order_id:
```

TDu: I don't think this will work, the if statement will never be true.

account_invoice_section_sale_order/models/sale_order.py
TDu: I used 10 as an increment as it seems to be common for `sequence` in Odoo.
Also it works well with the increment below

rafaelbn:
Nice module!

Don't miss https://github.com/OCA/account-invoicing/tree/13.0/sale_order_invoicing_grouping_criteria :smile:

account_invoice_section_sale_order/models/sale_order.py
i-vyshnevska: maybe I don't understand well the intention 'so' should hold id sale_order of the previous line? why do not group lines from scratch and just add a section line to each group

account_invoice_section_sale_order/models/sale_order.py
```diff
@@ -0,0 +1,46 @@
...
+                    section_lines.append(
+                        (
+                            0,
+                            0,
+                            {
+                                "name": so.name,
```
jbaudoux: The section should include the SO customer reference (buyer PO number)

TDu: I fixed the PR accordingly.

PR #745, opened 2020-06-30
[11.0][FIX] account_global_discount: link line taxes to discount move line

chienandalu:
cc @Tecnativa TT24526

### DIFF
--- a/account_global_discount/README.rst
+++ b/account_global_discount/README.rst
@@ -53,9 +53,10 @@ You can assign global discounts to partners as well:
 Known issues / Roadmap
 ======================
 
-* Global Discount move lines are created for a common base amount. If that
-  wasn't the case, we should split discount move lines between bases and
-  assign proper taxes accordingly.
+* Not all the taxes combination can be compatible with global discounts, as
+  the generated journal items won't be correct for taxes declarations. An error
+  is raised in that cases.
+* Currently, taxes in invoice lines are mandatory with global discounts.
 
 Bug Tracker
 ===========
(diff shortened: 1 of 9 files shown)


pedrobaeza (approved):

chienandalu:
please review @carlosdauden

carlosdauden (changes requested):
It doesn't work as expected.

Selección_053

Selección_054

pedrobaeza:
@carlosdauden please check my latest changes

chienandalu (commented):
Tested @pedrobaeza changes :+1: 
@carlosdauden Please re-review

SoniaViciana:
This PR isn't ok, at the accounting entry of discount, I miss the detail of 349 model on intra-community invoices:

image

Please, could you check it @chienandalu ?

Thanks a lot!

pedrobaeza:
@SoniaViciana this is because you need a glue module between both. It's the problem of having this anti-nature way of giving discounts. We have been forced to do it in a customer, but we DON'T RECOMMEND these global discounts as it needs a lot of patching in successive things, like this 349 side effect.

pedrobaeza:
@SoniaViciana @chienandalu I think I have found  method for not needing that glue modules in most cases. Can you please check with latest push about the 349?

pedrobaeza:
Well, not sure if we need to change 349 method:

https://github.com/OCA/l10n-spain/blob/245ff2527aabd2670c99325a056651fef5cfce3f/l10n_es_aeat_mod349/models/account_invoice.py#L22

and doing this in invoice_line_move_line_get instead. And even in that case, it would depend on the call chain...

I'm giving up now on this, as our customer only uses global discounts for national operations (and in very few cases).

chienandalu:
Yes, looks like a glue module is still needed for that case: https://github.com/OCA/l10n-spain/blob/245ff2527aabd2670c99325a056651fef5cfce3f/l10n_es_aeat_mod349/models/account_invoice.py#L35-L36

Anyway, let's move forward

PR #747, opened 2020-07-06
[11.0][ADD] account_invoice_split: new module

robinkeunen:
This module allows splitting an invoice. I took the (free) code from [here](https://apps.odoo.com/apps/modules/10.0/split_account_invoice/) and did the migration from 10.0 to 11.0 (I sent them an email to warn).

To split an invoice:
- select the lines you want to split on the invoice
- click "split invoice" button 
- a tree view is displayed with the origin invoice and the new invoice

The origin invoice is copied to create the new invoice. I then
1. unlink from origin invoice the lines marked as "split"
2. unkink from split invoice the lines not marked as "split"

Question: should I allow to. split the invoice only if it is in draft state?


account_invoice_split/models/account_invoice.py
```diff
@@ -0,0 +1,66 @@
...
+class AccountInvoiceInherit(models.Model):
+    _inherit = "account.invoice"
+
+    split_id = fields.Many2one(
+        string="Split From",
+        comodel_name="account.invoice",
+        help="INV Split From Ref:",
+    )
+
+    def btn_split_quotation(self):
```
mclaeysb: I don't understand where this button can be clicked.

account_invoice_split/models/account_invoice.py
```diff
@@ -0,0 +1,66 @@
...
+        """
+        Define function to split invoice when we click on button
+        :return: New INV view
+        """
+        self.ensure_one()
+
+        # split invoice
+        lines_to_split = [line for line in self.invoice_line_ids if line.split]
```
mclaeysb: I'm wondering how these lines will get their `split=True` value. I don't see anything that sets it - but maybe I'm overlooking.

mclaeysb (commented):
Super interesting code! Had some questions...

account_invoice_split/models/account_invoice.py
robinkeunen: in account_invoice_view.xml line 17 :-)

account_invoice_split/models/account_invoice.py
robinkeunen: It's a field on the line that is displayed on the invoice form :-)

robinkeunen:
```xml
            <xpath expr="//field[@name='invoice_line_ids']//tree/field[last()]"
                   position="before">
                <field name="split"
                       groups="account.group_show_checkbox_button"/>
            </xpath>
```

kittiu:
Interesting, but I have some a question.

* Is it ok with OCA to copy (free) code from app store, and make it into OCA? If it is declared free license there, can we just get it (notify or not), with credit to the original author of course. @dreispt do you know?

kittiu:
> Question: should I allow to. split the invoice only if it is in draft state?

I would think so, if the invoice is opened, it will break the control isn't it? :)

Q: What happen if the splitting invoice is created from sales order / purchase order. Will it have problem with those documents in term of consistency.

robinkeunen:
Screenshot 2020-07-20 at 16 18 18

Good catch, the products moved in the second invoice are not marked as invoiced in the sale order (grapes in the screenshot). I'll try to fix that.

robinkeunen:
Hi @kittiu, I only depend on account here. Is it better to add "sale" in the dependencies or should i create a glue module to update the `invoice_lines` field of `sale.order.line`?

kittiu:
> Hi @kittiu, I only depend on account here. Is it better to add "sale" in the dependencies or should i create a glue module to update the `invoice_lines` field of `sale.order.line`?

AFAIK, glue module in this repo will be good.

robinkeunen:
I added a glue module for sale to keep the link between sale order lines and invoice lines. 

The link between purchase order line and invoice line is not affected since the field purchase_line_id is set to copy=True (default):

```python
# purchase/models/account_invoice.py
class AccountInvoiceLine(models.Model):
    """ Override AccountInvoice_line to add the link to the purchase order line it is related to"""

    _inherit = "account.invoice.line"

    purchase_line_id = fields.Many2one(
        "purchase.order.line",
        "Purchase Order Line",
        ondelete="set null",
        index=True,
        readonly=True,
    )
```

apollet (approved):

HaraldPanten (changes requested):
Could you check runbot, please?

robinkeunen:
The [logs](https://runbot2.odoo-community.org/runbot/static/build/3435090-747-1c99f5/logs/job_20_test_all.txt) seem to be ok. It says "manually killed" on ["build"](https://runbot.odoo-community.org/runbot/build/3435090) page. I clicked "force rebuild", I'll see what it says.

robinkeunen:
@HaraldPanten I don't know what I can do about the runbot being killed by admin. Can you give me pointers?

dreispt:
> Is it ok with OCA to copy (free) code from app store, and make it into OCA? If it is declared free license there, can we just get it (notify or not), with credit to the original author of course. @dreispt do you know?

No, it is not OK.
We need the original author to agree and sign the CLA.
We had problems in the past from contributors not following this, and the code had to be deleted from the OCA repos.

dreispt (changes requested):
I'm sorry, we need the original authors to approve this and sign the CLA.

robinkeunen:
I'll send another email 🤞

HaraldPanten:
How could we solve Runbot issue, @dreispt ?

carmenbianca:
Looking at this PR at the moment, so bumping before it goes stale :)

PR #752, opened 2020-07-14
[FIX] account_invoice_supplier_ref_unique: Fix Test Script

kranokporn:
Test script error in account_invoice_supplier_ref_unique module.
Because in the test script, Import InvoiceTestCommon is not available in Odoo, so it must be changed to AccountTestInvoicingCommon.

cc : @kittiu


kittiu:
Same as here, https://github.com/OCA/account-invoicing/pull/754
Close this for now.

PR #753, opened 2020-07-14
[FIX] account_invoice_refund_link: Fix foreignkey exception when a credit note invoice has done from sale order.

sergio-teruel:
With this moduled installed, if you try to do an refund invoice from sale order, the method action_switch_invoice_into_refund_credit_note called from sale order removes all original lines.

see https://github.com/odoo/odoo/blob/bbbb11f3fd10bbb0f1086920f291aa9c7c2052be/addons/account/models/account_move.py#L2409

CC @Tecnativa

ping @carlosdauden @ernestotejeda

### DIFF
--- a/account_invoice_refund_link/__init__.py
+++ b/account_invoice_refund_link/__init__.py
@@ -1,4 +1,5 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from . import models
+from . import wizards
 from .hooks import post_init_hook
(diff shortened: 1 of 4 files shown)


sergio-teruel:
Tests are failing due to 
home/travis/build/OCA/account-invoicing/account_invoice_supplier_ref_unique/tests/test_account_invoice_supplier_ref_unique.py", line 8, in <module>
from odoo.addons.account.tests.invoice_test_common import InvoiceTestCommon

Fixed on https://github.com/OCA/account-invoicing/pull/754

carlosdauden (approved):

pedrobaeza:
I think the method should be the contrary: only do the traverse of the lines if a context is detected, and pass that context only when doing the refund from the wizard. That way, you avoid that other modules that uses the same method don't fail.

pedrobaeza (approved):

pedrobaeza:
Please pass pre-commit

sergio-teruel:
Done!! and installed pre-commit in repo

PR #755, opened 2020-07-14
[13.0]]ADD] Add account_invoicing_mode

TDu:
Add three modules helping on automatically invoicing customers.

The base module `account_invoice_base_invoicing_mode` does not actually
do anything but adds a selection field on partner to allow assigning an
invoicing mode to a customer.

The two other modules add specific invoicing mode.

The module `account_invoice_mode_monthly` creates monthly invoices for
customer on a specific day (configruation is in Accounting Settings)

The module `account_invoice_mode_at_shipping` creates invoices on the
shipping of the goods.

Those modules use queue_job to generate and validate invoices.


jgrandguillaume (approved):
Functional test, looks good. Thx for the job

account_invoice_mode_monthly/data/ir_cron.xml
```diff
@@ -0,0 +1,18 @@
...
+    <record forcecreate="True" id="ir_cron_generate_monthly_invoice" model="ir.cron">
+        <field name="name">Generate Monthly Invoices</field>
+        <field eval="True" name="active" />
+        <field name="user_id" ref="base.user_root" />
+        <field name="interval_number">1</field>
+        <field name="interval_type">days</field>
+        <field name="numbercall">-1</field>
+        <field eval="False" name="doall" />
+        <field name="model_id" ref="model_sale_order" />
+        <field name="code">model.cron_generate_montly_invoices()</field>
```
jbaudoux:
```suggestion
        <field name="code">model.cron_generate_monthly_invoices()</field>
```

account_invoice_mode_monthly/models/sale_order.py
```diff
@@ -0,0 +1,100 @@
...
+from odoo.osv.expression import OR
+
+from odoo.addons.queue_job.job import job
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    @api.model
+    def cron_generate_montly_invoices(self):
```
jbaudoux:
```suggestion
    def cron_generate_monthly_invoices(self):
```

account_invoice_mode_monthly/models/sale_order.py
```diff
@@ -0,0 +1,100 @@
...
+            )
+        sales = self.search(
+            [
+                ("invoice_status", "=", "to invoice"),
+                ("partner_invoice_id", "=", partner.id),
+                ("order_line.qty_to_invoice", ">", 0),
+            ]
+        )
+        # By default grouped by partner/currency. Refund are not generated
+        invoices = sales._create_invoices(grouped=partner.one_invoice_per_order)
```
jbaudoux: Do not group sales having different Payment Term or Payment Mode
It would have been nice to be able to set a specific SO as "one_invoice_per_order"

account_invoice_mode_monthly/models/sale_order.py
```diff
@@ -0,0 +1,100 @@
...
+        for invoice in invoices:
```
jbaudoux, line 68: Please also add a block for refunds where order_line.qty_to_invoice < 0 and call _create_invoice with final=True

jbaudoux (changes requested):
Some remarks (typo, refunds)

account_invoice_base_invoicing_mode/models/sale_order.py
```diff
@@ -0,0 +1,15 @@
...
+
+from odoo import fields, models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    invoicing_mode = fields.Selection(related="partner_invoice_id.invoicing_mode")
+
+    # def _create_invoices(self, grouped=False, final=False):
```
jbaudoux: To remove

account_invoice_base_invoicing_mode/models/sale_order.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2020 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo import fields, models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    invoicing_mode = fields.Selection(related="partner_invoice_id.invoicing_mode")
```
jbaudoux, line 10: Remove or set readonly

jbaudoux (changes requested):

account_invoice_mode_at_shipping/models/stock_picking.py
```diff
@@ -0,0 +1,32 @@
...
+        )
+        for picking in picking_to_invoice:
+            picking.with_delay()._invoicing_at_shipping()
+        return res
+
+    @job(default_channel="root.invoice_at_shipping")
+    def _invoicing_at_shipping(self):
+        self.ensure_one()
+        sale_order_ids = self._get_sales_order_to_invoice()
+        invoices = sale_order_ids._create_invoices()
```
jbaudoux: Can you also take care of passing grouped=partner.one_invoice_per_order in the same logic as you did above ?

account_invoice_mode_at_shipping/models/stock_picking.py
```diff
@@ -0,0 +1,32 @@
...
+        for invoice in invoices:
+            invoice.with_delay()._validate_invoice()
+        return invoices
+
+    def _get_sales_order_to_invoice(self):
+        return self.sale_id
```
jbaudoux: self.mapped("move_lines.group_id.sale_id")

jbaudoux (commented):
Some additional remarks to be compatible with the module https://github.com/OCA/stock-logistics-workflow/tree/13.0/stock_picking_group_by_partner_by_carrier

account_invoice_base_invoicing_mode/tests/test_base_invoicing_mode.py
```diff
@@ -0,0 +1,11 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+
+from odoo.tests.common import SavepointCase
+
+
+class TestBaseInvoicingMode(SavepointCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
```
mmequignon: No tests here, is it normal?

account_invoice_mode_monthly/models/sale_order.py
```diff
@@ -0,0 +1,100 @@
...
+            ("invoicing_mode_monthly_day_todo", ">", today.day),
+        ]
+        if today.day == last_day_of_this_month:
+            domain = OR(
+                [domain_last_month, domain_this_month, domain_last_day_of_month]
+            )
+        else:
+            domain = OR([domain_last_month, domain_this_month])
+
+        return self.env["res.company"].search(domain)
```
mmequignon, line 106: I feel like that method is kind of complicated for what it does.
Would it be easier to compute the next invoicing day at the end of the invoicing, and return companies where `("next_invoice_day", "=", "fields.date.today()")` ?

account_invoice_base_invoicing_mode/tests/test_base_invoicing_mode.py
TDu: There is no test because it is the base module that does not anything without a specific mode.
Maybe could be removed ?

account_invoice_mode_monthly/models/sale_order.py
TDu: The complication was added because you never know when a cron is run. A daily cron could just not have run for multiple days.

account_invoice_mode_monthly/models/sale_order.py
TDu: I have changed the domain above with `qty_to_invoice` != 0 and set the final to True. That should do the trick

account_invoice_mode_monthly/models/sale_order.py
mmequignon: Yes, then for any company with `("next_invoice_day", "<=", "field.date.today()")`, that method should be run.
You just have to change the operator. Once that method is run on a record, you can update the date on it.
Except for the special case you mentioned, you shouldn't have some `next_invoice_day` inferior to today, am I right ?

mmequignon: I probably don't have the full picture of it, though.

mmequignon: Also, it may be complicated to compute the next month given a day.

pedrobaeza:
I think this module is not correctly focused by several reasons:

- The aim is to invoice sales order, so the name and the repo are not good.
- It's overlapping `sale_automatic_workflow` flow.

IMO, you should extend that module for adding the monthly/dayly mechanism, but using the same base methods.

account_invoice_mode_monthly/data/ir_cron.xml
TDu: Fixed

account_invoice_mode_monthly/models/sale_order.py
TDu: fixed

account_invoice_mode_monthly/models/sale_order.py
TDu: @jbaudoux I changed the machinery so sales orders with different payment terms are not grouped into the same invoice.
This is done before the creation of the jobs (by the cron). And can be easily extended.
Also for the payment mode this part of an other OCA module so will not be done here

account_invoice_base_invoicing_mode/models/sale_order.py
TDu: Set readonly

account_invoice_mode_at_shipping/models/stock_picking.py
TDu: Done

account_invoice_mode_at_shipping/models/stock_picking.py
TDu: Done

account_invoice_base_invoicing_mode/models/sale_order.py
```diff
@@ -0,0 +1,12 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo import fields, models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    invoicing_mode = fields.Selection(
+        related="partner_invoice_id.invoicing_mode", readonly=True
```
simahawk:
```suggestion
        related="partner_invoice_id.invoicing_mode"
```

account_invoice_mode_at_shipping/models/stock_picking.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def action_done(self):
+        res = super().action_done()
+        picking_to_invoice = self.filtered(
+            lambda r: r.sale_id.partner_invoice_id.invoicing_mode == "at_shipping"
+            and r.picking_type_code == "outgoing"
+        )
+        for picking in picking_to_invoice:
+            picking.with_delay()._invoicing_at_shipping()
+        return res
```
simahawk: Avoid looping twice and make it more readable and customizable:
```suggestion
        for picking in picking_to_invoice:
            if picking._invoice_at_shipping():
                picking.with_delay()._invoicing_at_shipping()
        return res
        
    def _invoice_at_shipping(self):
        """Check if current record must be invoiced at shipping."""
        return (
            self.picking_type_code == "outgoing" and
            self.sale_id.partner_invoice_id.invoicing_mode == "at_shipping"
        )
```

account_invoice_mode_at_shipping/models/stock_picking.py
```diff
@@ -0,0 +1,35 @@
...
+
+    @job(default_channel="root.invoice_at_shipping")
+    def _invoicing_at_shipping(self):
+        self.ensure_one()
+        sale_order_ids = self._get_sales_order_to_invoice()
+        partner = sale_order_ids.partner_invoice_id
+        invoices = sale_order_ids._create_invoices(
+            grouped=partner.one_invoice_per_order
+        )
```
simahawk:
```suggestion
        sale_orders = self._get_sales_order_to_invoice()
        partner = sale_orders.partner_invoice_id
        invoices = sale_orders._create_invoices(
            grouped=partner.one_invoice_per_order
        )
```

account_invoice_mode_at_shipping/tests/test_invoice_mode_at_shipping.py
```diff
@@ -0,0 +1,77 @@
+# Copyright 2020 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo.tests.common import SavepointCase
+
+
+class TestInvoiceModeAtShipping(SavepointCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
```
simahawk, line 10:
```suggestion
        super().setUpClass()
        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
```
you could move this and the creation of sale order in the base class (which is not used here).
Otherwise, remove the base class.

account_invoice_mode_at_shipping/tests/test_invoice_mode_at_shipping.py
```diff
@@ -0,0 +1,77 @@
...
+        for picking in self.so1.picking_ids:
+            for line in picking.move_lines:
+                line.quantity_done = line.product_uom_qty
+                picking.action_assign()
+                picking.with_context(test_queue_job_no_delay=True).button_validate()
+        self.assertEqual(len(self.so1.invoice_ids), 1)
+        self.assertEqual(self.so1.invoice_ids.state, "posted")
+
+    def test_invoice_not_created_at_shipping(self):
+        """Check that an invoice is created when goods are shipped."""
```
simahawk:
```suggestion
        """Check that an invoice is not created when goods are shipped."""
```

account_invoice_mode_monthly/models/sale_order.py
simahawk: @TDu ?

account_invoice_mode_monthly/models/sale_order.py
```diff
@@ -0,0 +1,106 @@
...
+                ("company_id", "in", companies.ids),
+            ],
+            ["partner_invoice_id"],
+            groupby=self._get_groupby_fields_for_monthly_invoicing(),
+            lazy=False,
+        )
+        for saleorder_group in saleorder_groups:
+            saleorder_ids = self.search(saleorder_group["__domain"]).ids
+            self.with_delay()._generate_invoices_by_partner(saleorder_ids)
+        companies.write({"invoicing_mode_monthly_last_execution": datetime.now()})
```
simahawk, line 46: why don't you compute and store the next invoice day instead? Your search query will be much more simple I think, no?

simahawk (changes requested):

account_invoice_base_invoicing_mode/models/sale_order.py
TDu: Well, I just noticed that it was first declared without the readonly, then it was added later on after a review.
I guess there is some good reason to have it or not !?

simahawk: readonly is true by default on v13 so you must specify it when you don't want it.

TDu: Oh yes you are correct so it was unnecessary code :+1:

account_invoice_mode_at_shipping/models/stock_picking.py
```diff
@@ -0,0 +1,39 @@
...
+        self.ensure_one()
+        return (
+            self.picking_type_code == "outgoing"
+            and self.sale_id.partner_invoice_id.invoicing_mode == "at_shipping"
+        )
+
+    @job(default_channel="root.invoice_at_shipping")
+    def _invoicing_at_shipping(self):
+        self.ensure_one()
+        sales_order = self._get_sales_order_to_invoice()
```
i-vyshnevska: I think you may need to filter sale order here if there are invocable lines, otherwise, it will raise an error 
https://github.com/odoo/odoo/blob/be6b59fb3bd52894524603ee81597d8ec289583a/addons/sale/models/sale.py#L629

account_invoice_mode_at_shipping/models/stock_picking.py
TDu: This has been changed

account_invoice_mode_at_shipping/models/stock_picking.py
TDu: This had been resolved

account_invoice_mode_at_shipping/tests/test_invoice_mode_at_shipping.py
TDu: I removed the base class

account_invoice_mode_at_shipping/models/stock_picking.py
TDu: I have added a check for it, thanks.

account_invoice_mode_at_shipping/models/stock_picking.py
```diff
@@ -0,0 +1,51 @@
...
+            and self.sale_id.partner_invoice_id.invoicing_mode == "at_shipping"
+        )
+
+    @job(default_channel="root.invoice_at_shipping")
+    def _invoicing_at_shipping(self):
+        self.ensure_one()
+        sales = self.env["sale.order"].browse()
+        # Filter out non invoicable sales order
+        for sale in self._get_sales_order_to_invoice():
+            if sale._get_invoiceable_lines():
```
jbaudoux, line 34: @TDu This should be part of the above method `_get_sales_order_to_invoice` that has to return only orders having invoiceable lines

account_invoice_base_invoicing_mode/tests/test_base_invoicing_mode.py
TDu: It has been removed

account_invoice_mode_at_shipping/models/stock_picking.py
TDu: done

account_invoice_mode_monthly/models/sale_order.py
TDu: The query may look complicated but it is covered by tests.
I am not convince that storing the next execution date would make things simpler. One of the reason is the day of invoicing in the month can be set in the configuration.
So now when changing that day the next invoicing day as too be computed but depending if the invoicing day as been done for the current month or not the computation would be different. (This is only one complication I can think of now)

TDu:
In previous comments there has been a suggestion by @pedrobaeza to include this into `sale_automatic_workflow` because the use case is overlapping  with it.
I don't think so myself, the monthly invoice mode is based on a daily cron, the option does not exist in `sale_automatic_workflow` and the at shipping invoicing is triggered by some specific code in the action_done of the stock_picking, which would not be easily implemented on top of `sale_automatic_workflow`.

Moreover the proposed change would also make this pr much more complex and maybe its UI element less intuitive to use.

And I personally think that this pr gives a good base to add different invoicing mode in the future.

account_invoice_mode_monthly/models/sale_order.py
TDu: I think this has been answered by this comment https://github.com/OCA/account-invoicing/pull/755#discussion_r625023369

pedrobaeza:
I don't remember right now my suggestion and the context, so I can't give you now an answer, but if you point me to it, I can agree or not on this.

TDu:
@pedrobaeza It is about this comment https://github.com/OCA/account-invoicing/pull/755#issuecomment-662846222 on the same PR.
We would like to get this merge and I felt your proposition still deserved an answer, yep that is one year ago.

sebalix (approved):
LG, a rebase to keep one commit / module is welcome

pedrobaeza:
OK, as you prefer then. I still encounter some common points:

- `sale_automatic_workflow` also has a cron that performs the task of invoicing.
- It contains the code to invoice.

What about the naming, being `sale_invoicing_mode` or similar?

TDu:
> OK, as you prefer then. I still encounter some common points:
> 
>     * `sale_automatic_workflow` also has a cron that performs the task of invoicing.
> 
>     * It contains the code to invoice.
> 
> 
> What about the naming, being `sale_invoicing_mode` or similar?

Well the base module is just a field on the partner really, so it is about invoicing first. So it could be used to invoice other things than a sale order if the business model requires that.

I would really like to keep the naming like it is.

pedrobaeza:
What about partner_invoicing_mode then? If not, I won't block it.

TDu:
> What about partner_invoicing_mode then? If not, I won't block it.

I actually like that name better, shorter and more explicit.
Maybe it could be renamed during the 14.0 migration

TDu:
@simahawk I squashed the commit and added the roadmap as discussed

simahawk:
@TDu check the build pls

simahawk (approved):
@jbaudoux @mmequignon good for you?

account_invoice_base_invoicing_mode/models/account_invoice.py
```diff
@@ -0,0 +1,15 @@
...
+
+from odoo import models
+
+from odoo.addons.queue_job.job import job, related_action
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    @job(default_channel="root.invoice_validation")
```
mmequignon, line 12: Isn't the @job decorator deprecated?
https://github.com/OCA/queue/pull/274

TDu: Yes it is, but it still works for v13.
It will be changed on the v14 migration.

mmequignon (approved):

jbaudoux (approved):

simahawk:
WTH?

image

sbidoul:
If you click on the link in the error message, you'll see that travis is indeed red (a transient GitHub error). It is not the same commit as the one just above for a good reason I don't remember right now.

image

simahawk:
let's try again 
/ocabot merge patch

PR #754, opened 2020-07-14
[13.0][FIX] account_invoice_supplier_ref_unique: fix tests

sergio-teruel:
cc @Tecnativa

ping @carlosdauden , I not sure by generic template

PR #757, opened 2020-07-15
[13.0][MIG] account_invoice_pricelist: Migration to 13.0

manuelregidor:
Migration of module with some new features:
- Discounts based on pricelists are shown on invoice lines.
- Multicurrency is supported and prices get automatically converted.


manuelregidor:
@HaraldPanten @ValentinVinagre Can you check it out, please?

manuelregidor:
@pedrobaeza, the module has not passed the Travis test and it seems that there is a problem with the module 'account_invoice_supplier_ref_unique'. Do you know if there is something wrong with that module? My module's tests have worked fine locally. Thank you.

HaraldPanten (approved):
It's functionally working fine. 👍

kittiu:
https://github.com/OCA/account-invoicing/pull/754 merged, you may rebase.

manuelregidor:
> #754 merged, you may rebase.

Done.

manuelregidor:
@kittiu could you review, please?

kittiu:
@manuelregidor I am looking through the commit. Correct me if I am wrong, but I think for migration, we will have 2 commits.

1. black, isort, prettier: this should not have any change in code, not even the version number yet, just the formatting. For example, here, I see file deletion, and change of some code, i.e., account.invoice -> account.move...etc.
https://github.com/OCA/account-invoicing/pull/757/commits/37444b84b8aec7459015a17e6cf93f939f8a4736
2. migration: this we change codes. And so, we expect to see the diff.

Also, since you have new code (which is great), can you add test too? :)

Can you check please? Thank you!

manuelregidor:
> @manuelregidor I am looking through the commit. Correct me if I am wrong, but I think for migration, we will have 2 commits.
> 
> 1. black, isort, prettier: this should not have any change in code, not even the version number yet, just the formatting. For example, here, I see file deletion, and change of some code, i.e., account.invoice -> account.move...etc.
>    [37444b8](https://github.com/OCA/account-invoicing/commit/37444b84b8aec7459015a17e6cf93f939f8a4736)
> 2. migration: this we change codes. And so, we expect to see the diff.
> 
> Also, since you have new code (which is great), can you add test too? :)
> 
> Can you check please? Thank you!

Hi @kittiu 

Yes, you're right about the issues with the commits. They've been sorted.

As regards the tests, I added 7 new tests to check the added functionalities. Please, let me know if you find functionalities or parts of the code that need to be tested.

Thank you

account_invoice_pricelist/README.rst
```diff
@@ -0,0 +1,90 @@
+===============================
```
kittiu, line 1: You can remove this file. ocabot will auto create it for v13 when merged.

account_invoice_pricelist/__manifest__.py
```diff
@@ -2,11 +2,11 @@
 
 {
     "name": "Account - Pricelist on Invoices",
-    "version": "12.0.1.0.3",
+    "version": "13.0.1.0.0",
     "summary": "Add partner pricelist on invoices",
     "category": "Accounting & Finance",
     "author": "GRAP," "Therp BV," "Tecnativa," "Odoo Community Association (OCA)",
-    "website": "https://github.com/OCA/account-invoicing",
+    "Sygel" "website": "https://github.com/OCA/account-invoicing",
```
kittiu:
```suggestion
    "website": "https://github.com/OCA/account-invoicing",
```

account_invoice_pricelist/models/account_move.py
```diff
@@ -0,0 +1,238 @@
...
+                self.partner_id
+                and self.type in ("out_invoice", "out_refund")
+                and self.partner_id.property_product_pricelist
+            ):
+                self.pricelist_id = self.partner_id.property_product_pricelist
+                self._calculate_currency()
+            return result
+
+    @api.onchange("pricelist_id")
+    def _calculate_currency(self):
```
kittiu:
```suggestion
    def _set_pricelist_currency(self):
```
Non-blocking. IMHO, to understand what the code does. WDYT?

account_invoice_pricelist/models/account_move.py
```diff
@@ -0,0 +1,238 @@
...
+    pricelist_id = fields.Many2one(
+        comodel_name="product.pricelist",
+        string="Pricelist",
+        readonly=True,
+        states={"draft": [("readonly", False)]},
+    )
+
+    @api.constrains("pricelist_id", "currency_id")
+    def _check_currency(self):
+        for sel in self.filtered(lambda a: a.pricelist_id and a.is_invoice()):
```
kittiu, line 19:
```suggestion
        for sel in self.filtered("pricelist_id"):
```
IMHO, what you did is making sense, even for non-invoice. Even for Journal Entry, we won't show this pricelist. But when it does, we should block it too. WDYT?

account_invoice_pricelist/models/account_move.py
```diff
@@ -0,0 +1,238 @@
...
+                and self.partner_id.property_product_pricelist
+            ):
+                self.pricelist_id = self.partner_id.property_product_pricelist
+                self._calculate_currency()
+            return result
+
+    @api.onchange("pricelist_id")
+    def _calculate_currency(self):
+        if (
+            self.is_invoice()
```
kittiu, line 41:
```suggestion
```
Non-blocking, same as above, so I think we don't need to check is_invoice. WHYT?

account_invoice_pricelist/README.rst
manuelregidor: Done

account_invoice_pricelist/__manifest__.py
manuelregidor: Done

account_invoice_pricelist/models/account_move.py
manuelregidor: Done

account_invoice_pricelist/models/account_move.py
```diff
@@ -0,0 +1,238 @@
...
+    def _onchange_product_id_account_invoice_pricelist(self):
+        # invoice_id ha pasado a llamarse move_id
+        if not self.move_id.pricelist_id or not self.move_id.partner_id:
+            return
+        self.with_context(check_move_validity=False).update(
+            {"price_unit": self._get_price_with_pricelist()}
+        )
+
+    @api.onchange("product_uom_id")
+    def _onchange_uom_id(self):
```
kittiu, line 79: Will this method collide with the above `_onchange_product_id_account_invoice_pricelist` for "product_uom_id" ?
Or why can't we include logic here in _get_price_with_pricelist?

kittiu:
@manuelregidor I have seen your effort to complete things that was missing, and want to thank you about it.
However, I think that the code is quite long (I saw that you try to get the logic from, i.e., sales module). Have you tried to optimized/shorten the code?
If you finished it, or is certain that it has to be. Please let me know, I will review it deeper.
Thanks again!

account_invoice_pricelist/models/account_move.py
manuelregidor: I would leave it as it is an extra check, but let me know if you think it's better the way you've suggested and I'll apply the changes.

account_invoice_pricelist/models/account_move.py
manuelregidor: Same as above.

account_invoice_pricelist/models/account_move.py
manuelregidor: Done. I had forgotten to delete "product_uom_id" from _onchange_product_id_account_invoice_pricelist onchange decorator.

manuelregidor:
> @manuelregidor I have seen your effort to complete things that was missing, and want to thank you about it.
> However, I think that the code is quite long (I saw that you try to get the logic from, i.e., sales module). Have you tried to optimized/shorten the code?
> If you finished it, or is certain that it has to be. Please let me know, I will review it deeper.
> Thanks again!

@kittiu I can include that in the roadmap, if that's ok, so it can be done in the future as an improvement. Thank you.

manuelregidor:
@kittiu Is it OK if I include the possible optimization in the roadmap? Thank you

kittiu:
> @kittiu Is it OK if I include the possible optimization in the roadmap? Thank you

@manuelregidor I can't judge this. @pedrobaeza what do you think?

pedrobaeza:
Can you summarize the situation?

kittiu:
> Can you summarize the situation?

https://github.com/OCA/account-invoicing/pull/757#issuecomment-665721922

The functional test works fine. But I think the code can be more...revised. And the contributor like it to be on roadmap.

Note: IMHO, that OCA wants to keep the higher level of quality. So I would suggest it to be revised than roadmap.

pedrobaeza:
Well, this PR is a migration, so any improvement outside the migration of the module that is not of the like of the reviewers should be removed from the PR and discussed isolately, as if not, migration will be blocked forever or forced to accept that improvements with a future promise.

HaraldPanten:
IMO I think that this PR could be approved and we could include that code refactoring/improvement in the Roadmap.

After this, we/you could open a new issue to discuss which code improvements should be done.

pedrobaeza:
I think the problem is that this PR already includes the improvements, but they are debatable due to the optimization. Did I understand correctly the problem?

kittiu:
> be blocked forever

I am more confused with your statement, may be my reading skill is not very well :p

What I understand from the statement, as rules of thumbs. As MIG, this PR should be just migration without any improvement yet. Which means the improvement that make this PR more complex than the original should be in anohter PR, those are,
```
* Discounts based on pricelists are shown on invoice lines.
* Multicurrency is supported and prices get automatically converted.
```
The improvement can be on another PR. This will ensure this PR can pass easily.

kittiu:
@HaraldPanten, @manuelregidor we all welcome the improvements and appreciate that. I am trying to learn what is best way to keep up with OCA standard. :)

kittiu:
> I think the problem is that this PR already includes the improvements, but they are debatable due to the optimization. Did I understand correctly the problem?

Yep

pedrobaeza:
There's no OCA standard here, as we can't fine grain that way, but I'm understanding that you, @kittiu, don't feel comfortable with the improvements done here outside the migration itself, so I think that it's better to limit this PR to the pure migration, and let that improvements for a future PR.

kittiu:
So, my comment here is to do a pure [MIG] first? And I think it can get passed quickly, then do the enhancement PR, with or without the optimization, then we can continue from there.

WDYT?

pedrobaeza:
Yes, that's what I'm suggesting, although I would understand than the original contributor can be a bit annoyed for this extra work.

HaraldPanten:
> Yes, that's what I'm suggesting, although I would understand than the original contributor can be a bit annoyed for this extra work.

If we have to apply this changes, for sure this will delay everything.

We took in consideration how sale pricelists work for sales orders, to make a "similar" functionality for invoices. @kittiu which changes do you think that should be applied to improve this PR?  Maybe this will be easier than leaving a "simple migration" PR here and open a new [IMP] PR for separate.

What do you think?

kittiu:
> > Yes, that's what I'm suggesting, although I would understand than the original contributor can be a bit annoyed for this extra work.
> 
> If we have to apply this changes, for sure this will delay everything.
> 
> We took in consideration how sale pricelists work for sales orders, to make a "similar" functionality for invoices. @kittiu which changes do you think that should be applied to improve this PR? Maybe this will be easier than leaving a "simple migration" PR here and open a new [IMP] PR for separate.
> 
> What do you think?

I still think it is better to make separated PR. Another important point is, once the [ENH] PR is merged. It will be easier for others to help back port it to v12, v11, ...

This is just my opinion. But let's see what other people think.

pedrobaeza:
If not in a separate PR, at least in a separate commit. See for example what I have done for contract: https://github.com/OCA/contract/pull/533

HaraldPanten:
Thank you for your comments.

@kittiu , We'll make 2 separate commits, each one with it's description.

I think that this will make the process clear enough. In the other hand, the improvement is a compatibility with multi-currency, as pricelists in sales orders do.

Do you think that this will be close enough to OCA standard?

kittiu:
Please, that will be good.

HaraldPanten:
> Please, that will be good.

Sure,

We'll do that in the next few days 👍

account_invoice_pricelist/models/account_move.py
```diff
@@ -0,0 +1,243 @@
...
+        for sel in self.filtered(lambda a: a.pricelist_id and a.is_invoice()):
+            if sel.pricelist_id.currency_id != sel.currency_id:
+                raise UserError(
+                    _("Pricelist and Invoice need to use the same currency.")
+                )
+
+    @api.onchange("partner_id", "company_id")
+    def _onchange_partner_id_account_invoice_pricelist(self):
+        if self.is_invoice():
+            result = super(AccountMove, self)._onchange_partner_id()
```
bealdav, line 28: Hi thanks for this contrib. No need Class name and self if it's the same than current class in python3 ;-)

manuelregidor:
Hi, @kittiu There are now two commits: one with the migration and another one with the improvement. Could you have a look? Thank you.

account_invoice_pricelist/models/account_move.py
kittiu: Agree with this comment, in many places, you can just use `super()._xxx`

kittiu (approved):
Thank you for taking your time to make separated commits.

HaraldPanten:
> Thank you for taking your time to make separated commits.

Hi @kittiu thank you for your reviews.

Could we merge this one, then?

PR #759, opened 2020-07-16
[12.0][ADD] New module account_invoice_auto_send_mail

victoralmau:
Allows you to automate the sending of customer sales invoices and rectifiers to send by email.

It will be necessary that at the level of each newspaper (related to type = sales) we define the fields:
- Email template: Email template to be used to send each of the invoices
- Author: If you want to establish it, it will replace the user_id of the invoice
- Invoice days: Defines the number of days that must pass between the invoice and the current date for them to be sent.


account_invoice_auto_send_mail/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
```
kittiu, line 1: @victoralmau thank you for your contribution!

For the __manifest__.py, you can make it more complete if you want, ie.,
https://github.com/OCA/mis-builder/blob/12.0/mis_builder/__manifest__.py

kittiu:
I also see that you are using your branch 12.0 to do this PR.
Personally, I would think it will be better if you branch it out from the origin/12.0 (OCA), i.e., to 12.0-add-account_invoice_auto_send_mail
The branch name will have more meaning too. WDYT?

kittiu:
Please see the red travis, I think it require some attention.

kittiu:
Normally, PR should have 1 commit only. You may do many commits, but at the end, please squash them into one.
Or otherwise, what I like is to use
> git commit --amend

account_invoice_auto_send_mail/README.rst
```diff
@@ -0,0 +1 @@
+Allows you to automate the sending of customer sales invoices and rectifiers to send by email
```
kittiu: You actually don't need README.rst, just make sure the content in "readme" folder is right. When this PR is merged, README.rst will auto generated for you.
A good example of "readme" https://github.com/OCA/mis-builder/tree/12.0/mis_builder/readme

account_invoice_auto_send_mail/readme/USAGE.rst
```diff
@@ -0,0 +1,4 @@
+It will be necessary that at the level of each newspaper (related to type = sales) we define the fields:
```
kittiu: Frankly, I do not quite understand the explaination here. (i.e., what is "newspaper").
Can you try review it again? Thank you.

account_invoice_auto_send_mail/models/account_journal.py
```diff
@@ -0,0 +1,20 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class AccountJournal(models.Model):
+    _inherit = 'account.journal'
+
+    invoice_mail_template_id = fields.Many2one(
```
kittiu, line 9: Hello, we we really need to have the email template per Journal? Is this the need?
I would prefer if we could minimize the complexity of account.journal, which is an important object, just for this feature.

Possible to put it in Configuration window instaead? Or if there is really need to be at journal level, may be a separated table to keep the info?

For this point, not sure if there is a common practice @OCA/accounting-maintainers

account_invoice_auto_send_mail/models/account_invoice.py
```diff
@@ -0,0 +1,88 @@
...
+            else:
+                if days_difference >= self.journal_id.invoice_mail_days:
+                    send_invoice = True
+            # send_invoice
+            if send_invoice:
+                self.account_invoice_auto_send_mail_item_real(self.journal_id.invoice_mail_template_id, self.journal_id.invoice_mail_template_id_author_id.partner_id)
+
+    @api.model
+    def cron_account_invoice_auto_send_mail(self):
+        account_invoice_ids = self.env['account.invoice'].search(
```
kittiu:
```suggestion
        invoices = self.env['account.invoice'].search(
```
since the result is recordset, I would prefer no `_ids`

account_invoice_auto_send_mail/models/account_invoice.py
```diff
@@ -0,0 +1,88 @@
...
+            [
+                ('state', 'in', ('open', 'paid')),
+                ('type', 'in', ('out_invoice', 'out_refund')),
+                ('journal_id.invoice_mail_template_id', '!=', False),
+                ('date_invoice_send_mail', '=', False)
+             ], order="date_invoice asc", limit=200
+        )
+        if len(account_invoice_ids) > 0:
```
kittiu:
```suggestion
        if account_invoice_ids:
```

account_invoice_auto_send_mail/models/account_invoice.py
```diff
@@ -0,0 +1,88 @@
...
+            count = 0
+            for account_invoice_id in account_invoice_ids:
+                count += 1
+                # cron_account_invoice_auto_send_mail_item
+                account_invoice_id.cron_account_invoice_auto_send_mail_item()
+                # logger_percent
+                percent = (float(count) / float(len(account_invoice_ids))) * 100
+                percent = "{0:.2f}".format(percent)
+                _logger.info(str(percent) + '% (' + str(count) + '/' + str(len(account_invoice_ids)) + ')')
```
kittiu: Please use only proper string injection where possible., I.e., '%s/%s' % (x, y), and not x + '/' + y

account_invoice_auto_send_mail/models/account_journal.py
```diff
@@ -0,0 +1,20 @@
...
+        comodel_name='mail.template',
+        domain=[('model_id.model', '=', 'account.invoice')],
+        string='Invoice mail template id'
+    )
+    invoice_mail_template_id_author_id = fields.Many2one(
```
kittiu: Non-blocking, this is a little long variable name, isn't it? Can be something more simpler? ;)

account_invoice_auto_send_mail/models/account_invoice.py
```diff
@@ -0,0 +1,88 @@
...
+            mail_compose_message_obj_vals['attachment_ids'] = return_onchange_template_id['value']['attachment_ids']
+        # update
+        mail_compose_message_obj.update(mail_compose_message_obj_vals)
+        # send_mail_action
+        mail_compose_message_obj.send_mail_action()
+        # other
+        self.date_invoice_send_mail = datetime.today()
+
+    @api.one
+    def cron_account_invoice_auto_send_mail_item(self):
```
kittiu:
```suggestion
    def cron_account_invoice_auto_send_mail_item(self):
        self.ensure_one()
```
Try not to use @api.con, it is deprecated (in v13), and not recommended to use.

account_invoice_auto_send_mail/models/account_invoice.py
```diff
@@ -0,0 +1,88 @@
...
+        if self.type in ['out_invoice', 'out_refund'] and self.date_invoice_send_mail is False and self.state in ['open', 'paid']:
+            current_date = fields.Datetime.from_string(str(datetime.today().strftime("%Y-%m-%d")))
```
kittiu: If you want to get today, you can use `fields.Date.context_today(self)`.
Please double check if it suitable for you.

account_invoice_auto_send_mail/models/account_invoice.py
```diff
@@ -0,0 +1,88 @@
...
+        # update
+        mail_compose_message_obj.update(mail_compose_message_obj_vals)
+        # send_mail_action
+        mail_compose_message_obj.send_mail_action()
+        # other
+        self.date_invoice_send_mail = datetime.today()
+
+    @api.one
+    def cron_account_invoice_auto_send_mail_item(self):
+        if self.type in ['out_invoice', 'out_refund'] and self.date_invoice_send_mail is False and self.state in ['open', 'paid']:
```
kittiu:
```suggestion
        if self.type in ['out_invoice', 'out_refund'] and not self.date_invoice_send_mail and self.state in ['open', 'paid']:
```
Are we saying this feature is only for customer invoices? Why not make it generic for all?

account_invoice_auto_send_mail/models/account_invoice.py
```diff
@@ -0,0 +1,88 @@
...
+class AccountInvoice(models.Model):
+    _inherit = 'account.invoice'
+
+    date_invoice_send_mail = fields.Datetime(
+        string='Date invoice send mail'
+    )
+
+    @api.one
+    def account_invoice_auto_send_mail_item_real(self, mail_template_id, author_id):
+        _logger.info('Operations account_invoice_auto_send_mail_item_real invoice ' + str(self.id))
```
kittiu: Please use only proper string injection where possible., I.e., '%s/%s' % (x, y), and not x + '/' + y

account_invoice_auto_send_mail/models/account_invoice.py
```diff
@@ -0,0 +1,88 @@
...
+
+class AccountInvoice(models.Model):
+    _inherit = 'account.invoice'
+
+    date_invoice_send_mail = fields.Datetime(
+        string='Date invoice send mail'
+    )
+
+    @api.one
+    def account_invoice_auto_send_mail_item_real(self, mail_template_id, author_id):
```
kittiu:
```suggestion
    def account_invoice_auto_send_mail_item_real(self, mail_template_id, author_id):
        self.ensure_one()
```

account_invoice_auto_send_mail/models/account_invoice.py
```diff
@@ -0,0 +1,88 @@
...
+        _logger.info('Operations account_invoice_auto_send_mail_item_real invoice ' + str(self.id))
+
+        mail_compose_message_vals = {
+            'author_id': self.user_id.partner_id.id,
+            'record_name': self.number,
+        }
+        # Author_id (journal_id)
+        if author_id.id > 0:
```
kittiu:
```suggestion
        if author:
```
I would prefer to name the variable, with `_id`, `_ids`, it it keeps number. And without `_id`, `_ids` for recordsets.

account_invoice_auto_send_mail/models/account_invoice.py
```diff
@@ -0,0 +1,88 @@
...
+            mail_compose_message_vals['author_id'] = author_id.id
+
+        mail_compose_message_obj = self.env['mail.compose.message'].with_context().sudo().create(mail_compose_message_vals)
```
kittiu:
```suggestion
        mail_compose_message_obj = self.env['mail.compose.message'].sudo().create(mail_compose_message_vals)
```
No need with_context() if we don't add anything.

account_invoice_auto_send_mail/models/account_invoice.py
```diff
@@ -0,0 +1,88 @@
...
+        return_onchange_template_id = mail_compose_message_obj.onchange_template_id(mail_template_id.id, 'comment', 'account.invoice', self.id)
```
kittiu: I would prefer some shorter variable name, i..e, res = {...}

kittiu:
Please also try to make tests scripts too. There are bunch of samples you can find here, it is quite easy. Just copy from others :p.
Module with tests are likely to get migrated easier in the future too.

All in all, these are my initial comments, others will have better comment. Thank you for joining force OCA.

kittiu:
@OCA/accounting-maintainers I also have questions. Email is quite generic, are there other similar modules?

victoralmau:
> I also see that you are using your branch 12.0 to do this PR.
> Personally, I would think it will be better if you branch it out from the origin/12.0 (OCA), i.e., to 12.0-add-account_invoice_auto_send_mail
> The branch name will have more meaning too. WDYT?

I think it would have been a very good idea, I take note for the following contributions.

victoralmau:
> Normally, PR should have 1 commit only. You may do many commits, but at the end, please squash them into one.
> Or otherwise, what I like is to use
> 
> > git commit --amend

Last week I have uploaded several addons for the first time and I was not very clear about the process, many times the commits refer to solving the errors indicated by travis

account_invoice_auto_send_mail/__manifest__.py
victoralmau: Thanks, I save it for future addons

account_invoice_auto_send_mail/README.rst
victoralmau: Ok i delete the file

account_invoice_auto_send_mail/readme/USAGE.rst
victoralmau: All the addons had them developed in Spanish and on some occasions I have done the translations automatically to go faster, but in this case it is a big mistake, sorry.

account_invoice_auto_send_mail/models/account_journal.py
victoralmau: Ok, i change it to invoice_mail_author_id

account_invoice_auto_send_mail/models/account_journal.py
victoralmau: Hello, several weeks ago I had the addon developed regarding ir.config.parameter but it did not allow configuring which invoices you wanted to send and / or if we wanted to send different emails (to solve this "problem" I thought that defining the fields at account_journal level would be good idea).

victoralmau:
> Please also try to make tests scripts too. There are bunch of samples you can find here, it is quite easy. Just copy from others :p.
> Module with tests are likely to get migrated easier in the future too.
> 
> All in all, these are my initial comments, others will have better comment. Thank you for joining force OCA.

Thank you very much for all your comments, I have made all the suggested changes.
I am going to check other addons to try to add tests since it is something I have never done.

victoralmau:
https://github.com/OCA/account-invoicing/pull/768

PR #770, opened 2020-08-05
[13.0] [MIG] account_invoice_mass_sending

misern2:
Standard migration


juliap98 (commented):
Remains on pending and doesn't send any invoice.

misern2:
@juliap98  I think it's because a misconfiguration of the queue job on your local environement

jguenat:
Hello @misern2 ,
I opened an PR for migration to 14.0, we could join force for review :) 
Do you want me to review this PR or maybe we close one of the 2 and then port it ?
What do you think is the best?

PR #772, opened 2020-08-13
[12.0][FIX] account.journal type values in wizard's constant

renatonlima:
Fix account.journal.type selection values in wizard, in Odoo 12.0 version there aren't the values 'sale_refund', 'purchase_refund', https://github.com/odoo/odoo/blob/12.0/addons/account/models/account.py#L423


marcelsavegnago (approved):
I really wondered about those types of journals that I haven't seen anywhere. I believe it must be something from previous versions.

rvalyi:
> I really wondered about those types of diaries that I haven't seen anywhere. I believe it must be something from previous versions.

Yes, previous versions https://github.com/odoo/odoo/blob/8.0/addons/account/account_invoice.py#L35

rvalyi (approved):

mbcosta (approved):

PR #777, opened 2020-08-26
Review the module description to be more precise

alexandreduf:
It's not clear what that module does and how it does it. Is it possible to give more information?
Thanks!


legalsylvain:
same comment as here : https://github.com/OCA/web/pull/1670

sergiocorato:
I think it is better to open an issue and comment there, so someone can create a more useful PR.

alexandreduf:
> I think it is better to open an issue and comment there, so someone can create a more useful PR.

@legalsylvain 
Thank you very much. I didn't know how to give my input on this module. I will open an issue!
Regards,

PR #778, opened 2020-08-27
[11.0][FIX] Queue Job import now inside a try/except block

n17:
- Fixes an issue where `ImportError: No module named 'queue_job'` appears even if the `account_invoice_validation_queued` module is not installed.
- A similar approach is used in this same module here: https://github.com/OCA/account-invoicing/blob/11.0/account_invoice_validation_queued/models/account_invoice.py#L6


bodi000 (approved):

n17:
@pedrobaeza Since you're the maintainer at least for the `14.0` branch, do you think this could be merged instead of closing the PR?

pedrobaeza:
The branch CI seems too old. It should be changed to GitHub actions I'm afraid.

PR #779, opened 2020-08-31
[FIX] None as a possible value for _state_to?

flachica:
I don't understand why None is needed as a value for state_to. I have tested the module without that value and it works.


kittiu:
Looks like this was fixed the same here - https://github.com/OCA/account-invoicing/pull/789
May be this one can closed?

AaronHForgeFlow:
Actually this PR was created before. However it is best if the title of the PR has the Odoo version and the module name, otherwise it is very complex to search.

AaronHForgeFlow:
I think this one was created before and should be merged but can you change the commit message to something more clear? something like

```
[FIX] account_move_tier_validation
* incorrect _state_to attribute causing RecursionError
```

PR #783, opened 2020-09-07
[13.0][IMP] account_invoice_validation_queued: Avoid m2m table collapse

pedrobaeza:
Having other m2m pointing to queue.job will make them to share the same table and get akward results.

We avoid it specifying a table. It includes migration script for moving data from one table to the other.

@Tecnativa TT22079

### DIFF
--- a/account_invoice_validation_queued/migrations/13.0.2.0.0/post-migration.py
+++ b/account_invoice_validation_queued/migrations/13.0.2.0.0/post-migration.py
@@ -1,0 +1,33 @@
+# Copyright 2020 Tecnativa - Pedro M. Baeza
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    if not openupgrade.table_exists(env.cr, "account_move_queue_job_rel"):
+        return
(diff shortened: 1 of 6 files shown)


account_invoice_validation_queued/migrations/13.0.2.0.0/post-migration.py
```diff
@@ -0,0 +1,33 @@
...
+        FROM account_move_queue_job_rel rel
+        JOIN queue_job qj ON qj.id = rel.job_id
+        WHERE qj.method_name = 'action_invoice_open_job'
+        """,
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        DELETE FROM account_move_queue_job_rel
+        WHERE id IN (
```
sergio-teruel: Many2many table relations has not column id...
Maybe... 
WHERE job_id IN (
            SELECT qj.id....

sergio-teruel (changes requested):

account_invoice_validation_queued/migrations/13.0.2.0.0/post-migration.py
pedrobaeza: Yes, totally. Fixed.

carlosdauden (approved):

sergio-teruel (approved):

PR #784, opened 2020-09-08
[10.0 ][ADD] sale_stock_restocking_fee_invoicing: New addon to charge r…

lmignon:
…eturn fees in case of return of goods by a customer.


sale_stock_picking_return_fees_invoicing/models/stock_move.py
```diff
@@ -0,0 +1,38 @@
...
+    )
+
+    @api.multi
+    def action_done(self):
+        result = super(StockMove, self).action_done()
+        move_ids_to_charge_by_so = defaultdict(list)
+        for move in self:
+            if not move.charge_customer_stock_return:
+                continue
+            if not move.origin_returned_move_id:
```
jbaudoux: What happen if you make a return of a return? Shouldn't you check for is_customer_return ?

sale_stock_picking_return_fees_invoicing/README.rst
```diff
@@ -0,0 +1,102 @@
...
+* Specify on the customer contact form that fees must be charged by default
+  in case of return of goods. This value is used as default value into the
+  stock return wizard;
+* Use the checkboxes into the stock return wizard.
+
+For each returned product, if fees must be charged, once the return picking
+is done, a new line is added into the origin sale order to charge fees. The
+quantity charged is always 1 whatever the number of items of the same product
+returned.
+
```
jbaudoux: Worth adding that in your implementation 1 fee line is added per returned SO line. And not 1 fee line per return.

sale_stock_picking_return_fees_invoicing/models/sale_order.py
```diff
@@ -0,0 +1,57 @@
...
+        product_id = self.company_id.customer_return_fees_product_id
+        if not product_id:
+            raise UserError(
+                _(
+                    "No product configured for customer return fees. "
+                    "Please fix the configuration into stock settings or "
+                    "contact you administrator."
+                )
+            )
+        description = _("Return fees for %s %s %s") % (
```
jbaudoux: This should be in the language of the customer

jbaudoux (commented):
Some question/remarks

sale_stock_picking_return_fees_invoicing/models/stock_move.py
lmignon: `charge_customer_stock_return` will never be True in this case. We don't need to put in place a double check

sale_stock_picking_return_fees_invoicing/README.rst
lmignon: No, we add a line by return not by so line. If you purchase 2 product and return the product 1 item at a time (1 pickings) you'll be charged twice

sale_stock_picking_return_fees_invoicing/models/sale_order.py
lmignon: indeed

pedrobaeza:
Please avoid using plural in module name as stated in the contribution guidelines: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#modules

lmignon:
@pedrobaeza I know but in this case AFAIK the meaning of the plural form is not the same as the singular form.
fee =~ cost
fees =~ charge 
IMO the plural form is the one appropriate to describe the functionality. Perhaps other people with a better level of english can validate/invalidate the correct use of the term fees.

pedrobaeza:
OK, I have no other better suggestion for now.

lmignon:
@pedrobaeza I agree with the remark and I am ready to change if my understanding is wrong or if another term is more appropriate. Finding a good module name is something I always find difficult. For a while I had also thought about `customer_return_fees_invoicing`.

pedrobaeza:
imagen

:rofl:

gdgellatly:
Maybe use the word restocking somwhere. In the English speaking world that is what we call them.

A Restocking Fee by the way, not plural. https://dictionary.cambridge.org/dictionary/english/restocking-fee

gdgellatly:
Hi I did a quick review. For the most part the structure is as expected, but in the countries and companies I have worked (so all high value B2B - not retail, UK/AU/NZ), restocking fees behave pretty much the same as order deposits. A percentage of the value of goods returned. 

Usually how we handle them in Odoo, we just put a discount on the credit note line for each product line.

Not blocking, just feedback, maybe for a todo or something to basically emulate the sale order deposit interface because fixed charge just won't work for a very large number of companies/countries.

lmignon:
Hi @gdgellatly 
Thank you for your review and feedback.
@pedrobaeza @gdgellatly What do you thing about 'restocking_fee_invoicing'?

I will also add a comment in the roadmap describing a potential evolution of the module so that it can support different return policy. Actually we support a fixed price per returned products line, but it will be nice to support a percentage of the value of the returned products.

pedrobaeza:
I would still keep the `sale_stock` part (so `sale_stock_restocking_fee_invoicing`) for having the technical prefix for "grouping" the module conveniently. OK for the roadmap. Thanks for this.

lmignon:
> I would still keep the `sale_stock` part (so `sale_stock_restocking_fee_invoicing`) for having the technical prefix for "grouping" the module conveniently. OK for the roadmap. Thanks for this.

OK I'll do the changes

lmignon:
@pedrobaeza @gdgellatly The module has been renamed and the naming into the source files adapted. In the same time I renamed the name of the method called to charge restocking fee into the sale_order to be agnostic of the invoicing policy applied for restocking fees github.com/OCA/account-invoicing/pull/784/files#diff-0b351e68bac2442912f5344d57a56478R66

sale_stock_restocking_fee_invoicing/data/product_product.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo noupdate="1">
+
+    <record id="product_restocking_fee" model="product.product">
+        <field name="name">Restocking fee</field>
+        <field name="list_price">100</field>
+        <field name="type">service</field>
+        <field name="sale_ok" eval="False"/>
+        <field name="invoice_policy">order</field>
```
pedrobaeza, line 10: Add `<field name="company_id" eval="False"/>` for making multi-company compliant by default.

sale_stock_restocking_fee_invoicing/models/res_company.py
```diff
@@ -0,0 +1,33 @@
...
+
+class ResCompany(models.Model):
+
+    _inherit = "res.company"
+
+    restocking_fee_product_id = fields.Many2one(
+        "product.product",
+        string="Restocking Fee Product",
+        ondelete="restrict",
+        default=lambda a: a._default_restocking_fee_product_id(),
```
pedrobaeza, line 16: This default will be assigned before data is loaded, so it will False. You can use a post-init hook for filling it.

sale_stock_restocking_fee_invoicing/models/stock_move.py
```diff
@@ -0,0 +1,38 @@
...
+
+    @api.multi
+    def action_done(self):
+        result = super(StockMove, self).action_done()
+        move_ids_to_charge_by_so = defaultdict(list)
+        for move in self:
+            if not move.charge_restocking_fee:
+                continue
+            if not move.origin_returned_move_id:
+                continue
```
pedrobaeza: Isn't shorter this?:
```suggestion
        for move in self.filtered(lambda x: x.charge_restocking_fee and x.origin_returned_move_id):
```

houssine78:
> 
> 
> imagen
> 
> 🤣

https://thecodinglove.com/when-im-searching-for-a-meaningful-variable-name

houssine78:
> 
> 
> imagen
> 
> 🤣

imagen

sale_stock_restocking_fee_invoicing/models/stock_move.py
lmignon: IMO, the readability is better without filtered. In this case, Moreover in this case, We don't need to reduce the recordset on which the process is done.

pedrobaeza: Well, at least join both conditions:
```suggestion
        for move in self:
            if not move.charge_restocking_fee or not move.origin_returned_move_id:
                continue
```

gdgellatly: Matter of style I guess. Personally I'd just go with

```python
if move.charge.. and move.origin..:
 do_stuff()
```
I get sometimes it works doing a continue if there is a massive block of code underneath with more branches, but it is 2 lines.

pedrobaeza: Yeah, indeed this is not blocking. I don't like that style though, as it implies an extra level of indentation for your main code, although it avoids uncovered lines in tests. We never have the ideal one!

lmignon:
Closing since it will never be merged....

PR #785, opened 2020-09-21
[13.0][ADD] account_invoice_payment_warranty

Saran440:
This module provide a feature to warranty for retention during invoice / payment. And return it at later time.

At first, We need configure `Enable Invoice's Warranty on Payment`
Next, Create customer invoice (like a performance bond) by check `Payment Warranty`
Peek 2020-09-21 17-36

------------------------------------

**As time arrived to return the warranty amount.**

Create the new Vendor Bill and Select the partner.
Return Warranty field will list only uncleared retained amount for this partner.
Peek 2020-09-21 17-37


dreispt (approved):

ps-tubtim (approved):

Saran440:
it error test script from module `purchase_invoicing_no_zero_line`

PR #791, opened 2020-10-02
[11.0][IMP] account_invoice_change_currency: query

danielduqma:
In databases with so many records in `mail_message` table, performing a search over `mail.tracking.value` with ORM takes a while. Replacing it with a query will be significantly faster.


hugosantosred (approved):

danielduqma:
Can anyone take a look of this change? We have experienced a lot of performance issues in large clients.

alv97:
Any news about this merge?

PR #798, opened 2020-10-15
[IMP] account_invoice_transmit_method: Improve mails, Integrate on send and tests

etobella:
This PR replaces #743 with the changes spoken with @alexis-via


pedrobaeza:
cc @OSevangelist

PR #803, opened 2020-10-27
[11.0][FIX] account_global_discount: Missing decorator @api.one

igallart:
The _compute_amount method does not have the `api.one`, so when
it is executed on more than one record, a singleton error occurs.

### DIFF
--- a/account_global_discount/models/account_invoice.py
+++ b/account_global_discount/models/account_invoice.py
@@ -120,12 +120,7 @@ class AccountInvoice(models.Model):
         """Trigger global discount lines to recompute all"""
         return self._onchange_invoice_line_ids()
 
-    @api.depends('invoice_line_ids.price_subtotal', 'tax_line_ids.amount',
-                 'tax_line_ids.amount_rounding', 'currency_id', 'company_id',
-                 'date_invoice', 'type',
-                 'invoice_global_discount_ids', 'global_discount_ids')
-    def _compute_amount(self):
-        super()._compute_amount()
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Please cherry-pick 4ebbcafcad492d6fcd3957d0b8a7b0f72e3807f3 instead.

account_global_discount/models/account_invoice.py
```diff
@@ -150,6 +145,15 @@ def _compute_amount(self):
         self.amount_total_signed = self.amount_total * sign
         self.amount_untaxed_signed = amount_untaxed_signed * sign
 
+    @api.depends('invoice_line_ids.price_subtotal', 'tax_line_ids.amount',
```
igallart, line 148: It is necessary? So it would not be right?

`@api.depends('invoice_global_discount_ids', 'global_discount_ids')`

pedrobaeza: The rest of the fields also conditionates the computation, so they are needed AFAIK

igallart: But when you inherit a function with depends, you just need to add the new dependencies. Has that changed?

pedrobaeza: Not sure right now, as I have got inconsistent behaviors on this through versions. Any way, this is out of scope, as this PR is for assuring the compute on a multi-record.

igallart: Ok, thanks.

PR #807, opened 2020-11-02
[14.0][ADD] account_global_discount_amount module

chafique-delli:
This module allows you to enter an amount for a global discount to be applied to the invoice.
Discount lines will be automatically created per tax group.
When creating and modifying an invoice with a global discount, all total amounts will be checked to ensure that the discount has been applied correctly. If not, an automatic correction of the invoice will be made.


rvalyi:
@chafique-delli @sebastienbeau it seems there is a double effort here, see this generic module being migrated to v13 by Tecnativa https://github.com/Tecnativa/server-backend/tree/13.0-mig-base_global_discount/base_global_discount
with its PR here https://github.com/Tecnativa/server-backend/tree/13.0-mig-base_global_discount

pedrobaeza:
And this one is not preserving commit history.

chafique-delli:
@rvalyi , @pedrobaeza , I have added an explanation of what the module will do.
Concerning the name of the module, I will take it if you have a more evocative name to propose.

pedrobaeza:
Sorry, I got confused this module with `account_global_discount` that we maintain and we are migrating in v13.

sebastienbeau:
Hi @pedrobaeza it's not the same module but having you feedback will be a pleasure

In our case we need to put a discount amount (not a percentage but an **amount**) on sale order and/or invoice.

We need to split this amount (by taxes), so we were thinking of adding extra (sale.order.line /account.invoice.line) like the carrier one. More over since V13 the invoice and move are merge adding new line with the type discount seem for use the simpliest solution.

Maybe we should try to add this feature directly in your module ? Instead of doing something else (the issue is that we are going to create a lot of global.discount)

Did you plan to migrate account_global_discount in V13 ? how to do plan to support this discount line ? still in the global.discount.line or directly like an account.move.line ?

pedrobaeza:
My colleague @chienandalu and me thought initially to add both kind of discounts in `base_global_discount` and subsequent modules: fixed and percentage. We finally removed it as we didn't need such fixed type and it will complicate a bit the logic of the module. But both share similar challenges. I still think that if you need to add fixed global discounts (I discourage though both kind of global discounts in business and implementation perspective), let's add it on the existing modules.

What is the main problem? The distribution of the discount across lines. You can't create an extra line for such discount (or at least not only one), as account and taxes may vary from line to line. So you need then to spread everything proportionally. In percentage discount, that proportionality is simple: apply the same percentage everywhere. On fixed discount, you need to add a bit more logic to spread it over lines, but it can be done with such addition.

Then the challenge is to get to proper invoice and journal items data. We use the intermediate model for easing this task, and touch on totals and tax computations, and on move generation.

My colleague @ernestotejeda is now analyzing how to port this to v13, and we will have soon this migration done. I haven't dug too much, but now that the move line generation is immediate, I think we can create as many extra move lines as needed for balancing the amounts, hidden as journal items, but not sure if the code, that is very convoluted, will allow us this for the computed totals, and maybe the extra model is still needed.

Anyway, these changes will be the same for you once you have made the spread over the lines (proportionally by line amount over the fixed amount?), so I recommend that if you can wait, adapt these set of modules for adding the new type of discount.

account_global_discount_amount/models/account_move.py
```diff
@@ -0,0 +1,114 @@
...
+
+from odoo import api, models
+
+
+class AccountMove(models.Model):
+    _inherit = ["account.move", "discount.mixin"]
+    _name = "account.move"
+
+    @api.model_create_multi
+    def create(self, vals):
```
sebastienbeau: Becareful it is not vals but list_vals, you can create several move at the same time.

account_global_discount_amount/models/account_move.py
```diff
@@ -0,0 +1,114 @@
...
+
+class AccountMoveLine(models.Model):
+    _inherit = ["account.move.line", "discount.line.mixin"]
+    _name = "account.move.line"
+
+    def _create_discount_lines(self, move):
+        amount_untaxed = move.amount_untaxed
+        # create discount lines by tax line
+        with_tax_lines = move.line_ids.filtered(lambda x: x.tax_line_id)
+        for line in with_tax_lines:
```
sebastienbeau: It seem that we create on line discount line per invoice line

account_global_discount_amount/models/account_move.py
```diff
@@ -0,0 +1,114 @@
...
+        discount_line_vals.update(
+            {
+                "move_id": move.id,
+                "partner_id": move.partner_id.id,
+                "quantity": 1,
+                "tax_ids": tax_ids,
+                "account_id": discount_line.account_id.id,
+            }
+        )
+        discount_line.with_context(check_move_validity=False).unlink()
```
sebastienbeau: Why do you call unlink and the create,  a create with anonchange call will be simplier

account_global_discount_amount/models/discount_mixin.py
```diff
@@ -0,0 +1,65 @@
...
+    global_discount_ok = fields.Boolean(
+        string="Global Discount OK",
+        compute="_compute_global_discount_ok",
+        store=True,
+        help="The application of the global discount amount is right",
+    )
+
+    def _get_line_ids_by_model(self):
+        if self._name == "account.move":
+            line_ids = self.invoice_line_ids
```
sebastienbeau: Assigned a variable is not needed, directly retirn self.invoice_line_ids

account_global_discount_amount/models/discount_mixin.py
```diff
@@ -0,0 +1,65 @@
...
+            return line_ids
+
+    @api.depends("global_discount_amount", "amount_tax", "amount_total")
+    def _compute_global_discount_ok(self):
+        for record in self:
+            global_discount_ok = True
+            line_ids = record._get_line_ids_by_model()
+            if (
```
sebastienbeau: Code is too complexe

pedrobaeza:
#967 should be taken into account, as previous version is not fully correct.

sebastienbeau (changes requested):

account_global_discount_amount/models/discount_mixin.py
sebastienbeau: first get the total discount
discount_amount = sum(line_ids.filtered(lambda x: x.is_discount_line).price_subtotal)

Then compare with global discount

account_global_discount_amount/tests/test_account_global_discount_amount.py
```diff
@@ -0,0 +1,247 @@
+# Copyright 2020 Akretion France (http://www.akretion.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo.tests.common import SavepointCase
+
+
+class TestAccountGlobalDiscountAmount(SavepointCase):
```
sebastienbeau: You can reuse this : https://github.com/odoo/odoo/blob/14.0/addons/account/tests/common.py

sebastienbeau: can you add a test where you have 3 lines (one with a taxe A and then 2 with the taxe B)

account_global_discount_amount/models/account_move.py
chafique-delli: @sebastienbeau , it is corrected

account_global_discount_amount/models/account_move.py
chafique-delli: @sebastienbeau , No, you create a discount line by tax and not by invoice line

account_global_discount_amount/models/account_move.py
chafique-delli: @sebastienbeau , it is corrected

account_global_discount_amount/models/discount_mixin.py
chafique-delli: @sebastienbeau , it is corrected

account_global_discount_amount/models/discount_mixin.py
chafique-delli: @sebastienbeau , it's done

account_global_discount_amount/tests/test_account_global_discount_amount.py
chafique-delli: @sebastienbeau , it's done for the test where you have 3 lines (one with a taxe A and then 2 with the taxe B)

rousseldenis:
@chafique-delli @sebastienbeau What's the status of this ?

chafique-delli:
@sebastienbeau , Can you update the status of the P.R.

sebastienbeau:
@rousseldenis Module is ready for review.
So basically the idea of the module is to support a global amount of discount, this amount will generate several invoice line (with a product discount) and the total amount is split proportionally between each line of the invoice.

sebastienbeau (approved):

sebastienbeau:
Note I have approve only to remove my blocking review, please review and approve this work before any merge

PR #808, opened 2020-11-03
[13][MIG] account_invoice_transmit_method

TDu:
There is already #682 that has been opened (for a long time) for this migration but despite all the effort from @alexis-via it is still not merged. So I opened this one to help.
This PR adds the following improvements to the migrations :
* Fix the lintiing of XML files
* Improve the domain on the `transmit_method_id` of the `account.move`.

Also because the pre-commit hooks have been updated on the repository some fixes are necessary on
* `account_menu_invoice_refund`


jgrandguillaume (approved):
Functional test all good

simahawk (commented):
https://github.com/OCA/account-invoicing/pull/808/commits/33720920581d7dad40000005e7bbf1f45d04e23f
should be ok now https://github.com/OCA/maintainer-quality-tools/pull/680
if the conf is updated in the repo

account_invoice_transmit_method/models/account_move.py
```diff
@@ -0,0 +1,82 @@
...
+    @api.model
+    def create(self, vals):
+        if (
+            "transmit_method_id" not in vals
+            and vals.get("type")
+            and vals.get("partner_id")
+        ):
+            partner = self.env["res.partner"].browse(vals["partner_id"])
+            if vals["type"] in ("out_invoice", "out_refund"):
+                vals["transmit_method_id"] = (
```
simahawk, line 77: I'd make it computed and writeable instead, and drop this override.

account_invoice_transmit_method/tests/test_invoice_transmit_method.py
```diff
@@ -0,0 +1,28 @@
+# Copyright 2017-2020 Akretion France (http://www.akretion.com/)
+# @author: Alexis de Lattre <alexis.delattre@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo.tests.common import TransactionCase
+
+
+class TestAccountInvoiceTransmitMethod(TransactionCase):
+    def test_create_invoice(self):
```
simahawk, line 14: disable tracking pls

account_invoice_transmit_method/views/partner.xml
```diff
@@ -0,0 +1,49 @@
...
+<odoo>
+    <record id="view_partner_property_form" model="ir.ui.view">
+        <field
+            name="name"
+        >Add Invoice Transmit Methods on partner form view (account tab)</field>
+        <field name="model">res.partner</field>
+        <field name="inherit_id" ref="account.view_partner_property_form" />
+        <field name="arch" type="xml">
+            <field name="property_payment_term_id" position="after">
+                <field name="customer_invoice_transmit_method_id" widget="selection" />
```
simahawk, line 16: any good reason to use `selection`? If you have several records here this is going to be not so user friendly.

account_menu_invoice_refund/__manifest__.py
```diff
@@ -12,6 +12,5 @@
     "depends": ["account"],
     "data": ["views/account_invoice_view.xml"],
     "installable": True,
-    "development_status": "beta",
```
simahawk, line 15: see my comment

simahawk (changes requested):

account_invoice_transmit_method/views/partner.xml
alexis-via: Well, it would be surprising to have many transmit methods ; we usually only have a few. That's why I put widget=selection.

alexis-via (approved):

account_invoice_transmit_method/models/account_move.py
TDu: Did not do it but added a comment in the code

account_invoice_transmit_method/tests/test_invoice_transmit_method.py
TDu: Done

account_menu_invoice_refund/__manifest__.py
TDu: @simahawk It still fails when I remove that commit and rebase the PR, so I guess this repository has not been updated ?
Maybe @sbidoul can help ?

simahawk: I guess we miss this https://github.com/OCA/queue/commit/eaf498011f7065076d212c401051e349d5b8db69

sbidoul: @Tdu @simahawk we have the latest pylint version here. I suspect it is case sensitive. Try with `Beta`.

TDu: @sbidoul Thanks for your help, indeed `development_status` is case sensitive

TDu:
@simahawk Hello, could you update your review ?

simahawk (approved):

PR #813, opened 2020-11-13
[FIX] Do not recalculate price if no product is set

CasVissers-360ERP:
If no product is set the price should not be updated.

Before:
Line - No product - Qty = 1 - Price = 10
On update Qty -> Price 0

After:
Line - No product - Qty = 1 - Price = 10
On update Qty -> Price 10

@kittiu


kittiu:
Agree :+1:

kittiu (approved):

PR #815, opened 2020-11-17
[13.0][MIG] account_global_discount: Migration to 13.0

ernestotejeda:
cc @Tecnativa TT26281

Main changes:
- Adapt code to v13.0 (account.invoice -> account.move, account.invoice.line -> account.move.line)


pedrobaeza:
cc @sebastienbeau @chafique-delli please reconsider include the fixed amount inside this module instead of creating a new one, and check to make use of the same methods.

pedrobaeza:
@ernestotejeda Travis is red

account_global_discount/readme/USAGE.rst
```diff
@@ -0,0 +1,13 @@
+To use this module, you need to:
+
+#. Go to *Settings > Parameters > Global Discounts*
+#. Add a new discount that can be either by percentage or fixed amount.
```
victoralmau: It is not possible to apply a fixed discount right? What is defined is only in % (it is for clarification and if appropriate change the USAGE.rst)

account_global_discount/models/account_move.py
```diff
@@ -0,0 +1,319 @@
...
+            # Apply onchanges
+            tax_line._onchange_amount_currency()
+            tax_line._onchange_balance()
+
+    def _set_global_discounts_by_tax(self):
+        """Create invoice global discount lines by taxes combinations and
+        discounts.
+        """
+        self.ensure_one()
+        invoice_global_discounts = self.env["account.invoice.global.discount"]
```
victoralmau:
```suggestion
```
IMO these line it's unnecessary.

victoralmau (approved):
Good job!! Only some little comments.

pedrobaeza:
@sebastienbeau @chafique-delli have you checked this one? Do you think the fixed amount can be coupled here? I think so with not so many effort.

pedrobaeza (approved):

pedrobaeza:
As nobody more is commenting about this, let's merge it:

/ocabot merge nobump

PR #817, opened 2020-11-18
[13.0][ADD] account_move_force_removal: Allow force removal account moves

victoralmau:
Allow force removal invoice

Please @pedrobaeza  review it.

@Tecnativa TT26819


account_invoice_force_removal/__manifest__.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2020 Tecnativa - Víctor Martínez
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+{
+    "name": "Account Invoice Force Removal",
+    "summary": """Allow force removal invoice""",
+    "version": "13.0.1.0.0",
+    "category": "Account",
+    "license": "LGPL-3",
```
pedrobaeza: We use LGPL only for website modules.

```suggestion
    "license": "AGPL-3",
```

You need to change all header comments.

account_invoice_force_removal/__manifest__.py
```diff
@@ -0,0 +1,14 @@
...
+    "website": "http://github.com/OCA/dms",
```
pedrobaeza:
```suggestion
    "website": "http://github.com/OCA/account-invoicing",
```

account_invoice_force_removal/__manifest__.py
```diff
@@ -0,0 +1,14 @@
...
+    "author": "Tecnativa, Odoo Community Association (OCA)",
+    "depends": ["account"],
+    "application": True,
```
pedrobaeza:
```suggestion
```

account_invoice_force_removal/models/account_move.py
```diff
@@ -0,0 +1,16 @@
...
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    def unlink(self):
+        """Allow to remove moves related to invoices."""
+        for item in self:
+            if item.type in ["out_invoice", "out_refund", "in_invoice", "in_refund"]:
+                if item.state == "cancel" and item.name != "/":
+                    item.write({"name": "/"})
```
pedrobaeza: Is it ' /'  the default value that allows to unlink or is it `False`?

account_invoice_force_removal/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+Allows you to delete canceled invoices
```
pedrobaeza:
```suggestion
This module allows you to delete canceled invoices, which is prevented in Odoo by default.

You must care yourself about the numbering sequence doing this.
```

account_invoice_force_removal/readme/USAGE.rst
```diff
@@ -0,0 +1,4 @@
+Go to Invoicing > Customers> Invoices and enter an invoice.
```
pedrobaeza: Missing list markers: `#. `

pedrobaeza (changes requested):

account_invoice_force_removal/models/account_move.py
victoralmau: Is '/', https://github.com/odoo/odoo/blob/13.0/addons/account/models/account_move.py#L1742

pedrobaeza: OK, but seeing this, we can use the context `force_delete` and not doing any write.

account_invoice_force_removal/tests/test_move.py
```diff
@@ -0,0 +1,52 @@
+# Copyright 2020 Tecnativa - Víctor Martínez
+# License AGPL-3 - See https://www.gnu.org/licenses/agpl-3.0.html
+
+from odoo.exceptions import UserError
+from odoo.tests import Form
+from odoo.tests.common import SavepointCase
+
+
+class TestMOve(SavepointCase):
```
sergio-teruel: Fix typo

account_invoice_force_removal/tests/test_move.py
```diff
@@ -0,0 +1,52 @@
...
+        invoice = Form(
+            cls.env["account.move"].with_context(
+                default_type="out_invoice", default_company_id=cls.env.company.id
+            )
+        )
+        invoice.partner_id = cls.partner
+        invoice.journal_id = cls.journal
+        with invoice.invoice_line_ids.new() as line_form:
+            line_form.name = cls.product.name
+            line_form.product_id = cls.product
```
sergio-teruel: If you want, for invoices the product is not mandatory, so we can save code lines and you do not need create a product

sergio-teruel (changes requested):
small comments

account_invoice_force_removal/models/account_move.py
```diff
@@ -0,0 +1,16 @@
...
+
+from odoo import models
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    def unlink(self):
+        """Allow to remove moves related to invoices."""
+        for item in self:
```
carlosdauden:
```suggestion
        cancelled_moves = self.filtered(lambda m: m.state == "cancel")
        super(AccountMove, cancelled_moves.with_context(force_delete=True)).unlink()
        return super(AccountMove, self - cancelled_moves).unlink()
```

account_invoice_force_removal/models/account_move.py
```diff
@@ -0,0 +1,16 @@
...
+            if item.type in ["out_invoice", "out_refund", "in_invoice", "in_refund"]:
```
carlosdauden: Why limit this feature to only invoice types?

carlosdauden (changes requested):

account_invoice_force_removal/tests/test_move.py
victoralmau: Although `product_id` isn't mandatory but help to auto-define other fields (`account_id` for example).

account_invoice_force_removal/models/account_move.py
victoralmau: Use these suggestion to improve code. Thanks!

account_invoice_force_removal/models/account_move.py
victoralmau: Addon is limited to invoices moves because addon (`account_invoice_force_removal`) is related to invoices.

account_invoice_force_removal/tests/test_move.py
sergio-teruel: Yes!! but account_id if the has not a product Odoo get it from Journal Default Credit Account

account_invoice_force_removal/models/account_move.py
```diff
@@ -0,0 +1,19 @@
...
+    _inherit = "account.move"
+
+    def unlink(self):
+        """Allow to remove moves related to invoices."""
+        types_allow = ["out_invoice", "out_refund", "in_invoice", "in_refund"]
+        moves = self.filtered(lambda m: m.type in types_allow).filtered(
+            lambda m: m.state == "cancel"
+        )
+        # update invoice name
+        moves.filtered(lambda m: m.name != "/").write({"name": "/"})
```
carlosdauden: Why do you write records that they will be removed?
```suggestion
```

carlosdauden (changes requested):

account_invoice_force_removal/models/account_move.py
carlosdauden: You can change addon name if that is the problem 😉

account_invoice_force_removal/models/account_move.py
```diff
@@ -0,0 +1,19 @@
...
+
+from odoo import models
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    def unlink(self):
+        """Allow to remove moves related to invoices."""
+        types_allow = ["out_invoice", "out_refund", "in_invoice", "in_refund"]
```
carlosdauden: In case that limit to move types you can define this out of method to be easily extended.
Or the type list could be defined as system parameter.
Those are suggestions only, It isn't a required changes.

account_invoice_force_removal/models/account_move.py
victoralmau: Ok, I'm going to change the name of the addon to `account_move_force_removal`, do you all agree? @pedrobaeza @sergio-teruel

pedrobaeza:
Let's allow it for any journal entry, but then, apart from the module name, it should be moved to OCA/account-financial-tools. Another code comment coming inline.

account_invoice_force_removal/models/account_move.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    types_allow_unlink = ["out_invoice", "out_refund", "in_invoice", "in_refund"]
+
+    def unlink(self):
+        """Allow to remove moves related to invoices."""
+        moves = self.filtered(lambda m: m.type in self.types_allow_unlink).filtered(
```
pedrobaeza: Don't do this `filtered`, but directly call `super` with the context. Or the rest of the checks (like the posted one) is ignored also by the context?

victoralmau:
> OCA/account-financial-tools

Could do you move it to OCA/account-financial-tools repo?

pedrobaeza:
I'm afraid not, a new PR must be opened and you need to cherry-pick the commit in the other branch.

account_invoice_force_removal/models/account_move.py
carlosdauden: @pedrobaeza the unique check is this:
https://github.com/odoo/odoo/blob/13.0/addons/account/models/account_move.py#L1741-L1742

pedrobaeza: I see, then it doesn't serve as is, as this will allow to unlink posted entries. I will comment in the other PR. Sorry for the noise.

PR #465, opened 2018-12-17
[12.0] [MIG] account_invoice_fiscal_position_update

bjeficent:
Bump to version 12.0

PR #819, opened 2020-11-19
[FIX] account_invoice_fiscal_position_update: filter on invoice lines without display_type

alexis-via:
Fixes PR #465 when the module was migrated to v12

### DIFF
--- a/account_invoice_fiscal_position_update/models/account_invoice.py
+++ b/account_invoice_fiscal_position_update/models/account_invoice.py
@@ -27,7 +27,8 @@ class AccountInvoice(models.Model):
         lines_without_product = []
         fp = self.fiscal_position_id
         inv_type = self.type
-        for line in self.invoice_line_ids:
+        for line in self.invoice_line_ids.filtered(
+                lambda l: not l.display_type):
             if line.product_id:
                 account = line.get_invoice_line_account(
                     inv_type, line.product_id, fp, self.company_id)


legalsylvain (approved):
Makes sense.

thanks.

account_invoice_fiscal_position_update/models/account_invoice.py
```diff
@@ -27,7 +27,8 @@ def fiscal_position_change(self):
         lines_without_product = []
         fp = self.fiscal_position_id
         inv_type = self.type
-        for line in self.invoice_line_ids:
+        for line in self.invoice_line_ids.filtered(
+                lambda l: not l.display_type):
             if line.product_id:
```
pedrobaeza, line 32: Some people don't  like `filtered` by its performance, but the truth is that you have now 2 comparison at 2 different levels, so let's unify them:
```suggestion
        for line in self.invoice_line_ids:
            if line.product_id or not l.display_type:
```

pedrobaeza (commented):
I'm also wondering: does any section or note line contains `product_id`. By default no, so previous code was correct, but your problem is specific to another code that is creating such lines with product filled. Can it be?

alexis-via:
The previous code was not correct because of the "else" and the warning: it would display the warning on section and note lines.

pedrobaeza:
You are totally right. Do you know why Travis is red?

pedrobaeza:
Let's try:

/ocabot merge patch

chienandalu (approved):

HaraldPanten (approved):

pedrobaeza:
Fixed problems of the branch in:

08c15179867d90a4ae80e09c85dc9e002fc75388
513553134a7b47c7034bd45a1c30e8d9d9766d36
33afcb8f5ad71332a6f57cecb1fa1ee6848c27e2

/ocabot merge patch

PR #827, opened 2020-12-09
[FIX] account_invoice_fiscal_position_update: filter `display_type` lines

chienandalu:
In invoices with notes or sections we'll get an exception if we try to change the fiscal position. This fix solves it

cc @Tecnativa TT27224


pedrobaeza:
Duplicate of #819

PR #831, opened 2020-12-11
[ADD] account_approval_process and account_approval_process_with_hr

khaeusler:
We have implemented an account approval process for one of our customers. We would like to make this available to the OCA as well.

This addon provides any number of configurable and named stages for incoming and outgoing invoices/credit notes. For each stage, users/groups and departments (if hr is installed) can be defined, which are responsible for the respective stage. Confirmation is blocked until all stages have been successfully completed.
If validation is needed, the user can request it. If a mail template has been set with the stage, the users who have the right to validate this stage will be notified by mail.


kittiu:
@khaeusler thank you for contributing to OCA.
I am trying to understand what it does but the readme seem lacking. Would be helpful if you can explain more or post some pictures.
Thanks!

kittiu:
From reading the code, I think similar thing can be achieved by server-ux's base_tier_validation. Have you check module `account_move_tier_validation` in this repo?

khaeusler:
> From reading the code, I think similar thing can be achieved by server-ux's base_tier_validation. Have you check module `account_move_tier_validation` in this repo?

No, but I will check it later!

khaeusler:
> From reading the code, I think similar thing can be achieved by server-ux's base_tier_validation. Have you check module `account_move_tier_validation` in this repo?

Hi @kittiu,

I have looked at the addon account_move_tier_validation.
* Both addons implement validation.
* Both do not provide the same functionality.

e.g., account_move_tier_validation cannot send the notifications to departments. For inexperienced users, account_approval_process is also easier to use.

chandni299 (commented):
I think it is good to have `@api.one`  or `@api.multi`  deocrators on `def _compute_xxx()`  and `def _check_xxx()` methods.

khaeusler:
> I think it is good to have `@api.one` or `@api.multi` deocrators on `def _compute_xxx()` and `def _check_xxx()` methods.

In newer Odoo versions api.one is deprecated also api.multi

chandni299:
> > I think it is good to have `@api.one` or `@api.multi` deocrators on `def _compute_xxx()` and `def _check_xxx()` methods.
> 
> In newer Odoo versions api.one is deprecated also api.multi

yes. you right with v13 and v14 only AFIK

khaeusler:
> > > I think it is good to have `@api.one` or `@api.multi` deocrators on `def _compute_xxx()` and `def _check_xxx()` methods.
> > 
> > In newer Odoo versions api.one is deprecated also api.multi
> 
> yes. you right with v13 and v14 only AFIK

I wrote it that way right away because it simplifies migration to v13 and v14.
If you know that something will not be supported later, you can take that into consider when writing new code for older versions.

PR #832, opened 2020-12-15
[11.0][MIG] account_invoice_validation_workflow

dsolanki-initos:


dsolanki-initos:
We close this PR as we currently focus on the newer available version.

PR #844, opened 2021-01-11
[MIG] account invoicing mode: Migration to 14.0

phuctranfxvn:
This PR includes migration of 3 modules from this PR https://github.com/OCA/account-invoicing/pull/755:
- account_invoice_base_invoicing_mode
- account_invoice_mode_at_shipping
- account_invoice_mode_monthly


fshah-initos:
Hi @phuctranerp 
Please create seperate PR for other modules we use only one module in a PR.

tarbib:
Hello @TDu could you please review the PR? cc @jgrandguillaume

account_invoice_mode_at_shipping/models/stock_picking.py
```diff
@@ -3,29 +3,25 @@
...
+    def _action_done(self):
+        res = super()._action_done()
         for picking in self:
             if picking._invoice_at_shipping():
                 picking.with_delay()._invoicing_at_shipping()
         return res
 
     def _invoice_at_shipping(self):
-        """Check if picking must be invoiced at shipping."""
+        """Check if picking must be invoiced at shippin`g."""
```
TDu, line 18: `"""Check if picking must be invoiced at shipping."""`

TDu (approved):
LG, thanks for the migration.

TDu:
> Hi @phuctranerp
> Please create seperate PR for other modules we use only one module in a PR.

I think it is good like this in one PR, because they are tightly related.
And this is how they have been proposed for v13, where they have not been merged.

tarbib:
@TDu Hello, I just tested it on v14 but there's an issue :
Commercial partner : Company A => invoicing mode : at shipping
Contact : Company A, Contact 1 => accounting tab says that configuration is done on the parent.
If the partner_invoice_id from the SO is a contact then we do not have automatic invoicing at shipping.
The contact should inherit the value from the commercial partner.
Can you please take this case into consideration? cc @jcoux

TDu:
@tarbib Indeed the fields should be added to the commercial_fields of the res.partner
Something like this on the res.partner model
 ```
@api.model
 def _commercial_fields(self):
          return super()._commercial_fields() + [
              "invoicing_mode",
              "one_invoice_per_order",
          ]                  
```
That would be a good addition to those modules

simahawk (changes requested):
please rebase (well, replay all commits) on v13.0 as @TDu squashed commits.

simahawk:
@phuctranerp if you preferer we can merge 13.0 version in 14.0 and make it uninstallable as we've done for postlogistics PR :wink:

simahawk:
https://github.com/OCA/account-invoicing/pull/983

PR #848, opened 2021-01-17
[12.0][FIX] company_dependent misspelled in some fields

sergiocorato:


yostashiro (approved):

legalsylvain (approved):

pedrobaeza:
You should transfer existing values through a migration script

sergiocorato:
> You should transfer existing values through a migration script

Creating the same value for all existing companies?

pedrobaeza:
Check https://github.com/OCA/openupgradelib/blob/077068b62c2ab3ea2cb09634f1084fb5743d3067/openupgradelib/openupgrade.py#L2551 method

pedrobaeza (changes requested):
Migration script needed.

sergiocorato:
> Migration script needed.

Added

pedrobaeza:
Are you sure you can use the same field for the source and the target of the field?

ivantodorovich (approved):

PR #849, opened 2021-01-24
[14.0] [ADD] new module account_receipt_print

primes2h:
This module allows printing of sale and purchase receipts.


marcelofrare (approved):
Works ok for me

marcelofrare (approved):
Yes! more quickly!
Ok for me

account_receipt_print/i18n/it.po
```diff
@@ -0,0 +1,178 @@
...
+#. module: account_receipt_print
+#: model_terms:ir.ui.view,arch_db:account_receipt_print.report_receipt_document
+msgid "Tax ID"
+msgstr "Partita IVA"
+
+#. module: account_receipt_print
+#: model_terms:ir.ui.view,arch_db:account_receipt_print.report_receipt_document
+msgid "on"
+msgstr "su"
+
```
tafaRU: Are the following lines needed? Otherwise, could you please remove them?

account_receipt_print/report/a5_report.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <data noupdate="1">
+        <record id="paperformat_euro_a5" model="report.paperformat">
```
tafaRU: Where did you take the specifications to fill the following fields? Could you please share a link?

account_receipt_print/views/account_move_view.xml
```diff
@@ -0,0 +1,21 @@
...
+    <record id="view_move_form_receipt" model="ir.ui.view">
+        <field name="name">account.move.receipt</field>
+        <field name="model">account.move</field>
+        <field name="inherit_id" ref="account.view_move_form" />
+        <field name="arch" type="xml">
+            <button name="button_draft" position="after">
+                <button
+                    name="print_receipt"
+                    type="object"
+                    string="Print"
```
tafaRU: Why do you prefer adding an explicit button rather than show the report in the Print drop-down menu?
I think can confuse the user.

tafaRU (changes requested):
Also, trying the module I noticed that there are some problems with page break by using  `paperformat_euro_a5`.

See below some examples:

image
image

Could you please have a look?

The rest LGTM, thanks a lot!

enigmista91:
@primes2h  will you fix the comment?

account_receipt_print/i18n/it.po
primes2h: Done!

matteoopenf (approved):

matteoopenf:
E' probabile che devi fare rebase che anche l'altro giorno una pr era rossa in questo stesso modo mi sembra

account_receipt_print/views/account_move_view.xml
primes2h: It's a long story. :sweat_smile: 
Anyway, you are right. See https://github.com/PNLUG/account-invoicing/pull/1

matteoopenf:
guardando meglio è un import non andato a buon fine

matteoopenf (approved):

primes2h:
> guardando meglio è un import non andato a buon fine

Adesso è ok, grazie.

matteoopenf:
> > guardando meglio è un import non andato a buon fine
> 
> Adesso è ok, grazie.

riapprovata

account_receipt_print/report/a5_report.xml
primes2h: For margins and spacing I used a4 paperformat as a reference (proportional).
Anyway, I changed my mind about this. I think that a new paperformat should be added in a separate module,  a base one if needed. I have removed it accordingly.

enigmista91 (approved):

account_receipt_print/views/account_move_view.xml
primes2h: Done.

tafaRU (approved):

tafaRU (changes requested):
@primes2h is there any particular reason for creating `report_receipt_document` without inheriting from `account.report_invoice_document` (by setting [primary](https://github.com/odoo/odoo/blob/0ec6629de721042ed215e0e011fe38daddccb0d5/odoo/addons/base/models/ir_ui_view.py#L240-L251) mode)?

Similarly to what has been achieved in https://github.com/OCA/l10n-italy/blob/764df48a8334a4e3cfde71d9e339921de6b7d2d3/l10n_it_corrispettivi/views/account_report.xml#L24

primes2h:
> @primes2h is there any particular reason for creating `report_receipt_document` without inheriting from `account.report_invoice_document` (by setting [primary](https://github.com/odoo/odoo/blob/0ec6629de721042ed215e0e011fe38daddccb0d5/odoo/addons/base/models/ir_ui_view.py#L240-L251) mode)?
> 
> Similarly to what has been achieved in https://github.com/OCA/l10n-italy/blob/764df48a8334a4e3cfde71d9e339921de6b7d2d3/l10n_it_corrispettivi/views/account_report.xml#L24

Good point.
My first idea needed specific report format (margins etc.) to fit into a5 size, so a custom report would have been more flexible.
After recent changes I think we could opt to inherit report with primary set as you pointed out.

account_receipt_print/views/report_receipt.xml
```diff
@@ -0,0 +1,85 @@
...
+                </h2>
+            </xpath>
+
+            <div name="invoice_date" position="replace">
+                <div
+                    class="col-auto col-3 mw-100 mb-2"
+                    t-if="o.invoice_date"
+                    name="invoice_date"
+                >
+                    <strong>Receipt Date:</strong>
```
tafaRU: Why don't you only replace the `strong` tag?

account_receipt_print/views/report_receipt.xml
```diff
@@ -0,0 +1,85 @@
...
+            </p>
+
+            <p t-if="o.invoice_payment_term_id" position="attributes">
+                <attribute name="style">display: none;</attribute>
+            </p>
+
+            <xpath expr="//p[@t-if='o.narration']" position="before">
+                <t t-if="o.move_type in ('out_receipt', 'in_receipt')">
+                    <p class="mt32 mb16">
+                        <strong>Amount In Words:</strong>
```
tafaRU, line 51: Cool add! But I'm wondering about whether it's not better keep it similar to `report_receipt_document` and move it to a customization by using i.e https://github.com/OCA/l10n-italy/pull/2841 Or is there some particular reason I'm not seeing?

tafaRU (changes requested):

tafaRU:
@pedrobaeza I'm wondering about what name is more suited for this new module (see https://github.com/PNLUG/account-invoicing/blob/14.0-add_receipt_printing/account_receipt_print/README.rst for having its description) between these two: `account_receipt_print` and `account_move_receipt_print`. IMO it is the second one. What about you?
Could you please also explain the reason for your choice?

Thank you in advance.

pedrobaeza:
I would go for `account_receipt_report`. My logic behind this is to try to homogenize the nomenclature following this scheme:

area/app + subarea + function

Regards.

account_receipt_print/views/report_receipt.xml
primes2h: > Cool add! But I'm wondering about whether it's not better keep it similar to `report_receipt_document` and move it to a customization by using i.e [OCA/l10n-italy#2841](https://github.com/OCA/l10n-italy/pull/2841) 
> Or is there some particular reason I'm not seeing?

That PR has a different purpose.
In fact, showing amount in words is common in receipts, as they are generally a proof of payment.

account_receipt_print/views/report_receipt.xml
primes2h: Done, thank you!
https://github.com/OCA/account-invoicing/compare/00fcdb1764a06a49549fb511d6a2720e89c62679..d85285a94793734a27ac994506d4ee2ad15f149f

tafaRU:
@primes2h last thing: please rename the module to `account_receipt_report` according to https://github.com/OCA/account-invoicing/pull/849#issuecomment-1184554057
Thanks!

primes2h:
> I would go for `account_receipt_report`. 
> My logic behind this is to try to homogenize the nomenclature following this scheme:
> area/app + subarea + function

Did you mean `account_receipt_print` ? (first of two proposed by @tafaRU and current module name)
It fits right in the scheme, in fact the function is to print.

primes2h:
> > I would go for `account_receipt_report`.
> > My logic behind this is to try to homogenize the nomenclature following this scheme:
> > area/app + subarea + function
> 
> Did you mean `account_receipt_print` ? (first of two proposed by @tafaRU and current module name) It fits right in the scheme, in fact the function is to print.

@pedrobaeza

pedrobaeza:
`report` is because it's used in other modules:

https://odoo-community.org/shop/adr-products-report-5730
https://odoo-community.org/shop/valued-picking-linked-with-mrp-kits-5717
https://odoo-community.org/shop/stock-picking-product-barcode-report-5709
https://odoo-community.org/shop/stock-picking-report-undelivered-products-5629

but if the report exists and this only enables the print option, then go ahead with it.

PR #850, opened 2021-01-25
14.0 [MIG] account_invoice_pricelist: Migration to 14.0

hildickethan:
Standard migration to 14.0

Applied [this change](https://github.com/OCA/account-invoicing/commit/defd56407d9c719aab4923ea1ba18df19f54174c#diff-f63da6ff419afd7f33e73cf400ec73fe25d83624e5e9ab30c92e4d88f7a4a47c) from 12.0 as well


hildickethan:
I don't know what the test errors are caused by, from what I could tell the code works fine but I fail to see where the expected values come from. My guess is that the USD to EUR rates aren't what they were previously in 13.0 or whenever the module was migrated, I'm not sure how Odoo gets them

If anyone could shed some light on the tests it would be appreciated

kittiu:
You can apply this patch https://github.com/OCA/account-invoicing/pull/875

kittiu (approved):
LGTM

kittiu:
There are some change in the PR https://github.com/OCA/account-invoicing/pull/875
Can you take a look please.

hildickethan:
@kittiu since your PR is now merged, I re-applied the history of commits so everything should be ok now

account_invoice_pricelist/models/account_move.py
```diff
@@ -0,0 +1,228 @@
...
+            pricelist_item = PricelistItem.browse(rule_id)
+            while (
+                pricelist_item.base == "pricelist"
+                and pricelist_item.base_pricelist_id
+                and pricelist_item.base_pricelist_id.discount_policy
+                == "without_discount"
+            ):
+                price, rule_id = pricelist_item.base_pricelist_id.with_context(
+                    uom=uom.id
+                ).get_product_price_rule(product, qty, self.order_id.partner_id)
```
gdgellatly: where/what is this order_id field?

This seems basically a cut and paste of the sale order line code. Anyway it should probably be move_id.

gdgellatly (commented):
The thing I don't like about modules like this, is they cut and paste so much, and short of using a new sale/line object and setting up and calling the functions it is a bit unavoidable.

But then my preference is that in the tests that where the code is copied, you take a hash of the copied function and if that hash changes you get a failure in tests indicating that the function you copied / largely copied has received a fix. 

e.g.
```python
import hashlib
import inspect

from odoo.tests.common import TransactionCase

from odoo.addons.product.models.product_pricelist import Pricelist as upstream

# mostly compatible, only bug fixes different, except last entry which is base
# other hashes may be valid but not tested, more of an early warning system
# for changed behaviour

VALID_HASHES = ["3b8538c9c617daf752b5ce8d588468bb"]


class TestProductPricelistHash(TransactionCase):
    def test_upstream_file_hash(self):
        """Test that copied upstream function hasn't received fixes"""
        func = inspect.getsource(upstream._compute_price_rule).encode()
        func_hash = hashlib.md5(func).hexdigest()
        self.assertIn(func_hash, VALID_HASHES)
```

hildickethan:
I wil lfinish updating the hash check on monday, I need the order_id change for now

hildickethan:
@gdgellatly Sorry for the late response and changes, I implemented your hash check for the primarily copied function as well as changed that order_id field that has been wrong since that function was copied in 13.0

PR #852, opened 2021-01-26
[13.0][FIX] account_global_discount: Add option to discount apply when line contains only one tax with zero amount

victoralmau:
Add option to discount apply when line contains only one tax with zero amount.

@Tecnativa TT27911

### DIFF
--- a/account_global_discount/models/account_move.py
+++ b/account_global_discount/models/account_move.py
@@ -1,5 +1,6 @@
 # Copyright 2019 Tecnativa - David Vidal
 # Copyright 2020 Tecnativa - Pedro M. Baeza
+# Copyright 2021 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 from odoo import _, api, exceptions, fields, models
 from odoo.tools import config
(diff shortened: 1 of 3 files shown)


victoralmau:
Error happen in `account_invoice_pricelist` tests, @manuelregidor can you review it please?

account_global_discount/models/account_move.py
```diff
@@ -131,6 +132,33 @@ def _set_global_discounts_by_tax(self):
                     )
                 )
                 base = discount["base_discounted"]
+        # Check all moves with tax_ids defined to check if any not discount created
+        # (tax amount is zero and only one tax is aplly)
+        if self.line_ids.filtered("tax_ids"):
```
ernestotejeda: I think we can remove this 'if' statement

account_global_discount/models/account_move.py
```diff
@@ -131,6 +132,33 @@ def _set_global_discounts_by_tax(self):
...
+            for move in self.line_ids.filtered("tax_ids"):
+                key = tuple(move.mapped("tax_ids").ids)
+                if key in taxes_keys:
+                    if taxes_keys[key]:
```
ernestotejeda: These two lines can be written in one 'if' statement: `if taxes_keys.get("key"):`

ernestotejeda (approved):
Just two non-blocking details.

pedrobaeza (approved):

pedrobaeza:
We need to fix this problem, so I merge manually, and let the unrelated error in `account_invoice_pricelist`.

PR #859, opened 2021-02-03
[12.0][FIX] Invoice price_unit

renatonlima:
When we create an invoice from picking, there is a method called _get_price_unit_invoice to be overwrite by other modules to define a price_unit for invoice line but this method does not work because after it was called the price_unit is lost in _simulate_invoice_line_onchange call


marcelsavegnago (approved):

rvalyi (approved):

cubells (approved):
:+1: 

LGTM

Code review and test

rvalyi (approved):

PR #781, opened 2020-09-03
[12.0][ADD] account_portal_hide_invoice module

Khalid-SerpentCS:
Website Hide Invoice
====================

This module hide invoices on customer portal.

PR #870, opened 2021-02-23
[14.0] [ADD] new module portal invoice hide by default

primes2h:
This module hides "Invoices & Bills" item by default in portal when documents are not available, as it was in previous versions of Odoo.


CasVissers-360ERP:
@primes2h Isn't this similar to  #781 account_portal_hide_invoice? If so, isn't it better to migrate that module?

primes2h:
> @primes2h Isn't this similar to #781 account_portal_hide_invoice? If so, isn't it better to migrate that module?

@CasVissers-360ERP 
In fact it has a different purpose.
#781 hides invoices "always", this PR do it only when invoice count = 0 (as it was in previous Odoo versions).

It's still in draft because in 14.0 the view has been refactored and it's not so easy to add needed code with xpath.

PR #773, opened 2020-08-18
[12.0][MIG] account_invoice_tier_validation

patrickrwilson:
Migrating from version 11 to 12.

PR #871, opened 2021-02-26
[12.0][MIG] account_invoice_tier_validation

astirpe:
Continuation of https://github.com/OCA/account-invoicing/pull/773


jonerikceberio:
I have a problem at the end of the process: user A requests a Validation, user B validates it, an user A sees it as validated, so presses the Validate button on the invoice. However, the validation process first writes a bunch of fields in the invoice BEFORE the state change, as it creates the move first. Namely, in my case: 'date_invoice', 'date_due', 'move_id', 'date', 'move_name', 'number', 'invoice_number', 'reference'
So I see two options: 
- adding those fields to _get_under_validation_exceptions (sloppy)
- making the domain in the tier definition inspect invoices that have reviews and are not approved
Am I the only one or am I missing something here?

astirpe:
@jonerikceberio you are right!
With last commit I'm trying to solve the problem by using the `_get_under_validation_exceptions` with extra check that invoice is also in draft.

@AaronHForgeFlow you are the original author of this module, how would you solve the issue mentioned above?

AaronHForgeFlow:
Hi @astirpe . I think your solution is fine, it should work and the fields will be filled in the correct time. I am ccing @LoisRForgeFlow as the original author for the base_tier_validation just in case.

BTW I tested and it seems I cannot validate the invoice even if all the validations are accomplished:

image

Could you check?

astirpe:
@AaronHForgeFlow I cannot reproduce the error in Runbot. Is it maybe caused by any other module you installed in your environment? Would be interesting to know which is the field that is the cause of the failure (I assume there's some field that is being written while validating).

AaronHForgeFlow:
Actually I am getting that message in runbot. This is link to the invoice: http://3469034-871-eb62ef.runbot2-2.odoo-community.org/web?debug=1#id=12&action=290&model=account.invoice&view_type=form&menu_id=139 I tried to validate it with the demo user. The invoice seems already validated but I cannot continue. Am I missing something?

astirpe:
@AaronHForgeFlow  OK I see. Thanks! Unfortunately the error message doesn't say much, and I'm not able to reproduce locally. I could add some more details in the error message so we could better investigate about the fields are being written. If you have better ideas please let me know.

astirpe:
I opened https://github.com/OCA/server-ux/pull/325 to propose a better error message, hopefully it will be an help in understanding issues like this: https://github.com/OCA/account-invoicing/pull/871#issuecomment-791228153

account_invoice_tier_validation/models/account_invoice.py
```diff
@@ -0,0 +1,50 @@
...
+    _name = "account.invoice"
+    _inherit = ["account.invoice", "tier.validation"]
+    _state_from = ["draft"]
+    _state_to = ["open", "paid"]
+
+    @api.model
+    def _get_under_validation_exceptions(self):
+        res = super()._get_under_validation_exceptions()
+        if self.env.context.get("invoice_tier_validation_allow_write"):
+            res += ["move_id", "date", "move_name", "date_due", "date_invoice"]
```
yostashiro:
```suggestion
            res += ["move_id", "date", "move_name", "date_due", "date_invoice", "reference"]
```

I suppose this should fix this issue https://github.com/OCA/account-invoicing/pull/871#issuecomment-791408920, although I haven't checked the record in runbot.  I got the same error in my local environment and the reference field was the cause.

yostashiro (changes requested):

astirpe:
Thank you @yostashiro ! I applied your changes.

@AaronHForgeFlow would you check if the fix is working for you?

AaronHForgeFlow (approved):
I tested on runbot and checked that the issue is solved. :+1:

AdriaGForgeFlow (approved):
Functional test and code review :+1:

AaronHForgeFlow:
@yostashiro can you please update your review?

yostashiro (approved):
LGTM 👍

PR #875, opened 2021-03-03
[13.0][FIX] account_invoice_pricelist, fix test currency rate

kittiu:
Test was failed due to wrong USD currency rate (which I think is the problem of core odoo).

image

image

This PR ensure that unit test will be using the correct rate.

### DIFF
--- a/account_invoice_pricelist/tests/test_account_move_pricelist.py
+++ b/account_invoice_pricelist/tests/test_account_move_pricelist.py
@@ -218,6 +218,20 @@ class TestAccountMovePricelist(common.SavepointCase):
                 ],
             }
         )
+        # Fix currency rate of EUR -> USD to 1.5289
+        usd_currency = cls.env["res.currency"].search([("name", "=", "USD")])
+        usd_rates = cls.env["res.currency.rate"].search(
+            [("currency_id", "=", usd_currency.id)]
+        )
+        usd_rates.unlink()
+        cls.env["res.currency.rate"].create(
+            {
+                "currency_id": usd_currency.id,
+                "rate": 1.5289,
+                "create_date": "2010-01-01",
+                "write_date": "2010-01-01",
+            }
+        )
 
     def test_account_invoice_pricelist(self):
         self.invoice._onchange_partner_id_account_invoice_pricelist()


kittiu:
The bug here is the cause of many PR fail.

@pedrobaeza can we fast forward this?

pedrobaeza:
I see that this is subject to fail at certain hours around midnight. I propose the following solution:

- Remove all the existing rates.
- Put that rate with a very old date (`2000-01-01`).

kittiu:
> I see that this is subject to fail at certain hours around midnight.

How can it fail? (With create new rate as proposed)

pedrobaeza:
As you are not indicating any date, it takes "current date", but such current date can vary depending on the execution time, so we have certain variability linked to it. As the solution is very simple, I prefer to remove any possible fail condition.

kittiu:
> As you are not indicating any date, it takes "current date", but such current date can vary depending on the execution time, so we have certain variability linked to it. As the solution is very simple, I prefer to remove any possible fail condition.

Done as suggested.

account_invoice_pricelist/tests/test_account_move_pricelist.py
```diff
@@ -218,6 +218,20 @@ def setUpClass(cls):
                 ],
             }
         )
+        # Fix currency rate of EUR -> USD to 1.5289
+        cls.usd_currency = cls.env["res.currency"].search([("name", "=", "USD")])
```
pedrobaeza: Why storing this and the next one in `cls`?

kittiu: Not sure I understand well, correct me.
I think we need to set the USD rates to be used in all test_xxx. So, I put it in setUpClass.

pedrobaeza: Yes, but don't store it as class variable. Just use local variables:
```python
usd_currency = cls.env["res.currency"].search([("name", "=", "USD")])
rates = ....
```

kittiu: Oh ok, sorry for that :). Done.

pedrobaeza (approved):

pedrobaeza:
Fast tracking the fix:

/ocabot merge patch

PR #876, opened 2021-03-03
[12.0][FIX] account_invoice_triple_discount: subtotal computation

SimoRubi:
**Steps (see also added test case)**
1. In Sales > Configuration > Settings, enable 'discounts'
2. Create a sale order with one line having:
    - price unit = 25
    - quantity = 65
    - discount = 50
    - discount = 13

**Before this PR**
The line's subtotal is 707

**After this PR**
The line's subtotal is 706.88

This modification borrows the triple discount computation from https://github.com/OCA/sale-workflow/blob/361d4b039f281fe9293e4dcde30f5a8ce9c805bb/sale_triple_discount/models/sale_order_line.py and puts everything in abstract model, in order to be reused from sale order line (PR coming soon) and purchase order line.

**todo**
- [x] Default discount type from partner (https://github.com/OCA/account-invoicing/pull/876#discussion_r586281779)
- [ ] ~~PR for purchase order~~ Does not make sense until the Odoo issue https://github.com/odoo/odoo/issues/67168 is solved (found during implementation of https://github.com/OCA/sale-workflow/pull/1495)
- [x] PR for sale order: https://github.com/OCA/sale-workflow/pull/1495

### DIFF
--- a/account_invoice_triple_discount/models/line_triple_discount.py
+++ b/account_invoice_triple_discount/models/line_triple_discount.py
@@ -1,0 +1,100 @@
+# Copyright 2017 Tecnativa - David Vidal
+# Copyright 2017 Tecnativa - Pedro M. Baeza
+# Copyright 2021 Simone Rubino - Agile Business Group
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+from odoo.addons import decimal_precision as dp
+
+
+class LineTripleDiscount (models.AbstractModel):
+    _name = "line.triple_discount.mixin"
+    _description = "Add triple discount fields and logic"
+
+    discount = fields.Float()
+    discount2 = fields.Float(
+        'Discount 2 (%)',
+        digits=dp.get_precision('Discount'),
+    )
+    discount3 = fields.Float(
+        'Discount 3 (%)',
+        digits=dp.get_precision('Discount'),
+    )
+    discounting_type = fields.Selection(
+        string="Discounting type",
+        selection=[
+            ('additive', 'Additive'),
+            ('multiplicative', 'Multiplicative'),
+        ],
+        default="multiplicative",
+        required=True,
+        help="Specifies whether discounts should be additive "
+        "or multiplicative.\nAdditive discounts are summed first and "
+        "then applied.\nMultiplicative discounts are applied sequentially.\n"
+        "Multiplicative discounts are default",
+    )
+
+    _sql_constraints = [
+        ('discount2_limit', 'CHECK (discount2 <= 100.0)',
+         'Discount 2 must be lower than 100%.'),
+        ('discount3_limit', 'CHECK (discount3 <= 100.0)',
+         'Discount 3 must be lower than 100%.'),
+    ]
+
+    def _get_final_discount(self):
+        self.ensure_one()
+        if self.discounting_type == "additive":
+            return self._additive_discount()
+        elif self.discounting_type == "multiplicative":
+            return self._multiplicative_discount()
+
+    def _additive_discount(self):
+        self.ensure_one()
+        discount = sum(
+            [getattr(self, x) or 0.0 for x in self._discount_fields()]
+        )
+        if discount <= 0:
+            return 0
+        elif discount >= 100:
+            return 100
+        return discount
+
+    def _multiplicative_discount(self):
+        self.ensure_one()
+        discounts = [1 - (self[x] or 0.0) / 100
+                     for x in self._discount_fields()]
+        final_discount = 1
+        for discount in discounts:
+            final_discount *= discount
+        return 100 - final_discount * 100
+
+    def _discount_fields(self):
+        return ['discount', 'discount2', 'discount3']
+
+    @api.multi
+    def triple_discount_preprocess(self):
+        """Save the values of the discounts in a dictionary,
+        to be restored in postprocess.
+        Resetting discount2 and discount3 to 0.0 avoids issues if
+        this method is called multiple times.
+        Updating the cache provides consistency through recomputations."""
+        prev_values = dict()
+        for line in self:
+            prev_values[line] = dict(
+                discount=line.discount,
+                discount2=line.discount2,
+                discount3=line.discount3,
+            )
+            line.update({
+                'discount': line._get_final_discount(),
+                'discount2': 0.0,
+                'discount3': 0.0
+            })
+        return prev_values
+
+    @api.model
+    def triple_discount_postprocess(self, prev_values):
+        """Restore the discounts of the lines in the dictionary prev_values.
+        Updating the cache provides consistency through recomputations."""
+        for line, prev_vals_dict in list(prev_values.items()):
+            line.update(prev_vals_dict)
--- a/account_invoice_triple_discount/models/line_triple_discount.py
+++ b/account_invoice_triple_discount/models/line_triple_discount.py
@@ -79,6 +79,14 @@ class LineTripleDiscount (models.AbstractModel):
         this method is called multiple times.
         Updating the cache provides consistency through recomputations."""
         prev_values = dict()
+
+        # The newly computed discount might have
+        # more digits than allowed from field's precision,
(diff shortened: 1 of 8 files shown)


account_invoice_triple_discount/models/line_triple_discount.py
```diff
@@ -0,0 +1,106 @@
...
+# Copyright 2017 Tecnativa - Pedro M. Baeza
+# Copyright 2021 Simone Rubino - Agile Business Group
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+from odoo.addons import decimal_precision as dp
+
+
+class LineTripleDiscount (models.AbstractModel):
+    _name = "line.triple_discount"
```
legalsylvain: could you rename to ``line.triple.discount.mixin`` to mention explicitely that it is an abstract model ?

account_invoice_triple_discount/models/line_triple_discount.py
```diff
@@ -0,0 +1,106 @@
...
+    discount = fields.Float()
+    discount2 = fields.Float(
+        'Discount 2 (%)',
+        digits=dp.get_precision('Discount'),
+    )
+    discount3 = fields.Float(
+        'Discount 3 (%)',
+        digits=dp.get_precision('Discount'),
+    )
+    discounting_type = fields.Selection(
```
legalsylvain, line 23: I don't understand where this field can be changed ? Should not be a setting on the related ``res.partner`` ? I mean some suppliers are using ``additive`` method and other ``multiplicative`` method.

Don't you think ?

account_invoice_triple_discount/models/line_triple_discount.py
SimoRubi: Thanks for the review! Done
EDIT: I actually kept `triple_discount` instead of `triple.discount` because I don't think they make sense if separated, hope that's not a problem

account_invoice_triple_discount/models/line_triple_discount.py
SimoRubi: This can be changed directly on the account invoice line's form, but it can be interesting to inherit that from the partner.
I'm currently investigating why Travis is still 🔴 (that's why this is still a Draft), after that I can try to add this new behavior.

HaraldPanten:
Might it be related to https://github.com/OCA/purchase-workflow/issues/834 and my comment in https://github.com/OCA/account-invoicing/pull/733 ?

SimoRubi:
> Might it be related to [OCA/purchase-workflow#834](https://github.com/OCA/purchase-workflow/issues/834) and my comment in #733 ?

It might be, I didn't know purchase's version of triple discount but looks like every module is re-implementing that logic in its own way.
This PR aims to centralize the logic in the new mixin class that will then be inherited from lines of `account.invoice`, `sale.order` and `purchase.order`.

HaraldPanten:
> > Might it be related to [OCA/purchase-workflow#834](https://github.com/OCA/purchase-workflow/issues/834) and my comment in #733 ?
> 
> It might be, I didn't know purchase's version of triple discount but looks like every module is re-implementing that logic in its own way.
> This PR aims to centralize the logic in the new mixin class that will then be inherited from lines of `account.invoice`, `sale.order` and `purchase.order`.

OK!

account_invoice_triple_discount/models/line_triple_discount.py
SimoRubi: While implementing this feature I noticed that it is not as trivial as it seems because default value for invoice line cannot rely on the line's fields (default methods are `api.model`), so I'd need to save it in the context etc..
I think that deserves its own PR, and here let's focus on the described issue, is that ok for you?

BTW Now that Travis is 🟢, I'll mark this PR as ready for reviews

legalsylvain: well, I was more thinking on storing the discounting_type only on the ``res.partner`` or at least on the order/invoice method.
it will be a lighter implementation.
Do you think there are real cases when there are some line with ``additive`` method, and other with ``multiplicative`` ?

I mean, setting this value 50 times, if you have 50 lines will be a mess, in a UX point of view.

account_invoice_triple_discount/models/line_triple_discount.py
legalsylvain: between triple_discount and triple.discount : not a blocking point. matter of taste ;-)

account_invoice_triple_discount/models/line_triple_discount.py
SimoRubi: That is why I was talking about a default value for new lines: each of the 50 lines would have the default value inherited from the partner; and switching partner in the invoice would trigger the edit of all the invoice's lines.

I don't know about real cases of lines having different kinds of discounts but this is how sale orders + triple_discount currently work and it would be nice to have it in invoices too.

SimoneVagile (approved):

account_invoice_triple_discount/models/line_triple_discount.py
SimoRubi: I added the logic explained above, can you update your review?

SimoRubi:
@legalsylvain can you update your review?

Thanks!

SimoRubi:
Last commit solves the same issue as https://github.com/OCA/sale-workflow/pull/1618

SimoneVagile (approved):

SimoRubi:
@OCA/accounting-maintainers can this be merged or do I have to fix anything?
Thanks!

tafaRU:
@SimoRubi could you please have a look at https://github.com/SimoRubi/account-invoicing/pull/1 ?

SimoRubi:
@OCA/accounting-maintainers can this be merged or do I have to fix anything?
Thanks!

SimoRubi:
> @SimoRubi could you please have a look at [SimoRubi#1](https://github.com/SimoRubi/account-invoicing/pull/1) ?

I had a look and I think that cannot be merged as explained in https://github.com/SimoRubi/account-invoicing/pull/1#pullrequestreview-744019757

SimoRubi:
@OCA/accounting-maintainers can this be merged or do I have to fix anything?
Thanks!

sergioeix:
So this is happening on v13 #733 and i guess it happens on v14 too??

francesco-ooops:
@legalsylvain could you review this one again? thanks

account_invoice_triple_discount/models/line_triple_discount.py
legalsylvain: > I don't know about real cases of lines having different kinds of discounts but this is how sale orders + triple_discount currently work and it would be nice to have it in invoices too.

1) Well if it's not a use case, it should be changed I think. Don't you ? 


2) the problem with the current implementation is that it will create inconsistencies. I mean, it's possible to have "multiplicative" and invoice level and "additive" on line level.

3) you are adding new field on invoice / invoice.line. don't you think you should add a pre-migration script to populate correctly the database. I fear that an update of the current module on a production database will be slow. Don't you think ?

legalsylvain (changes requested):
Regarding Point 1, I think it can be changed later, but could you write a bullet in the ``ROADMAP.rst`` file to mention that topic ?
Also write a bullet to say that sale_triple_discount should be refactored to use new mixin introduced in that module.

https://github.com/OCA/sale-workflow/blob/14.0/sale_triple_discount/__manifest__.py

Regarding point 2 & 3, let me know your point of view.

francesco-ooops:
@SirTakobi can you take a look at the recent review?

SirTakobi:
> @SirTakobi can you take a look at the recent review?

Thanks for letting me know of https://github.com/OCA/account-invoicing/pull/876#pullrequestreview-1041453265.

I can't edit this PR because it is based on @SimoRubi's branch.
I'm currently not working on this topic so I won't be looking into this in the near future, if you want to make improvements or go on with the PR you can simply create another PR starting from https://github.com/SimoRubi/account-invoicing/tree/12.0-fix-account_invoice_triple_discount-subtotal_computation

PR #877, opened 2021-03-05
[13.0][FIX] purchase_stock_picking_return_invoicing create bill

qgroulard:
The context key 'create_bill' is currently ignored in the extension of 'action_view_invoice'.
Therefore it is not possible to create new invoices from a purchase order if it is already linked to one 'in_invoice' invoice.

This is an issue when you receive i.e. half of your order and bill this half.
Then when you receive the other half and want to bill it the 'Create bill' button will open the first invoice rather than a new one.


pedrobaeza:
I think #956 solves the problem in a simpler way.

PR #845, opened 2021-01-12
[MIG][14.0] account_invoice_force_number

SimoneVagile:

PR #880, opened 2021-03-06
[MIG][14.0] account invoice_force_number: Migration to 14.0

chusamo:
I've just realised that in Odoo 14 we could edit the name if we validate the invoice and set to draft.
I don't know if this module is usefull in this version, however I let the PR opened.


CasVissers-360ERP:
Double PR, see #845

chusamo:
> Double PR, see #845

Ups, thank you

PR #892, opened 2021-03-16
[13.0][MIG] sale_order_invoicing_queued: Change account.invoice model to account.move for stored queue jobs

sergio-teruel:
Script Openupgrade
cc @Tecnativa
ping @carlosdauden @pedrobaeza

PR #893, opened 2021-03-17
[13.0][FIX] account_billing: Prevent error in tests related to sale journal not exist

victoralmau:
Prevent error in tests related to sale journal not exist.
It's related to https://github.com/OCA/account-invoicing/pull/892

Please, @joao-p-marques and @CarlosRoca13 can you review it?.
@Tecnativa TT28825

### DIFF
--- a/account_billing/tests/test_account_billing.py
+++ b/account_billing/tests/test_account_billing.py
@@ -57,6 +57,9 @@ class TestAccountBilling(SavepointCase):
         cls.journal_purchase = cls.env["account.journal"].create(
             {"name": "Purchase", "type": "purchase", "code": "PURCHASE"}
         )
+        cls.journal_sale = cls.env["account.journal"].create(
+            {"name": "Sale", "type": "sale", "code": "SALE"}
+        )
 
         cls.inv_1 = cls.create_invoice(
             cls, amount=100, currency_id=cls.currency_eur_id, partner=cls.partner_id.id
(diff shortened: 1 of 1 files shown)


joao-p-marques (approved):

CarlosRoca13 (approved):

chienandalu (approved):

kittiu:
@victoralmau Do we actually need cls.journal_sale in test? I don't see it's use anywhere.

joao-p-marques:
> @victoralmau Do we actually need cls.journal_sale in test? I don't see it's use anywhere.

@kittiu Indeed, the variable is not used. However, it needs to be created to ensure a Sale type Journal exists in the environment and, when creating the invoice, one will be automatically assigned (whether it is the previously existing one or the newly created).

So, I'd say it is needed, but not mandatory to store it in a variable.

kittiu:
> > @victoralmau Do we actually need cls.journal_sale in test? I don't see it's use anywhere.
> 
> @kittiu Indeed, the variable is not used. However, it needs to be created to ensure a Sale type Journal exists in the environment and, when creating the invoice, one will be automatically assigned (whether it is the previously existing one or the newly created).
> 
> So, I'd say it is needed, but not mandatory to store it in a variable.

IC, so I think we could remove this line too.
https://github.com/OCA/account-invoicing/pull/893/files#diff-e4574fd2fb19093235a878ffecd917756e12503d4679a025d726b37931473a29L17

kittiu (approved):

PR #895, opened 2021-03-22
[13.0][MIG+IMP] account_invoice_date_due: Migration to 13.0 and refactor

joao-p-marques:
1. Migration
2. Refactor to use an editable field instead of a wizard
3. Add tests

@Tecnativa
TT28875

ping @victoralmau @pedrobaeza


account_invoice_date_due/models/__init__.py
```diff
@@ -0,0 +1,3 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from . import update_due_date
```
victoralmau:
```suggestion
from . import account_move
```

victoralmau (changes requested):

victoralmau (approved):

chienandalu (commented):
There's a travis error. I'm not sure it's related to this PR though...

joao-p-marques:
> There's a travis error. I'm not sure it's related to this PR though...

Thanks @chienandalu Indeed, it was due to the `onchange` I added, should be fixed now.

joao-p-marques:
@chienandalu I think I have fixed the handling of the `invisible`/`readonly` situations, integrating the standard Odoo mechanism for handling those fields. Can you check, please?

account_invoice_date_due/views/account_move_date_due.xml
```diff
@@ -0,0 +1,20 @@
...
+    <record id="view_move_form" model="ir.ui.view">
+        <field name="name">account.move.form</field>
+        <field name="model">account.move</field>
+        <field name="inherit_id" ref="account.view_move_form" />
+        <field name="arch" type="xml">
+            <field name="invoice_date_due" position="before">
+                <field name="is_invoice_date_due_editable" invisible="1" />
+            </field>
+            <field name="invoice_date_due" position="attributes">
+                <attribute name="readonly">0</attribute>
```
chienandalu: Wouldn't it be easier to play with fields declared multiple times and positive/negative groups? Something like this for the regular field:
```xml
<attribute name="groups">!account_invoice_date_due.allow_to_change_due_date</attribute>
```
And then:
```xml
<field name="invoice_date_due" position="after">
    <field name="invoice_date_due" groups="account_invoice_date_due.allow_to_change_due_date" attrs="{'readonly': your domain, 'invisible': another domain}" />
</field>
```

account_invoice_date_due/models/account_move.py
```diff
@@ -0,0 +1,28 @@
...
+        compute="_compute_is_invoice_date_due_editable"
+    )
+
+    def _compute_is_invoice_date_due_editable(self):
+        user_in_group = (
+            self.env.user
+            in self.env.ref("account_invoice_date_due.allow_to_change_due_date").users
+        )
+        for rec in self:
+            if user_in_group and rec.state == "draft":
```
chienandalu: This is wrong, isn't it?

chienandalu (changes requested):

joao-p-marques:
Thank you for the suggestions @chienandalu 
I think now it is correct :+1: 

I just have a question: should we make the `invoice_payment_term_id` field also editable after the invoice has been posted? The date field will be invisible if a payment term is set, so it wouldn't be possible to change the value.

chienandalu:
> I just have a question: should we make the invoice_payment_term_id field also editable after the invoice has been posted? The date field will be invisible if a payment term is set, so it wouldn't be possible to change the value.

What could you do in the previous version?

joao-p-marques:
> What could you do in the previous version?

With the wizard, you could change the due date even if a different payment term had been set.

pedrobaeza:
It's a difficult question, as letting the payment term to be editable will make very difficult to match previous AR/AP lines with the corresponding due dates, but also not letting to edit anything will make this to have less features.

Can you make due date always editable once confirmed no matter the initial data and propagate the date?

joao-p-marques:
Yes, but the problem is keeping the Odoo mechanism to hide the Due date 
field once a payment term is selected... You need to remove that 
payment term to further edit the due date field.

chienandalu:
> Yes, but the problem is keeping the Odoo mechanism to hide the Due date field once a payment term is selected... You need to remove that payment term to further edit the due date field.


But you're using a diferent field declaration in the view for the proper group, so it should not affect it, or does it?

joao-p-marques:
> But you're using a diferent field declaration in the view for the proper group, so it should not affect it, or does it?

Yes, but I kept the declaration of the `invisible` attribute on both of them, to keep the original mechanism. Another option is to make it always visible in another field in the view.

joao-p-marques:
@chienandalu @pedrobaeza I just added some changes with a slightly different flow, which IMHO is more clear. It is documented under USAGE and added in the tests. Can you check again please?

account_invoice_date_due/models/account_move.py
```diff
@@ -0,0 +1,24 @@
...
+    # Show invoice date due even when payment term is defined
+    invoice_date_due_payment_term = fields.Date(
+        related="invoice_date_due", string="Due Date Payment Term"
+    )
+
+    @api.onchange("invoice_date_due")
+    def _onchange_invoice_date_due(self):
+        # Set due date on account move lines
+        if self.invoice_date_due:
+            self.mapped("line_ids").date_maturity = self.invoice_date_due
```
chienandalu: In my functional test, I changed the due date of the invoice but the maturity date wasn't set. Any way, shouldn't you set it only for the debit lines?

joao-p-marques:
@chienandalu changes done

account_invoice_date_due/models/account_move.py
```diff
@@ -0,0 +1,30 @@
...
+    def _onchange_invoice_date_due_payment_term(self):
+        """Propagate from Payment term due date to original field"""
+        if self.invoice_date_due_payment_term:
+            self.invoice_date_due = self.invoice_date_due_payment_term
+
+    def write(self, vals):
+        res = super().write(vals)
+        # Propagate due date to move lines
+        # that correspont to the receivable/payable account
+        if "invoice_date_due" in vals and self.state in ("draft", "posted"):
```
chienandalu: I'd say you won't have payment term lines when the invoice is in draft
```suggestion
        if "invoice_date_due" in vals and self.state == "posted":
```

chienandalu (commented):
Just a little detail

account_invoice_date_due/models/account_move.py
joao-p-marques: Yes, it was just that, from my testing, the correspondent move lines that should be updated are created already while in draft mode. 
Should I just ignore the values while in draft and change only when posted?

chienandalu: The thing is that when you're in draft you'll work with the standard behavior. The edition of the due date should be allowed after posting the invoice

joao-p-marques: Ok, I think I was looking at the problem wrong :sweat_smile: 
I removed that condition following your suggestion and also removed the check from the tests (I now only check the propagation after having posted the invoice)

chienandalu (approved):

pedrobaeza (changes requested):
Trying on runbot, I have posted an invoice, and I'm not able to edit the due date:

Selección_007

joao-p-marques:
@pedrobaeza You need to add the user to the group that allows editing that field, as happened in v12:
image

I've tested again in the runbot and everything seems to work

pedrobaeza (approved):
Oh, stupid mistake not seeing the group.

PR #896, opened 2021-03-23
[12.0][IMP+FIX] Return Picking Invocing Process and mapping Invoice Type

mbcosta:
Included in Return Picking Wizard option to select if should be Invocing, fix mapping Invoice Type and test this process.

image

cc @renatonlima @rvalyi

### DIFF
--- a/stock_picking_invoicing/__manifest__.py
+++ b/stock_picking_invoicing/__manifest__.py
@@ -14,6 +14,7 @@
     ],
     "data": [
         "wizards/stock_invoice_onshipping_view.xml",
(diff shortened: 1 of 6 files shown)


renatonlima (approved):

marcelsavegnago (approved):
LGTM

rvalyi (approved):
well done @mbcosta !

rvalyi:
@mbcosta kudos also for the SavepointCase test move. As we have 12 tests in the same file it really makes less DB requests and saves some time.

PR #897, opened 2021-03-24
[14.0][MIG] sale_timesheet_invoice_description

clementmbr:
Basic migration adapting the tests and improving the check to catch sale.order.line's timesheets.


kittiu:
Here may be the fix, https://github.com/OCA/account-invoicing/pull/910/files by @JordiBForgeFlow 
I will make the cherry pick to fix account_billing.

rvalyi:
@clementmbr can you rebase please? the offending test has been fixed in #911

bosd:
I'm actually using this module. Works fine!! :+1: 
Lets get this one merged!
ping @clementmbr

clementmbr:
@kittiu @bosd @rvalyi sorry for the late reply, I made the rebase!

bosd (approved):
Migration LGTM!
I am using this in production.
I've encoutered a problem with this module. When there is no invoiceline descirption selected on a partner, it throws an internal server error.
 Imo, it would be better to have a default description selected, or an usefull error message.
  
 need to do some further testing on other versions and installs,as it might occure there as well. Will open a new issue, when I have more info. 
it is no stopper for the migration :)

bosd:
Im having problems on a database that timesheet lines get unlinked from the invoice.

Not sure if it is related to this pr. Or one of the other modules that are installed like tinesheet_sheet.

Will report back when more info is available.

bosd:
went trough the code again. The problem is likely not caused by this module, but in Odoo native.
@clementmbr  Does it need a rebase again before it can be merged?

bosd:
my bad. Did'nt notice there is a superseding pr.
https://github.com/OCA/account-invoicing/pull/1078

This one can be closed now.

jordancoest (approved):

JulianaPoudou (approved):

rvalyi:
@clementmbr can you check? I think https://github.com/OCA/account-invoicing/pull/1078 doesn't include your improvements...

pedrobaeza:
Already migrated.

PR #900, opened 2021-03-25
[12.0][IMP] account_global_discount: Set to zero global discount.

alfredoavanzosc:
invoice_global_discount_ids is correctly emptied but amount_global_discount and amount_untaxed_before_global_discounts are not being recomputed or set to zero, so the global discount amount is staying on invoice causing incorrect amount.

### DIFF
--- a/account_global_discount/models/account_invoice.py
+++ b/account_global_discount/models/account_invoice.py
@@ -54,6 +54,8 @@ class AccountInvoice(models.Model):
         """
         self.ensure_one()
         if not self.global_discount_ids:
+            self.amount_global_discount = 0.0
+            self.amount_untaxed_before_global_discounts = 0.0
             return
         invoice_global_discounts = self.env['account.invoice.global.discount']
         taxes_keys = {}
@@ -131,7 +133,9 @@ class AccountInvoice(models.Model):
         return self._onchange_invoice_line_ids()
 
     def _compute_amount_one(self):
-        if not self.invoice_global_discount_ids:
+        if not self.global_discount_ids:
+            self.amount_global_discount = 0.0
+            self.amount_untaxed_before_global_discounts = 0.0
             return
         round_curr = self.currency_id.round
         self.amount_global_discount = sum(


alfredoavanzosc:
@pedrobaeza , @alexis-via , review please.

pedrobaeza:
It should be @anajuaristi who should check if this fixes her problem.

anajuaristi:
Hi Pedro. Tested in
http://3473816-900-65a3b0.runbot1.odoo-community.org/?

Invoiced sale order with global discount --> On invoice:

Untaxed Amount Before Disc. | $ 37,500.00
-- | --
Global Discounts | $ -3,750.00
Untaxed Amount | $ 33,750.00
Tax | $ 5,062.50
Total | $ 38,812.50

After erasing the discount:
Untaxed Amount | $ 33,750.00
-- | --
Tax | $ 5,625.00
Total | $ 39,375.00

So, it seems the fix is not solving the problem.

The fields are not being recomputed pushing any button. but... By the moment I have found a simple way to set those fields to Zero, and it is... 
 - just define a new global discount 0%
 - Asign that discount to invoice --> Fields are perfectly recalculated
 - Erase 0% discount
So.. functionally it works but technically the error is still there. We will see it again next week.
Thank you:
Ana

anajuaristi:
I think it could be the same fix in line 50 we can see
```
    def _set_global_discounts_by_tax(self):
        """Create invoice global discount lines by taxes combinations and
        discounts.
        """
        self.ensure_one()
        if not self.global_discount_ids:
            return
```

Maybe it should also be:
```
    def _set_global_discounts_by_tax(self):
        """Create invoice global discount lines by taxes combinations and
        discounts.
        """
        self.ensure_one()
        if not self.invoice_global_discount_ids:
            self.amount_global_discount = 0.0
            self.amount_untaxed_before_global_discounts = 0.0
            return
```
We will try with this next week

greciosd:
Hello.

I recently used the latest version of the account_global_discount module for version 12 and got this error.
Is there any news about its correction?

Thanks.

MPoch-PlanetaTIC (approved):
I approve it.

anajuaristi (approved):

anajuaristi:
@pedrobaeza 
I also approved the PR. Let's see if it can be merged

pedrobaeza:
OK, I'm not using anymore 12.0, although it seems a bit weird, as pre-v13 resets the values of each compute to the default one, but it doesn't hurt:

/ocabot merge patch

PR #902, opened 2021-03-29
[12.0][FIX] remove from invoice lines with no quantity

sergiocorato:


bodedra (commented):
Instead of unlink, how about not include it during prepare invoice line?

sergiocorato:
> Instead of unlink, how about not include it during prepare invoice line?

I didn't found the method which prepare invoice lines, can you link it?

bodedra:
How about following?

In purchase module, purchase_order_change() method

Ref: https://github.com/odoo/odoo/blob/12.0/addons/purchase/models/account_invoice.py#L91

pedrobaeza:
This must be opt-in, as people may want to have them. It can be an option at the wizard level.

sergiocorato:
> This must be opt-in, as people may want to have them. It can be an option at the wizard level.

Done, thanks

pedrobaeza (approved):
Thanks for the feature.

bodedra (approved):
LG :+1:

pedrobaeza:
/ocabot merge minor

Does this apply to v13 as well?

PR #908, opened 2021-04-06
[14.0][MIG] account_invoice_date_due: Migration to 14.0

joao-p-marques:
@Tecnativa 
TT28729

ping @pedrobaeza @victoralmau @chienandalu


account_invoice_date_due/models/account_move.py
```diff
@@ -17,6 +17,22 @@ def _onchange_invoice_date_due_payment_term(self):
...
+        # Avoid recomputation of amount fields when we are only changing the
+        # invoice's due date
+        if self.env.context.get("bypass_compute_amount"):
+            return
+        return super()._compute_amount()
+
+    def onchange(self, values, field_name, field_onchange):
+        obj = self
+        if field_name == "invoice_date_due":
+            # A change in the invoice date due would trigger a recompute of the
```
chienandalu, line 30: Should we constraint this to the `posted` state?

joao-p-marques: I was just writing a question about this :smile: 

I had some issues due to the fact that the change of the date due after the invoice had been posted triggered a recompute of the amounts in the lines and was marking the invoice as paid. I don't know if this is the correct fix though, as what I observed was that the residual amount was being computed as 0.0 after the change of the due date (See https://github.com/odoo/odoo/blob/14.0/addons/account/models/account_move.py#L1316)

The constraint you mention makes sense, though, if we follow this approach.

chienandalu: I'd say so. Otherwise you're blocking the expected standard behavior. After that, as in core it's not possible to change the due date from the invoice, we must prevent to trigger that recompute: https://github.com/odoo/odoo/blob/14.0/addons/account/models/account_move.py#L491

joao-p-marques: Ok, I made the suggested change of limiting to the `posted` state.

I also pushed another commit to fix the tests of `account_billing`, that were failing due to the change in Odoo from a couple of weeks ago of limiting the journal types for the specific invoice types.

pedrobaeza:
Can you please solve the conflict?

joao-p-marques:
@pedrobaeza done. I had a commit that fixed the tests in `account_billing`, but someone did it in the main branch as well, so I just removed it here.

pedrobaeza:
Travis is red

joao-p-marques:
Thanks for the warning @pedrobaeza 
The tests in `account_billing` were failing after https://github.com/odoo/odoo/commit/fcaa54939e9a4f0dd5e47cd0ccffe7aa24bd451c
Should be fixed now.

Issue #914, opened 2021-04-14
[12.0] account_invoice_triple_discount Discount 2 and 3 missing in invoice view/report but are calculated correctly


antonionardella:
Hello,

a SO with three discounts as shown here:
![image](https://user-images.githubusercontent.com/7383572/114708703-fbe25200-9d2b-11eb-8988-6f464a3003c4.png)

does not keep the discount 2 and 3 in the invoice view and report
![image](https://user-images.githubusercontent.com/7383572/114708779-1ae0e400-9d2c-11eb-918e-e3f6adab44fb.png)
 
Although the discounts are correctly calculated

legalsylvain:
Hi. This module hide the values, if you're are not member of the group ``base.group_no_one``.

https://github.com/OCA/account-invoicing/blob/12.0/account_invoice_triple_discount/views/account_invoice_view.xml#L11

I think that is not good design, (and not consistent with the rest of Odoo).

Could you make a PR, removing the groups fields ? 

thanks !

PR #370, opened 2018-04-04
[11.0][MIG] account_invoice_triple_discount

xavierjimenez:
Migration of account_invoice_triple_discount module to version 11.0

PR #915, opened 2021-04-16
[12][account_invoice_triple_discount] removed groups="base.group_no_one" from all lines

antonionardella:
Solves issue https://github.com/OCA/account-invoicing/issues/914

### DIFF
--- a/account_invoice_triple_discount/views/account_invoice_view.xml
+++ b/account_invoice_triple_discount/views/account_invoice_view.xml
@@ -8,8 +8,8 @@
         <field name="arch" type="xml">
             <xpath expr="//field[@name='invoice_line_ids']//tree//field[@name='discount']"
                    position="after">
-                <field name="discount2" groups="base.group_no_one"/>
-                <field name="discount3" groups="base.group_no_one"/>
+                <field name="discount2"/>
+                <field name="discount3"/>
             </xpath>
         </field>
     </record>
@@ -22,8 +22,8 @@
         <field name="arch" type="xml">
             <xpath expr="//field[@name='invoice_line_ids']//tree//field[@name='discount']"
                    position="after">
-                <field name="discount2" groups="base.group_no_one"/>
-                <field name="discount3" groups="base.group_no_one"/>
+                <field name="discount2"/>
+                <field name="discount3"/>
             </xpath>
         </field>
     </record>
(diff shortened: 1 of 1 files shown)


legalsylvain (commented):
Hi. 

I personally agree with that move. I think it makes no sense to have this group.

in the other way it is not consistent with odoo core account module.

See : https://github.com/OCA/account-invoicing/pull/370#discussion_r179718037
And : original code : https://github.com/odoo/odoo/blob/12.0/addons/account/views/account_invoice_view.xml#L55

@pedrobaeza : any PoV ? 

thanks.

pedrobaeza:
If original `discount` field has such group, it's not homogeneous not to put it, but I think `sale` module adds another specific group for putting discounts (`sale.group_discount` or similar), that on v13, it's transferred to `account` module. `sale_triple_discount` can fill the gap as well.

cscocca:
Yes @pedrobaeza it's right, the invoice discount attributes (for all views) changed here in sale module:

https://github.com/odoo/odoo/blob/12.0/addons/sale/views/sale_views.xml#L1053.

with new group `sale.group_discount_per_so_line`

So these attributes has to be changed also for `discount2 `and `discount3 `in `sale_triple_discount ` module.

If I'll have time I'll do a PR as soon as possible, but if anyone can be before it is appreciated

sergiocorato (approved):

sergiocorato:
I think this fields should be visible if a specific group does not exist.
If sale is installed with this [PR](https://github.com/OCA/sale-workflow/pull/2297) `sale.group_discount_per_so_line` can be used.
@pedrobaeza @legalsylvain do you think this is acceptable?

pedrobaeza:
The discount field not being in v12 in the account module makes the thing difficult, so you decide what you prefer.

cscocca:
> I think this fields should be visible if a specific group does not exist. If sale is installed with this [PR](https://github.com/OCA/sale-workflow/pull/2297) `sale.group_discount_per_so_line` can be used. @pedrobaeza @legalsylvain do you think this is acceptable?

I think this is an acceptable solution but if sale is not installed then also original `discount` fields should haven't the group `base.group_no_one`, otherwise a standard user will see just `discount2 `and `discount3 ` without the main `discount`.

legalsylvain (approved):

rafaelbn (approved):

PR #918, opened 2021-04-20
 [FIX] account_invoice_merge: compatibility with account_invoice_margin module (OCA/margin-analysis repository)

legalsylvain:
**Context**

When both ``account_invoice_merge`` and ``account_invoice_margin`` (OCA/margin-analysis) are installed, the ``purchase_price`` of ``account.invoice.line`` is lost, when merging invoices, so the invoice (lines) margin are bad.

Ref : apps/deck/#/board/144/card/1401
CC : @quentinDupont 

CC : authors of both modules : @sergio-teruel, @pedrobaeza, @Cedric-Pigeon, @mourad-ehm

### DIFF
--- a/account_invoice_merge/models/account_invoice.py
+++ b/account_invoice_merge/models/account_invoice.py
@@ -28,7 +28,10 @@ class AccountInvoice(models.Model):
             'product_id', 'account_id', 'account_analytic_id',
             'uom_id'
         ]
-        for field in ['sale_line_ids']:
+        for field in [
+                'sale_line_ids',        # odoo/sale
+                'purchase_price',       # OCA/account_invoice_margin
+        ]:
             if field in self.env['account.invoice.line']._fields:
                 fields.append(field)
         return fields


pedrobaeza:
I refuse to use this `merge` modules, due to all these problems, and needing to merge is because a process before has failed in its purpose and workflow.

legalsylvain:
Sorry @pedrobaeza. I pinged you because I saw you as coauthor of one of the both modules. Sorry for the noice.

Side question : just to understand, how do you handle cases, when user generated invoices from sales (or purchase) without correct option, and then want to merge it ? do you ask to cancel all invoices, and generate new ones ?

legalsylvain:
Sorry @pedrobaeza. I pinged you because I saw you as coauthor of one of the both modules. Sorry for the noice.

Side question : just to understand, how do you handle cases, when user generated invoices from sales (or purchase) without correct option, and then want to merge it ? do you ask to cancel all invoices, and generate new ones ?

pedrobaeza:
I would say I have never been as coauthor of this one... and looking at the commit history, my intervention has only been on administrative commits (rename manifest, PEP8, make uninstallable, etc), but not a problem to explain you or to accept this if you need reviewers.

About the flow question, when invoices are incorrectly generated, the solution is to delete them, and regenerate from sales order, grouping or taking whole quantities into account.  Do you see any problem in such flow? Anyways, the merge option can't be applied over already validated invoices.

legalsylvain:
>  Do you see any problem in such flow?

In a first sight, no. I think it's maybe the better solution in V12.

The single minor problem I see is user experience. I think that is easier to select invoices and merge them, than select invoices, delete them, go to sale (/purchase) orders, select orders, invoice them. 

Anyway, thanks for your comments !

pedrobaeza:
The problem is that such merged invoices may have other potential side effects (like the one you are finding), so better to avoid them explaining the other flow. In my country, it's called "mejor una vez roja, que cientos coloradas" :stuck_out_tongue_winking_eye: 

About the design of the module, I would even prefer to use `openupgradelib.openupgrade_merge_records` than specific code, but depending on the business logic, default field updates strategy may not merge them, and you need to do adjusts anyway.

JordiADARIA:
> The problem is that such merged invoices may have other potential side effects (like the one you are finding), so better to avoid them explaining the other flow. In my country, it's called "mejor una vez roja, que cientos coloradas" 😜
> 
> About the design of the module, I would even prefer to use `openupgradelib.openupgrade_merge_records` than specific code, but depending on the business logic, default field updates strategy may not merge them, and you need to do adjusts anyway.

Hi Pedro, i think youre right, the problem with the merge invoices module is that it's merging "default" fields ( a determinated number of fileds) on invoice lines, and we'll find the same problem with any other module that adds on the future new fields on account invoice lines... as far as i can see, the openupgrade_merge_records.py code is trying to merge all the fields that exist on the model, and it'll be a better way to merge anything... 

@legalsylvain , did you check the code proposal from @pedrobaeza ?

legalsylvain:
> @legalsylvain , did you check the code proposal from @pedrobaeza ?

Absolutely not. my PR just solves a trivial compatibility between both modules, in the same design than before.
It is not in the scope to make a full refactor, even it could be interesting. 

Feel free to make another PR, if you're interested by introducing calls of openupgradelib functions.

In the meantime, could you review this PR ? Thanks !

pedrobaeza (approved):
Code review and for favoring the quick merge, but better to avoid the module as stated!

PR #927, opened 2021-05-04
[13.0][IMP] purchase_stock_picking_return_invoicing: New field to avoid compute quantity invoiced

carlosdauden:
@Tecnativa TT28557


purchase_stock_picking_return_invoicing/tests/test_purchase_stock_picking_return_invoicing.py
```diff
@@ -150,6 +150,10 @@ def test_purchase_stock_return_1(self):
         self.assertEqual(action["res_id"], inv_1.id)
         action2 = self.po.action_view_invoice_refund()
         self.assertEqual(action2["res_id"], inv_2.id)
+        # Test avoid_compute_qty_invoiced
+        inv_2.avoid_compute_qty_invoiced = True
+        inv_2.flush()
```
pedrobaeza, line 155: If you need to flush, this means something is not exactly correct in your depends.

pedrobaeza (changes requested):
Isn't this mechanism something similar to `purchase_force_invoiced`? Please explain why doing this

PR #912, opened 2021-04-13
[13.0] [FIX] account_invoice_fiscal_position_update: Filter only lines without comments or notes.

nicomacr:
Here if you create a note or a comment in the invoice lines, and you change the fpos an error occurs.
After this changes only non display type lines are taking into account.

PR #928, opened 2021-05-05
[FIX] account_invoice_fiscal_position_update: Mark to recompute line for taxes when change fpos

nicomacr:
If you have products with taxes, then when change the fpos, the aml taxes lines currently are not change, after this changes we force to create new aml for the correct taxes.

### DIFF
--- a/account_invoice_fiscal_position_update/models/account_invoice.py
+++ b/account_invoice_fiscal_position_update/models/account_invoice.py
@@ -48,11 +48,12 @@ class AccountMove(models.Model):
                     taxes = fp.map_tax(taxes)
 
                 line.tax_ids = [(6, 0, taxes.ids)]
+                line._onchange_mark_recompute_taxes()
 
                 line.account_id = account.id
             else:
                 lines_without_product.append(line.name)
-        self._recompute_dynamic_lines(recompute_tax_base_amount=True)
+        self.with_context(check_move_validity=False)._recompute_dynamic_lines()
 
         if lines_without_product:
             res["warning"] = {"title": _("Warning")}
--- a/account_invoice_pricelist/tests/test_account_move_pricelist.py
+++ b/account_invoice_pricelist/tests/test_account_move_pricelist.py
@@ -193,6 +193,7 @@ class TestAccountMovePricelist(common.SavepointCase):
         cls.invoice = cls.AccountMove.create(
             {
                 "partner_id": cls.partner.id,
+                "fiscal_position_id": cls.fiscal_position.id,
                 "type": "out_invoice",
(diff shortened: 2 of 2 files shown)


nicomacr:
ping @OCA/accounting-maintainers

HaraldPanten (approved):
It's functionally working fine.

@pedrobaeza could you have a look at this one, please? Without this patch, the module is missing an important part, as it's not recomputing taxes amount.

THX!

pedrobaeza:
Travis is red

HaraldPanten:
@pedrobaeza it seems that it's not related to this PR but account_invoice_pricelist. WDYT?

ValentinVinagre (approved):
LGTM 👍🏻 travis is failing by 'account_invoice_pricelist' module

pedrobaeza:
Yeah, but we shouldn't merge anything in a red branch, or errors will stack up.

HaraldPanten:
> Yeah, but we shouldn't merge anything in a red branch, or errors will stack up.

How can we solve this? As this FIX needs to be done

pedrobaeza:
You should greenify the branch.

HaraldPanten:
@nicomacr Can you solve this?

nicomacr:
> @nicomacr Can you solve this?

Yes i' ve already solved

alexis-via (approved):

alexis-via:
I made some tests on v13 with this PR:
- it works well for the account_id: it is proprely updated when the FP is changed
- it doesn't work well for taxes: when FP is changed, taxes are removed instead of being replaced.
- there is an error when there is a section line ; I guess this is fixed in your another PR: https://github.com/OCA/account-invoicing/pull/912

alexis-via:
I made a PR for v14 which seems to work (at last !!!) https://github.com/OCA/account-invoicing/pull/963
As you can see in the v14 code, I call both _onchange_mark_recompute_taxes() as you do, and also self.with_context(check_move_validity=False)._recompute_dynamic_lines() and then self._check_balanced(). I guess these methods are the same in v13 (I didn't check)

jjscarafia:
@nicomacr I believe that maybe this PR is working for us because we have [this revert](https://github.com/odoo/odoo/pull/67632) on our instances, isn't it?
We use that revert to make https://github.com/ingadhoc/account-invoicing/tree/13.0/account_invoice_tax keep working.

Maybe the new approach of @alexis-via could avoid the need of the revert?

nicomacr:
@alexis-via yes that approach works for this version too.

alexis-via:
@nicomacr OK, good news.

PR #929, opened 2021-05-06
[13.0][FIX] sale_order_invoicing_grouping_criteria: Addons compatibility

victoralmau:
Be sure return `account.move` (it's possible to return MagicMock according test from other addons: `sale_automatic_workflow`)

Please @pedrobaeza and @CarlosRoca13 

@Tecnativa TT29607


pedrobaeza:
This is a problem of the other addon, not this one. Such `MagicMock` thing should never be returned.

PR #932, opened 2021-05-08
[14.0][ADD] account_sequence_option

kittiu:
Base on https://github.com/OCA/server-tools/pull/2080


remi-filament (changes requested):
This module works fine, however it is missing data to be imported (there are a lot of configuration options that makes it very difficult to create those manually - readonly on model in base.sequence.option, create="0" on tree views, etc.)
But you probably know since it is still marked as WIP !
Thank you for your work

remi-filament:
Hi @kittiu  testing further this module, when trying to validate a new invoice, I have an error on constraint validation (function _constrains_date_sequence from account_move.py)
I have found a workaround to bypass this constraint validation if sequence is defined (I have pushed a PR on your repo here : https://github.com/ecosoft-odoo/account-invoicing/pull/1 ), let me know if it makes sense for you !

kittiu:
My intention is for each document types, user will not create base.sequence.option manually. And so we will need a module like this to do it.

I..e, if you want to manage sequence of purchase.order, you may create a module purchase_sequence_option for it first. https://github.com/OCA/purchase-workflow/pull/1173

I was just afraid that user will make something that will mess up with standard sequence. WDYT?

remi-filament:
Makes sense, then you really should add a data XML file then (and not only demo, which was the point of my comment), otherwise if no demo is installed you have nothing and you cannot create from UI :)

Maybe you should only rename demo directory in data (and update path in manifest), and add a noupdate in data tag so that admin can still update the sequences directly

kittiu:
> Makes sense, then you really should add a data XML file then (and not only demo, which was the point of my comment), otherwise if no demo is installed you have nothing and you cannot create from UI :)
> 
> Maybe you should only rename demo directory in data (and update path in manifest), and add a noupdate in data tag so that admin can still update the sequences directly

Yes, `base.sequence.option` should be date, but the `sequence.option` should be demo. I will check again. thanks for your notice.

kittiu:
@remi-filament ready for review.

Note: I will later squash commits into one commit (I think you name will persist).

account_sequence_option/data/account_sequence_option.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <data>
```
remi-filament: You may want to add a noupdate here in order to avoid changing use_sequence_option to False upon update of the module

remi-filament:
> Note: I will later squash commits into one commit (I think you name will persist).

Don't worry if my name does not persist, I should survive it :)

remi-filament:
Hi @kittiu I have found an error when trying to implement a payment : 

```
File "/opt/odoo/additional_addons/account_sequence_option/models/account_move.py", line 45, in _constrains_date_sequence
    record.super()._constrains_date_sequence()
Exception

The above exception was the direct cause of the following exception:

  File "/opt/odoo/odoo/odoo/http.py", line 640, in _handle_exception
  File "/opt/odoo/odoo/odoo/http.py", line 316, in _handle_exception
AttributeError: 'account.move' object has no attribute 'super'
```

There is a problem with that line 45 in account_move.py, you should not make a record.super() but a super(AccountMove, record) I believe, let me test and propose a PR on your code if it works for me

kittiu:
@remi-filament yes, I notice some problem when use with payment too. That's because it _compute_name even before all values the payment record is settled (for the 1st number). Odoo numbering is a bit tricky.

I have fixed it. Can you try again, thanks! You will need the update of `base_sequence_option` https://github.com/OCA/server-tools/pull/2080

alexis-via:
Seems to me that this solution doesn't handle multi-company. See my comment on the base PR OCA/server-tools#2080

kittiu:
I am thinking...
This module should be in account-financial-tools than account-invoicing. WDYT?

remi-filament (approved):
Working fine for me (I have not tested with multi company though), thanks @kittiu

ps-tubtim:
**Process**

1. Create customer invoice normally
2. Post the customer invoice then the customer invoice sequence is CINV/0002
3. Reset to draft the customer invoice
4. Post the customer invoice again then the customer invoice sequence is CINV/0003

Peek 2021-10-11 15-29

@kittiu 
IMHO, I think when the posted journal entry is reset to draft then post it again, the journal entry sequence should not change.

kittiu:
@ps-tubtim fixed. Now, we will not remove number when set to draft (same as normal v14 way)

HaraldPanten (approved):
Tested in our local environment and LGTM

HaraldPanten:
@kittiu can u rebase, please?

THX!

kittiu:
@HaraldPanten rebased.

HaraldPanten:
> I am thinking... This module should be in account-financial-tools than account-invoicing. WDYT?

This is the only question right now... WDYT @pedrobaeza @AaronHForgeFlow @sbidoul ?

pedrobaeza:
Yes, it may fit better, as this sequence is not only for invoices AFAIK.

HaraldPanten:
@kittiu Do you mind moving the PR to account financial tools, please? Just ping me and I will help you to get it merged.

THX!

kittiu:
Here you go https://github.com/OCA/account-financial-tools/pull/1299

HaraldPanten:
Superseded by https://github.com/OCA/account-financial-tools/pull/1299

PR #935, opened 2021-05-13
[13.0][FIX] account_invoice_warn_message, calculation when parent has message but partner not

Rad0van:
Wrong indentation causes the message set to False in case when partner's parent has a message set but the partner itself not.

Should I add a specific test just for that situation? Added.


Rad0van:
@sbidoul & @HviorForgeFlow could you have a look please? Thanx.

HviorForgeFlow (approved):
Code review LGTM :+1:

kittiu:
@Rad0van thanks for this fix.
Can you help some little things before we merge.
* Change name of title to `[13.0][FIX]  account_invoice_warn_message, calculation when parent has message but partner not`, this make future reference easier.
* Merge the 2 commits into one.

kittiu (approved):

dreispt (approved):

PR #936, opened 2021-05-14
[13.0][IMP]account_move_exception&account_move_post_block

JordiMForgeFlow:
Minor fixes and improvements:

- account_move_exception: fix view display and make them appear directly in the move form, as the exceptions will be only displayed when there are some.
- account_move_post_block: do not break the description of the created exception for better display (sorry for the nitpicking)

### DIFF
--- a/account_move_exception/views/account_view.xml
+++ b/account_move_exception/views/account_view.xml
@@ -44,21 +44,19 @@
                         /></bold>
                 </div>
             </header>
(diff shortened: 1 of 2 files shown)


AaronHForgeFlow (approved):
Makes sense. Thanks!

JordiBForgeFlow (approved):
:+1: functional review

Issue #937, opened 2021-05-17
[12.0] Creation of second invoice from purchase order is inhibited


sergiocorato:
This function [1] does not take in account `create_bill` variable in context as the original function [2], so after the issue of the first invoice, it does not create further invoices, only showing the first invoice generated.

[1] https://github.com/OCA/account-invoicing/blob/36cb648dc2bb1b2ca75cc822fdc9d8bd5d34a1ca/purchase_stock_picking_return_invoicing/models/purchase_order.py#L99

[2] https://github.com/odoo/odoo/blob/ac402ec49ffcf345b4017f38560aaa44307a41ac/addons/purchase/models/purchase.py#L413

PR #938, opened 2021-05-17
[12.0][FIX] purchase_stock_picking_return_invoicing create invoices after the first from purchase order

sergiocorato:
Solve #937

### DIFF
--- a/purchase_stock_picking_return_invoicing/models/purchase_order.py
+++ b/purchase_stock_picking_return_invoicing/models/purchase_order.py
@@ -95,13 +95,15 @@ class PurchaseOrder(models.Model):
         invoices = self.invoice_ids.filtered(
             lambda x: x.type == 'in_invoice'
         )
+        create_bill = self.env.context.get('create_bill', False)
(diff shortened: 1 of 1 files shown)


pedrobaeza:
But what happens without this patch if you click on Create? I think that's the intended behavior.

sergiocorato:
It creates an invoice with 0 qty as in original function (the button is in grey).
The intended behavior here is as the comment says `"Change super action for displaying only normal invoices."` so to hide the `in_refund` invoices, or not?
I don't think it was intended to forbid to invoice partially delivered goods anyway.

sergiocorato:
@OCA/accounting-maintainers review?

AaronHForgeFlow (approved):
Code review LGTM

kittiu:
@sergiocorato thank you for the fix. 
Before the merge, can you add name of this module `purchase_stock_picking_return_invoicing` in the subject. I think it is important to ease future search. Thank you!

kittiu (approved):

PR #874, opened 2021-03-02
[FIX] account_invoice_supplier_self_invoice: Fix report

etobella:
With the previous code, the header was not properly shown. Now it has been fixed
Previous view:
imagen

New view:
imagen

Also adding a copy=False on self_invoice_number_field

PR #947, opened 2021-05-26
[FIX] account_invoice_supplier_self_invoice: fix migration script from 12 to 13

albariera:
This PR fixes the migration script from 12 to 13. 
It also integrates PR #874 from @etobella to improve the report. 
Finally, I have also updated usage.


lmmar-ic (approved):
It's works perfect, the script run and the data mig to v13 without problem, and the module in v13 execute without problem.

etobella (approved):

etobella:
@OCA/accounting-maintainers Can this be merged as it is an issue with a migration script?

PR #956, opened 2021-06-14
[FIX] purchase_stock_picking_return_invoicing: consider context properly

chienandalu:
The top invoicing button uses a context to check if we want to create an
invoice or to view the current ones.

cc @Tecnativa TT30418

please review @pedrobaeza @joao-p-marques


pedrobaeza (approved):
Indeed

pedrobaeza:
Fast-tracking for fixing the bug:

/ocabot merge patch

Note that a similar patch is in #877 for 13.0, but I will propose this simpler one.

PR #958, opened 2021-06-17
[ADD] new module account_invoice_installment

britoederr:
The objective this module is add a tab in account move form view to display payables and receivables moves as installments with their main informations.


kittiu:
@britoederr thank you for your contribution.

I will not quite sure the name of this module make sense to understand what it does. Can you explain what is the main purpose of this module.

Also, IMHO, we should try avoid new tab, but instead use the smart button where possible (to avoid computing when not necessary), but it is just my opinion.

PR #960, opened 2021-06-22
[13.0][ADD] sale_line_refund_to_invoice_qty

JordiMForgeFlow:
### Sale Line Refund To Invoice Qty

This module allows the user to choose whether refunded quantities in credit
notes should be considered as quantities to be reinvoiced in the related sales,
thus deciding if they should be added or not from the quantity to invoice of
the related sales order lines.


AaronHForgeFlow:
Would it be possible to change the qty_to_invoice field instead the qty_invoiced? At the end this is the purpose of the module. You are interfere the qty_invoiced in order to get the proper value for the qty_to_invoice. Perhaps  it is best to just change the qty_to_invoice field, if possible.

JordiMForgeFlow:
I totally agree @AaronHForgeFlow. I have updated the code, thanks! :)

sale_line_credit_note_invoiced_qty/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+    "summary": """Allow deciding whether refunded quantity should be considered
+                as quantity to reinvoice""",
+    "version": "13.0.1.0.0",
+    "category": "Sales",
+    "website": "https://github.com/OCA/account-invoicing",
+    "author": "ForgeFlow, Odoo Community Association (OCA)",
+    "license": "LGPL-3",
+    "application": False,
+    "installable": True,
+    "depends": ["sale", "account"],
```
MiquelRForgeFlow:
```suggestion
    "depends": ["sale"],
```

MiquelRForgeFlow (approved):
LGTM

AaronHForgeFlow (approved):
Functional test. It works smoothly :)

AaronHForgeFlow (approved):

AaronHForgeFlow:
I see in standard the quantity invoiced is is reduced when creating refunds. For example, here I created two refunds, one have to be reinvoiced, the other refund not:
image

The quantity to invoice is correct, but for some reason the quantity invoiced shows qty =3??? it deducted the 2 refunds. I thought that standard Odoo never subtracts qty on the quantity invoiced, but it seems it does :S Perhaps we should also change the quantity_invoiced based if the refund has to be reinvoiced or not, don't you think?

JordiMForgeFlow:
> I see in standard the quantity invoiced is is reduced when creating refunds. For example, here I created two refunds, one have to be reinvoiced, the other refund not:
> image
> 
> The quantity to invoice is correct, but for some reason the quantity invoiced shows qty =3??? it deducted the 2 refunds. I thought that standard Odoo never subtracts qty on the quantity invoiced, but it seems it does :S Perhaps we should also change the quantity_invoiced based if the refund has to be reinvoiced or not, don't you think?

Yes, that is why I developed this module. We discussed this in previous comments and agreed to not interfere on the standard quantity invoiced computation.

The thing is that the quantity invoiced computed by Odoo is the "net" quantity invoiced, as it is deducting the refunded quantity. However, this behaviour causes that the quantity to invoice is again increased when quantities are refunded.

AaronHForgeFlow:
Ok! so then I think this is fine as it is now :)

AaronHForgeFlow (approved):
LGTM!

PR #961, opened 2021-06-22
[13.0][FIX] account_global_discount: Prevent error when add more taxes in multi-lines from invoices when global discount is empty

victoralmau:
Prevent error when add more taxes in multi-lines from invoices when global discount is empty.

Please @pedrobaeza and @chienandalu can you review it?

@Tecnativa TT30552

### DIFF
--- a/account_global_discount/models/account_move.py
+++ b/account_global_discount/models/account_move.py
@@ -105,7 +105,8 @@ class AccountMove(models.Model):
         taxes_keys = {}
         # Perform a sanity check for discarding cases that will lead to
         # incorrect data in discounts
-        for inv_line in self.invoice_line_ids.filtered(lambda l: not l.display_type):
+        _self = self.filtered("global_discount_ids")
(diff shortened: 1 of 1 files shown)


pedrobaeza:
What happens if you have global discounts and then remove them? I think you should still call the same methods, but not raise the error if no global discounts are set.

account_global_discount/models/account_move.py
```diff
@@ -105,7 +105,9 @@ def _set_global_discounts_by_tax(self):
         taxes_keys = {}
         # Perform a sanity check for discarding cases that will lead to
         # incorrect data in discounts
-        for inv_line in self.invoice_line_ids.filtered(lambda l: not l.display_type):
+        for inv_line in self.filtered("global_discount_ids").invoice_line_ids.filtered(
```
pedrobaeza: Don't do 3 times the same `filtered`, but store it in an intermediate variable.

pedrobaeza (approved):

chienandalu (approved):

HaraldPanten (approved):

PR #963, opened 2021-06-23
[14.0][MIG] account_invoice_fiscal_position_update

alexis-via:


account_invoice_fiscal_position_update/__manifest__.py
```diff
@@ -1,21 +1,22 @@
...
+    "version": "14.0.1.0.0",
+    "category": "Accounting",
     "license": "AGPL-3",
     "summary": "Changing the fiscal position of an invoice will auto-update "
     "invoice lines",
     "website": "https://github.com/OCA/account-invoicing",
     "author": "Julius Network Solutions,"
     "Akretion,"
     "Odoo Community Association (OCA)",
+    "maintainers": ["alexis-via"],
```
legalsylvain, line 19: thanks for maintaining that usefull module.

legalsylvain (approved):
Thanks. (diff review / no test).

kittiu:
Functional Testing, I found this error with stesp:

1. Create new invoice, without saving yet.
2. Add product line (with taxes), and change Fiscal Position
3. Error as following,
```
  File "/opt/odoo/custom/src/odoo/odoo/http.py", line 640, in _handle_exception
  File "/opt/odoo/custom/src/odoo/odoo/http.py", line 316, in _handle_exception
psycopg2.errors.SyntaxError: syntax error at or near ")"
LINE 8:             WHERE line.move_id IN ()
                                           ^
```
It will work ok, if the invoice is already saved in the first place.

I am not sure what is the desired working step. But if it need the record to be saved first, may be we an make Fiscal Position readyone at first?

account_invoice_fiscal_position_update/models/account_move.py
```diff
@@ -0,0 +1,73 @@
...
+                    taxes = fp.map_tax(taxes)
+
+                line.tax_ids = [(6, 0, taxes.ids)]
+                line._onchange_mark_recompute_taxes()
+                line.account_id = account.id
+            else:
+                lines_without_product.append(line.name)
+
+        self.with_context(check_move_validity=False)._recompute_dynamic_lines()
+        self._check_balanced()
```
hildickethan, line 55:
```suggestion
        self._recompute_dynamic_lines()
```
Not sure if there was a previous reasoning to separating these steps as the context is to stop it from doing the `_check_balanced` internally but then does it immeditately after, but I verified the error @kittiu found and by making this change it fixed it. The lines recalculated their taxes correctly before saving the invoice with each change of the fiscal position

ernestotejeda: I have tested this PR locally and the solution that @hildickethan  proposes has worked well for me and I think it is correct.
Also, I think that if we pass the paramenter `recompute_tax_base_amount=True` to that method (`_recompute_dynamic_lines`), we can remove `line._onchange_mark_recompute_taxes()` added a few lines above.
@alexis-via tell me if you plan to continue with this PR, if not, let me know to open a new one with those changes in order to move forward with this migration :). Thanks.

alexis-via: @ernestotejeda Feel free to superseed this PR, I'm too busy with other developments at the moment.

pedrobaeza:
Superseded by #1032

PR #968, opened 2021-07-09
[13.0][ADD] account_invoice_show_currency_rate: Show currency rate in invoices

victoralmau:
Show currency rate in invoices.

Please @pedrobaeza and @joao-p-marques can you review it?

@Tecnativa TT30632


account_invoice_show_currency_rate/README.rst
```diff
@@ -0,0 +1,85 @@
...
+
+.. contents::
+   :local:
+
+Usage
+=====
+
+o use this module, the company have to access rights Multi Currencies.
+
+#. Go to Invoicing > Configuration > Settings > Currencies > Multi-Currencies
```
joao-p-marques: This would probably fit better under `Configuration`. Also, maybe it could be clearer like this:
```suggestion
To use this module, the company must have Multi Currencies enabled.

#. Go to Invoicing > Configuration > Settings > Currencies > Multi-Currencies
```

account_invoice_show_currency_rate/models/account_move.py
```diff
@@ -0,0 +1,52 @@
...
+                total_debit = sum(item.line_ids.mapped("debit"))
+                item.currency_rate_amount = amount_currency_positive / total_debit
+            elif item.date and item.currency_id != item.company_id.currency_id:
+                rates = item.currency_id._get_rates(item.company_id, item.date)
+                item.currency_rate_amount = rates.get(item.currency_id.id)
+            else:
+                item.currency_rate_amount = 1
+
+    @api.onchange("date", "currency_id")
+    def _onchange_currency_account_invoice_show_currency_rate(self):
```
joao-p-marques: Where does this field come from? Do you mean `_onchange_currency`?

joao-p-marques (changes requested):
Code and functional review. I had some functional issues that I think might be related to my comments in the code.

account_invoice_show_currency_rate/models/account_move.py
victoralmau: Yes, these function is related to `_onchange_currency()`, we have call super-onchange and after `_compute_currency_rate_amount()` again. 
IMO it's necessary to customize onchange function to name.

pedrobaeza: This shouldn't be needed in a computed field. Add the fields in the depends instead.

account_invoice_show_currency_rate/README.rst
```diff
@@ -0,0 +1,85 @@
...
+.. |badge4| image:: https://img.shields.io/badge/weblate-Translate%20me-F47D42.png
+    :target: https://translation.odoo-community.org/projects/account-invoicing-13-0/account-invoicing-13-0-account_invoice_show_currency_rate
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/95/13.0
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+This module show currency rate applied in invoices.
```
pedrobaeza:
```suggestion
This module shows the currency rate applied in invoices, so you can visually verify what is going to be applied for the exchange, or which one was applied once converted to company currency.
```

account_invoice_show_currency_rate/README.rst
```diff
@@ -0,0 +1,85 @@
...
+
+**Table of contents**
+
+.. contents::
+   :local:
+
+Usage
+=====
+
+o use this module, the company have to access rights Multi Currencies.
```
pedrobaeza: I agree this should be in the configuration section, and with this rephrasing. Having steps is for differentiating the things you do on screen:
```suggestion
Enable the option for multiple currencies in your instance:

#. Go to *Invoicing > Configuration > Settings*
#. Search for or locate the section "Currencies".
#. Enable "Multi-Currencies" option there.
```

account_invoice_show_currency_rate/README.rst
```diff
@@ -0,0 +1,85 @@
...
+
+#. Go to Invoicing > Configuration > Settings > Currencies > Multi-Currencies
+#. Go to any draft invoice
```
pedrobaeza, line 41: Please always put whole steps. How do you get to any draft invoice? Do these instructions creating a new invoice. I'll explain later why.

account_invoice_show_currency_rate/README.rst
```diff
@@ -0,0 +1,85 @@
...
+#. Change the invoice currency
+#. A default currency rate, based on invoice date and the selected currency will be set
```
pedrobaeza:
```suggestion
#. The proper currency rate, based on the invoice date and the selected currency, will be shown.
```

account_invoice_show_currency_rate/README.rst
```diff
@@ -0,0 +1,85 @@
...
+
```
pedrobaeza, line 61: Add now usage instructions when the invoice already contains lines and there are totals to the division computation:

```suggestion
#. Add any invoice line.
#. Odoo has already generated the journal item lines with the rate applied, so the currency rate shown is get from the division between the amount in currency by the amount in company currency.

```

account_invoice_show_currency_rate/README.rst
```diff
@@ -0,0 +1,85 @@
...
+~~~~~~~
+
+* Tecnativa
+
+Contributors
+~~~~~~~~~~~~
+
+* `Tecnativa <https://www.tecnativa.com>`_:
+
+  * Víctor Martínez
```
pedrobaeza, line 86: Designer also counts here :wink:

account_invoice_show_currency_rate/models/account_move.py
```diff
@@ -0,0 +1,46 @@
...
+    @api.depends("date", "state", "company_id", "currency_id")
+    def _compute_currency_rate_amount(self):
+        """ It's necessary to define value according to some cases:
+        - Case A: Currency is equal to company currency (Value = 1)
+        - Case B: Move exist previously (posted) and get real rate according to lines
+        - Case C: Get expected rate (according to date) to show some value in creation.
+        """
+        for item in self:
+            if (
+                item.state == "posted"
```
pedrobaeza: I think the condition shouldn't be on the state, but if you have `amount_currency` or not.

account_invoice_show_currency_rate/models/account_move.py
```diff
@@ -0,0 +1,46 @@
...
+    _inherit = "account.move"
+
+    currency_rate_amount = fields.Float(
+        string="Rate amount",
+        compute="_compute_currency_rate_amount",
+        store=True,
+        digits=0,
+    )
+
+    @api.depends("date", "state", "company_id", "currency_id")
```
pedrobaeza: You should depend on the real fields that changes the computation:

* date
* amount_currency
* company_id
* currency_id

account_invoice_show_currency_rate/models/account_move.py
```diff
@@ -0,0 +1,46 @@
...
+                )
+                total_debit = sum(item.line_ids.mapped("debit"))
+                item.currency_rate_amount = amount_currency_positive / total_debit
+            elif item.date and item.currency_id != item.company_id.currency_id:
+                rates = item.currency_id._get_rates(item.company_id, item.date)
+                item.currency_rate_amount = rates.get(item.currency_id.id)
+            else:
+                item.currency_rate_amount = 1
+
+    def action_post(self):
```
pedrobaeza: This is not needed with the previous comment, but even in this code state, it shouldn't be as well, as you have included `state` in the depends list.

account_invoice_show_currency_rate/views/account_move_view.xml
```diff
@@ -0,0 +1,20 @@
...
+-->
+<odoo>
+    <record id="account_move_form" model="ir.ui.view">
+        <field name="name">account.move.form</field>
+        <field name="model">account.move</field>
+        <field name="inherit_id" ref="account.view_move_form" />
+        <field name="arch" type="xml">
+            <field name="currency_id" position="before">
+                <field
+                    name="currency_rate_amount"
```
pedrobaeza, line 14: This shouldn't shown when the currency is the same as the company currency.

account_invoice_show_currency_rate/views/account_move_view.xml
```diff
@@ -0,0 +1,20 @@
...
+                    readonly="True"
```
pedrobaeza: This is redundant. A computed field is already readonly by default.

account_invoice_show_currency_rate/views/account_move_view.xml
```diff
@@ -0,0 +1,20 @@
...
+                    groups="base.group_multi_currency"
```
pedrobaeza, line 15: Add `digits="[12,12]"` for proper visualization of whole rate.

account_invoice_show_currency_rate/models/account_move.py
```diff
@@ -0,0 +1,46 @@
...
+from odoo import api, fields, models
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    currency_rate_amount = fields.Float(
+        string="Rate amount",
+        compute="_compute_currency_rate_amount",
+        store=True,
```
pedrobaeza: Don't do it `store`, or all existing invoices will need to recompute its value on installation/migration. Any reason for doing it stored that I lose?

account_invoice_show_currency_rate/models/account_move.py
```diff
@@ -0,0 +1,46 @@
...
+                        "amount_currency"
+                    )
+                )
+                total_debit = sum(item.line_ids.mapped("debit"))
+                item.currency_rate_amount = amount_currency_positive / total_debit
+            elif item.date and item.currency_id != item.company_id.currency_id:
+                rates = item.currency_id._get_rates(item.company_id, item.date)
+                item.currency_rate_amount = rates.get(item.currency_id.id)
+            else:
+                item.currency_rate_amount = 1
```
pedrobaeza: Better to set 0 than 1, and at the beginning of the compute method for all in one shot: `self.currency_rate_amount = 0`

pedrobaeza (changes requested):

account_invoice_show_currency_rate/models/account_move.py
pedrobaeza: I'm now getting the reason for that. Although it's going to be hidden, `1` is OK, yes. Do the other thing of putting it at the beginning in one shot.

account_invoice_show_currency_rate/models/account_move.py
victoralmau: Ok, i have removed `store=True` because is not really necessary.

account_invoice_show_currency_rate/views/account_move_view.xml
victoralmau: I have added invisible domain but not work fine. Any idea about it?

account_invoice_show_currency_rate/models/account_move.py
```diff
@@ -0,0 +1,50 @@
...
+                    item.line_ids.filtered(lambda x: x.amount_currency > 0).mapped(
+                        "amount_currency"
+                    )
+                )
+                total_debit = sum(item.line_ids.mapped("debit"))
+                item.currency_rate_amount = amount_currency_positive / total_debit
+            elif (
+                not item.line_ids.filtered(lambda x: x.amount_currency > 0)
+                or item.state != "posted"
+            ) and item.currency_id != item.company_id.currency_id:
```
joao-p-marques: You end up repeating this condition here and above for the first `if`. To avoid it, you can convert this into:

```suggestion
           else:
               if item.currency_id != item.company_id.currency_id:
```

or you could store that info in a variable at the beginning to avoid having to make multiple calls

account_invoice_show_currency_rate/models/account_move.py
```diff
@@ -0,0 +1,50 @@
...
+        - Case B: Move exist previously (posted) and get real rate according to lines
+        - Case C: Get expected rate (according to date) to show some value in creation.
+        """
+        for item in self:
+            item.currency_rate_amount = 1
+            if item.state == "posted" and item.line_ids.filtered(
+                lambda x: x.amount_currency > 0
+            ):
+                amount_currency_positive = sum(
+                    item.line_ids.filtered(lambda x: x.amount_currency > 0).mapped(
```
joao-p-marques: Also, as this value `item.line_ids.filtered(lambda x: x.amount_currency > 0)` is the same you use above for the comparison, you can store it in a variable to avoid the multiple calls…

account_invoice_show_currency_rate/README.rst
```diff
@@ -0,0 +1,103 @@
...
+Enable the option for multiple currencies in your instance:
+
+#. Go to Invoicing > Configuration > Settings > Currencies > Multi-Currencies
+#. Go to any draft invoice
+#. Change the invoice currency
+#. The proper currency rate, based on the invoice date and the selected currency, will be shown.
+#. Add any invoice line.
+#. Odoo has already generated the journal item lines with the rate applied, so the currency rate shown is get from the division between the amount in currency by the amount in company currency.
+
+It's really necessary to exists some rates about not company currency distinct to 1.0, for example from EUR:
```
joao-p-marques:
```suggestion
Some rates must be defined (and be distinct to 1.0) for currencies different from the company default.
```

joao-p-marques (changes requested):
Code and functional review 👍 

Apart from this, maybe try to improve the functional documentation, as suggested by @pedrobaeza

account_invoice_show_currency_rate/models/account_move.py
```diff
@@ -0,0 +1,48 @@
...
+        "show_currency_rate_amount",
+    )
+    def _compute_currency_rate_amount(self):
+        """ It's necessary to define value according to some cases:
+        - Case A: Currency is equal to company currency (Value = 1)
+        - Case B: Move exist previously (posted) and get real rate according to lines
+        - Case C: Get expected rate (according to date) to show some value in creation.
+        """
+        for item in self:
+            item.currency_rate_amount = 1
```
pedrobaeza: You can assign in one shot:
```suggestion
        self.currency_rate_amount = 1
```

account_invoice_show_currency_rate/tests/test_account_move.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def test_01_invoice_usd(self):
+        self.partner.property_product_pricelist = self.pricelist_usd
+        invoice = self._create_invoice(self.currency_usd)
+        self.assertEqual(invoice.state, "posted")
+        self.assertEqual(len(invoice.invoice_line_ids), 1)
+        self.assertTrue(
+            invoice.invoice_line_ids.filtered(lambda x: x.product_id == self.product)
+        )
+        self.assertEqual(invoice.currency_id, self.currency_usd)
```
pedrobaeza: Don't check things that doesn't belong to this module.

account_invoice_show_currency_rate/tests/test_account_move.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def test_02_invoice_eur(self):
+        self.partner.property_product_pricelist = self.pricelist_eur
+        invoice = self._create_invoice(self.currency_eur)
+        self.assertEqual(invoice.state, "posted")
+        self.assertEqual(len(invoice.invoice_line_ids), 1)
+        self.assertTrue(
+            invoice.invoice_line_ids.filtered(lambda x: x.product_id == self.product)
+        )
+        self.assertEqual(invoice.currency_id, self.currency_eur)
```
pedrobaeza: Don't check things that doesn't belong to this module.

pedrobaeza (approved):

pedrobaeza:
@joao-p-marques please re-check

account_invoice_show_currency_rate/models/account_move.py
```diff
@@ -0,0 +1,47 @@
...
+        - Case C: Get expected rate (according to date) to show some value in creation.
+        """
+        self.currency_rate_amount = 1
+        for item in self.filtered("show_currency_rate_amount"):
+            lines = item.line_ids.filtered(lambda x: x.amount_currency > 0)
+            if item.state == "posted" and lines:
+                amount_currency_positive = sum(lines.mapped("amount_currency"))
+                total_debit = sum(item.line_ids.mapped("debit"))
+                item.currency_rate_amount = amount_currency_positive / total_debit
+            elif not lines or item.state != "posted":
```
joao-p-marques:
```suggestion
            else:
```

joao-p-marques (commented):
Just a little comment, otherwise looks good :+1:

account_invoice_show_currency_rate/models/account_move.py
victoralmau: Thanks, but is not correct because we ONLY change "default value" (1) in these case.

joao-p-marques: Ok, but this condition is the negation of the one in the `if`, so the result will be the same, right?

victoralmau: Not exactly, in these case we only apply when not lines OR state != 'posted', is not else.

yajo: Actually it is the same, so better just put the else.

joao-p-marques: Ok, but your condition `not lines or item.state != "posted"` is the [logic negation](https://en.wikipedia.org/wiki/Negation) of `item.state == "posted" and lines`, so the result here will be the same. Changing it to `else` avoids the extra check in the code. If this is not the desired behavior, then the current condition might be wrong...

pedrobaeza:
@joao-p-marques please refresh your review

joao-p-marques (approved):

PR #969, opened 2021-07-12
[12.0][IMP] stock_picking_invoicing: add action_invoice_draft method

marcelsavegnago:
When the invoice is canceled the picking status is changed to 2binvoiced. Therefore, I added the method that when returning the invoice to the draft status, the system returns the picking to the Invoiced status.

ping @mileo @renatonlima @rvalyi


mileo (approved):
LGTM

rvalyi (approved):
good one @marcelsavegnago !

PR #971, opened 2021-07-16
[13.0][MIG] account_invoice_merge from 12.0

acsonefho:
Migration of `account_invoice_merge` from v`12.0`

Sorry for the `flake8: noqa: C901` but this function is quite hard to refactor and the goal of this PR is to migrate the module. Not to improve it.


kittiu:
Functional review, I found that the merged document combine name "/" from all invoices

image

And as it it no just "/", I post it and the number is kept as "/ /".

image

PR #975, opened 2021-07-23
[14.0][ADD] account_move_original_partner

SilvioC2C:
This module adds a new field on `account.move` that allows showing every original SO partner (if the invoice is created from sale orders).


account_move_ordering_contacts/models/account_move.py
```diff
@@ -0,0 +1,26 @@
+# Copyright 2021 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo import api, fields, models
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    ordering_contact_ids = fields.Many2many(
```
simahawk: the name of the module and the field seems wrong. It should be something like `orginal_partner_ids` or `origin_partner_ids` or.... you got the point ;)

account_move_ordering_contacts/models/account_move.py
```diff
@@ -0,0 +1,26 @@
...
+        comodel_name="res.partner",
+        relation="account_move_ordering_contact_rel",
+        column1="account_move_id",
+        column2="res_partner_id",
+        compute="_compute_ordering_contact_ids",
+        store=True,
+        string="Ordering Contacts",
+    )
+
+    @api.depends("move_type", "invoice_line_ids.sale_line_ids.order_id.partner_id")
```
simahawk: are you sure this deeply nested relation depends will work?

simahawk (changes requested):

account_move_ordering_contacts/models/account_move.py
SilvioC2C: @simahawk both module and fields are now renamed

account_move_ordering_contacts/models/account_move.py
SilvioC2C: @simahawk ran some tests, it works as supposed. However, I found that the field is left empty upon module installation, because of this: https://github.com/odoo/odoo/blob/18362722/odoo/models.py#L2598-L2605
Since method `update_db()` returns `None` if the field column type is not defined, and we're initializing a `Many2many` field, the field is never set as "to recompute". Therefore, it won't be recomputed upon existing records when the module is installed.
Any idea about how to fix this? I was thinking about overriding the `init` method, check whether the table `account_move_original_partner_rel` exists and whether it's empty or not; if exists but it's empty, then the field should be recomputed. Could it work?

simahawk: I'd go for a post init hook

account_move_ordering_contacts/models/account_move.py
simahawk: module name should be singular ;)

account_move_original_partner/hooks.py
```diff
@@ -0,0 +1,11 @@
+# Copyright 2021 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo import SUPERUSER_ID, api
+
+
+def post_init_hook(cr, registry):
+    with api.Environment.manage():
+        env = api.Environment(cr, SUPERUSER_ID, {})
```
simahawk, line 13: a comment explaining why would help ;)

simahawk (approved):

tarbib (approved):
Functional approval :+1:

account_move_original_partner/hooks.py
```diff
@@ -0,0 +1,25 @@
...
+        force_compute_original_partners(env)
+
+
+def force_compute_original_partners(env):
+    """
```
simahawk: FTR https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

account_move_original_partner/hooks.py
```diff
@@ -0,0 +1,25 @@
...
+    Force compute original partners: field is not automatically recomputed
+    upon module install, so we need to force it
+    """
+    domain = [("move_type", "=", "out_invoice")]
+    inv = env["account.move"].search(domain)
+    _logger.info(f"Force-compute original partners on {len(inv)} invoices...")
```
simahawk:
```suggestion
    _logger.info("Force-compute original partners on %s invoices", len(inv))
```
Use logger string interpolation instead. See https://docs.python.org/3/howto/logging.html#optimization

tarbib:
@simahawk Hello! Could you approve and merge please?
Thanks

PR #976, opened 2021-07-26
[13.0][FIX] account_invoice_tax_required: Restore multi-record support in action_post

joao-p-marques:
`action_post` should be able to accept a recordset. The core function does it and other implementations do it as well:
https://github.com/odoo/odoo/blob/13.0/addons/account/models/account_move.py#L2350
https://github.com/OCA/purchase-workflow/blob/13.0/purchase_work_acceptance/models/account_move.py#L41

This restores that behavior here to solve integration tests.

@Tecnativa

ping @pedrobaeza @sergio-teruel

### DIFF
--- a/account_invoice_tax_required/models/account_move.py
+++ b/account_invoice_tax_required/models/account_move.py
@@ -44,6 +44,8 @@ class AccountMove(models.Model):
                 config["test_enable"],
             )
         )
-        if self.type != "entry" and (force_test or not skip_test):
-            self._test_invoice_line_tax()
+        if force_test or not skip_test:
+            for record in self:
+                if record.is_invoice():
+                    record._test_invoice_line_tax()
         return super().action_post()


account_invoice_tax_required/models/account_move.py
```diff
@@ -44,6 +44,7 @@ def action_post(self):
                 config["test_enable"],
             )
         )
-        if self.type != "entry" and (force_test or not skip_test):
-            self._test_invoice_line_tax()
+        for record in self:
+            if record.type != "entry" and (force_test or not skip_test):
```
pedrobaeza: This should be `if record.is_invoice()` (not your fault, but it's not correct either).

joao-p-marques: Ok, thanks! I also changed the if condition to avoid having to check `force_test or not skip_test` for each iteration, when it can be done at the start.

pedrobaeza (approved):

pedrobaeza:
Let's fast-track it as needed bugfix:

/ocabot merge patch

Needed for v14 too?

joao-p-marques:
Thanks :+1: 

> Needed for v14 too?

Yes, I will forward port

joao-p-marques:
Oh no, sorry, it is not needed after all: https://github.com/OCA/account-invoicing/blob/14.0/account_invoice_tax_required/models/account_move.py#L47

PR #977, opened 2021-07-26
[13.0][ADD] account_move_tier_validation_approver

MateuGForgeFlow:
This module allows to assign a default user for approval associated to a supplier.
cc @LoisRForgeFlow

### DIFF
--- a/account_move_tier_validation_approver/__manifest__.py
+++ b/account_move_tier_validation_approver/__manifest__.py
@@ -1,0 +1,16 @@
+# Copyright 2021 ForgeFlow, S.L.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+{
+    "name": "Account Move Tier Validation Approver",
+    "version": "13.0.1.0.0",
+    "author": "ForgeFlow, Odoo Community Association (OCA)",
+    "category": "Accounting",
+    "license": "LGPL-3",
+    "depends": ["account_move_tier_validation"],
+    "data": [
+        "views/account_move_views.xml",
+        "views/res_partner_views.xml",
+        "views/res_config_settings_views.xml",
+    ],
+    "installable": True,
+}
--- a/account_move_tier_validation_approver/models/account_move.py
+++ b/account_move_tier_validation_approver/models/account_move.py
@@ -1,0 +1,31 @@
+# Copyright 2021 ForgeFlow, S.L.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    approver_id = fields.Many2one("res.users", string="Responsible for Approval")
+
+    @api.onchange("partner_id")
+    def _onchange_partner_approver_id(self):
+        if self.partner_id:
+            self.approver_id = self.partner_id.approver_id.id
+
+    def post(self):
+        for move in self:
+            require_approver_in_vendor_bills = (
+                move.company_id.require_approver_in_vendor_bills
+            )
+            if (
+                move.is_purchase_document(include_receipts=True)
+                and require_approver_in_vendor_bills
+                and not move.approver_id
+            ):
+                raise UserError(
+                    _("It is mandatory to indicate a Responsible for Approval")
+                )
+        return super(AccountMove, self).post()
--- a/account_move_tier_validation_approver/models/res_company.py
+++ b/account_move_tier_validation_approver/models/res_company.py
@@ -1,0 +1,15 @@
+# Copyright 2020 ForgeFlow, S.L.
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
+
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    require_approver_in_vendor_bills = fields.Boolean(
+        string="Require approver in vendor bills"
+    )
+    validation_approver_tier_definition_id = fields.Many2one(
+        comodel_name="tier.definition", string="Bill approval tier definition"
+    )
--- a/account_move_tier_validation_approver/models/res_config_settings.py
+++ b/account_move_tier_validation_approver/models/res_config_settings.py
@@ -1,0 +1,41 @@
+# Copyright 2020 ForgeFlow, S.L.
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
+
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    require_approver_in_vendor_bills = fields.Boolean(
+        string="Require Approver In Vendor Bills",
+        help="Requires adding an approver before a vendor bill can be posted.",
+        related="company_id.require_approver_in_vendor_bills",
+        readonly=False,
+    )
+
+    def set_values(self):
+        tier_definition = self.company_id.validation_approver_tier_definition_id
+        if not tier_definition:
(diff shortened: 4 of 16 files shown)


account_move_tier_validation_approver/init_hooks.py
```diff
@@ -0,0 +1,56 @@
...
+try:
+    from openupgradelib import openupgrade
+except Exception:
+    from odoo.tools import sql as openupgrade
+
+from odoo import SUPERUSER_ID, api
+
+_field_renames = [
+    ("res.partner", "res_partner", "approval_user_id", "approver_id"),
+]
```
LoisRForgeFlow: this init hook doesn't belong here I think.

account_move_tier_validation_approver/models/account_move.py
```diff
@@ -0,0 +1,36 @@
...
+            self.approver_id = self.partner_id.approver_id.id
+
+    def post(self):
+        params = self.env["ir.config_parameter"].sudo()
+        require_approver_in_vendor_bills = bool(
+            params.get_param(
+                "account_move_tier_validation_approver"
+                ".require_approver_in_vendor_bills",
+                default=False,
+            )
```
LoisRForgeFlow: move to a company-dependent setting

account_move_tier_validation_approver/models/tier_review.py
```diff
@@ -0,0 +1,17 @@
...
+from odoo import api, models
+
+
+class TierReview(models.Model):
+    _inherit = "tier.review"
+
+    @api.depends("reviewer_ids")
+    def _compute_todo_by(self):
+        for review in self:
+            rec = self.env[review.model].browse(review.res_id)
```
LoisRForgeFlow: you should not inherit this method but use the new feature you proposed https://github.com/OCA/server-ux/pull/371

LoisRForgeFlow (changes requested):

account_move_tier_validation_approver/models/account_move.py
```diff
@@ -0,0 +1,31 @@
...
+    approver_id = fields.Many2one("res.users", string="Responsible for Approval")
+
+    @api.onchange("partner_id")
+    def _onchange_partner_approver_id(self):
+        if self.partner_id:
+            self.approver_id = self.partner_id.approver_id.id
+
+    def post(self):
+        require_approver_in_vendor_bills = (
+            self.company_id.require_approver_in_vendor_bills
```
LoisRForgeFlow: take the company from the move in the loop below, move.company_id.

account_move_tier_validation_approver/README.rst
```diff
@@ -0,0 +1,38 @@
+=====================================
```
LoisRForgeFlow: Move readme to proper fragments.

account_move_tier_validation_approver/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2021 ForgeFlow, S.L.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+{
+    "name": "Account Move Tier Validation Approver",
+    "version": "13.0.1.0.0",
+    "author": "ForgeFlow, Odoo Community Association (OCA)",
+    "category": "Accounts",
```
LoisRForgeFlow: "Accounting"

account_move_tier_validation_approver/models/res_company.py
```diff
@@ -0,0 +1,16 @@
...
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    require_approver_in_vendor_bills = fields.Boolean(
+        string="Require approver in vendor bills"
+    )
+
+    validation_approver_tier_definition_id = fields.Many2one(
+        comodel_name="tier.definition", string="Tier Definition"
```
LoisRForgeFlow: Change the string

account_move_tier_validation_approver/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+    "license": "LGPL-3",
+    "depends": ["account_move_tier_validation", "base_tier_validation_formula"],
```
LoisRForgeFlow: base_tier_validation_formula is not needed.

account_move_tier_validation_approver/models/res_config_settings.py
```diff
@@ -0,0 +1,42 @@
...
+                    }
+                )
+                self.company_id.write(
+                    {"validation_approver_tier_definition_id": tier_definition.id}
+                )
+            else:
+                tier_definition.action_unarchive()
+        else:
+            tier_definition.action_archive()
+        super(ResConfigSettings, self).set_values()
```
LoisRForgeFlow: you always have to return something.

account_move_tier_validation_approver/models/res_config_settings.py
```diff
@@ -0,0 +1,42 @@
...
+    )
+
+    def set_values(self):
+        tier_definition = self.company_id.validation_approver_tier_definition_id
+        if self.require_approver_in_vendor_bills:
+            if not tier_definition and not tier_definition.active:
+                field = self.env["ir.model.fields"].search(
+                    [("model", "=", "account.move"), ("name", "=", "approver_id")]
+                )
+                tier_definition = self.env["tier.definition"].create(
```
LoisRForgeFlow: you need to add a propoer domain as well, otherwise this is affecting customer invoices too.

account_move_tier_validation_approver/models/res_config_settings.py
```diff
@@ -0,0 +1,42 @@
...
+    require_approver_in_vendor_bills = fields.Boolean(
+        string="Require Approver In Vendor Bills",
+        help="Requires adding an approver before a vendor bill can be posted.",
+        related="company_id.require_approver_in_vendor_bills",
+        readonly=False,
+    )
+
+    def set_values(self):
+        tier_definition = self.company_id.validation_approver_tier_definition_id
+        if self.require_approver_in_vendor_bills:
```
LoisRForgeFlow, line 37: can you check in odoo if there is a way to only execute this when the setting is changes, with this code this is executed everytime any setting is changed, related or unralated.

LoisRForgeFlow (changes requested):

account_move_tier_validation_approver/models/res_config_settings.py
```diff
@@ -0,0 +1,43 @@
...
+            if not tier_definition and not tier_definition.active:
```
kittiu: Non-blocking,
Will it better to use xml data to load the required tier.definition? to lesser code and as a noupdate, it can be more configurable.

account_move_tier_validation_approver/models/tier_definition.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2021 ForgeFlow, S.L.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from odoo import fields, models
+
+
+class Partner(models.Model):
```
MiquelRForgeFlow: ...

account_move_tier_validation_approver/models/res_config_settings.py
LoisRForgeFlow: If you create it by xml, you are a bit constrained, aren't you? either you create one per company or a global one in the xml. After the latest changes I think the code is quite simple, we are always creating a tier def per company and just archiving or unarchiving it using the setting.

Could you check if it looks good to you now?

Thanks for reviewing as always :)

LoisRForgeFlow (approved):
LGTM

MiquelRForgeFlow (approved):

AdriaGForgeFlow (approved):
LGTM :+1:

PR #978, opened 2021-07-28
[13.0][FIX] account_invoice_date_due: restore multi-record handling in write()

joao-p-marques:
415bb0b4c2eea0435fbd6431dad8b239216c90dd introduced a bug disallowing to `write()` in a multi-record set.
The bug is reproduced when doing mass edits.
This restores that possibility.

@Tecnativa
TT31218

ping @pedrobaeza @victoralmau

### DIFF
--- a/account_invoice_date_due/models/account_move.py
+++ b/account_invoice_date_due/models/account_move.py
@@ -21,10 +21,11 @@ class AccountMove(models.Model):
         res = super().write(vals)
         # Propagate due date to move lines
         # that correspont to the receivable/payable account
-        if "invoice_date_due" in vals and self.state == "posted":
-            payment_term_lines = self.line_ids.filtered(
-                lambda line: line.account_id.user_type_id.type
-                in ("receivable", "payable")
-            )
-            payment_term_lines.write({"date_maturity": vals["invoice_date_due"]})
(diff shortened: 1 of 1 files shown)


account_invoice_date_due/models/account_move.py
```diff
@@ -21,10 +21,14 @@ def write(self, vals):
...
         # that correspont to the receivable/payable account
-        if "invoice_date_due" in vals and self.state == "posted":
-            payment_term_lines = self.line_ids.filtered(
-                lambda line: line.account_id.user_type_id.type
-                in ("receivable", "payable")
-            )
-            payment_term_lines.write({"date_maturity": vals["invoice_date_due"]})
+        if "invoice_date_due" in vals:
+            for record in self:
+                if record.state == "posted":
```
pedrobaeza:
```suggestion
            for record in self.filtered(lambda x: x.state == "posted"):
``

joao-p-marques: Thanks! Change made

pedrobaeza (approved):

pedrobaeza:
Let's fast-track it for avoiding the singleton error:

/ocabot merge patch

PR #985, opened 2021-08-09
[14.0][ADD] account_mail_autosubscribe

ivantodorovich:


simahawk (changes requested):
same remark as https://github.com/OCA/sale-workflow/pull/1693

simahawk (approved):

PR #986, opened 2021-08-11
[13.0][IMP] purchase_batch_invoicing: improve wizard tree view

ChristianSantamaria:
Simplify tree view to see the main data
imagen

### DIFF
--- a/purchase_batch_invoicing/views/purchase_order_view.xml
+++ b/purchase_batch_invoicing/views/purchase_order_view.xml
@@ -13,4 +13,24 @@
             </xpath>
         </field>
     </record>
+    <record id="purchase_batch_invoicing_tree_view" model="ir.ui.view">
+        <field name="name">purchase.order.tree (purchase_batch_invoicing)</field>
+        <field name="model">purchase.order</field>
+        <field eval="999" name="priority" />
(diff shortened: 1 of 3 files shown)


purchase_batch_invoicing/views/purchase_order_view.xml
```diff
@@ -13,4 +13,24 @@
             </xpath>
         </field>
     </record>
+    <record id="purchase_batch_invoicing_tree_view" model="ir.ui.view">
+        <field name="name">purchase.order.tree (purchase_batch_invoicing)</field>
+        <field name="model">purchase.order</field>
+        <field eval="20" name="priority" />
```
pedrobaeza: Put an even higher priority (`999` or so)

pedrobaeza (changes requested):
Don't bump module version manually. Bot will do it. I'm wondering if putting the widget `many2many_tags` would be enough.

ChristianSantamaria:
Yes, adding the many2many_tags widget to it would be an option but I think it's fine to see this data

pedrobaeza (approved):

marcelsavegnago (approved):

PR #990, opened 2021-08-23
[14.0][MIG] account_global_discount

omar7r:
Depends on https://github.com/OCA/server-backend/pull/131


account_global_discount/models/account_move.py
```diff
@@ -234,33 +236,35 @@ def _compute_amount_one(self):
...
                 self.amount_untaxed, self.company_id.currency_id, self.company_id, date
             )
-        sign = self.type in ["in_refund", "out_refund"] and -1 or 1
-        self.amount_total_company_signed = amount_total_company_signed * sign
+        sign = self.move_type in ["in_refund", "out_refund"] and -1 or 1
         self.amount_total_signed = self.amount_total * sign
         self.amount_untaxed_signed = amount_untaxed_signed * sign
 
     @api.depends(
+        "line_ids.matched_debit_ids.debit_move_id.move_id.payment_id.is_matched",
```
rousseldenis: Why all these additions ?

remi-filament (approved):
Functional test OK
Thanks for this migration !

omar7r:
Hi, 

I found an improvement we can do in this addon, account move lines that represent global discounts have the mark of "exclude_from_invoice_tab", that's correct, but, it makes that these lines don't appear in invoice analysis view, doing this information erronous. 
If in the "_where" function of this object, I add " OR global_discount_item = true" these lines appear and the result will be correct, but this change depends that the "_where" function isn't inherited for other modules, if the function was inherited it will fail. Any solution for allowing this without breaking potentially the inheritance? Maybe a monkey patching, replacing the original definition?

pedrobaeza:
Do you mean on the SQL view? Properly injected, it should be compatible. Maybe a PoC here?

omar7r:
Yes, the original function is:

```
@api.model
def _where(self):
    return '''
        WHERE move.move_type IN ('out_invoice', 'out_refund', 'in_invoice', 'in_refund', 'out_receipt', 'in_receipt')
            AND line.account_id IS NOT NULL
            AND NOT line.exclude_from_invoice_tab
        '''
```
I can extend in the end with " OR global_discount_item = true" but if other module add other statements in this function, will depend on load order that works.

The PoC of monkey patching would be:
```
import odoo.addons.account.report.account_invoice_report as invoice_report_file

def _where_replaced(self):
    return '''
        WHERE move.move_type IN ('out_invoice', 'out_refund', 'in_invoice', 'in_refund', 'out_receipt', 'in_receipt')
            AND line.account_id IS NOT NULL
            AND (NOT line.exclude_from_invoice_tab OR global_discount_item = true)
    '''


invoice_report_file.AccountInvoiceReport._where = _where_replaced
```

pedrobaeza:
But monkey-patching is not needed. Odoo handles such proper inheritance. You only have to override the methods. Example: https://github.com/OCA/margin-analysis/blob/13.0/account_invoice_margin/report/account_invoice_report.py

omar7r:
Something like that?:
```
def _where(self):
    where_str = super()._where()
    return where_str.replace("NOT line.exclude_from_invoice_tab",
                             "(NOT line.exclude_from_invoice_tab OR "
                             "global_discount_item = true)")
```
It's important that the new statement will be in this position

pedrobaeza:
Yeah, I think that can work. Put parenthesis for assuring the evaluation order.

omar7r:
Done, thanks @pedrobaeza

francesco-ooops:
@rousseldenis can this be merged?

francesco-ooops:
@omar7r account_invoice_pricelist tests are failing with latest commit

omar7r:
@francesco-ooops I will try to do a rebase, but the error is in the account_invoice_pricelist module's test not this.

francesco-ooops:
> @francesco-ooops I will try to do a rebase, but the error is in the account_invoice_pricelist module's test not this.

looks good now! @rousseldenis @pedrobaeza could we proceed with merge?

pedrobaeza:
Cherry-pick extra fix+imp commits in 13.0 in 72aec197574cf5b9f89abce2b094aa1de8ba5026, as they apply as well. Thanks for them.

PR #992, opened 2021-08-25
[13.0][FIX] purchase_stock_picking_return_invoicing: some fixes

MiquelRForgeFlow:
Some migration fixes.

### DIFF
--- a/purchase_stock_picking_return_invoicing/models/purchase_order.py
+++ b/purchase_stock_picking_return_invoicing/models/purchase_order.py
@@ -1,4 +1,4 @@
-# Copyright 2017 Eficent Business and IT Consulting Services
+# Copyright 2017 ForgeFlow S.L. (https://www.forgeflow.com)
 # Copyright 2018 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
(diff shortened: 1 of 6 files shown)


AaronHForgeFlow (approved):
Code review :+1:

pedrobaeza:
Please explain the fixes.

MiquelRForgeFlow:
Basically, I adapted methods action_view_invoice_refund (and action_view_invoice) to the action_view_invoice of core (v13). That's a bad migration issue. The fixes in the view are self explanatory. And the fixes in the test, it's because when action_view_invoice button is used to create an invoice, a context is passed in the xml view, but not when only is used to view the invoice (same case for refunds).

purchase_stock_picking_return_invoicing/models/purchase_order.py
```diff
@@ -68,37 +68,60 @@ def action_view_invoice_refund(self):
         action = self.env.ref("account.action_move_in_refund_type")
         result = action.read()[0]
         create_refund = self.env.context.get("create_refund", False)
+        # Invoice_ids may be filtered depending on the user. To ensure we get all
+        # invoices related to the purchase order, we read them in sudo to fill the
+        # cache.
+        self.sudo()._read(["invoice_ids"])
```
pedrobaeza, line 72: I don't get the need for this

purchase_stock_picking_return_invoicing/models/purchase_order.py
```diff
@@ -68,37 +68,60 @@ def action_view_invoice_refund(self):
...
         # choose the view_mode accordingly
         if len(refunds) > 1 and not create_refund:
             result["domain"] = "[('id', 'in', " + str(refunds.ids) + ")]"
         else:
-            res = self.env.ref("account.invoice_supplier_form", False)
-            result["views"] = [(res and res.id or False, "form")]
+            res = self.env.ref("account.view_move_form", False)
+            form_view = [(res and res.id or False, "form")]
+            if "views" in result:
+                result["views"] = form_view + [
```
pedrobaeza, line 88: I don't get as well this, as we should only have form view, not anything more.

purchase_stock_picking_return_invoicing/models/purchase_order.py
```diff
@@ -68,37 +68,60 @@ def action_view_invoice_refund(self):
...
+                    (state, view) for state, view in action["views"] if view != "form"
+                ]
+            else:
+                result["views"] = form_view
+            # Do not set an invoice_id if we want to create a new bill.
+            if not create_bill:
+                result["res_id"] = invoices.id or False
+        result["context"]["default_invoice_origin"] = self.name
```
pedrobaeza, line 123: Why not joining the equal code in a method that is called by both buttons and change only specific part?

purchase_stock_picking_return_invoicing/views/purchase_view.xml
```diff
@@ -13,7 +13,6 @@
                     name="action_view_invoice_refund"
                     string="Create Refund"
                     type="object"
-                    class="oe_stat_button"
```
pedrobaeza, line 16: Why removing this class here and not in the other one?

purchase_stock_picking_return_invoicing/models/purchase_order.py
MiquelRForgeFlow: Because the core method `action_view_invoice` also has it.

pedrobaeza: Then put that and an absolute link to the reference in Odoo core.

purchase_stock_picking_return_invoicing/models/purchase_order.py
MiquelRForgeFlow: Yeah, probably, but who knows if a module extension can need it. The core method `action_view_invoice` has it. But I can remove it if you insist.

purchase_stock_picking_return_invoicing/models/purchase_order.py
MiquelRForgeFlow: Because is not possible. The action and the context in both are distinct.

purchase_stock_picking_return_invoicing/models/purchase_order.py
pedrobaeza: Can you reuse then such core code for not repeating everything? And only change what needed.

purchase_stock_picking_return_invoicing/models/purchase_order.py
pedrobaeza: I see a lot of duplicated lines. Please check better.

purchase_stock_picking_return_invoicing/views/purchase_view.xml
MiquelRForgeFlow: In version v11, the button only appears once and it has it. But in v12 and onwards, the button appears duplicated and has double utility. The case with "create" don't have this class while the case "only view" has it.

purchase_stock_picking_return_invoicing/models/purchase_order.py
MiquelRForgeFlow: https://github.com/odoo/odoo/blob/13.0/addons/purchase/models/purchase.py#L448

pedrobaeza: I mean to put it in the code:
```suggestion
        # Mimic upstream code
        # https://github.com/odoo/odoo/blob/3572ce3a06bad/addons/purchase/models/purchase.py#L448
        self.sudo()._read(["invoice_ids"])
```

purchase_stock_picking_return_invoicing/models/purchase_order.py
MiquelRForgeFlow: See the core method here https://github.com/odoo/odoo/blob/13.0/addons/purchase/models/purchase.py#L430 and feel free to check if both methods can reuse this method in some way just to avoid the verbatim copy.

purchase_stock_picking_return_invoicing/models/purchase_order.py
MiquelRForgeFlow: ah, ok

purchase_stock_picking_return_invoicing/models/purchase_order.py
MiquelRForgeFlow: I wasn't able to avoid repeating :S

pedrobaeza:
Please move it later to v14, but the smart-button "Create refund" should be changed to a regular button in the header.

PR #998, opened 2021-09-03
[11.0][ADD] account_menu_invoice_refund: Backport from 12.0

GerardoGomez:
Backport from 12.0


AaronHForgeFlow:
Thank you for the PR. The commit history here is missing, can you add it? In order to do that you just need to be even to version 11.0 and then do the `git format-patch --keep-subject --stdout origin/11.0..origin/12.0 -- account_menu_invoice_refund | git am -3 --keep `thing. Then commit the backport changes

GerardoGomez:
> Thank you for the PR. The commit history here is missing, can you add it? 

Thanks for your comment @AaronHForgeFlow. I was not sure if it was necessary to do it in the backports

account_menu_invoice_refund/__init__.py
```diff
@@ -1,2 +1,3 @@
 # Copyright 2019 Ecosoft Co., Ltd (http://ecosoft.co.th/)
+# © 2021 FactorLibre - Gerardo Gomez-Caminero <gerardo.gomez@factorlibre.com>
```
dreispt: Please remove copyright claims.
Copyright applies to original work author only, and porting the code is not enough to qualify for authorship.

dreispt (changes requested):

account_menu_invoice_refund/__init__.py
RodrigoBM: Removed

PR #999, opened 2021-09-03
[13.0][FIX] purchase_stock_picking_return_invoicing: Use Form in tests to avoid crash test

sergio-teruel:
Crash test due to this Odoo commit
OCA/OCB@8ca10a8
The wizard lines have the field uom_id related to move.product_uom and readonly=False, if you call directly to wiz.onchage_picking_id a write in stock move is executed.

cc @Tecnativa
ping @carlosdauden @CarlosRoca13


pedrobaeza:
It seems not enough I'm afraid.

sergio-teruel:
It's red because other modules in tehe same repo have the same issue... And I have done in distinct PR

pedrobaeza:
Well, next time please do everything in the same PR but different commits, as this way we have to merge one manually with red CI, no version bump, etc.

PR #1008, opened 2021-09-28
[13.0][IMP] account_invoice_show_currency_rate: Make sure the currency is correct in the main company

victoralmau:
Make sure the currency is correct in the main company

Please @pedrobaeza and @joao-p-marques can you review it?

@Tecnativa TT30631

### DIFF
--- a/account_invoice_show_currency_rate/tests/test_account_move.py
+++ b/account_invoice_show_currency_rate/tests/test_account_move.py
@@ -9,8 +9,10 @@ class TestAccountMove(common.SavepointCase):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
-        cls.currency_usd = cls.env.ref("base.USD")
-        cls.currency_eur = cls.env.ref("base.EUR")
+        usd = cls.env.ref("base.USD")
+        eur = cls.env.ref("base.EUR")
+        cls.currency = cls.env.ref("base.main_company").currency_id
+        cls.currency_extra = eur if cls.currency == usd else usd
         cls.account_tax = cls.env["account.tax"].create(
             {"name": "0%", "amount_type": "fixed", "type_tax_use": "sale", "amount": 0}
         )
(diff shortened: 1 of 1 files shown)


account_invoice_show_currency_rate/tests/test_account_move.py
```diff
@@ -11,6 +11,7 @@ def setUpClass(cls):
         super().setUpClass()
         cls.currency_usd = cls.env.ref("base.USD")
         cls.currency_eur = cls.env.ref("base.EUR")
+        cls.env.ref("base.main_company").write({"currency_id": cls.currency_usd.id})
```
pedrobaeza: This change is forbidden by a constraint in Odoo itself, so this is not the correct thing. Several options:

1. Create a new company, which can be too much.
2. Change the currency of the company by SQL as a hackish woraround.
3. Have the secondary currency in a variable depending on the company currency, and do the rest of the tests according it.

account_invoice_show_currency_rate/tests/test_account_move.py
```diff
@@ -50,8 +51,8 @@ def setUpClass(cls):
             {"name": "Pricelist EUR", "currency_id": cls.currency_eur.id}
         )
         # Create custom rates to USD + EUR
-        cls._create_currency_rate(cls, cls.currency_usd, "2000-01-01", 1.0)
-        cls._create_currency_rate(cls, cls.currency_eur, "2000-01-01", 2.0)
+        cls._create_currency_rate(cls, cls.currency_usd, fields.Date.today(), 1.0)
```
pedrobaeza: Can you avoid variable dates?

victoralmau: It has been changed in order to standardize it and not need to define a date. Similar to what the https://github.com/OCA/purchase-workflow/blob/13.0/purchase_stock_picking_show_currency_rate addon does in tests.
Do you think it is better to define those specific dates?

account_invoice_show_currency_rate/tests/test_account_move.py
victoralmau: Ok, I have made the changes according to option 3,

joao-p-marques (approved):

account_invoice_show_currency_rate/tests/test_account_move.py
pedrobaeza: The approach is to have the less variable data possible. That's why we also avoid demo data. Some side effects derived from using variable dates: error in tests on leap years and around midnight.

pedrobaeza (approved):

PR #1021, opened 2021-10-15
[14.0][ADD] account_move_line_accounting_description

ajaniszewska-dev:


account_move_line_accounting_description_purchase/tests/test_purchase_order_description.py
```diff
@@ -0,0 +1,76 @@
...
+        )
+        cls.po_2_line = cls.po_2.order_line
+        cls.po_2.button_confirm()
+        cls.po_2.button_approve()
+
+    def test_purchase_order_line_name(self):
+
+        # For 1st PO check order line is same as product description
+
+        self.assertEqual(self.po_1_line.name, self.product_1.accounting_description)
```
ajaniszewska-dev: @grindtildeath  assertion is failing here, what did i missed? tried to put pdb in` _prepare_account_move_line()` and it seems to not be called at all.

```
 File "/home/travis/build/OCA/account-invoicing/setup/account_move_line_accounting_description_purchase/odoo/addons/account_move_line_accounting_description_purchase/tests/test_purchase_order_description.py", line 68, in test_purchase_order_line_name

    self.assertEqual(self.po_1_line.name, self.product_1.accounting_description)

AssertionError: 'Virtual Interior Design' != 'Product1_acc_desc'

- Virtual Interior Design

+ Product1_acc_desc
```

grindtildeath: @ajaniszewska-dev `_prepare_account_move_line` is called when you generate the vendor bill from the PO. And only on the `account.move.line` we want the accounting description to be used for name, not on `purchase.order.line` :wink:

yankinmax (approved):

account_move_line_accounting_description/models/account_move.py
```diff
@@ -0,0 +1,15 @@
+from odoo import api, fields, models
```
grindtildeath, line 4: Missing header (on many other files too)

account_move_line_accounting_description_sale/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2021 Camptocamp
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Acccount Move Accounting Sale Order Description",
+    "version": "14.0.1.0.0",
+    "summary": "Adds an 'accounting description' on sale order",
```
grindtildeath:
```suggestion
    "summary": "Consider accounting description when invoicing sale order",
```

account_move_line_accounting_description_sale/__manifest__.py
```diff
@@ -0,0 +1,15 @@
...
+    "author": "Camptocamp, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/account-invoicing",
+    "license": "AGPL-3",
+    "category": "Accounting & Finance",
+    "depends": ["sale", "product", "account_move_line_accounting_description"],
+    "data": [],
```
grindtildeath: You could remove empty key and set module as auto install. (also for the glue to purchase module)

account_move_line_accounting_description_sale/models/sale_order.py
```diff
@@ -0,0 +1,13 @@
+from odoo import models
+
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    def _prepare_invoice_line(self, **optional_values):
+        line = super(SaleOrderLine, self)._prepare_invoice_line(**optional_values)
```
grindtildeath:
```suggestion
        line = super()._prepare_invoice_line(**optional_values)
```
You don't need this in py3 ;)

account_move_line_accounting_description_sale/models/sale_order.py
```diff
@@ -0,0 +1,13 @@
...
+        line["external_name"] = line["name"]
+        product = self.env["product.product"].search([("id", "=", line["product_id"])])
```
grindtildeath:
```suggestion
        product_id = line.get("product_id")
        if product_id:
            product = self.env["product.product"].browse(line["product_id"])
```
`browse` is better than `search` if you already have the id.
Also if the line is a note or a section `line["product_id"]` might throw a `KeyError` here.

account_move_line_accounting_description_purchase/models/purchase_order.py
```diff
@@ -0,0 +1,13 @@
...
+
+class PurchaseOrderLine(models.Model):
+    _inherit = "purchase.order.line"
+
+    def _prepare_account_move_line(self, move=False):
+        res = super()._prepare_account_move_line(move)
+        res["external_name"] = res["name"]
+        product = self.env["product.template"].search([("id", "=", res["product_id"])])
+        if product.accounting_description:
+            res["name"] = product.accounting_description
```
grindtildeath: Same as above or below ;)

account_move_line_accounting_description_purchase/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2021 Camptocamp
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Acccount Move Accounting Purchase Description",
+    "version": "14.0.1.0.0",
+    "summary": "Adds an 'accounting description' on purchase order",
```
grindtildeath: To fix (cf suggestion for other module)

account_move_line_accounting_description/views/account_move.xml
```diff
@@ -0,0 +1,16 @@
...
+        <field name="name">account_view_move_form</field>
+        <field name="model">account.move</field>
+        <field name="inherit_id" ref="account.view_move_form" />
+        <field name="arch" type="xml">
+             <xpath expr="//field[@name='invoice_line_ids']//tree" position="inside">
+                <field name="external_name" />
+            </xpath>
+            <xpath expr="//field[@name='invoice_line_ids']//form" position="inside">
+                <field name="external_name" />
+            </xpath>
```
grindtildeath: Could you also set the name column invisible for invoice line view?

grindtildeath (changes requested):
- The change in invoice report (to use `external_name` instead of `name` is missing)
- Could you add a test for the onchange in account_move_line_accounting_description? Tip: Use Form class in tests instead of `create` to simulate the UI (it also plays onchange functions automatically)

account_move_line_accounting_description/views/account_move.xml
ajaniszewska-dev: @grindtildeath thanks, PR updated with your remarks.

account_move_line_accounting_description/tests/test_account_move_line_description.py
```diff
@@ -0,0 +1,73 @@
...
+            lambda x: x.product_id
+        )
+        self.assertTrue(self.product_1.accounting_description)
+        self.assertEqual(
+            inv_line_with_product.name, self.product_1.accounting_description
+        )
+
+        # self.assertEqual(
+        #     inv_line_with_product.name, inv_line_with_product.external_name
+        # )
```
ajaniszewska-dev: @grindtildeath  any idea why `external_name` is `False` in test? (In UI it is properly set to product's name when changing product so I'd expect same here). Due to that, test is failing - same with l.71-73.

grindtildeath: I would say, check with a pdb in your onchange. Ff the `name` is set to `accounting_description` if means your onchange must be played and there's probably a `False` value in the `name` beforehand on this line: `line.external_name = line.name`?

account_move_line_accounting_description/tests/test_account_move_line_description.py
```diff
@@ -0,0 +1,73 @@
...
+from odoo.tests.common import SavepointCase
+
+
+class TestAccountLineDescription(SavepointCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.partner_1 = cls.env.ref("base.res_partner_1")
+        cls.product_1 = cls.env.ref("product.product_product_1")
+        cls.product_2 = cls.env.ref("product.product_product_2")
```
grindtildeath: The test would be much easier to read if you call your variables for what you expect (e.g. `cls.product_with_acc_desc` and `cls.product_without_acc_desc`) :wink:

account_move_line_accounting_description/tests/test_account_move_line_description.py
ajaniszewska-dev: @grindtildeath  it wasnt saved after saving form, that's why. Changed field to related to product_id.name instead - seems to be working fine. No idea why form didnt save this field normally.

grindtildeath:
@ajaniszewska-dev Please check https://github.com/ajaniszewska-dev/account-invoicing/pull/1

grindtildeath (approved):

kittiu:
@ajaniszewska-dev
Can you provide a decent readme for each module? 
Thanks!

PR #1014, opened 2021-10-07
[14.0][MIG] Module stock_picking_invoicing

mbcosta:
Migration module stock_picking_invoicing to v14, the PR https://github.com/OCA/account-invoicing/pull/1010 are missing the last commits made in v13:

image

cc @rvalyi @renatonlima @mileo @zar21

PR #1025, opened 2021-10-19
[14.0][ADD] Module sale_stock_picking_invoicing

mbcosta:
Module sale_stock_picking_invoicing, based in https://github.com/OCA/l10n-brazil/tree/12.0/l10n_br_sale_stock , depends https://github.com/OCA/account-invoicing/pull/1014.

This module extends Stock Picking Invoicing implementation to Sale, you can define the Sale Creation Invoice Policy, if the invoice should create from Sale Order or Stock Picking, in this case, the information used at invoice come from Sale Order e.g.: Price Unit, Payment Terms, and others fields. Similar feature exist in v8 core.

Implementation questions when "Sale Creation Invoice Policy" is "Stock Picking" :

* When Products of Sale Order has lines where "Product Type" is "Storable Product" the button "Create Invoice" in Sale Order become invisible even when Product "Invoice Policy" defined as "Ordered quantities",  is this right? Or it should be possible create invoice from Sale Order in this case? My undestand for now is as "Sale Creation Invoice Policy" is general rule should be has priority over the "Invoice Policy"  defined in each Product.

* There are two similar buttons to create invoice from Sale Order, one just appear in status Sale the other appear in all status but will be open with Down Payments option selected. Should be keep the second button to allow Down Paymens?

*  In the case of Sale Order has lines with Product Type "Storable Product" and "Service" will be create two separated Invoices, one created from Sale Order an other from Stock Picking, should be try to join and create only one? Or it should be a Parameter in the system?

As said this module is based in l10n_br_sale_stock, by this reason I like to ask if the License Header and the CONTRIBUTORS.MD should be keeped as the same based files?

It is possible reference mutiple sale lines in only one invoice line https://github.com/odoo/odoo/blob/14.0/addons/sale/models/account_move.py#L34, but there is a error when Posted Invoice:
 Issue https://github.com/odoo/odoo/issues/77028
 PR with FIX https://github.com/odoo/odoo/pull/77195
By this reason today the module not allow grouping stock moves when has sale_line_id, the PR above solve the problem, waiting merge to make useful here.

cc @renatonlima @rvalyi @mileo @gabrielcardoso21 @marcelsavegnago


sale_stock_picking_invoicing/models/res_company.py
```diff
@@ -0,0 +1,20 @@
...
+
+from odoo import _, fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    sale_create_invoice_policy = fields.Selection(
+        selection=[
+            ("sale_order", _("Sale Order")),
```
kittiu: Not sure, we need translation in selection field. _("...")

mbcosta:
Thanks @kittiu by your review, I'm close this PR because in my view it's important confirm with the mainteners of the repo if there are interest in include this functionality here and I think we lost the "window" to made this extraction from Brazilian Localization and will be better check this case in the future v16.

mbcosta:
Reopen because as talked in https://github.com/OCA/account-invoicing/pull/1672 the question about if this implementation would has interest outside Brazilian Localization project was answered, and to keep the isolation and modularization of stock_picking_invoicing will be better another module to integrate the functionality of create Invoice from Picking when it's linked to a Sale Order, some points about the implementation:

* Module is based in l10n_br_sale_stock version 14.0 https://github.com/OCA/l10n-brazil/tree/14.0/l10n_br_sale_stock the original License Header, Maintainers and Contributors was kept and also include the license in XML files ( in the beginning of the PR was based in v12.0 )

* It's possible to link one Invoice Line with mutiple Sale Lines, the main problem( the error described above in method action_post ) was solved https://github.com/OCA/OCB/blob/14.0/addons/sale/models/account_move.py#L15 but there is other problem the field qty_invoiced in sale.order.line show the quantity of invoice line without consider, in this case, that the value is the sum of others sale lines https://github.com/odoo/odoo/blob/14.0/addons/sale/models/sale.py#L1230, what can make confuse the user about the real Invoiced Quantity:

image

So for keep compatible with the Brazilian Localization I think will be better check this possibility to Grouping Sale Lines in the future in another PR or in versions v16, v17 or the newest.

* Test to check qty_invoiced commented https://github.com/akretion/account-invoicing/blob/14.0-NEW-sale_stock_picking_invoicing/sale_stock_picking_invoicing/tests/test_sale_stock.py#L274 to avoid error with the module sale_line_refund_to_invoice_qty https://github.com/OCA/account-invoicing/blob/14.0/sale_line_refund_to_invoice_qty/models/sale.py#L20 when run the tests as CI and both modules are install, so in the case of use the both modules should be consider this problem.

* This module should has integration with account_payment_mode https://github.com/OCA/bank-payment/tree/14.0/account_payment_mode as done in Brazilian Localization? To avoid a "glue" module.

The review and tests of people outside Brazilian Localization are important to know if there are any change that can make the module more compatible with another cases of use.

cc @renatonlima @rvalyi @mileo @gabrielcardoso21 @marcelsavegnago @kittiu @Christian-RB

mbcosta:
About important changes made in the last two commits:

* Get fields from Sale methods to prepare invoice and invoice lines to avoid the necessity of 'glue' modules and make the Invoice created by Picking equals to created by Sale. 

Other modules can included new fields in Sale Order and include this fields in the dict of creation Invoice from sale, for example:

account_payment_sale https://github.com/OCA/bank-payment/blob/14.0/account_payment_sale/models/sale_order.py#L41
sale_commission https://github.com/OCA/commission/blob/14.0/sale_commission/models/sale_order.py#L64 

Example with demo data, account_payment_sale and sale_commission modules installed:

image

image

image

image

image

image

image

*  Make module compatible with 'Down Payments' case.
Initially to force the creation of Invoice from Stock Picking when the "Sale Create Invoice Policy" was as stock_picking, the implementation turned the button 'Create Invoice' in Sale Order view invisible, but this block the case of "Down Payments", now in this case the _get_invoiceable_lines method just return the Products with Type Service and if there is no Service raise user error. Example with demo data:

image

image

image

image

image

image

image

Case where "Sale Create Invoice Policy"  defined as stock_picking

image

Create made in Stock Picking

image

image


In order to avoid errors with the tests made by CI when the enviroment are Demo the default "Sale Create Invoice Policy"  defined as "Sale Order", so for test this case you need to change in Company view or Sale Settings for "Stock Picking"

image

The commits are separated to let brazilian community check this changes in a easy way, but if necessary I can merge all commits and let only one.

rvalyi:
thank you for the hard work @mbcosta I promiss I'll look at that soon. If I forget please ping me again. We have to take the opportunity to modularize stock_invoice_picking indeed...

sale_stock_picking_invoicing/readme/INSTALL.rst
```diff
@@ -0,0 +1,4 @@
+This module depends on:
+
+* sale_management
+* stock_picking_invoicing
```
rvalyi, line 5: not the same list as in the manifest

sale_stock_picking_invoicing/__manifest__.py
```diff
@@ -0,0 +1,30 @@
...
+{
+    "name": "Sales Stock Picking Invocing",
+    "category": "Warehouse Management",
+    "license": "AGPL-3",
+    "author": "Akretion, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/account-invoicing",
+    "version": "14.0.1.0.0",
+    "maintainers": ["mbcosta", "renatonlima"],
+    "depends": [
+        "sale_management",
```
rvalyi, line 16: are you sure sale_management is required? If yes could you please just explain why? (and eventually leave a short comment why here)

rvalyi (commented):
@mbcosta I would say it's a bit late for us to focus on v14 unfortunately... I mean the module is quite large and we are not going to sell any v14 project anymore for instance nor migrate customer to v14 and I think people on v14 can work their way the way it is now in the Brazilian localization...
Would it be possible to try to make the equivalent PR on v16? I know this might be some work... But honestly it has more chance to enter in v16 than v14 that is already being phased out...

rvalyi:
I understand you re-opened this because https://github.com/OCA/account-invoicing/pull/1672 is for 14.0 ...
Well that sucks... Cause as you explain, if we merge #1672  we would make it even harder to make things modular in the future so merging it is bad. I'll make my best to find time and see if we can get this one merged then.

sale_stock_picking_invoicing/models/res_company.py
```diff
@@ -0,0 +1,32 @@
...
+        #  Create of Invoice by Sale Order using sale.advance.payment.inv object
+        #  is necessary let default policy as sale_order
+        #  TODO: Is there other form to avoid this problem?
+        result = "stock_picking"
+        module_base = self.env["ir.module.module"].search([("name", "=", "base")])
+        if module_base.demo:
+            result = "sale_order"
+        return result
+
+    sale_create_invoice_policy = fields.Selection(
```
rvalyi, line 23: sale_invoicing_policy would be a better name

rvalyi: sale_invoicing_policy would be a much better name

sale_stock_picking_invoicing/models/res_company.py
```diff
@@ -0,0 +1,32 @@
...
+        selection=[
+            ("sale_order", "Sale Order"),
+            ("stock_picking", "Stock Picking"),
+        ],
+        string="Sale Create Invoice Policy",
```
rvalyi: change the variable name and drop the string

sale_stock_picking_invoicing/models/res_company.py
```diff
@@ -0,0 +1,32 @@
...
+        help="Define, when Product Type are not service, if Invoice"
+        " should be create from Sale Order or Stock Picking.",
```
rvalyi: s/be create/be created/
s/or Stock Picking/or from StockPicking/

sale_stock_picking_invoicing/models/res_company.py
```diff
@@ -0,0 +1,32 @@
...
+        default=_default_sale_create_invoice_policy,
```
rvalyi: change to default_sale_invoicing_policy

sale_stock_picking_invoicing/models/res_config_settings.py
```diff
@@ -0,0 +1,13 @@
...
+# @author Magno Costa <magno.costa@akretion.com.br>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    sale_create_invoice_policy = fields.Selection(
```
rvalyi: change to sale_invoicing_policy

sale_stock_picking_invoicing/models/sale_order.py
```diff
@@ -0,0 +1,35 @@
...
+        ):
+            new_lines = lines.filtered(lambda ln: ln.product_id.type != "product")
+            if new_lines:
+                # Case lines with Product Type 'service'
+                lines = new_lines
+            else:
+                # Case only Products Type 'product'
+                raise UserError(
+                    _(
+                        "When 'Sale Create Invoice Policy' is defined as"
```
rvalyi: change to When 'Sale Invoicing Policy' is defined...

rvalyi (changes requested):
I reviewed a part of the module (all except the wirzard) and I suggest a few English changes. As for the wizard I'll review later, but eventually that's cool if you could check if all the comments are still required or if some might be dropped.

Actually the module isn't that large, I'm confident we can get it merged in 14.0

cc @marcelsavegnago @antoniospneto @renatonlima

sale_stock_picking_invoicing/__manifest__.py
mbcosta: Necessary to access Sale Menu
image

image

By some reason the sale module create menuitem with attribute active="False" https://github.com/OCA/OCB/blob/14.0/addons/sale/views/sale_views.xml#L8 and sale_management change for active="True" https://github.com/OCA/OCB/blob/14.0/addons/sale_management/views/sale_management_views.xml#L4

mbcosta:
Thanks @rvalyi for review, I made the changes, about the wizards it's important to check if the Refund case should not get values from "_prepare_invoice" and "_prepare_invoice_lines" of Sale module as I did, and in the _create_invoice method https://github.com/akretion/account-invoicing/blob/14.0-NEW-sale_stock_picking_invoicing/sale_stock_picking_invoicing/wizards/stock_invoice_onshipping.py#L149 I use as reference the original Sale method https://github.com/OCA/OCB/blob/14.0/addons/sale/models/sale.py#L681 to get and include the Down Payments.

sale_stock_picking_invoicing/models/sale_order.py
```diff
@@ -0,0 +1,35 @@
...
+            new_lines = lines.filtered(lambda ln: ln.product_id.type != "product")
+            if new_lines:
+                # Case lines with Product Type 'service'
+                lines = new_lines
+            else:
+                # Case only Products Type 'product'
+                raise UserError(
+                    _(
+                        "When 'Sale Invoicing Policy' is defined as"
+                        "'Stock Picking' the Invoice only can be created"
```
rvalyi: s/only can be/can only be/

sale_stock_picking_invoicing/models/sale_order.py
```diff
@@ -0,0 +1,35 @@
...
+                        " from Stock Picking, if necessary you can change"
+                        " this in Company or Sale Settings."
```
rvalyi: s/from Stock Picking/from the Stock Picking/
s/change this in Company/in the Company/

sale_stock_picking_invoicing/models/sale_order.py
```diff
@@ -0,0 +1,35 @@
...
+            self.company_id.sale_invoicing_policy == "stock_picking"
+            and model != "stock.picking"
+        ):
+            new_lines = lines.filtered(lambda ln: ln.product_id.type != "product")
+            if new_lines:
+                # Case lines with Product Type 'service'
+                lines = new_lines
+            else:
+                # Case only Products Type 'product'
+                raise UserError(
```
mileo, line 28: Could you just continue and don't add the product?

How we gonna deal with this example?

image

image

mbcosta: Thanks @mileo for review, the UserError just appear when there are only Lines with Product Type 'product' if there are any Service line the Invoice will be create, it's a way to inform the user about the "Sale Invoicing Policy" ( similar to the message when Product must be Delivered to be able Create the Invoice ), in the case of Services the 'product' must be defined as Type 'service'

image

Demo data has a Sale Order of the case of Product and Service, for now in this case will be create two Invoices, one with the Service created from Sale Order and the other from Picking, 

image

image

Service Invoice

image

Try to create Invoice from Sale Order, there lines but the policy chosen are Stock Picking

image

Create from Stock Picking

image

image

image

image

This is the expected behavior in brazilian localization, but I think this can be analise, debate and if necessary change, for example:

- If Policy are Stock Picking create only one Invoice from Picking( get Services Lines in the same way of 'Down Payments' )
- If there are Product and Service lines ignore the "Sale Invoicing Policy" and allow Create from Sale Order
- Other ideas

mbcosta:
Last force-push was fixed to get the Section and Note sale lines when create from Picking, as done when created from Sale Order

image

image

Grouping Invoice 

image

image


I'm looking to fix the error in the tests, it's seems not be related with this PR 
https://github.com/OCA/account-invoicing/actions/runs/8546887501/job/23418068649?pr=1025#step:8:725

2024-04-03 23:37:04,698 563 ERROR odoo odoo.addons.account_invoice_pricelist.tests.test_account_move_pricelist: FAIL: TestAccountMovePricelist.test_account_invoice_fixed_pricelist_without_discount_secondary_currency
  File "/__w/account-invoicing/account-invoicing/account_invoice_pricelist/tests/test_account_move_pricelist.py", line 319, in test_account_invoice_fixed_pricelist_without_discount_secondary_currency
    self.assertAlmostEqual(invoice_line.price_unit, 65.41)
AssertionError: 100.0 != 65.41 within 7 places

I change the sale_stock_picking_invoicing Tests to be  "Post Install" to check if something affect the account_invoice_pricelist, but the error persist, the tests of sale_stock_picking_invoicing run without 
 error https://github.com/OCA/account-invoicing/actions/runs/8546887501/job/23418068649?pr=1025#step:8:965

mbcosta:
Force push to:

* Include in Demo Data Sections and Notes lines and show field Inconterms
* When create grouped Sale Orders join fields "ref", "payment_reference" as done in sale module https://github.com/OCA/OCB/blob/14.0/addons/sale/models/sale.py#L754 field "invoice_origin" already been referenced from Pickings and added field "narration"( join field note from sale orders)
* Resequencing Lines to keep same sequence as Sale Orders and put Down Payments at the end of Invoice

mbcosta:
[FIX] Refund case don't get Section, Note and Down Payments lines
[REF] When Pickings has Sale with 'Down Payments' let user choose if 'Deduct down payments' as possible in Sale wizard, removed hardcode

image

sale_stock_picking_invoicing/models/res_company.py
```diff
@@ -0,0 +1,19 @@
...
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    sale_invoicing_policy = fields.Selection(
+        selection=[
+            ("sale_order", "Sale Order"),
+            ("stock_picking", "Stock Picking"),
+        ],
+        help="Define, when Product Type are not service, if Invoice"
```
kevinkhao: Suggestion to list specific effects of the selection field:
- If set to Sale Order, keep native Odoo behaviour for creation of invoices from Sale Orders
- If set to Stock Picking, disallow creation of invoices from Sale Orders

sale_stock_picking_invoicing/models/stock_rule.py
```diff
@@ -0,0 +1,14 @@
...
+# @author Magno Costa <magno.costa@akretion.com.br>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class StockRule(models.Model):
+    _inherit = "stock.rule"
+
+    def _get_custom_move_fields(self):
```
kevinkhao: should this be in stock_picking_invoicing instead ?

sale_stock_picking_invoicing/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This module extends Stock Picking Invoicing implementation to Sale, you can define the 'Sale Invoicing Policy', if the invoice should be created from Sale Order or from Stock Picking, in this case, the information used at invoice come from Sale Order e.g.: Price Unit, Payment Terms, and others fields.
```
kevinkhao: This is missing the one of the functionalities of this module : for stock.moves, override price calculation that is present in stock_picking_invoicing, with the native Sale Order Line price calculation, same for the partner_id

sale_stock_picking_invoicing/wizards/stock_invoice_onshipping.py
```diff
@@ -12,37 +12,88 @@ class StockInvoiceOnshipping(models.TransientModel):
...
+                refs.add(sale_values["ref"])
+                narration.add(sale_values["narration"])
+
+                # Original dict from sale module, fields uncomment will be remove
+                vals_to_remove = {
+                    # "ref": self.client_order_ref or '',
+                    "move_type",
+                    # "narration": self.note,
+                    "currency_id",
+                    # "campaign_id": self.campaign_id.id,
```
kevinkhao: As the module is not in a draft state, I think these comments should be removed, maybe concatenate everything in one line as a comment for the different options you might need for reference

sale_stock_picking_invoicing/tests/test_sale_stock.py
```diff
@@ -0,0 +1,429 @@
...
+        Test Sale Order with product and service
+        """
+
+        sale_order_2 = self.env.ref(
+            "sale_stock_picking_invoicing.main_company-sale_order_2"
+        )
+        # Necessary to get the currency
+        sale_order_2.onchange_partner_id()
+        sale_order_2.action_confirm()
+        self.assertTrue(sale_order_2.state == "sale")
```
kevinkhao: a lot of these assertions seem to be testing Odoo native functionality and not relevant to this module, only things that are impacted by this module should be tested

sale_stock_picking_invoicing/models/res_company.py
mbcosta: Thanks @kevinkhao for your review, sorry for take time to answer, I needed to check a similar module [l10n_br_purchase_stock](https://github.com/OCA/l10n-brazil/tree/14.0/l10n_br_purchase_stock) adapt stock_picking_invoicing but for Purchase, for comparison and see if something could was missing here. About your suggestion I included your describe in the Field and README

sale_stock_picking_invoicing/models/stock_rule.py
mbcosta: Yes PR https://github.com/OCA/account-invoicing/pull/1782 should FIX

sale_stock_picking_invoicing/readme/DESCRIPTION.rst
mbcosta: I tried to include your description, let me know if now it's better

sale_stock_picking_invoicing/wizards/stock_invoice_onshipping.py
mbcosta: I separated Fields do Get and to Remove, but let as reference, if necessary I can remove it

sale_stock_picking_invoicing/tests/test_sale_stock.py
mbcosta: I tried to remove all redundant tests

mbcosta:
Force-Push to remove file **sale_stock_picking_invoicing/models/stock_rule.py**, because it should be in **stock_picking_invocing** as suggest in the above Review, by this reason the CI tests fail here so this PR now depends of PR https://github.com/OCA/account-invoicing/pull/1782 ( already with two Approves )

rvalyi:
@mbcosta #1782 was merged. I also re-launched the failed tests as the date pricelist issue in the CI seems fixed.

mbcosta:
Thanks @rvalyi with rebase the PR back to be green, I took a time to check the PR again to be sure everything work as well with [Extract PR of l10n_br_sale_stock](https://github.com/OCA/l10n-brazil/pull/2955)

sale_stock_picking_invoicing/models/stock_picking.py
```diff
@@ -0,0 +1,18 @@
...
+
+from odoo import models
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    def _get_partner_to_invoice(self):
+        partner_id = super()._get_partner_to_invoice()
+        partner = self.env["res.partner"].browse(partner_id)
```
rvalyi: you can avoid this browse here if you compare partner_id with self.sale_id.partner_invoice_id.id directly 2 lines below in `if partner != self.sale_id.partner_invoice_id: `
(partner_invoice_id is required so no need to check it's not False/None)

rvalyi (changes requested):

sale_stock_picking_invoicing/models/stock_picking.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    def _get_partner_to_invoice(self):
+        partner_id = super()._get_partner_to_invoice()
+        if self.sale_id:
+            if partner_id != self.sale_id.partner_invoice_id.id:
+                partner_id = self.sale_id.partner_invoice_id.id
```
rvalyi: in fact, you can simplify further to just:

```
        if self.sale_id:
            partner_id = self.sale_id.partner_invoice_id.id
```
if it's the same it will affect partner_id but it won't hurt performance and the code will be simpler

or eventually something like:

`partner_id = self.sale_id.partner_invoice_id.id if self.sale_id else super()._get_partner_to_invoice()`

rvalyi (changes requested):

sale_stock_picking_invoicing/models/stock_picking.py
mbcosta: Done, the error in the tests is the problem we saw yesterday the Tests of module **account_invoice_pricelist** fail at a specific Date Time

rvalyi: OK, the PR seems OK now. I will wait a few hours, trigger the tests again and approve once everything is green again. Thank you for the awesome work and sorry for the delay. Would be great if you could port it to 16.0 as well... (you can skip v15)

mbcosta: I didn't saw your comment and just make a force push, I have tried to find where is the problem but not success, the clue is the problem related with Date Time, after UTC midnight the tests back to work

rvalyi (approved):
OK for me.

mbcosta:
Thanks @rvalyi for your review, no problems for delay, this was important to make a better implementation

antoniospneto (approved):
LGTM

rvalyi:
and finally...

/ocabot merge nobump

sale_stock_picking_invoicing/models/stock_picking.py
mbcosta: @rvalyi I decide to make v15 migration https://github.com/OCA/account-invoicing/pull/1826 just to avoid the possibility of a problem for the cases of someone by some reason still using v15, after merge I'll port to v16

rvalyi: Ok, perfect

PR #1027, opened 2021-10-27
[FIX] account_invoice_supplier_self_invoice: Set ref as self_invoice_number

etobella:
Without this change if a self invoice was created, no ref was defined. That was a problem when using account_payment_order, as it required a communication text.

A migration script is added in order to fix old data.


AaronHForgeFlow (approved):
code review LGTM

albariera (approved):
LGTM :+1:

HviorForgeFlow (approved):
LGTM :+1:

PR #1050, opened 2021-12-02
[14.0][IMP] account_invoice_section_sale_order: custom section name and refactor

grindtildeath:
Contains:
 - Possibility to customize the invoice section name using python expression with object and time variables
 - Refactor to allow extension module to group invoice lines with different model than sale.order

TODO:

 - [x] Improve readme
 - [x] Add views
 - [x] Fix help strings
 - [x] Check new sequence assignation is not messed up with refactor
 - [x] Run pre-commit

### DIFF
--- a/account_invoice_section_sale_order/models/account_move.py
+++ b/account_invoice_section_sale_order/models/account_move.py
@@ -1,0 +1,40 @@
+# Copyright 2021 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo import _, api, models
+from odoo.exceptions import UserError
+
+
+class AccountMove(models.Model):
+
+    _inherit = "account.move"
+
+    def _get_ordered_invoice_lines(self):
+        """Sort invoice lines according to the section ordering"""
+        return self.invoice_line_ids.sorted(
+            key=self.env["account.move.line"]._get_section_ordering()
+        )
+
+
+class AccountMoveLine(models.Model):
+
+    _inherit = "account.move.line"
+
+    def _get_section_group(self):
+        """Return the section group to be used for a single invoice line"""
+        self.ensure_one()
+        return self.mapped(self._get_section_grouping())
+
+    def _get_section_grouping(self):
+        """Defines the grouping relation from the invoice lines to be used.
+
+        Meant to be overriden, in order to allow custom grouping.
+        """
+        invoice_section_grouping = self.company_id.invoice_section_grouping
+        if invoice_section_grouping == "sale_order":
+            return "sale_line_ids.order_id"
+        raise UserError(_("Unrecognized invoice_section_grouping"))
+
+    @api.model
+    def _get_section_ordering(self):
+        """Function to sort invoice lines before grouping"""
+        return lambda r: r.mapped(r._get_section_grouping())
--- a/account_invoice_section_sale_order/models/sale_order.py
+++ b/account_invoice_section_sale_order/models/sale_order.py
@@ -1,57 +1,67 @@
 # Copyright 2020 Camptocamp SA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+from collections import OrderedDict
 
 from odoo import models
+from odoo.tools.safe_eval import safe_eval, time
 
 
 class SaleOrder(models.Model):
     _inherit = "sale.order"
 
     def _create_invoices(self, grouped=False, final=False, date=None):
-        """Add sections by sale order in the invoice line.
+        """Add sections by groups in the invoice line.
 
-        Order the invoicing lines by sale order and add lines section with
-        the sale order name.
-        Only do this for invoices targetting multiple sale order
+        Order the invoicing lines by groups and add lines section with
+        the group name.
+        Only do this for invoices targetting multiple groups
         """
         invoice_ids = super()._create_invoices(grouped=grouped, final=final, date=date)
         for invoice in invoice_ids:
-            if len(invoice.line_ids.mapped("sale_line_ids.order_id.id")) == 1:
+            if (
+                len(invoice.line_ids.mapped(invoice.line_ids._get_section_grouping()))
+                == 1
+            ):
                 continue
-            so = None
             sequence = 10
+            move_lines = invoice._get_ordered_invoice_lines()
+            # Group move lines according to their sale order
+            section_grouping_matrix = OrderedDict()
+            for move_line in move_lines:
+                group = move_line._get_section_group()
+                section_grouping_matrix.setdefault(group, []).append(move_line.id)
+            # Prepare section lines for each group
             section_lines = []
-            lines = self._get_ordered_invoice_lines(invoice)
-            for line in lines:
-                if line.sale_line_ids.order_id and so != line.sale_line_ids.order_id:
-                    so = line.sale_line_ids.order_id
+            for group, move_line_ids in section_grouping_matrix.items():
+                if group:
                     section_lines.append(
                         (
                             0,
                             0,
                             {
-                                "name": so._get_saleorder_section_name(),
+                                "name": group._get_invoice_section_name(),
                                 "display_type": "line_section",
                                 "sequence": sequence,
                             },
                         )
                     )
                     sequence += 10
-                line.sequence = sequence
-                sequence += 10
+                for move_line in self.env["account.move.line"].browse(move_line_ids):
+                    move_line.sequence = sequence
+                    sequence += 10
             invoice.line_ids = section_lines
-
         return invoice_ids
 
-    def _get_ordered_invoice_lines(self, invoice):
-        return invoice.line_ids.sorted(
-            key=lambda r: r.sale_line_ids.order_id.id
-        ).filtered(lambda r: not r.exclude_from_invoice_tab)
-
-    def _get_saleorder_section_name(self):
+    def _get_invoice_section_name(self):
         """Returns the text for the section name."""
         self.ensure_one()
-        if self.client_order_ref:
+        naming_scheme = (
+            self.partner_invoice_id.invoice_section_name_scheme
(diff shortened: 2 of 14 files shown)


ajaniszewska-dev (approved):

ivantodorovich (approved):

account_invoice_section_sale_order/models/account_move.py
```diff
@@ -0,0 +1,33 @@
...
+class AccountMoveLine(models.Model):
+
+    _inherit = "account.move.line"
+
+    def _get_section_group(self):
+        return self.mapped(self._get_section_grouping())
+
+    @api.model
+    def _get_section_grouping(self):
+        invoice_section_grouping = self.env.company.invoice_section_grouping
```
i-vyshnevska: this should relly on the company from invoice not from env

account_invoice_section_sale_order/models/sale_order.py
```diff
@@ -16,42 +18,53 @@ def _create_invoices(self, grouped=False, final=False):
         """
         invoice_ids = super()._create_invoices(grouped=grouped, final=final)
         for invoice in invoice_ids:
-            if len(invoice.line_ids.mapped("sale_line_ids.order_id.id")) == 1:
+            if (
+                len(invoice.line_ids.mapped(invoice.line_ids._get_section_grouping()))
```
TDu, line 22:
```suggestion
                len(invoice.line_ids._get_section_group())
```
Maybe

account_invoice_section_sale_order/models/account_move.py
```diff
@@ -0,0 +1,33 @@
...
+
+class AccountMoveLine(models.Model):
+
+    _inherit = "account.move.line"
+
+    def _get_section_group(self):
+        return self.mapped(self._get_section_grouping())
+
+    @api.model
+    def _get_section_grouping(self):
```
TDu, line 27: I don't think we could have multiple grouping in one invoice so could be moved to account.move model.

account_invoice_section_sale_order/models/sale_order.py
```diff
@@ -16,42 +18,53 @@ def _create_invoices(self, grouped=False, final=False):
...
+                            "name": group._get_invoice_section_name(),
+                            "display_type": "line_section",
+                            "sequence": sequence,
+                        },
                     )
-                    sequence += 10
-                line.sequence = sequence
+                )
                 sequence += 10
+                for move_line in self.env["account.move.line"].browse(move_line_ids):
```
TDu, line 49: If above (line 34 and 36) the move_line record is added to the grouping instead of the id we would not need to browse here again. and we could just loop on move_line_ids to update the sequence

simahawk: interpolating can be slow on large recordsets. In this case is fine. Also, these records will be loaded from cache.

account_invoice_section_sale_order/models/sale_order.py
```diff
@@ -16,42 +18,53 @@ def _create_invoices(self, grouped=False, final=False):
...
+            move_lines = invoice._get_ordered_invoice_lines()
+            # Group move lines according to their sale order
+            section_grouping_matrix = OrderedDict()
+            for move_line in move_lines:
+                group = move_line._get_section_group()
+                group_move_line_ids = section_grouping_matrix.get(group)
+                if not group_move_line_ids:
+                    section_grouping_matrix[group] = [move_line.id]
+                else:
+                    group_move_line_ids.append(move_line.id)
```
simahawk:
```suggestion
                section_grouping_matrix.setdefault(group, []).append(move_line.id)
```

account_invoice_section_sale_order/tests/test_invoice_group_by_sale_order.py
```diff
@@ -80,21 +80,25 @@ def setUpClass(cls):
...
         lines = (
             invoice_ids[0]
             .line_ids.sorted("sequence")
             .filtered(lambda r: not r.exclude_from_invoice_tab)
         )
-        for idx, line in enumerate(lines):
-            self.assertEqual(line.name, result[idx])
+        for line in lines:
+            self.assertEqual(line.name, result[line.sequence][0])
+            self.assertEqual(line.display_type, result[line.sequence][1])
```
simahawk, line 110: would be nice to see a test on custom grouping

account_invoice_section_sale_order/models/account_move.py
simahawk: AFAIR the point is that a line could come from different delivery orders

simahawk (changes requested):

account_invoice_section_sale_order/models/account_move.py
simahawk: seems legit

simahawk:
remember to squash some commits when done

account_invoice_section_sale_order/models/sale_order.py
TDu: Interpolating seems to be a kind of estimation at finding new data point are we sure this is related ?
I am surprised browsing for a record is not slower.
Also about performance for this, the implementation goes from one  loop to 3 indented loops.

simahawk: well, I assumed you wanted to collect move line line `mapping[group] |= line` but if you just append the to a list is fine.

account_invoice_section_sale_order/models/account_move.py
grindtildeath: Actually it's the move lines that are grouped by section, so IMO it makes sense to define this function on this model.

account_invoice_section_sale_order/models/sale_order.py
grindtildeath: I'm not sure if the performance is really degraded through this refactor and would be interesting to know. As browsing should read from the cache since we already had to read these records, I chose to keep the ids because unions of recordset is usually more costly.

Then from my POV the code is still quite readable between:
1. Grouping of lines
2. Preparation of section lines to create
3. Update of the existing lines sequence
4. Creation of section lines

If performance is really an issue, I guess we can revisit it.

account_invoice_section_sale_order/tests/test_invoice_group_by_sale_order.py
grindtildeath: Done is separated module :wink:

account_invoice_section_sale_order/models/account_move.py
grindtildeath: Is there really a case where we could have something different between the invoice we produce and the company from env? I used this because I defined the function as `api.model` to avoid relying on actual move lines records

i-vyshnevska: in multi-company mode, if you would have a cron it will be called under main company and invoices may belong to different 
I always prefer to use a company from the obj if there is a special field for it instead of env

account_invoice_section_sale_order/tests/test_invoice_group_by_sale_order.py
simahawk: sure, but if you don't install the other module and run tests for it, you'll never know if it's broken ;)

Maybe add a TODO like "TODO: test this thing here as for now is tested on only in module xyz"

account_invoice_section_sale_order/models/account_move.py
grindtildeath: That's true especially after my issue on queue jobs, fixed.

account_invoice_section_sale_order/models/sale_order.py
grindtildeath: I reverted this change, because we need a different function to get the grouping for invoice lines, and to get the group for one invoice line. (I added `ensure_one`) to avoid breaking this.

i-vyshnevska (approved):

account_invoice_section_sale_order/tests/test_invoice_group_by_sale_order.py
```diff
@@ -113,3 +115,57 @@ def test_create_invoice_no_section(self):
...
+        is possible by redefining these functions"""
+        demo_user = self.env.ref("base.user_demo")
+        admin_user = self.env.ref("base.partner_admin")
+        orders = self.order1_p1 + self.order2_p1
+        orders.write({"user_id": admin_user.id})
+        sale_order_3 = self.order1_p1.copy({"user_id": demo_user.id})
+        sale_order_3.order_line[0].name = "order 3 line 1"
+        sale_order_3.order_line[1].name = "order 3 line 2"
+        sale_order_3.action_confirm()
+        section_grouping_func = "odoo.addons.account_invoice_section_sale_order.models.account_move.AccountMoveLine._get_section_grouping"  # noqa
```
simahawk: maybe move these paths to constant var on the top?

grindtildeath:
@simahawk Updated, commits squashed, can you please update your review? :pray:

account_invoice_section_sale_order/tests/test_invoice_group_by_sale_order.py
```diff
@@ -1,16 +1,25 @@
 # Copyright 2020 Camptocamp SA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+import mock
 
+from odoo.exceptions import UserError
 from odoo.tests.common import SavepointCase
 
+SECTION_GROUPING_FUNCTION = "odoo.addons.account_invoice_section_sale_order.models.account_move.AccountMoveLine._get_section_grouping"  # noqa
```
simahawk, line 8:
```suggestion
SECTION_GROUPING_FUNCTION = (
    "odoo.addons.account_invoice_section_sale_order"
    ".models.account_move.AccountMoveLine._get_section_grouping"
)
```

simahawk (approved):

PR #1051, opened 2021-12-02
[14.0] Create module account_invoice_section_picking

grindtildeath:
Depends on:
 - [x] #1050 

TODO:
 - [x] Add more tests
 - [x] Add readme
 - [x] Run pre-commit
 - [x] Add test dependency


ajaniszewska-dev (approved):
check travis, code LG

ivantodorovich:
Travis is red

account_invoice_section_picking/models/account_move_line.py
```diff
@@ -0,0 +1,46 @@
...
+
+    _inherit = "account.move.line"
+
+    def _get_section_group(self):
+        group = super()._get_section_group()
+        # If product is invoiced by delivered quantities:
+        #  - filter on done pickings to avoid displaying backorders
+        #  - Remove pickings linked with same sale order lines if an invoice
+        #    was created after its date_done
+        invoice_section_grouping = self.env.company.invoice_section_grouping
```
i-vyshnevska: same comment about company

account_invoice_section_picking/models/stock_picking.py
```diff
@@ -0,0 +1,25 @@
...
+
+    _inherit = "stock.picking"
+
+    def _get_invoice_section_name(self):
+        """Returns the text for the section name."""
+        section_names = []
+        for pick in self:
+            naming_scheme = (
+                self.partner_id.invoice_section_name_scheme
+                or self.company_id.invoice_section_name_scheme
```
simahawk: :fire:
```suggestion
            naming_scheme = (
                pick.partner_id.invoice_section_name_scheme
                or pick.company_id.invoice_section_name_scheme
```

account_invoice_section_picking/tests/common.py
```diff
@@ -0,0 +1,29 @@
+# Copyright 2021 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from odoo.tests.common import Form, SavepointCase
+
+
+class TestAccountInvoiceSectionPickingCommon(SavepointCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
```
simahawk, line 9:
```suggestion
        super().setUpClass()
        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
```

account_invoice_section_picking/tests/test_account_invoice_section_picking.py
```diff
@@ -0,0 +1,97 @@
...
+    def _create_order(cls, order_line_name=None):
+        order_form = Form(cls.env["sale.order"])
+        order_form.partner_id = cls.partner_1
+        with order_form:
+            with order_form.order_line.new() as line_form:
+                line_form.product_id = cls.product_1
+                line_form.product_uom_qty = 5
+                if order_line_name is not None:
+                    line_form.name = order_line_name
+        return order_form.save()
```
simahawk:
```suggestion
```
already on common class no?

account_invoice_section_picking/tests/test_account_invoice_section_picking_backorder.py
```diff
@@ -0,0 +1,168 @@
...
+        super().setUp()
+        self._monkey_patch_cursor_freeze_time()
+        self.addCleanup(self._reset_monkey_patch_cursor_freezetime)
+
+    def _reset_monkey_patch_cursor_freezetime(self):
+        self.env.cr.execute = self.env.cr.standard_exec
+
+    def _monkey_patch_cursor_freeze_time(self):
+        """Monkey patch cursor to replace SQL TIME/DATE keywords"""
+        # TODO: Extract this part into a fork of freezegun for odoo?
```
simahawk, line 32: I'd move it to its own package and provide a decorator and a ctx manager to be used alone. Then we put it in a new repo like `psql-freeze-time` or alike in the OCA org.

simahawk (changes requested):

account_invoice_section_picking/tests/test_account_invoice_section_picking_backorder.py
grindtildeath: I'll try to do this as soon as I have time, but I guess we'll have to do something like `odoo-freezegun` because I'm not sure we can do it in psycopg2, since we need it only when using a test cursor.

Anyway, I'll welcome your help on this :wink:

account_invoice_section_picking/__manifest__.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2021 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+{
+    "name": "Acccount Invoice Section Picking",
+    "version": "14.0.1.0.1",
```
i-vyshnevska:
```suggestion
    "version": "14.0.1.0.0",
```

account_invoice_section_picking/models/account_move_line.py
```diff
@@ -0,0 +1,47 @@
...
+        #    was created after its date_done
+        invoice_section_grouping = self.company_id.invoice_section_grouping
+        if (
+            invoice_section_grouping == "delivery_picking"
+            and self.product_id.invoice_policy == "delivery"
+        ):
+            last_invoice_line = self.search(
+                [
+                    ("sale_line_ids", "in", self.sale_line_ids.ids),
+                    ("id", "not in", self.ids),
```
i-vyshnevska:
```suggestion
                    ("id", "!=", self.id),
```
the result will be the same but this is more precise presentation of what is happening here

account_invoice_section_picking/models/account_move_line.py
```diff
@@ -0,0 +1,47 @@
...
+                ],
+                order="create_date",
+                limit=1,
+            )
+            pickings_already_invoiced = self.env["stock.picking"].search(
+                [
+                    ("date_done", "<=", last_invoice_line.create_date),
```
i-vyshnevska, line 34:
```suggestion
                    ("date_done", "<=", last_invoice_line.create_date),
```
this condition doesn't guarantee that this picking was invoiced, sale.order.line and fields like qty_to_invoice is more reliable data if we need concider only invoiced pickings

grindtildeath: @i-vyshnevska Do you see any use case where this code would not give the expected results?

i-vyshnevska: invoice can be canceled for some reason and a new one created, partial deliveries, partial invoices
create_date - is rather a technical field for the functional side there are other separate fields 
and having such difficulties to have a proper time in the test case is confirming that the functional side tries to rely on a side that shouldn't be here

grindtildeath: I added an extra domain condition to avoid selecting a canceled invoice. :+1: 
Then, partial deliveries are actually covered through this (it's the whole point of this override).
I agree that a reduction of the quantity on the invoice lines could bring issues, so I did state this in the readme as a known issue, because I don't see any possible way handle this use case properly.
Finally, I would have preferred to be able to avoid relying on `create_date` but we cannot rely on the `date` because it's only set on the posting. Then as the `qty_invoiced` is already taken into account when an invoice is in draft, the only other possibility to know what was delivered after the last invoice was generated would have been to define a generated date on the invoice, which would have been the same than the `create_date`. So I guess it's ok to use it in such a case and I also did state it in the readme.

account_invoice_section_picking/models/account_move_line.py
```diff
@@ -0,0 +1,48 @@
...
+        #    was created after its date_done
+        invoice_section_grouping = self.company_id.invoice_section_grouping
+        if (
+            invoice_section_grouping == "delivery_picking"
+            and self.product_id.invoice_policy == "delivery"
+        ):
+            last_invoice_line = self.search(
+                [
+                    ("sale_line_ids", "in", self.sale_line_ids.ids),
+                    ("state", "!=", "cancel"),
```
ajaniszewska-dev:
```suggestion
                    ("parent_state", "!=", "cancel"),
```

as discussed on slack. Tested locally - works fine.

ajaniszewska-dev (changes requested):

test_requirements.txt
```diff
@@ -0,0 +1 @@
+git+https://github.com/OCA/account-invoicing@refs/pull/1050/head#subdirectory=setup/account_invoice_section_sale_order
```
simahawk: this has to be dropped. Anyway. The file name is wrong: should test-requirements.txt :wink:

simahawk (commented):
can you squash pls?

grindtildeath:
@simahawk @i-vyshnevska @ajaniszewska-dev Can you please update your reviews? :pray:

ajaniszewska-dev (approved):

simahawk (approved):

PR #1054, opened 2021-12-04
[14.0][MIG] sale_order_invoicing_queued: Migration to 14.0

iwkse:
@pedrobaeza That's the porting using the XML instead of the decorator, but it doesn't seem to work. I hope the PR so that we can proceed with the module,
Thank you!


iwkse:
@pedrobaeza please when it's possible give me a feedback
Thank you

sale_order_invoicing_queued/data/job_queue.xml
```diff
@@ -0,0 +1,8 @@
+<odoo noupdate="1">
+  <record id="job_function_sale_order_invoicing_action_done" model="queue.job.function">
+    <field name="model_id" ref="sale.model_sale_order" />
+    <field name="method">create_invoices_job</field>
+    <field name="related_action" eval='{"func_name": "custom_related_action"}' />
```
pedrobaeza: Is this one correct? (I don't know the syntax)

pedrobaeza (approved):

sale_order_invoicing_queued/data/job_queue.xml
iwkse: That's from the example provided in the README file

pedrobaeza: But I think you should provide a valid action name. If not, I would say to remove it and it should take the default one.

pedrobaeza: I was talking about this line, not the whole file.

pedrobaeza:
Travis is red

iwkse:
> Travis is red
Yes, when removing the action it requires the channel_id configuration I think. I'm going to fix that asap

iwkse:
> Travis is red

fixed it

pedrobaeza (approved):

HaraldPanten (changes requested):
Couldn't make it work in runbot. Steps:

1- I Created a sales order with an invoiceable line.
2- I went to the tree view to create the invoice and I selected to enqueue it in the wizard.
3- The job seems to be enqueued, but it never gets finished and the invoice is not created.

Am I missing anything?

THX

pedrobaeza:
Runbot is not configured for running jobs, so that's the problem. It's not something of this module.

HaraldPanten (approved):
LGTM 👍

PR #1055, opened 2021-12-08
[14.0][FIX]  account_invoice_payment_retention: compute_amount on payment wizard

Saran440:
Move on `account_payment.py` (model) to `account_payment_register.py` (wizard)
and Fix compute_amount should be call super() before dos something.

### DIFF
--- a/account_invoice_payment_retention/models/__init__.py
+++ b/account_invoice_payment_retention/models/__init__.py
@@ -3,4 +3,3 @@
 from . import res_config_settings
 from . import res_company
 from . import account_move
-from . import account_payment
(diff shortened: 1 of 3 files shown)


kittiu:
@Saran440 can you describe the error without this fix?

kittiu:
There module by itself, is working OK, but has problem when use with `account_payment_multi_deduction` and `l10n_th_account_tax_multi`. The fix here make it works in case with / without the module.

kittiu:
FFW

/ocabot merge patch

kittiu (approved):

PR #1058, opened 2021-12-09
[14.0][FIX] account_invoice_payment_retention, return retention with …

kittiu:
Note: this bug occur when use with `account_payment_multi_deduction`

image

In the picture, the Return Retention (account.payment) was multi deduct with multiple invoice. And so, this fix, the warning will occure on invoice posting.


ps-tubtim (approved):

Saran440 (approved):
LGTM :+1:

PR #1059, opened 2021-12-10
[14.0][MIG] account_invoice_validation_queued

iwkse:


iwkse:
@pedrobaeza ready

account_invoice_validation_queued/migrations/14.0.2.0.0/post-migration.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    if not openupgrade.table_exists(env.cr, "account_move_queue_job_rel"):
+        return
+    openupgrade.logged_query(
+        env.cr,
+        """
+        INSERT INTO account_move_validation_job_rel
```
pedrobaeza: Why is this needed? Can't you preserve the old name or rename the table in pre?

iwkse: @pedrobaeza means like that?

pedrobaeza: Not really. Now I see both post and pre migration scripts doing the insert. My question about why is this needed still unanswered, and without it, I can't recommend you the actions to take.

iwkse: Right. That's something from 13.0 and I believe it can be removed. I've pushed without the migration folder.
Thank you @pedrobaeza

pedrobaeza (approved):

iwkse:
@pedrobaeza There's the README which should be updated as it doesn't reflect the actual steps to see the validation popup.
I've seen the information seems to be wrong also for the 13.0, so I'm a bit confused to understand how to find such popup.
Is such documentation valid for 12.0?
Could you please help to clarify it?
Thank you!

pedrobaeza:
The action entry is called "Post entries" instead.

iwkse:
@pedrobaeza README updated!

francesco-ooops (approved):
From "usage":

If any of the selected invoices are not in draft state, you will receive an error.

This is not working - it is't working on 13 either, while it's ok on 12. I suggest to change description as the feature doesn't seem so important (posted invoices do not get enqueued anyway)

For the rest, functional review ok for what job queue allows

iwkse:
> From "usage":
> 
> If any of the selected invoices are not in draft state, you will receive an error.
> 
> This is not working - it is't working on 13 either, while it's ok on 12. I suggest to change description as the feature doesn't seem so important (posted invoices do not get enqueued anyway)
> 
> For the rest, functional review ok for what job queue allows

@pedrobaeza what do you think?

pedrobaeza:
Proceed the way you think is better.

iwkse:
@pedrobaeza PR updated with changes in USAGE.rst

iwkse:
@pedrobaeza could you try now merging?

thank you

PR #1075, opened 2022-01-11
[14.0][FIX] account_invoice_section_sale_order: Fix import in test_invoice_group_by_sale_order

albingilles:
Fix import in test_invoice_group_by_sale_order causing issue while executing in only python3 environment


albingilles:
Hi @simahawk, would it be possible for you to check this PR and merge it ?

I've just added a conditional import of mock in the tests which I found in other part of Odoo unit test as we were having build problem because our Docker image only ships with python3.

Thanks !

PR #1076, opened 2022-01-11
[13.0][FIX] account_invoice_pricelist: Avoid division by zero

mrgoetz:
Module account_invoice_pricelist:
When calculating the discount, pricelists discount_policy not
"with_discount", the percentage of the discount is calculated.
If base_price is 0.0 we receive an error.

So we try to avoid this by checking the base_price.


AaronHForgeFlow (commented):
LGTM can you please add the module name in the title of this PR? and also in the commit message. It should be "[FIX]account_invoice_pricelist: avoid division by zero" or something similar

ferran-S73 (approved):

hildickethan (approved):

rafaelbn:
@AaronHForgeFlow please could you re-review?

AaronHForgeFlow (approved):

AaronHForgeFlow:
I see there is a FW port to 16: https://github.com/OCA/account-invoicing/pull/1485 What about 14.0 and 15.0? is that done already?

AaronHForgeFlow:
I see the fix is already in 14.0 but not in 15.0, would you create the forward port?

AaronHForgeFlow:
Sorry, I did  not realize the commit history is not correct and it has 3 commits. Apologies.

AaronHForgeFlow:
Just did the PW port to v15 here: https://github.com/OCA/account-invoicing/pull/1641

PR #1079, opened 2022-01-18
[13.0][FIX] account_invoice_fixed_discount: Ensure price_unit is different …

HviorForgeFlow:
It could happen that a fixed discount is equal to total line product price. This would lead to Zero-division error when trying to calculate the standard discount based on price_unit.

CC @ForgeFlow


AaronHForgeFlow (commented):
if vals.get("price_unit") is 0.0 it can also be ZeroDivisionError, right?

PR #1069, opened 2022-01-03
[13.0][ADD] account_invoice_origin_link and account_invoice_sale_origin_link

OriolVForgeFlow:
This module adds a clickable link to "Source Documents" in Invoices.
@ForgeFlow @JordiMForgeFlow

PR #1081, opened 2022-01-18
[13.0][FIX] account_invoice_show_currency_rate: Round currency_rate_amount to avoid many decimal places 

victoralmau:
Round currency_rate_amount to avoid many decimal places (1.9974358974358974 for example).

Related to and indirectly discovered at https://github.com/OCA/account-invoicing/pull/1069.

Please @ernestotejeda and @pedrobaeza can you review it?

@Tecnativa

### DIFF
--- a/account_invoice_show_currency_rate/models/account_move.py
+++ b/account_invoice_show_currency_rate/models/account_move.py
@@ -34,7 +34,9 @@ class AccountMove(models.Model):
             if item.state == "posted" and lines:
                 amount_currency_positive = sum(lines.mapped("amount_currency"))
                 total_debit = sum(item.line_ids.mapped("debit"))
-                item.currency_rate_amount = amount_currency_positive / total_debit
+                item.currency_rate_amount = item.currency_id.round(
+                    amount_currency_positive / total_debit
+                )
             else:
                 rates = item.currency_id._get_rates(item.company_id, item.date)
                 item.currency_rate_amount = rates.get(item.currency_id.id)
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -1,0 +1,1 @@
+freezegun


account_invoice_show_currency_rate/models/account_move.py
```diff
@@ -34,7 +34,10 @@ def _compute_currency_rate_amount(self):
             if item.state == "posted" and lines:
                 amount_currency_positive = sum(lines.mapped("amount_currency"))
                 total_debit = sum(item.line_ids.mapped("debit"))
-                item.currency_rate_amount = amount_currency_positive / total_debit
+                item.currency_rate_amount = round(
```
pedrobaeza: Better use `item.currency_id.round`

test-requirements.txt
```diff
@@ -0,0 +1 @@
+freezegun
```
pedrobaeza, line 1: Why this?

victoralmau: This is necessary to avoid the runboat action error indicating that dependency is missing.

pedrobaeza: OK, it seems a bug of it, but not going to argue right now.

account_invoice_show_currency_rate/models/account_move.py
```diff
@@ -34,7 +34,10 @@ def _compute_currency_rate_amount(self):
             if item.state == "posted" and lines:
                 amount_currency_positive = sum(lines.mapped("amount_currency"))
                 total_debit = sum(item.line_ids.mapped("debit"))
-                item.currency_rate_amount = amount_currency_positive / total_debit
+                item.currency_rate_amount = item.currency_id.round(
+                    (amount_currency_positive / total_debit),
+                    item.currency_id.decimal_places,
```
pedrobaeza: I don't think this is the syntax of the method. It already knows the decimal places.

pedrobaeza (approved):

pedrobaeza:
Let's fast track it:

/ocabot merge patch

PR #1082, opened 2022-01-19
[14.0][ADD] account_invoice_line_unit_price_untaxed

rousseldenis:


account_invoice_line_unit_price_no_tax/models/account_move_line.py
```diff
@@ -0,0 +1,21 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class AccountMoveLine(models.Model):
+
+    _inherit = "account.move.line"
+
+    price_unit_no_taxes = fields.Float(
```
legalsylvain: This field shloud have the same precision as the original unit price field, don't you think ?
https://github.com/odoo/odoo/blob/15.0/addons/account/models/account_move.py#L3496

legalsylvain (commented):
Thanks for this module.
Lgtm. (Code review / no test)

However, regarding the name of the module and the field, I think that "vat_excluded" is more explicit (and consistent with other naming) than "no_tax" and "no_taxes".

rousseldenis:
> However, regarding the name of the module and the field, I think that "vat_excluded" is more explicit (and consistent with other naming) than "no_tax" and "no_taxes".

Thanks for your review. But no, you can have other taxes that are not vat. This should stay as is.

legalsylvain:
> Thanks for your review. But no, you can have other taxes that are not vat. This should stay as is.

I didn't understood. The field you introduced is an untaxed price unit. Or did I missed something ?

```             line.price_unit_no_taxes = tot["total_excluded"]```

**Comment by rousseldenis** [comment]
> > Thanks for your review. But no, you can have other taxes that are not vat. This should stay as is.
> 
> I didn't understood. The field you introduced is an untaxed price unit. Or did I missed something ?
> 
> ` line.price_unit_no_taxes = tot["total_excluded"]`

Yes. but I spoke about field naming. Speaking about 'vat' is too limitative.

**Comment by legalsylvain** [comment]
You're right on that point, thanks.

However, I still think that other name could be better : 
- See ``account`` module with ``amount_untaxed``, ``amount_tax`` and ``amount_total`` (https://github.com/odoo/odoo/blob/15.0/addons/account/models/account_move.py#L231)
 - see ``sale`` module with same naming https://github.com/odoo/odoo/blob/15.0/addons/sale/models/sale_order.py#L233

I propose to name the new field ``price_unit_untaxed`` (and the module ``account_invoice_line_price_unit_untaxed``)


``rgrep "no_tax.*field" ./ --include=*.py`` return not occurence in all odoo master
``rgrep "untaxed.*field" ./ --include=*.py`` return many occurences in odoo master

**Comment by rousseldenis** [comment]
FYI there is also https://github.com/OCA/sale-workflow/pull/1877

File: account_invoice_line_unit_price_untaxed/tests/test_account_move_line_untaxed.py [hunk 2]
```diff
@@ -140,3 +140,41 @@ def test_invoice_with_product_company2(self):
...
+        precision = self.env["decimal.precision"].search(
+            [("name", "=", "Product Price")]
+        )
+        precision.write({"digits": 3})
+        self.product.list_price = 118.573
+        with Form(self.invoice) as invoice_form:
+            with invoice_form.invoice_line_ids.new() as invoice_line:
+                invoice_line.product_id = self.product
+        invoice_line = first(self.invoice.invoice_line_ids)
+        self.assertEqual(invoice_line.price_unit, 118.573)
```
acsonefho:
```
# Get the precision of your field (into setUpClass(...) )
cls.precision = cls.env["decimal.precision"].precision_get(self._fields["price_unit"]._digits)
...
self.assertAlmostEquals(invoice_line.price_unit, 118.573, places=self.precision)
self.assertAlmostEquals(invoice_line.price_unit_untaxed, 118.573, places=self.precision)
```

account_invoice_line_unit_price_untaxed/tests/test_account_move_line_untaxed.py
```diff
@@ -140,3 +140,41 @@ def test_invoice_with_product_company2(self):
...
+
+    def test_invoice_with_decimal_precision(self):
+        """
+        Set tax with price include False
+        Set the product price to 118.573
+        Set product price precision to 3
+        Check prices are equivalent (price_unit == price_unit_untaxed) with
+        same
+        """
+        self.tax2.price_include = False
```
acsonefho, line 155: IMO you should also ensure that no currency are created. Even if you do a `new(...)` it's better to ensure the behaviour won't change.

acsonefho (changes requested):
Minor changes! But GJ!

account_invoice_line_unit_price_untaxed/tests/test_account_move_line_untaxed.py
rousseldenis: @acsonefho I don't remember why you said this :sweat_smile:

rousseldenis:
@legalsylvain @acsonefho Could you update your review ?

@OCA/accounting-maintainers

acsonefho (approved):

bosd (approved):
Quick code review, LGTM :+1:

PR #1095, opened 2022-02-09
[13.0][FIX][BUG] Get the current company should be made with self.env.company 

mbcosta:
As reported in OCA maillist after v13 to get the current company should be made by self.env.company and not self.env.user.company_id, reference https://www.odoo.com/documentation/13.0/developer/howtos/company.html

cc @gurneyalex @rvalyi


AaronHForgeFlow (approved):
Code review only. LG!

rvalyi (approved):

PR #1102, opened 2022-02-16
[14.0][FIX][BUG] Get the current company should be made with self.env.company

mbcosta:
As reported in OCA maillist after v13 to get the current company should be made by self.env.company and not self.env.user.company_id, reference https://www.odoo.com/documentation/13.0/developer/howtos/company.html, the same was did in v13 https://github.com/OCA/account-invoicing/pull/1095

cc @rvalyi @AaronHForgeFlow


legalsylvain (approved):
thanks.

AaronHForgeFlow (approved):
LG!

PR #776, opened 2020-08-25
[13.0][MIG] account_invoice_change_currency

Saran440:
- In version 13 message tracking has changed. you have to save new invoice before change currency.
- action force rate change `reconciled` to `to_check` because account.move don't have a reconciled field
- add context `check_move_validity` in function action_account_change_currency

PR #1104, opened 2022-02-24
[MIG] account_invoice_change_currency: Migration to 14.0

rolandojduartem:
[MIG] account_invoice_change_currency: Migration to 14.0

Migration from https://github.com/OCA/account-invoicing/pull/776

**Workthrough:**

**Changes**

Now, the account_invoice_change_currency avoid the mail tracking using account.move, res.currency
and new fields to handle the different calculations.

**New fields**

- original_currency_id: Store the base currency when the invoice is created or the conversion is called for the first time. This is used to calculate conversion from this currency.
- is_original_currency: Check if current currency is the base currency. This is used to hide custom rate field in the form view.
- original_price_unit: Store price unit from every line when the invoice is created or the conversion is called for the first time to use it to convert the amount in the new currency.


rolandojduartem:
I am still working on it....

rolandojduartem:
Hi, @luisg123v, could you review this pull request, please?

CC @hugho-ad

rolandojduartem:
Also, @CarmenMiranda, could you review this pull request, please?

CC @hugho-ad

account_invoice_change_currency/models/account_change_currency.py
```diff
@@ -130,14 +136,19 @@ def get_last_currency_id(self, skip_update_currency=False):
...
 FROM mail_tracking_value as mtv INNER JOIN mail_message AS mm
-ON mtv.mail_message_id = mm.id """
+ON mtv.mail_message_id = mm.id INNER JOIN
+ir_model_fields AS field
+ON field.id = mtv.field """
         if skip_update_currency:
             query += " AND mm.subtype_id != %s "
-            params = (subtype_id.id, tuple(self.message_ids.ids))
+            params = (subtype_id.id, self._name, tuple(self.message_ids.ids))
         else:
```
CarmenMiranda: Hi @rolandojduartem This `else` can be gone if the `params` are set before the `if skip_update_currency`

account_invoice_change_currency/tests/test_account_invoice_change_currency.py
```diff
@@ -146,17 +155,17 @@ def test_change_invoice_currency(self):
...
         self.assertEqual(
-            float_compare(inv.amount_total, expected_value, 1),
+            float_compare(inv.amount_total, expected_value, 0),
             0,
             "Total amount of invoice does not equal to expected value!!!",
         )
 
     def test_change_validated_invoice_currency(self):
         inv = self.create_simple_invoice(fields.Date.today())
+        inv.track_created_invoice()
```
CarmenMiranda: If you are calling this method in the `create` why do you need to call it after the creation?

rolandojduartem: I removed it. Done!

account_invoice_change_currency/models/account_change_currency.py
rolandojduartem: Done!

CarmenMiranda (approved):
Hi @rolandojduartem I did the review of your commits and technically LGTM 👍🏼

account_invoice_change_currency/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2017 Komit <http://komit-consulting.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Account Invoice - Change Currency",
+    "version": "14.0.1.0.0",
```
luisg123v, line 5: This is being done in the wrong commit, it's in the black/prettier one.

Actually, I think that commit should be removed, as black was already run when migrated to 13.0.

rolandojduartem:
Thanks, @CarmenMiranda. @luisg123v, could you review this PR?

account_invoice_change_currency/models/account_change_currency.py
```diff
@@ -0,0 +1,311 @@
...
+            )
+            .ids
+        )
+        last_values = (
+            self.env["mail.tracking.value"]
+            .sudo()
+            .search(
+                [
+                    ("mail_message_id", "in", self.message_ids.ids),
+                    ("field", "in", ir_model_fields_ids),
```
luisg123v:
```suggestion
                    ("field.name", "in", ["rate", "currency_id"]),
```

account_invoice_change_currency/models/account_change_currency.py
```diff
@@ -0,0 +1,311 @@
...
+                ],
+                limit=2,
+                order="write_date desc, id desc",
+            )
+        )
+        if len(last_values) == 2:
```
luisg123v: Why?

account_invoice_change_currency/models/account_change_currency.py
```diff
@@ -0,0 +1,311 @@
...
+            currency_value, rate_value = sorted(
```
luisg123v: This seems too complicated

account_invoice_change_currency/models/account_change_currency.py
```diff
@@ -0,0 +1,311 @@
...
+            force = not state.new_value_integer
+            self._track_force_rate(force)
+
+    def get_force_rate_state(self):
+        self.ensure_one()
+        subtype_id = self.env.ref("account_invoice_change_currency.mt_force_rate")
+        ir_model_fields = self.env["ir.model.fields"]
+        ir_model_fields_id = (
+            ir_model_fields.sudo()
+            .search([("name", "=", "to_check"), ("model", "=", self._name)])
```
luisg123v: `ir.model.fields` has a method `get(model_name, field_name)` that does this for you

account_invoice_change_currency/models/account_change_currency.py
```diff
@@ -0,0 +1,311 @@
...
+            .id
+        )
+        domain = [
+            ("mail_message_id", "in", self.message_ids.ids),
+            ("mail_message_id.subtype_id", "=", subtype_id.id),
+            ("field", "=", ir_model_fields_id),
```
luisg123v: Same as above

account_invoice_change_currency/models/account_change_currency.py
```diff
@@ -0,0 +1,311 @@
...
+            ],
+        ]
+        self.message_post(
+            subtype_id=self.env.ref("account_invoice_change_currency.mt_force_rate").id,
+            tracking_value_ids=tracking_value_ids,
+        )
+
+    def track_created_invoice(self):
+        track = self.env["mail.tracking.value"]
+        today = fields.Date.today()
```
luisg123v:
```suggestion
        today = fields.Date.context_today(self)
```

account_invoice_change_currency/models/account_change_currency.py
```diff
@@ -0,0 +1,311 @@
...
+        for invoice in self.filtered(lambda x: x.state == "draft").with_context(
+            {"check_move_validity": False}
```
luisg123v: You should avoid passing context as dict, as it overrides current context:

```suggestion
            check_move_validity=False
```

account_invoice_change_currency/models/account_change_currency.py
```diff
@@ -0,0 +1,311 @@
...
+                    "res.currency",
+                ),
+            ],
+        ]
+        self.message_post(
+            subtype_id=self.env.ref("account_invoice_change_currency.mt_force_rate").id,
+            tracking_value_ids=tracking_value_ids,
+        )
+
+    def track_created_invoice(self):
```
luisg123v: Why is this required?

account_invoice_change_currency/tests/test_account_invoice_change_currency.py
```diff
@@ -0,0 +1,334 @@
...
+                    "user_type_id",
+                    "=",
+                    self.env.ref("account.data_account_type_current_assets").id,
+                )
+            ],
+            limit=1,
+        )
+
+    def create_journal_type(self, type_string=None):
+        if not type_string:
```
luisg123v: Since the default value is a string, it may be specified as method default param

account_invoice_change_currency/tests/test_account_invoice_change_currency.py
```diff
@@ -0,0 +1,334 @@
...
+                (
+                    "user_type_id",
+                    "=",
+                    self.env.ref("account.data_account_type_current_assets").id,
+                )
+            ],
+            limit=1,
+        )
+
+    def create_journal_type(self, type_string=None):
```
luisg123v: Journal type?

account_invoice_change_currency/tests/test_account_invoice_change_currency.py
```diff
@@ -0,0 +1,334 @@
...
+                "padding": 6,
+                "company_id": self.env.ref("base.main_company").id,
+            }
+        )
+        account_journal_type = self.env["account.journal"].create(
+            {
+                "name": "Journal",
+                "code": type_string,
+                "type": type_string,
+                "sequence": sequence.id,
```
luisg123v: Sequence is an integer, not an ID of an `ir.sequence` record

account_invoice_change_currency/tests/test_account_invoice_change_currency.py
```diff
@@ -0,0 +1,334 @@
...
+        )
+        return account_journal_type
+
+    def create_simple_invoice(
+        self, date=False, context=None, inv_type=None, journal=None
+    ):
+        if not context:
+            context = {}
+        if not journal:
+            journal = self.create_journal_type(type_string="purchase")
```
luisg123v: No need to create a new journal every time an invoice is created, you could create a new one in the `setUp` method

account_invoice_change_currency/tests/test_account_invoice_change_currency.py
```diff
@@ -0,0 +1,334 @@
...
+        ]
+        invoice = (
+            self.env["account.move"]
+            .with_user(self.manager)
+            .with_context(**context)
+            .create(
+                {
+                    "partner_id": 1,
+                    "move_type": inv_type or "in_invoice",
+                    "journal_id": journal.id,
```
luisg123v: Why this change is necessary?

FYI if you pass `default_move_type` as context, default journal will be computed automatically.

account_invoice_change_currency/tests/test_account_invoice_change_currency.py
```diff
@@ -0,0 +1,334 @@
...
+        eur = self.env.ref("base.EUR")
+        inv.write({"currency_id": usd.id, "custom_rate": custom_rate})
+        inv._onchange_currency_change_rate()
+        inv.write({"custom_rate": custom_rate})
+        inv.action_account_change_currency()
+        expected_value = before_curr._convert(
+            before_amount, usd, inv.company_id, fields.Date.today()
+        )
+        self.assertEqual(
+            float_compare(inv.amount_total, expected_value, 0),
```
luisg123v, line 176: Why?

account_invoice_change_currency/data/ir_server_actions.xml
```diff
@@ -0,0 +1,18 @@
+<odoo noupdate="1">
```
luisg123v: No need to remove XML header.

Same for other XML files

account_invoice_change_currency/__manifest__.py
```diff
@@ -0,0 +1,18 @@
...
+    "category": "Accounting & Finance",
+    "summary": "Allows to change currency of Invoice by wizard",
+    "author": "Komit Consulting, Odoo Community Association (OCA)",
```
luisg123v: Please add Vauxoo as author (actually that's the original one) and myself as maintainer:

```suggestion
    "author": "Vauxoo, Komit Consulting, Odoo Community Association (OCA)",
    "maintainers": ["luisg123v"],
```

luisg123v (changes requested):

account_invoice_change_currency/__manifest__.py
rolandojduartem: Done!

account_invoice_change_currency/models/account_change_currency.py
rolandojduartem: Done!

account_invoice_change_currency/models/account_change_currency.py
rolandojduartem: Before, on Odoo 12 and 13, mail tracking just showed the new value, but on Odoo 14, mail tracking shows old value and new value and if I keep that line, mail tracking would show 0 to new value in every change of currency, but with this change, mail tracking would show old_value to new_value.

account_invoice_change_currency/models/account_change_currency.py
rolandojduartem: I made a change with this, you could review it

account_invoice_change_currency/models/account_change_currency.py
rolandojduartem: The change that I applied makes this code unnecessary

account_invoice_change_currency/tests/test_account_invoice_change_currency.py
rolandojduartem: Changes that I applied makes this code unnecessary

account_invoice_change_currency/tests/test_account_invoice_change_currency.py
rolandojduartem: Changes are done!

account_invoice_change_currency/tests/test_account_invoice_change_currency.py
rolandojduartem: This code makes a comparison between 2004.64 vs 2004.67

account_invoice_change_currency/data/ir_server_actions.xml
rolandojduartem: Done!

account_invoice_change_currency/__manifest__.py
rolandojduartem: Done!

account_invoice_change_currency/models/account_change_currency.py
rolandojduartem: Done!

account_invoice_change_currency/models/account_change_currency.py
rolandojduartem: The bug I commented once was that the first currency change did not detect the last value when the invoice was created. In Odoo 12 this worked fine, but in Odoo 13 and Odoo 14 it didn't detect the currency when the invoice was created. This function adds this functionality back.

rolandojduartem:
Hi, @luisg123v, I applied some changes and I responded some observation, could you review this pull request, please?

account_invoice_change_currency/models/account_change_currency.py
```diff
@@ -0,0 +1,174 @@
+# Copyright 2017-2018 Vauxoo
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+from odoo import api, fields, models
+from odoo.tools import float_compare
+
+
+class AccountMoveLine(models.Model):
```
luisg123v: Move below `AccountMove`, as this one is not the primary class

account_invoice_change_currency/models/account_change_currency.py
```diff
@@ -0,0 +1,174 @@
...
+    _inherit = "account.move.line"
+    original_price_unit = fields.Monetary(
+        default=lambda self: self.price_unit,
```
luisg123v: Defaults are run over an empty recordset, because they're run when the record doesn't exist already. So trying to access any field will retrieve always falsy values, in this case 0.0

account_invoice_change_currency/models/account_change_currency.py
```diff
@@ -0,0 +1,174 @@
...
+        help="""
+        Save price unit from every line in the invoice when the currency back to
+        original state""",
+    )
+
+    def create(self, values):
+        lines = super().create(values)
+        for line in lines:
+            line.write({"original_price_unit": line.price_unit})
```
luisg123v: Don't trigger a write just after a create, you only need to update values:

```suggestion
        values.setdefault("original_price_unit", values.get("price_unit"))
```

account_invoice_change_currency/models/account_change_currency.py
```diff
@@ -0,0 +1,174 @@
...
+        default=1,
+        help="Set new currency rate to apply on the invoice\n."
+        "This rate will be taken in order to convert amounts between the "
+        "currency on the invoice and last currency",
+    )
+    created_invoice_currency = fields.Many2one(
+        "res.currency",
+        help="""Save the original currency from the company""",
+    )
+    created_invoice_rate = fields.Float(help="""Save previous custom rate""")
```
luisg123v: Don't use triple quotes, they will preserve spaces and newlines, and we don't want those in translations. Just do:

```
help="this is a very long line "
"that continues in the next one.",
```

The same for all other help messages

account_invoice_change_currency/models/account_change_currency.py
```diff
@@ -0,0 +1,174 @@
...
+    _inherit = "account.move"
+
+    custom_rate = fields.Float(
+        digits=(12, 6),
+        default=1,
+        help="Set new currency rate to apply on the invoice\n."
+        "This rate will be taken in order to convert amounts between the "
+        "currency on the invoice and last currency",
+    )
+    created_invoice_currency = fields.Many2one(
```
luisg123v: Missing `_id` suffix.

Check all fields, many2one fields should have the suffix.

account_invoice_change_currency/models/account_change_currency.py
```diff
@@ -0,0 +1,174 @@
...
+        "res.currency",
+        help="""Save the original currency from the company""",
```
luisg123v: Company has nothing to do with the currency, it could be any currency

account_invoice_change_currency/models/account_change_currency.py
```diff
@@ -0,0 +1,174 @@
...
+    )
+    created_invoice_rate = fields.Float(help="""Save previous custom rate""")
+    last_currency = fields.Many2one(
+        "res.currency",
+        help="""Save the original currency from the company""",
```
luisg123v: Help doesn't match

account_invoice_change_currency/models/account_change_currency.py
```diff
@@ -0,0 +1,174 @@
...
+    )
+    old_rate = fields.Float(help="""Save previous custom rate""")
+    is_forced_rate = fields.Boolean(default=False, help="""Save force rate state""")
```
luisg123v: `False` is already the default value

account_invoice_change_currency/models/account_change_currency.py
```diff
@@ -0,0 +1,174 @@
...
+    is_first_currency_equal_to_currency_id = fields.Boolean(
```
luisg123v: This is the opposite, not m2o but has suffix

account_invoice_change_currency/models/account_change_currency.py
```diff
@@ -0,0 +1,174 @@
...
+            if from_currency != currency:
+                rate = new_rate
+            invoice.last_currency = currency
+            invoice.old_rate = new_rate
+            for line in invoice.invoice_line_ids:
+                line.price_unit *= rate
+            invoice._recompute_dynamic_lines(recompute_all_taxes=True)
+
+    @api.onchange("currency_id", "invoice_date")
+    def _onchange_currency_change_rate(self):
```
luisg123v: I think no onchange is required

account_invoice_change_currency/models/account_change_currency.py
```diff
@@ -0,0 +1,174 @@
...
+                    self.currency_id,
+                    self.company_id,
+                    self.invoice_date or today,
+                ),
+            }
+        )
+
+    @api.depends("created_invoice_currency", "currency_id")
+    def _compute_is_first_currency_equal_to_currency_id(self):
+        self.is_first_currency_equal_to_currency_id = (
```
luisg123v: `for move in self:`

account_invoice_change_currency/models/account_change_currency.py
```diff
@@ -0,0 +1,174 @@
...
+        return self.is_forced_rate
+
+    def _track_force_rate(self, force=True):
+        self.write(
+            {
+                "is_forced_rate": force,
+                "currency_id": self.currency_id,
+            }
+        )
+        self.currency_id.write({"rate": self.custom_rate})
```
luisg123v: Why?

account_invoice_change_currency/views/account_move_views.xml
```diff
@@ -0,0 +1,65 @@
...
+                        class="oe_inline"
+                    />
+                </div>
+            </xpath>
+            <xpath expr="//div[@name='journal_div']" position="after">
+                <label for="custom_rate" groups="base.group_multi_currency" />
+                <div groups="base.group_multi_currency">
+                    <field
+                        name="is_first_currency_equal_to_currency_id"
+                        attrs="{'invisible': True, 'readonly': True}"
```
luisg123v: Just invisible is enough, no need attrs nor readonly.

BTW, I think this field is not required

account_invoice_change_currency/data/ir_server_actions.xml
```diff
@@ -0,0 +1,19 @@
...
+    <record id="toggle_force_custom_rate_action" model="ir.actions.server">
+        <field name="name">Toggle custom force rate</field>
+        <field name="model_id" ref="account.model_account_move" />
+        <field name="state">code</field>
+        <field name="code">
+for inv in records:
+    inv._toggle_forced_rate()</field>
+    </record>
+    <function name="create_action" model="ir.actions.server">
```
luisg123v: remove, you only need to provide `binding_model_id` field inorder for the action to be bind

account_invoice_change_currency/models/__init__.py
```diff
@@ -0,0 +1 @@
+from . import account_change_currency
```
luisg123v:
```suggestion
from . import account_move
```

account_invoice_change_currency/models/account_change_currency.py
```diff
@@ -0,0 +1,174 @@
...
+        return lines
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    custom_rate = fields.Float(
+        digits=(12, 6),
+        default=1,
+        help="Set new currency rate to apply on the invoice\n."
```
luisg123v:
```suggestion
        help="Set new currency rate to apply on the invoice.\n"
```

account_invoice_change_currency/models/account_change_currency.py
```diff
@@ -0,0 +1,174 @@
...
+        is equal to currency when the invoice was created""",
+    )
+
+    @api.model
+    def create(self, values):
+        invoice = super(AccountMove, self).create(values)
+        if self.env.context.get("force_rate"):
+            invoice._toggle_forced_rate()
+        invoice._onchange_currency_change_rate()
+        invoice.write(
```
luisg123v: Same as invoice line, don't write on creation

account_invoice_change_currency/models/account_change_currency.py
```diff
@@ -0,0 +1,174 @@
...
+        help="""Save a boolean value to recognize if current currency
+        is equal to currency when the invoice was created""",
+    )
+
+    @api.model
+    def create(self, values):
+        invoice = super(AccountMove, self).create(values)
+        if self.env.context.get("force_rate"):
+            invoice._toggle_forced_rate()
+        invoice._onchange_currency_change_rate()
```
luisg123v: Don't run onchanges manually

account_invoice_change_currency/models/account_change_currency.py
```diff
@@ -0,0 +1,174 @@
...
+        )
+        return invoice
+
+    def action_account_change_currency(self):
+        precision = self.env["decimal.precision"].precision_get("Payment Terms")
+        today = fields.Date.context_today(self)
+        for invoice in self.filtered(lambda x: x.state == "draft").with_context(
+            check_move_validity=False,
+        ):
+            from_currency = invoice.last_currency
```
luisg123v: This could be added as a filtered condition

account_invoice_change_currency/models/account_change_currency.py
```diff
@@ -0,0 +1,174 @@
...
+                new_rate = rate_skip
+            if (
+                from_currency == currency
+                and new_rate
+                and old_rate
+                and float_compare(new_rate, old_rate, precision_digits=precision) == 0
+            ):
+                if not invoice.custom_rate or old_rate == new_rate:
+                    continue
+                if float_compare(new_rate, old_rate, precision_digits=precision) == 0:
```
luisg123v: Can be joined with the above conditional

account_invoice_change_currency/models/account_change_currency.py
```diff
@@ -0,0 +1,174 @@
+# Copyright 2017-2018 Vauxoo
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
```
luisg123v: Manifest says AGPL

luisg123v (changes requested):
Squash te migration and refactor commits

rafaelbn:
Hello @rolandojduartem could you rebase please and check what travis says? Thanks

account_invoice_change_currency/models/account_change_currency.py
rolandojduartem: Done!

account_invoice_change_currency/models/account_change_currency.py
rolandojduartem: Done!

account_invoice_change_currency/models/account_change_currency.py
rolandojduartem: In this case I could not apply this change because of inheritance issues.

account_invoice_change_currency/models/account_change_currency.py
rolandojduartem: Done!

account_invoice_change_currency/models/account_change_currency.py
rolandojduartem: Done!

account_invoice_change_currency/models/account_change_currency.py
rolandojduartem: Done!

account_invoice_change_currency/models/account_change_currency.py
rolandojduartem: Done!

account_invoice_change_currency/models/account_change_currency.py
rolandojduartem: Done!

account_invoice_change_currency/models/account_change_currency.py
rolandojduartem: Done!

account_invoice_change_currency/models/account_change_currency.py
rolandojduartem: Done!

rolandojduartem:
Hi, @luisg123v, May you review it, please? I realized that travis CI failed because of account_invoice_pricelist.

cc @hugho-ad

account_invoice_change_currency/models/account_move.py
```diff
@@ -0,0 +1,93 @@
...
+        digits=(12, 6),
+        default=1,
+        help="Set new currency rate to apply on the invoice.\n"
+        "This rate will be taken in order to convert amounts between the "
+        "currency on the invoice and last currency",
+    )
+    last_rate = fields.Float(digits=(12, 6), default=1, help="Stores the last rate.")
+    created_invoice_currency_id = fields.Many2one(
+        "res.currency",
+        help="Store the original currency from the company",
```
luisg123v: Not the company, but the one used to create the invoice. If the company's currency is USD but the invoice was created in EUR, this field will be set to EUR

account_invoice_change_currency/models/account_move.py
```diff
@@ -0,0 +1,93 @@
...
+        compute="_compute_is_original_currency",
+        help="Check if current currency is the original currency",
+    )
+
+    @api.model
+    def create(self, values):
+        values.setdefault("created_invoice_currency_id", values.get("currency_id"))
+        values.setdefault("last_currency_id", values.get("currency_id"))
+        values.setdefault("last_rate", values.get("custom_rate"))
+        return super(AccountMove, self).create(values)
```
luisg123v: Always use new-style super

```suggestion
        return super().create(values)
```

account_invoice_change_currency/models/account_move.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def action_account_change_currency(self):
+        today = fields.Date.context_today(self)
+        for invoice in self.filtered(
+            lambda x: x.state == "draft" and x.last_currency_id
+        ).with_context(
```
luisg123v: Better to assign this to a variable

Or assign filtered to a variable and use with_context in for

account_invoice_change_currency/models/account_move.py
```diff
@@ -0,0 +1,93 @@
...
+                        line.price_unit, to_currency, invoice.company_id, invoice_date
+                    )
+            invoice._recompute_dynamic_lines(recompute_all_taxes=True)
+            invoice.update(
+                {"last_currency_id": to_currency, "last_rate": invoice.custom_rate}
+            )
+
+    @api.onchange("currency_id", "invoice_date")
+    def _onchange_currency_change_rate(self):
+        date = self.invoice_date or fields.Date.today()
```
luisg123v: `context_today`

account_invoice_change_currency/models/account_move.py
```diff
@@ -0,0 +1,93 @@
...
+        last_currency = self.last_currency_id
+        if (
+            not (self.currency_id and last_currency)
```
luisg123v: three or's would be clearer IMHO

account_invoice_change_currency/models/account_move.py
```diff
@@ -0,0 +1,93 @@
...
+        "res.currency",
+        help="Store the original currency from the company",
+    )
+    last_currency_id = fields.Many2one(
+        "res.currency",
+        help="Store the last currency applied in the invoice",
+    )
+    is_original_currency = fields.Boolean(
+        default=False,
+        store=True,
```
luisg123v: You don't need it stored

account_invoice_change_currency/models/res_currency.py
```diff
@@ -0,0 +1,27 @@
...
+    def _convert(self, from_amount, to_currency, company, date, round=True):
+        custom_rate = self.env.context.get("custom_rate")
+        last_rate = self.env.context.get("last_rate")
+        if (
+            self == to_currency
+            and custom_rate
+            and last_rate
+            and custom_rate != last_rate
+        ):
+            from_amount = from_amount * custom_rate / last_rate
```
luisg123v: Missing rounding

account_invoice_change_currency/models/res_currency.py
```diff
@@ -0,0 +1,27 @@
...
+            currency_rates[to_currency.id] = custom_rate
+        return currency_rates
+
+    def _convert(self, from_amount, to_currency, company, date, round=True):
+        custom_rate = self.env.context.get("custom_rate")
+        last_rate = self.env.context.get("last_rate")
+        if (
+            self == to_currency
+            and custom_rate
+            and last_rate
```
luisg123v: This shouldn't be used here, this only make sense under invoice context.

account_invoice_change_currency/models/res_currency.py
```diff
@@ -0,0 +1,27 @@
...
+
+class ResCurrency(models.Model):
+    _inherit = "res.currency"
+
+    def _get_rates(self, company, date):
+        currency_rates = super()._get_rates(company, date)
+        custom_rate = self.env.context.get("custom_rate")
+        to_currency = self.env.context.get("to_currency")
+        if custom_rate and to_currency:
+            currency_rates[to_currency.id] = custom_rate
```
luisg123v: Not enough I think, because you need to provide rate for the other currency explicitly, so you ensure it's 1.0. Otherwise, other one could be retrieved by Odoo

account_invoice_change_currency/models/account_move_line.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2017-2018 Vauxoo
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo import fields, models
+
+
+class AccountMoveLine(models.Model):
+    _inherit = "account.move.line"
+    original_price_unit = fields.Monetary(
```
luisg123v, line 9: Leave a blank line between `_inherit` and fields

account_invoice_change_currency/models/account_move_line.py
```diff
@@ -0,0 +1,17 @@
...
+        help="Save price unit from every line in the invoice "
+        "when the currency back to original state",
+    )
+
+    def create(self, values):
+        lines = super().create(values)
+        for line in lines:
+            line.original_price_unit = line.price_unit
```
luisg123v: Remember to don't trigger a write within a create

account_invoice_change_currency/views/account_move_views.xml
```diff
@@ -0,0 +1,63 @@
...
+                            ],
+                            'readonly': [
+                                '|',
+                                ('state', '!=', 'draft'),
+                                ('is_original_currency', '=', True),
+                            ]
+                        }"
+                    />
+                    <button
+                        aria-label="Opened"
```
luisg123v: Update currency

account_invoice_change_currency/views/account_move_views.xml
```diff
@@ -0,0 +1,63 @@
...
+<!-- Copyright 2017 Komit <http://komit-consulting.com>
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+<odoo>
+    <record id="view_move_form" model="ir.ui.view">
+        <field name="name">account.move.form.inherit</field>
+        <field name="model">account.move</field>
+        <field name="inherit_id" ref="account.view_move_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='currency_id']" position="attributes">
+                <attribute name="invisible">1</attribute>
```
luisg123v: You're hiding the field and then adding it again. You could use an xpath position="move" instead

account_invoice_change_currency/models/account_move.py
```diff
@@ -0,0 +1,93 @@
...
+        help="Set new currency rate to apply on the invoice.\n"
+        "This rate will be taken in order to convert amounts between the "
+        "currency on the invoice and last currency",
+    )
+    last_rate = fields.Float(digits=(12, 6), default=1, help="Stores the last rate.")
+    created_invoice_currency_id = fields.Many2one(
+        "res.currency",
+        help="Store the original currency from the company",
+    )
+    last_currency_id = fields.Many2one(
```
luisg123v: Are you sure this field is required?

luisg123v (changes requested):
- Edit commit's description to remove:  
  `[REF] account_invoice_change_currency: Refactor the code to avoid mail tracking use`
- PR's description seems outdated, edit it to match commit's description

account_invoice_change_currency/models/account_move.py
rolandojduartem: Done!

account_invoice_change_currency/models/account_move.py
rolandojduartem: Done!

account_invoice_change_currency/models/account_move.py
rolandojduartem: Done!

account_invoice_change_currency/models/account_move.py
rolandojduartem: Done!

account_invoice_change_currency/models/account_move_line.py
rolandojduartem: Done!

account_invoice_change_currency/models/account_move.py
rolandojduartem: Done!

account_invoice_change_currency/models/account_change_currency.py
rolandojduartem: Done!

rolandojduartem:
Hi @luisg123v, I followed your recommendations, may you review this pull request, please?

cc @hugho-ad

account_invoice_change_currency/models/res_currency.py
```diff
@@ -0,0 +1,23 @@
...
+    def _get_rates(self, company, date):
+        """
+        Inheritance to use the provided custom rate by user
+        instead of the rate from odoo.
+        """
+        currency_rates = {}
+        custom_rate = self.env.context.get("custom_rate")
+        to_currency = self.env.context.get("to_currency")
+        if custom_rate and to_currency:
+            from_currency = self.filtered(lambda x: x != to_currency)
```
luisg123v: You can use dict comprehension here:

```
return {
    currency.id:
    custom_rate if currency == to_currency else 1.0
    for currency in self
}
```

In addition, return so you don't need an else

account_invoice_change_currency/models/account_move.py
```diff
@@ -0,0 +1,93 @@
...
+        was created to the current currency. The custom rate field is
+        editable, but it will not change if custom rate is zero or the
+        current currency and the original currency are the same
+        """
+        for invoice in self:
+            date = invoice.invoice_date or fields.Date.context_today(invoice)
+            from_currency = (
+                invoice.created_invoice_currency_id or invoice.company_currency_id
+            )
+            if not invoice.currency_id:
```
luisg123v: Compute methods should always set a value. If any value is missing (including company), set it to 1.0. The if should be the first thing inside the loop

account_invoice_change_currency/models/account_move_line.py
```diff
@@ -0,0 +1,19 @@
...
+
+    original_price_unit = fields.Monetary(
+        help="Store price unit from every line when the "
+        "invoice was created to use it when a conversion "
+        "between currencies is necessary",
+    )
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        for val in vals_list:
```
luisg123v:
```suggestion
        for vals in vals_list:
```

account_invoice_change_currency/i18n/es.po
```diff
@@ -0,0 +1,73 @@
+# Translation of Odoo Server.
+# This file contains the translation of the following modules:
+# 	* account_invoice_change_currency
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: Odoo Server 14.0\n"
```
luisg123v: Don't worry about this, all these files will be re-generated by Weblate automatically from the template. Only change template

account_invoice_change_currency/README.rst
```diff
@@ -0,0 +1,104 @@
...
+   !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+
+.. |badge1| image:: https://img.shields.io/badge/maturity-Beta-yellow.png
+    :target: https://odoo-community.org/page/development-status
+    :alt: Beta
+.. |badge2| image:: https://img.shields.io/badge/licence-AGPL--3-blue.png
+    :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+    :alt: License: AGPL-3
+.. |badge3| image:: https://img.shields.io/badge/github-OCA%2Faccount--invoicing-lightgray.png?logo=github
+    :target: https://github.com/OCA/account-invoicing/tree/14.0/account_invoice_change_currency
```
luisg123v: README is autogenerated, no need to update it

luisg123v (changes requested):
I think these are my final comments :smile: 

In commit description, put fields as a list, i.e. just add "- " before field name

```
- field1: description
- field2: description
```

account_invoice_change_currency/models/res_currency.py
rolandojduartem: Done!

account_invoice_change_currency/models/account_move.py
rolandojduartem: Done!

account_invoice_change_currency/models/account_move_line.py
rolandojduartem: Done!

account_invoice_change_currency/i18n/es.po
rolandojduartem: Done!

account_invoice_change_currency/README.rst
rolandojduartem: Done!

rolandojduartem:
Hi, @luisg123v, may you review this PR?

cc @hugho-ad

account_invoice_change_currency/models/account_move.py
```diff
@@ -0,0 +1,93 @@
...
+    def _compute_currency_change_rate(self):
+        """
+        Compute the custom rate from the original currency when the invoice
+        was created to the current currency. The custom rate field is
+        editable, but it will not change if custom rate is zero or the
+        current currency and the original currency are the same
+        """
+        for invoice in self:
+            if not invoice.currency_id or not invoice.company_id:
+                invoice.custom_rate = 1.0
```
luisg123v, line 73: Missing continue after this line

luisg123v (changes requested):

rolandojduartem:
@luisg123v may you review this PR?

luisg123v (approved):
@hugho-ad @moylop260,

What do you think?

moylop260:
LGTM only check the CI red, please

rolandojduartem:
Hi @luisg123v I made some changes to handle the invoices that were created before installing this module. Please, may you review them?

PS: CI is red but no for these changes.

account_invoice_change_currency/models/account_move.py
```diff
@@ -0,0 +1,93 @@
...
+        currency when the invoice was created to the current currency
+        using the custom rate and recompute all taxes.
+        """
+        today = fields.Date.context_today(self)
+        invoices = self.filtered(lambda x: x.state == "draft").with_context(
+            check_move_validity=False,
+        )
+        for invoice in invoices:
+            if not invoice.base_currency_id:
+                invoice.update({"base_currency_id": invoice.currency_id})
```
luisg123v: Use update only within onchanges, otherwise write

account_invoice_change_currency/models/account_move.py
```diff
@@ -0,0 +1,93 @@
...
+                invoice.line_ids._set_base_price_unit()
+                continue
```
luisg123v: Why?

account_invoice_change_currency/models/account_move.py
```diff
@@ -0,0 +1,93 @@
...
+        was created to the current currency. The custom rate field is
+        editable, but it will not change if custom rate is zero or the
+        current currency and the original currency are the same
+        """
+        for invoice in self:
+            if not invoice.currency_id or not invoice.company_id:
+                invoice.custom_rate = 1.0
+                continue
+            date = invoice.invoice_date or fields.Date.context_today(invoice)
+            from_currency = invoice.base_currency_id or invoice.company_currency_id
```
luisg123v: Shouldn't it be this way?

```suggestion
            from_currency = invoice.base_currency_id or invoice.currency_id
```

account_invoice_change_currency/models/account_move_line.py
```diff
@@ -0,0 +1,24 @@
+# Copyright 2017-2018 Vauxoo
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo import api, fields, models
+
+
+class AccountMoveLine(models.Model):
+    _inherit = "account.move.line"
+
+    base_price_unit = fields.Monetary(
```
luisg123v: I think base is not clear enough, as it could be confused with the base price (price without taxes)

account_invoice_change_currency/models/account_move_line.py
```diff
@@ -0,0 +1,24 @@
...
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        for vals in vals_list:
+            vals.setdefault("base_price_unit", vals.get("price_unit"))
+        return super().create(vals_list)
+
+    def _set_base_price_unit(self):
+        for line in self:
+            line.update({"base_price_unit": line.price_unit})
```
luisg123v: Same here, write

account_invoice_change_currency/models/account_move.py
```diff
@@ -0,0 +1,93 @@
...
+        store=True,
+        readonly=False,
+        compute="_compute_currency_change_rate",
+        help="Set new currency rate to apply on the invoice.\n"
+        "This rate will be taken in order to convert amounts between the "
+        "currency on the invoice and last currency",
+    )
+    base_currency_id = fields.Many2one(
+        "res.currency",
+        help="Store the base currency when the invoice is created or the"
```
luisg123v: Since field names and translations are being modified, you should update POT file to reflect those changes

luisg123v (changes requested):

account_invoice_change_currency/models/account_move.py
rolandojduartem: Done!

account_invoice_change_currency/models/account_move.py
rolandojduartem: Done!

account_invoice_change_currency/models/account_move.py
rolandojduartem: company_currency_id was a behavior from v12.0/v13.0, I changed it to currency_id. What do you think about it?

account_invoice_change_currency/models/account_move_line.py
rolandojduartem: Done!

account_invoice_change_currency/models/account_move_line.py
rolandojduartem: Done!

account_invoice_change_currency/models/account_move.py
rolandojduartem: Done!

rolandojduartem:
Hi, @luisg123v I applied the recommended changes and I added tests to increase the coverage.

cc @hugho-ad @moylop260

rolandojduartem:
Hi @luisg123v friendly reminder of this PR.

account_invoice_change_currency/tests/test_account_invoice_change_currency.py
```diff
@@ -0,0 +1,320 @@
...
+            )
+        )
+        return invoice
+
+    def test_set_original_price_unit(self):
+        inv = self.create_simple_invoice(fields.Date.today())
+        self.assertEqual(
+            inv.line_ids.mapped("original_price_unit"),
+            inv.line_ids.mapped("price_unit"),
+            "Original price unit of invoice does" "not equal to expected value!!!",
```
luisg123v:
```suggestion
            "Original price unit of invoice does not match native unit price",
```

account_invoice_change_currency/tests/test_account_invoice_change_currency.py
```diff
@@ -0,0 +1,320 @@
...
+                    "invoice_date": date,
+                    "currency_id": self.env.ref("base.EUR").id,
+                    "invoice_line_ids": invoice_lines,
+                    "state": "draft",
+                }
+            )
+        )
+        return invoice
+
+    def test_set_original_price_unit(self):
```
luisg123v: You don't need a new test for this, an invoice is being created in several tests already

account_invoice_change_currency/tests/test_account_invoice_change_currency.py
```diff
@@ -0,0 +1,320 @@
...
+            float_compare(inv.amount_total, expected_value, 1),
+            0,
+            "Total amount of invoice does not equal to expected value!!!",
+        )
+        # Test if there are no currency
+        inv.write({"currency_id": False})
+        self.assertEqual(
+            inv.custom_rate,
+            1.0,
+            "Custom rate of invoice does not equal to expected value!!!",
```
luisg123v: Same as above, equal -> match and no "!", no need to scream :)

account_invoice_change_currency/tests/test_account_invoice_change_currency.py
```diff
@@ -0,0 +1,320 @@
...
+            before_amount,
+            "Amount must remain the same, because None change was made",
+        )
+
+    def test_old_invoices(self):
+        # This simulate an old invoice (no stored original values)
+        res_currency = self.env["res.currency"]
+        inv = self.create_simple_invoice(inv_type="out_invoice")
+        inv.write({"original_currency_id": False})
+        for line in inv.line_ids:
```
luisg123v: Neither loops nor conditionals should be used in tests, as we should know exactly the number of records.

In this case, it's not even necessary because you could write to all lines at once

account_invoice_change_currency/tests/test_account_invoice_change_currency.py
```diff
@@ -0,0 +1,320 @@
...
+            line.write({"original_price_unit": False})
+        self.assertEqual(
+            inv.original_currency_id,
+            res_currency,
```
luisg123v: You don't need an empty recordset to compare, just use `assertFalse`

account_invoice_change_currency/tests/test_account_invoice_change_currency.py
```diff
@@ -0,0 +1,320 @@
...
+        self.assertEqual(
+            inv.line_ids.mapped("original_price_unit"),
+            [0.0 for line in inv.line_ids],
+            "There are original price units in the invoice",
+        )
+        inv.action_account_change_currency()
+        self.assertEqual(
+            inv.original_currency_id,
+            inv.currency_id,
+            "Original currency of invoice is" "not equal to expected value!!!",
```
luisg123v: Same here, apply comments pointed out above

account_invoice_change_currency/tests/test_account_invoice_change_currency.py
```diff
@@ -0,0 +1,320 @@
...
+        )
+        self.assertEqual(
+            inv.line_ids.mapped("original_price_unit"),
+            inv.line_ids.mapped("price_unit"),
+            "Original price units of invoice are" "not equal to expected value!!!",
```
luisg123v: And here

luisg123v (changes requested):

rolandojduartem:
Hi @luisg123v changes are done!

luisg123v (approved):
LGTM :+1: 

@moylop260  and @hugho-ad,

What do you think?

FYI it's now green, previous red was unrelated.

Regards,

moylop260:
@rolandojduartem 


Could you rebase the branch in order to get a new commit and trigger the CI in order to get the same error as merged one, please?

Also, could you have a look to the following error, please?
 - https://github.com/OCA/account-invoicing/runs/6427348061?check_suite_focus=true#step:8:478

rolandojduartem:
Hi @luisg123v and @moylop260, these red tests are from account_invoice_line_description

moylop260:
Yes

moylop260:
Yes, I just rebuilt the stable build in order to check if it is already failing even without this change:
 - https://github.com/OCA/account-invoicing/actions/runs/2290815681

moylop260:
I confirm it is even failing in stable

So the red is unrelated

Manually merging it

@rolandojduartem 

Even could you make an issue with the current error in CI and ping authors/maintainers please?

PR #1117, opened 2022-03-15
[FIX][14.0][account_invoice_section_sale_order] fix passing default_journal_id when invoicing

sebastienbeau:
If an other module pass in the context a specific journal for the invoice (passing default_journal_id in the context).
The invoice generation is broken.
Indeed the line created do not have any account_id specify and odoo will add it with the default_get method (in account/model/move.py). So it will try to create a line_section with an account and the sql constraint will raise an error.

Fix this incompatibility and add a test

### DIFF
--- a/account_invoice_section_sale_order/models/sale_order.py
+++ b/account_invoice_section_sale_order/models/sale_order.py
@@ -42,6 +42,10 @@ class SaleOrder(models.Model):
                                 "name": group._get_invoice_section_name(),
                                 "display_type": "line_section",
                                 "sequence": sequence,
+                                # see test: test_create_invoice_with_default_journal
+                                # forcing the account_id is needed to avoid
+                                # incorrect default value
+                                "account_id": False,
                             },
                         )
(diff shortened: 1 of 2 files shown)


legalsylvain (approved):

TDu (approved):

AaronHForgeFlow:
What would be the account in case you don't pass a specific journal? The one from the default_get? or just False?

rafaelbn:
@sebastienbeau please, could you answer here https://github.com/OCA/account-invoicing/pull/1117#issuecomment-1068811022 ? Thank you

sebastienbeau:
> What would be the account in case you don't pass a specific journal? The one from the default_get? or just False?

If you do not pass a specific journal and do not set False to the account value.

It will use the default get but in that case as there is not default_journal_id the default_get will return False, so this do not raise any issue

AaronHForgeFlow (approved):
Ok, thanks LGTM

PR #1127, opened 2022-04-05
[12.0][IMP] account_invoice_refund_line_selection: prepare refund vals

Abranes:
Call the Odoo base function `_prepare_refund` because this function implements that logic and some other modules inherit it to add more data.

For example the module `account_invoice_pricelist` in this same repository which adds `pricelist_id` to `account.invoice` and inherits `_prepare_refund` to pass the `pricelist_id` when you create a refund. With the current code we miss the pricelist when we create a refund but with this changes we preserve this value and all others.


etobella:
It was done this way because there are some fields that needed to be recalcualted (like taxes, I am not sure if this will work in all cases :sob: )

fuentes73 (approved):

PR #1130, opened 2022-04-06
Added copier from template

etobella:
Passing from Travis to GH Actions


etobella:
@LoisRForgeFlow @AaronHForgeFlow @JordiBForgeFlow There is license issues here. Can you review? Is it right for you?

AaronHForgeFlow (approved):
It is ok to me, perhaps we can make the base_tier_validation LGPL as it is a base module for others to use but I guess it is not an issue now.

LoisRForgeFlow (approved):
:+1: to relicense

PR #1131, opened 2022-04-06
[FIX] account_invoice_refund_line_selection: Business fields were missing

etobella:
When using with sale, sale_order_lines were lost

How to test the issue, create a Sale Order and Invoice it. If you create a refund from the invoice, the sale order lines are marked as to invoice, but it will not happen if using the partial refund


albariera (approved):
LGTM :+1:

AaronHForgeFlow (approved):
Code review LGTM

PR #1135, opened 2022-04-11
[FIX] account_invoice_section_sale: fix invoice total when using currency

sebastienbeau:
In case that you invoice sale order that are not in the same currency of the company the total amount on the invoice is wrong.

Indeed lines do not have the same currency (section line have the company currency) and so odoo will show as total the total in the company currency instead of the total in the invoice currency

see logic here : https://github.com/odoo/odoo/blob/01d4cefd20961a6518b6d20b88f0f42638fa16a1/addons/account/models/account_move.py#L1421


sebastienbeau:
@TDu @legalsylvain @AaronHForgeFlow

adrienpeiffer (approved):
LGTM

mourad-ehm (approved):

PR #1140, opened 2022-04-14
[13.0][IMP] account_product_move

ntsirintanis:


ntsirintanis:
@NL66278 @thomaspaulb @lfreeke @gfcapalbo

account_move_journal_template/models/account_move.py
```diff
@@ -0,0 +1,116 @@
...
+            "type": "ir.actions.act_window",
+            "res_model": "account.move",
+            "view_mode": "tree,form",
+            "target": "current",
+            "domain": [("id", "in", self.journal_entry_ids.ids)],
+        }
+
+    def post(self):
+        res = super().post()
+        all_journals = self.env["account.move"]
```
NL66278: This is extremely confusing. With the name `journals` you would expect a recordset of "account.journal". Please rename to something like `moves`.

NL66278: Or maybe `extra_moves` to distinguish from the moves in self.

ntsirintanis:
@NL66278 the main problem here is that I am not sure how to reverse the automatically created journal entries, when reversing the account.move that generated them...

account_move_journal_template/models/account_move_journal_template.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def _check_balanced(self):
+        """ An adaptation of account.move._check_balanced()"""
+        moves = self.filtered(lambda move: move.journal_item_ids)
+        if not moves:
+            return
+        self.env["account.move.journal.template.item.line"].flush(
+            self.env["account.move.journal.template.item.line"]._fields
+        )
+        self._cr.execute(
```
NL66278: The currency on the journal can override the currency of the company. Maybe best to setup a for this in self iteration and test template for template. Then determine currency from journal and company. Inject the currency.decimal_places as a variable in the SQL statement. On the plus side this will safe some joins. All this rounding and manipulation would not be needed if Odoo used proper decimal fields instead of floats to store monetary amounts.

account_move_journal_template/models/account_move_journal_template.py
```diff
@@ -0,0 +1,116 @@
...
+            sums = [res[1] for res in query_res]
+            raise UserError(
+                _(
+                    "Cannot create unbalanced journal entry. "
+                    "Ids: %s\nDifferences debit - credit: %s"
+                )
+                % (ids, sums)
+            )
+
+    def write(self, vals):
```
NL66278: If you can change the product template attached to the move template, you also will have to remove the link from the old product template, not only add a link to the new.

account_move_journal_template/models/account_move_journal_template.py
```diff
@@ -0,0 +1,116 @@
...
+    @api.model_create_multi
+    def create(self, vals_list):
+        records = super().create(vals_list)
+        records._check_balanced()
+        for this in records:
+            this.product_tmpl_id.journal_tmpl_id = this
+        return records
+
+
+class ProductTemplate(models.Model):
```
NL66278: Put this in a separate file

account_move_journal_template/models/account_move_journal_template.py
```diff
@@ -0,0 +1,116 @@
...
+    _inherit = "product.template"
+
+    journal_tmpl_id = fields.Many2one(
+        comodel_name="account.move.journal.template", string="Journal Template",
+    )
+
+
+class AccountMoveJournalTemplateItemLine(models.Model):
```
NL66278: And this in a separate file as well

account_move_journal_template/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,3 @@
+To configure this module, you need to:
+
+#. Install it.
```
NL66278: Well you need to go to the product (templates) and add a journal entry templates to them...

account_move_journal_template/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+Automatically create Journal entries and items when posting out-invoices.
```
NL66278: This always happens when posting invoices. make somehow clear that it are extra moves dependent on the products in the invoice lines.

account_move_journal_template/readme/CONFIGURE.rst
NL66278: OK, this is taken up in the Usage

NL66278:
I am not quite happy with the name of the module, nor with the names of some of the models. I think the central concept of this module is that certain products bring with them certain journal items (account move lines).

So I would propose:
account_product_move_template for the module and for the tamplate linked to the product (template)
account_product_move_template_line for the lines (items) in this template.

Maybe even better to get rid of the _template part. The logic of the module seems to be that invoicing certain products, inescapably results in extra account moves that are fully determined by those products and the defined product moves. No option for the user to override these, so in fact it is not really a "template".

account_move_journal_template/models/account_move.py
```diff
@@ -0,0 +1,116 @@
+# Copyright 2022 Therp BV <https://therp.nl>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from odoo import _, fields, models
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    journal_entry_ids = fields.One2many(
```
NL66278: If I understand correctly this is the link FROM the Invoice TO the extra moves generated for some products. Please add a help="blahblah" to that effect.

account_move_journal_template/models/account_move.py
```diff
@@ -0,0 +1,116 @@
...
+        comodel_name="account.move", inverse_name="journal_entry_id",
+    )
+    journal_entry_id = fields.Many2one(comodel_name="account.move")
```
NL66278: And again if I understand correctly this is the link FROM the generated extra move TO the original invoice. Please add a help= also here.

account_move_journal_template/models/account_move.py
NL66278: And maybe rename field to product_move_ids?

account_move_journal_template/models/account_move.py
NL66278: And maybe rename field to invoice_move_id?

account_move_journal_template/models/account_move.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def button_draft(self):
```
NL66278: need to make sure also that a move generated from a product cannot be set to draft, unless the invoice move has been set to draft. An equivalent check is needed also for cancel.

account_move_journal_template/models/account_move.py
```diff
@@ -0,0 +1,116 @@
...
+            "domain": [("id", "in", self.journal_entry_ids.ids)],
+        }
+
+    def post(self):
+        res = super().post()
+        all_journals = self.env["account.move"]
+        for move in self:
+            if move.type not in ["out_invoice", "out_refund"]:
+                continue
+            # Remove previous lines
```
NL66278: Why only the lines? And maybe, if each post action results in a full regeneration, should lines (and moves) not be deleted / ulinked when setting an invoice to draft?

account_move_journal_template/models/account_move.py
```diff
@@ -0,0 +1,116 @@
...
+            move.journal_entry_ids.line_ids.unlink()
+            product_tmpl_ids = move.mapped(
+                "invoice_line_ids.product_id.product_tmpl_id"
+            )
+            for tmpl in product_tmpl_ids:
```
NL66278: I think it will be much easier if the code follows the structure of the data, something like (pseudocode):
```
DELETE existing extra moves if any (should not happen)
FOR move in self:
   FOR invoice_line IN move.line_ids:
        tmpl = invoice_line.product_id.product_templ_id
        IF NOT tmpl..account_product_move_id:
            continue
         CREATE extra move for product template
         FOR product_move_line IN tmpl..account_product_move_id.line_ids:
             CREATE extra line for product_move_line
         extra_move.post()
```

No need to search for existing extra moves. No need to keep extra moves in a variable. No need to look up the qiuantity for the invoice line product, as you will have this in the invoice_line record...

NL66278:
@ntsirintanis For the credit invoices I think you can override reverse_moves in the account_move_reversal wizard:
https://github.com/odoo/odoo/blob/13.0/addons/account/wizard/account_move_reversal.py#L65

You can call the super on this and then loop over all the linked moves (if any) and reverse them as well.

Of course this supposes that Credit Invoices go through this wizard, which is the case anyway if users use the Add Credit Note button on the invoice.

There is also a _reverse_move method on account.move itself. It might be necesarry to do the same thing there. I think it is used in reversing reconciliations.

account_move_journal_template/models/account_move.py
ntsirintanis: Better keep the moves here, and add/remove lines between state transitions. if an extra move is cancelled or in draft, it won't partake into anything really. User can delete them if they feel like it.

account_move_journal_template/models/account_move.py
ntsirintanis: The problem with this one is that we need an extra_move per journal_id of account_product_move_line.

account_move_journal_template/models/account_move_journal_template.py
ntsirintanis: I will need to revisit this, because it is not yet completely clear to me what to do

ntsirintanis:
tests are failing in another module https://github.com/OCA/account-invoicing/runs/6654881995

account_move_journal_template/models/account_move_journal_template.py
ntsirintanis: Yes, @NL66278 I do not really understand what your comment is about. This method is there to ensure that debit-credit is balanced, and it is called whenever we create or write to account_product_move.journal_item_ids. It is basically what happens to account.move too.
What I can further do on this is to get rid of journal_id per account_product_move_line, add it to account_product_move instead, and impose a uniqueness constraint on the combination of product_tmpl_id and journal_id for account_product_move records.

account_product_move/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+Create extra Journal entries and items when posting out-invoices, that depend on the products in the invoice lines.
```
thomaspaulb, line 1: I feel this is a more clear description:

```
This module creates extra Journal Entries for outgoing invoices, depending on the products sold in the invoice lines. To specify the products for which this should happen, the user creates Account Product Move records.
```

Do you agree?

What I'm also missing here is a backstory about the use case in which this can be beneficial, illustrated by an example, along the lines of:

```
For example, a user could decide to create an account product move record with lines of 10 cents for a cone, and 15 cents for ice cold delivery, on an Ice Cream product. This would then lead to extra journal entries of 10 and 15 cents being created, signifying....
```

When you finish the complete text, could you then concatenate it with the USAGE text and put it on the description of the PR on top, so that it's clear and easy to see what's being worked on here?

account_product_move/readme/USAGE.rst
```diff
@@ -0,0 +1,9 @@
+To use this module, you need to:
+
+#. Go to *Accounting > Accounting > Journals > Journal Templates*.
```
thomaspaulb, line 3: This is now `Extra Product Moves`.

account_product_move/readme/USAGE.rst
```diff
@@ -0,0 +1,9 @@
...
+#. Create a new Account Product Move record and give it a name.
+#. Create product template lines.
+#. Create Product Move Lines for these products.
+#. Create a new out invoice with product lines.
+#. Post the out invoice.
+#. If any of the products in invoice lines has an Account Product Move record set-up, exist a new journal entry will be added per line in Account Product Move Lines of Account Product Move, holding these Account Product Move Lines.
```
thomaspaulb, line 9: Maybe the last three steps more clear as:

#. Create a new outgoing invoice with invoice lines mentioning some of these products.
#. Confirm the invoice.
#. A new journal entry will be added per line in Account Product Move Lines of Account Product Move, holding these Account Product Move Lines.

account_product_move/models/account_product_move.py
```diff
@@ -0,0 +1,99 @@
...
+        string="Product",
+        help="Journal items will be created for this product",
+    )
+    journal_item_ids = fields.One2many(
+        comodel_name="account.product.move.line",
+        inverse_name="product_move_id",
+        copy=True,
+        string="Extra Journal Items",
+        help="Journal items to be added in new journal entry",
+    )
```
thomaspaulb, line 31: Am I correct in assuming that an `account.product.move` record would end up being create as a move (Journal Entry) and the lines as underlying move lines (Journal Item)?

account_product_move/models/account_move.py
```diff
@@ -0,0 +1,108 @@
...
+                    continue
+                for item in product_move.journal_item_ids:
+                    extra_move = move._get_or_create_extra_move(item)
+                    move._create_journal_entry_item(
+                        invoice_line.quantity, extra_move, item
+                    )
+                    # TODO: revisit this
+                    if extra_move in extra_moves:
+                        continue
+                    extra_moves += extra_move
```
thomaspaulb, line 65: I think you can just do: `extra_moves |= extra_move` to achieve this

account_product_move/views/account_product_move.xml
```diff
@@ -0,0 +1,96 @@
...
+    </record>
+    <!-- Action and menu -->
+    <record model="ir.actions.act_window" id="account_product_move_act_window">
+        <field name="name">Account Product Move</field>
+        <field name="res_model">account.product.move</field>
+        <field name="view_mode">tree,form</field>
+    </record>
+    <menuitem
+        id="menu_product_move"
+        name="Extra Product Moves"
```
thomaspaulb, line 90: I still feel it's a bit of a confusing name, but I can't think of anything better. Maybe Product-related Journal Entry templates.

account_product_move/readme/DESCRIPTION.rst
ntsirintanis: @thomaspaulb yes you are right about this, definitely should be clearer. I will add way more text when the requirements are completely clear to me, because at this very moment, they are still in variable state.

account_product_move/models/account_product_move.py
ntsirintanis: yes, that is the plan

account_product_move/readme/DESCRIPTION.rst
thomaspaulb: Yea i guess that starts with formulating the clients use case, maybe it ends up another solution is better. What does the extra move signify?

ntsirintanis: extra_move = a new journal_entry, created per product in invoice.invoice line, for account_product_move records configured for that product.

 The journal_items of this journal_entry will be the account_product_move_lines of that account_product_move record.

This is the plan for the time being.

thomaspaulb: @ntsirintanis Thats a technical description of what it is, but what we need is an odoo-functional description of what it signifies in the ledger, in the language of the use case.

ntsirintanis: @lfreeke this one's for you.

account_product_move/models/account_move.py
```diff
@@ -0,0 +1,108 @@
...
+            # Remove previous lines
+            move.product_move_ids.line_ids.unlink()
+            # Remove previous journal entries
+            move.product_move_ids.with_context(force_delete=True).unlink()
+            for invoice_line in move.invoice_line_ids:
+                product_move = invoice_line.product_id.product_tmpl_id.journal_tmpl_id
+                if not product_move:
+                    continue
+                for item in product_move.journal_item_ids:
+                    extra_move = move._get_or_create_extra_move(item)
```
NL66278, line 58: Did you not just delete any existing extra moves? So why get or create? Create should be sufficient?

account_product_move/models/account_move.py
```diff
@@ -0,0 +1,108 @@
...
+                    if extra_move in extra_moves:
+                        continue
+                    extra_moves += extra_move
+            for extra_move in extra_moves:
+                extra_move.action_post()
+            return res
+
+    def _get_or_create_extra_move(self, item):
+        move_model = self.env["account.move"]
+        vals = {
```
NL66278, line 72: if you need to search existing moves at all (see previous comment), all the moves with the right invoice_move_id would be sufficient. What if the invoice date changed after creating the extra moves? Anyway would the existing moves linked to the invoice not be in self.product_move_ids?

account_product_move/models/account_product_move_line.py
```diff
@@ -0,0 +1,17 @@
...
+class AccountProductMoveLine(models.Model):
+    _name = "account.product.move.line"
+    _description = "Items for journal entry for given products"
+
+    debit = fields.Monetary(default=0.0)
+    credit = fields.Monetary(default=0.0)
+    account_id = fields.Many2one(
+        comodel_name="account.account", ondelete="cascade", required=True
+    )
+    journal_id = fields.Many2one(comodel_name="account.journal", required=True)
```
NL66278, line 15: Should definitely be on the account.product.move level. You cannot have lines of the same move in different journals. probably the same goes for currency_id.

account_product_move/models/account_product_move_line.py
```diff
@@ -0,0 +1,17 @@
...
+    currency_id = fields.Many2one(related="journal_id.currency_id")
+    product_move_id = fields.Many2one(comodel_name="account.product.move")
```
NL66278, line 17: Make this the first field, because this is really part of a `parent==>chldren` relationship, So in the child the most important thing is the reference back to the parent....

account_product_move/models/product_template.py
```diff
@@ -0,0 +1,11 @@
+# Copyright 2022 Therp BV <https://therp.nl>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from odoo import fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    journal_tmpl_id = fields.Many2one(
```
NL66278, line 9: If you make the relation between product.template and account.product.move a Many2many, you can have several moves in a product (for instance to support different currencies or journals), but you can also re-use the account.product.move for several products. For instance if they all have the same cost, or the same tax, or the same whatever moves.

NL66278:
@ntsirintanis Can you close this one and create a new PR based on the branch in the Therp repository?

PR #189, opened 2016-10-04
WIP 8.0 account invoice line default account merged

jbaudoux:

PR #23, opened 2014-10-20
8.0 account invoice line default account

jbaudoux:
Porting this MP from lp to github
https://code.launchpad.net/~bcim/account-invoicing/7.0-account_invoice_line_default_account

LP description: Extracted account_invoice_line_default_account module version 6.1 from account-financial-tools project and ported to 7.0. I'm pushing this migrated module to this account-invoicing project as this module is for invoicing and is not a financial tool

Ported to 8.0

PR #1143, opened 2022-04-21
[12.0][MIG] account_invoice_line_default_account

SirTakobi:
Migrating/Adding **account_invoice_line_default_account** to `12.0`.

This poor module has a troubled history: as far as I have found, it was brought from launchpad's `6.1` (unmerged) to `8.0` with https://github.com/OCA/account-invoicing/pull/23, then https://github.com/OCA/account-invoicing/pull/189, and finally https://github.com/OCA/account-invoicing/pull/356 (currently still open).

Then a migration to `10.0` happened in https://github.com/OCA/account-invoicing/pull/351 (currently still open).

I tried to put together something like a consistent commit history and this is the result, with a last commit for migrating to `12.0`.

Additional notes:

- I added some tests :tada:
- ~~There is a fixup commit because I couldn't find where the `__openerp__` was removed between https://github.com/OCA/account-invoicing/pull/356 and https://github.com/OCA/account-invoicing/pull/351, I didn't squash it in order not to mix authorships, but I can do it if @jbaudoux agrees~~
- About the roadmap:
  - For `account.invoice.line._onchange_account_id` I already tried substituting `write` with `update` but that results in not updating the partner (can be verified with the tests),
  - For `account.invoice.line.account_id`, currently the module assumes the previous `default` method is the one defined in **account** but that might not be True; retrieving the field's `default` inside the defined method is useless because it has already been changed to the defined method

  If you have any suggestion it is welcome

Looking at the linked PRs and commits, @jbaudoux you might want to have a look :)


account_invoice_line_default_account/models/account_invoice_line.py
```diff
@@ -0,0 +1,58 @@
...
+        elif invoice_type in ['out_invoice', 'out_refund']:
+            partner = self.env['res.partner'].browse(partner_id)
+            if partner.property_account_income:
+                return partner.property_account_income.id
+        return self._default_account()
+
+    account_id = fields.Many2one(default=_account_id_default)
+
+    @api.onchange('account_id')
+    def _onchange_account_id(self):
```
jbaudoux: maybe better to use a different name to not have to call super

account_invoice_line_default_account/models/account_invoice_line.py
```diff
@@ -0,0 +1,58 @@
...
+        if not self.account_id or not self.partner_id or self.product_id:
+            return super(AccountInvoiceLine, self)._onchange_account_id()
+        if self._context.get('journal_id'):
```
jbaudoux: self.env.context

jbaudoux (commented):
Thanks for adding tests.

Some minor remarks.

I remember when I ported to v10 that lot of code had to be rewritten, that's why I squashed all the old commits. I don't think it is useful to recover all those old commits and I would squash them all into a single commit or at least reduce them a lot.

account_invoice_line_default_account/models/account_invoice_line.py
SirTakobi: Thanks for the review!
I can see no harm in calling `super`, actually it looks to me like a common thing to call `super` when overriding a method; why should this case be treated in a different way?

account_invoice_line_default_account/models/account_invoice_line.py
SirTakobi: Thanks, done :)

SirTakobi:
> I remember when I ported to v10 that lot of code had to be rewritten, that's why I squashed all the old commits. I don't think it is useful to recover all those old commits and I would squash them all into a single commit or at least reduce them a lot.

@jbaudoux I squashed most of the commits, and added the **\_\_openerp__.py** removal in `10.0` migration, let me know if it looks better now.
Just as reference, I have kept the previous history in https://github.com/SirTakobi/account-invoicing/tree/12.0-mig-account_invoice_line_default_account-bkp

Can you update your review? Github is not allowing me to re-request you to do it

> /ocabot migration account_invoice_line_default_account

@rafaelbn thanks, I'm sorry I forgot to do those things

account_invoice_line_default_account/models/account_invoice_line.py
jbaudoux: Before `api.onchange`, there could be only a single method called on an onchange. With `api.onchange` you can have multiple different methods. The conditions are not to test if super has to be called or not but if this method has to be executed. So I find the current implementation hard to read.
Nevertheless it will work. Just a minor remark.

SirTakobi: That makes perfect sense: overriding the method is really needed only when you want to edit the method's behavior, but that does not have to be the way to go with the `onchange` methods.
I have done this change too, with a bit of explanation in the comments, let me know if it looks better now :)

jbaudoux: Looks clear now

account_invoice_line_sequence/__manifest__.py
```diff
@@ -16,6 +16,9 @@
     'data': ['views/account_invoice_view.xml',
              'views/report_invoice.xml'],
     'depends': ['account'],
+    'excludes': [
```
jbaudoux: Why the 2 modules cannot work together?

SirTakobi: I wrote it in the commit message:
> This module is not compatible with account_invoice_line_sequence because them both add to the context of invoice lines in invoice form view

jbaudoux (approved):
LGTM

rafaelbn (approved):

.travis.yml
```diff
@@ -23,8 +23,10 @@ env:
 
   matrix:
   - LINT_CHECK="1"
-  - TESTS="1" ODOO_REPO="OCA/OCB"
-  - TESTS="1" ODOO_REPO="odoo/odoo" MAKEPOT="1"
+  - TESTS="1" ODOO_REPO="OCA/OCB" EXCLUDE="account_invoice_line_default_account"
+  - TESTS="1" ODOO_REPO="OCA/OCB" EXCLUDE="account_invoice_line_sequence"
+  - TESTS="1" ODOO_REPO="odoo/odoo" MAKEPOT="1" EXCLUDE="account_invoice_line_default_account"
+  - TESTS="1" ODOO_REPO="odoo/odoo" MAKEPOT="1" EXCLUDE="account_invoice_line_sequence"
```
eLBati: Can runboat be informed about this?

https://runboat.odoo-community.org/api/v1/builds/bf7e8fb88-cda7-4c02-a0c2-5717acb6d8bf/init-log

eLBati (changes requested):
Proposing https://github.com/SirTakobi/account-invoicing/pull/1 to avoid basing on context and thus incompatibilities

.travis.yml
SirTakobi: Runboat currently does not support conflicting addons, see https://github.com/sbidoul/runboat/issues/36

SirTakobi:
> Proposing [SirTakobi#1](https://github.com/SirTakobi/account-invoicing/pull/1) to avoid basing on context and thus incompatibilities

@eLBati there are some things that do not really convince me in https://github.com/SirTakobi/account-invoicing/pull/1.

Instead of working on the linked PR, can we go on with the merge of this migration alone? Since we already have enough approvals, this module might finally be merged for the first time! :smile:
Then you will be able to propose that PR directly to `12.0`.
Unless you think there is something in this PR that might be blocking for the merge.

SirTakobi:
@OCA/accounting-maintainers can this be merged or do I have to change something? Thanks!

eLBati:
> Unless you think there is something in this PR that might be blocking for the merge

Uhm yes, this https://github.com/SirTakobi/account-invoicing/pull/1

SirTakobi:
> > Unless you think there is something in this PR that might be blocking for the merge
> 
> Uhm yes, this [SirTakobi#1](https://github.com/SirTakobi/account-invoicing/pull/1)

Now that you explained those changes, I merged them :)

Just for future reference, my doubts were about choosing the onchange on `product_id` field specifically, but we clarified that that onchange is called on any line creation from UI; so the onchange can do the job that was previously (when onchanges were not called on line creation) performed from the line's view using the context.

SirTakobi:
@eLBati can you update your review now? Thanks!

eLBati (approved):
Thanks

SirTakobi:
@OCA/accounting-maintainers can this be merged? There are a few happy reviewers already :smile:

SirTakobi:
@OCA/accounting-maintainers can this be merged? Thanks!

SirTakobi:
@OCA/accounting-maintainers can this be merged? Thanks!

PR #1145, opened 2022-04-29
[14.0] [MIG] account_invoice_fix_tax_rounding

TheMule71:
A lot changed since 12.0, so it's basicly a rewrite.

Further investigation has been done on the root cause. It turns out to be related to floating point quirks, that the Decimal class (in python) solves. See https://docs.python.org/3/library/decimal.html and https://docs.python.org/3/tutorial/floatingpoint.html

This addresses a rare case when a sum (e.g. the total VAT) happens to be barely the right amount to be rounded up instead of down, e.g, with precision 2, 0.005. This is rounded up to 0.01. Anything below that, like 0.00499999999, is rounded down to 0.00.

It is possible that numbers ending with a 5 in the decimal precision + 1 position get represented in python, as floating point numbers slightly higher or lower than the actual value. Higher isn't a problem, but lower causes rounding problems.

When global rouding is on, in methods like `compute_all()`, `_recompute_tax_lines()` (in `_recompute_dynamic_lines()`), tax amounts are still computed line by line, but with higher precision and without rounding, which, from a purely mathematical standpoint, is correct. But unfortunately floating point math isn't always correct, mathematically.

We mitigate the problem by actually rounding the amounts at a higher precision before the final rounding happens. This
turns e.g. 0.0049999999 into 0.0050000 first (extra zeros at the end intented), which later gets rounded to 0.01 (if precision is 2).


andreampiovesana (approved):
go

matteoopenf (approved):
functional test ok

rafaelbn:
Hello! 😄 I've never notice this problem. Any case, and obviously not blocking!! 🟢 Just a question? @sergiocorato @TheMule71 did you report this BUG to Odoo in order they could fix it? I guess yes... and that why you did a module to fix a BUG. Is curiosity about the problem I missed. Thank you a lot.

matteoopenf:
Hi, 

We have the same problem.
We have also open a ticket to Odoo, but Odoo not provide properly response and the ticket stil open, we wait response, but until Odoo response we need a solution.
We have a lot e-invoice not accepted from SDI. 



Matteo Mircoli
Developer | ***@***.*** 

> Il giorno 10 mag 2022, alle ore 21:30, Rafael Blasco ***@***.***> ha scritto:
> 
> ﻿
> Hello! 😄 I've never notice this problem. Any case, and obviously not blocking!! 🟢 Just a question? @sergiocorato @TheMule71 did you report this BUG to Odoo in order they could fix it? I guess yes... and that why you did a module to fix a BUG. Is curiosity about the problem I missed. Thank you a lot.
> 
> —
> Reply to this email directly, view it on GitHub, or unsubscribe.
> You are receiving this because you commented.

-- 

_
Openforce Srls Unipersonale *| www.openforce.it 
<http://www.openforce.it/> |*__ 02 21119472 | Via Tiziano,32 20145 Milano_

matteoopenf:
@OCA/accounting-maintainers  is it possible merge?
Thank you

matteoopenf:
HI @TheMule71, 

I retry the module in my local env, but when I valorize the field:
image


the value is not keep, is it normal?

Thank you

andreampiovesana:
i suggest to change the default to 0,001    
tax_max_diff_global_rounding_method = fields.Float(
        default=0.001,
        required=True,
        string="Tax calculation max diff amount for rounding method",
    )

matteoopenf (approved):

matteoopenf:
@OCA/accounting-maintainers  is possibile to merge?

andreampiovesana (approved):
merge

matteoopenf:
@pedrobaeza  is it possibile for you merge?
thank you

pedrobaeza:
CI is red

matteoopenf:
@TheMule71  have you any idea?

TheMule71:
> @TheMule71 have you any idea?

Probably #1157

matteoopenf:
@pedrobaeza  the CI red is not for this PR

pedrobaeza:
Please open an issue about that and tell it to the `account_invoice_line_description` maintainers.

matteoopenf:
@pedrobaeza  now is it possible do /ocabot rebase?

matteoopenf:
> /ocabot rebase

now ready to merge

matteoopenf:
@pedrobaeza  is it now ready to merge right?
thank you for the help

pedrobaeza:
A PSC must review it:

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#review

matteoopenf:
> A PSC must review it:
> 
> https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#review

thanks

matteoopenf:
@OCA/accounting-maintainers  can you review?

rafaelbn:
Hello @alexis-via , what do you think about this? You opinion in this could really interesting.

@matteoopenf , one think you can do is to make a Twit like this one: https://twitter.com/alexisdlattre/status/1535174658768654338

matteoopenf:
@alexis-via  can you review?

matteoopenf:
@alexis-via  any chance ?

elvise:
@pedrobaeza can we merge ? :)

francesco-ooops:
@sergiocorato @tafaRU what's your opinion on this?

francesco-ooops:
@TheMule71 how can we proceed with this? it's more than 4 months since we've been trying to get a review from @OCA/accounting-maintainers , with no success :(

sergiocorato:
To review this PR I need a test, as it's possible that this module is no more useful.

PR #1158, opened 2022-05-16
[12.0][FIX]account_global_discount: Don't round tax amounts.

jesusVMayor:
Round only base in get_taxes_values function


pedrobaeza:
Can you please add a regression test to prove the problem and its solution?

PR #1159, opened 2022-05-16
[IMP] account_invoice_supplier_ref_unique: Allows duplicate invoice number if they are from different fiscal years

gcobachoesment:
[IMP] account_invoice_supplier_ref_unique: Allows duplicate invoice number if they are from different fiscal years


gcobachoesment:
This change is not necessary as of version 13 because it can be handled through the reference field.

PR #1165, opened 2022-05-26
[14.0][ENH] account_invoice_payment_retention: add retention computation method

ps-tubtim:
This PR adds the retention computation method. In case compute retention by percent, you can select the retention method for retention computation.

- If you select **Untexed Amount**,  the retention amount will compute from the _untaxed amount._
- If you select **Total**,  the retention amount will compute from the _total amount_.

Screenshot from 2022-05-26 16-25-51


Saran440 (approved):
Run on production :+1:

newtratip (approved):
OK For me :+1:

PR #1171, opened 2022-06-04
[15.0][MIG] account_invoice_supplier_ref_unique

JuanyDForgeflow:
This module checks that a supplier invoice/refund is not entered twice. This is important because if you enter twice the same supplier invoice, there is also a risk that you pay it twice !

This module adds a constraint on supplier invoice/refunds to check that (commercial_partner_id, supplier_invoice_number) is unique, without considering the case of the supplier invoice number.


MiquelRForgeFlow:
please rebase

JuanyDForgeflow:
> please rebase

Done! Thanks for the notice

MiquelRForgeFlow (approved):

account_invoice_supplier_ref_unique/models/account_move.py
```diff
@@ -0,0 +1,80 @@
...
+        # OVERRIDE
+        if not default_values_list:
+            default_values_list = [{} for move in self]
+        for move, default_values in zip(self, default_values_list):
+            if (
+                move
+                and move.is_purchase_document(include_receipts=True)
+                and default_values.get("ref")
+            ):
+                default_values.update({"ref": ""})
```
BT-anieto: I would put it in an else condition since it does not make sense to run this code if there is not default_values_list

BT-anieto (approved):
LGTM. I left a comment because I think that code could be slightly improved, but not critical.

BT-anieto:
Hi! :) Could you solve the wrong checks so that it can be merged?

account_invoice_supplier_ref_unique/models/account_move.py
JuanyDForgeflow: done

AaronHForgeFlow (approved):
LG!

PR #1173, opened 2022-06-09
[15.0][MIG] account_move_exception

JasminSForgeFlow:
Standard Migration

@ForgeFlow

- Depends on https://github.com/OCA/server-tools/pull/2355


MiquelRForgeFlow:
Now depends on:

- [x] https://github.com/OCA/server-tools/pull/2366

MiquelRForgeFlow:
please rebase

JasminSForgeFlow:
> please rebase

Done, Thanks

MiquelRForgeFlow:
Depends on:

- [x] https://github.com/OCA/server-tools/pull/2373

MiquelRForgeFlow:
Please rebase :)

JasminSForgeFlow:
> Please rebase :)

Done, Thanks

MiquelRForgeFlow (approved):

JuanyDForgeflow (approved):
LGTM!

AaronHForgeFlow (changes requested):
This menu in invoicing settings does not take you anywhere. I am full accounting features and exception manager, but it does nothing. Also, it seems to me that two menus should be shown and there is only one:

JasminSForgeFlow:
> This menu in invoicing settings does not take you anywhere. I am full accounting features and exception manager, but it does nothing. Also, it seems to me that two menus should be shown and there is only one: 

@AaronHForgeFlow I have changed the menu to the same level as management.
Now it will show both menu

Thanks

account_move_exception/views/account_view.xml
```diff
@@ -0,0 +1,91 @@
...
+            name="domain"
+        >[('model', 'in', ['account.move', 'account.move.line'])]</field>
+        <field
+            name="context"
+        >{'active_test': False, 'default_model' : 'account.move'}</field>
+    </record>
+    <menuitem
+        id="menu_account_exception_parent"
+        parent="account.menu_finance_configuration"
+        sequence="20"
```
AaronHForgeFlow, line 18: Can you add the group exception manager here? Other wise I get an error when accessing and trying to modify an exception: 

```
This operation is allowed for the following groups:
	- Exception manager
```

If I am not exception manager I should not have access

JasminSForgeFlow: Internal user have read access so there will no error

AaronHForgeFlow: But it is ugly to throw a exception like this one:

account_move_exception

I think it makes sense to restrict the group

JasminSForgeFlow: Please try after hard refresh, you won't get create/edit button

image

image

AaronHForgeFlow: I reproduced that in runboat, all the changes are available there I think. When do you restricted that?

In any case I think this is a bad decision when creating the module in previous versions and the group must be assigned to the menu items, only the mangers should see and edit, IMHO. Indeed the base_exception module restricts the security group.

JasminSForgeFlow: @AaronHForgeFlow I have added a group for the menu

Thanks

AaronHForgeFlow (approved):
Thank you @JasminSForgeFlow I will keep this open some days more just in case someone else disagrees with the changes

PR #1179, opened 2022-06-16
[ADD]account_invoice_discount_display_amount: Show total discount on invoices

manuelregidor:
This module shows the total discount amount and the amount without discount on invoices.


manuelregidor:

invoice_discount_display_amount/models/account_move_line.py
```diff
@@ -0,0 +1,40 @@
...
+    discount_total = fields.Monetary(
+        compute="_compute_discount_amount", string="Discount Total", store=True
+    )
+    price_total_no_discount = fields.Monetary(
+        compute="_compute_discount_amount", string="Total Without Discount", store=True
+    )
+
+    @api.depends("discount", "price_total")
+    def _compute_discount_amount(self):
+        for line in self:
```
ValentinVinagre: You should only take into account those that are invoice lines

ValentinVinagre (changes requested):

invoice_discount_display_amount/models/account_move_line.py
```diff
@@ -0,0 +1,40 @@
...
+            price_total_no_discount = 0
+            discount_total = 0
+            if not line.discount:
```
ValentinVinagre: can be added as a filtered in the for and make a massive change to those that are not invoice lines and do not have a discount

invoice_discount_display_amount/models/account_move.py
```diff
@@ -0,0 +1,38 @@
...
+        string="Total Without Discount",
+        currency_field="currency_id",
+        store=True,
+    )
+
+    @api.depends(
+        "invoice_line_ids.discount_total", "invoice_line_ids.price_total_no_discount"
+    )
+    def _compute_discount_total(self):
+        for invoice in self:
```
ValentinVinagre: you should filter those account.move that are not invoices and directly indicate that the value of the fields is 0.0 with a massive change.

ValentinVinagre (approved):
👯‍♂️

ValentinVinagre:
@HaraldPanten could you review?

HaraldPanten (approved):
Functional review

ValentinVinagre:
@AaronHForgeFlow could you review?

PR #1180, opened 2022-06-16
[IMP] account_invoice_change_currency: Speed up module installation.

rolandojduartem:
Initializing column custom_rate on table account_move for the improvement in performance to avoid long duration in databases with thousand of moves using a hook.

### DIFF
--- a/account_invoice_change_currency/hooks.py
+++ b/account_invoice_change_currency/hooks.py
@@ -1,0 +1,29 @@
+import logging
+
+from odoo import tools
+
+_logger = logging.getLogger(__name__)
+
+
+def pre_init_hook(cr):
+    """
+    Initializing column custom_rate on table account_move
+    for the improvement in performance to avoid long
+    duration in databases with thousand of moves
+    """
+    _logger.info("Initializing column custom_rate on table account_move")
+    tools.create_column(
+        cr=cr,
+        tablename="account_move",
+        columnname="custom_rate",
+        columntype="numeric",
+        comment="Custom Rate",
+    )
+    cr.execute(
+        """
+        UPDATE
+            account_move
+        SET
+            custom_rate = 1.0
+        """
+    )
--- a/account_invoice_change_currency/__init__.py
+++ b/account_invoice_change_currency/__init__.py
@@ -1,3 +1,4 @@
 # Copyright 2017 Komit <http://komit-consulting.com>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 from . import models
+from .hooks import pre_init_hook
--- a/account_invoice_change_currency/__manifest__.py
+++ b/account_invoice_change_currency/__manifest__.py
@@ -13,5 +13,6 @@
     "data": [
         "views/account_move_views.xml",
     ],
+    "pre_init_hook": "pre_init_hook",
     "installable": True,
 }


account_invoice_change_currency/hooks.py
```diff
@@ -0,0 +1,29 @@
...
+_logger = logging.getLogger(__name__)
+
+
+def pre_init_hook(cr):
+    """
+    Initialize custom_rate in account_move for the
+    improvement in performance to avoid long duration
+    in databases with thousand of moves
+    """
+    _logger.info("Initialize custom_rate field in account_move")
```
luisg123v:
```suggestion
    _logger.info("Initializing column custom_rate on table account_move")
```

luisg123v (changes requested):

rolandojduartem:
@luisg123v, ready!, please, may you review it?

account_invoice_change_currency/hooks.py
```diff
@@ -0,0 +1,29 @@
...
+    Initializing column custom_rate on table account_move
+    for the improvement in performance to avoid long
+    duration in databases with thousand of moves
+    """
+    _logger.info("Initializing column custom_rate on table account_move")
+    tools.create_column(
+        cr=cr,
+        tablename="account_move",
+        columnname="custom_rate",
+        columntype=fields.Float.type,
```
luisg123v:
```suggestion
        columntype=fields.Float.column_type[1],,
```

To follow the same way as Odoo [does to create columns](https://github.com/odoo/odoo/blob/da6bd90eed55186e00e330f390c4d0d0b351167c/odoo/fields.py#L841)

luisg123v (changes requested):
CC @moylop260

luisg123v:
Title: Speed up module installation

flotho (changes requested):
same request as @luisg123v otherwise, LGTM

account_invoice_change_currency/hooks.py
rolandojduartem: Take a look what I did, this direct solution was not working for me.

rolandojduartem:
Hi, @luisg123v and @flotho, may you review this PR?

cc @moylop260

account_invoice_change_currency/hooks.py
```diff
@@ -0,0 +1,29 @@
...
+    Initializing column custom_rate on table account_move
+    for the improvement in performance to avoid long
+    duration in databases with thousand of moves
+    """
+    _logger.info("Initializing column custom_rate on table account_move")
+    tools.create_column(
+        cr=cr,
+        tablename="account_move",
+        columnname="custom_rate",
+        columntype=fields.Float().column_type[1],
```
luisg123v: Why parenthesis?

This is creating a new field, this is not what I asked

luisg123v:
@rolandojduartem remember to edit also PR title

account_invoice_change_currency/hooks.py
rolandojduartem: Well, without parenthesis gives you "property object is not subscriptable"

luisg123v: I see. That's because float fields [don't use always the same value](https://github.com/odoo/odoo/blob/a66bed194cfd1adf8dd4030448fc5ccd9dc5d3c1/odoo/fields.py#L1313), because it depends on whether digits is defined.

Since we know in advance the field has digits, let's use `numeric`

rolandojduartem: Yes, but also I tried adding digits as the custom_rate field, but I got the same result without it 'double precision' and odoo says this is faster for most operations https://github.com/odoo/odoo/blob/94f74aed2b9986f56569b2d4e56e363bf0c2592a/odoo/fields.py#L1307. Is it a good idea to write 'numeric'?

rolandojduartem: @luisg123v Done!

rolandojduartem:
@luisg123v, please, may you review it?

cc @moylop260

luisg123v (approved):
LGTM :+1:

moylop260 (approved):
👍

luisg123v:
@rolandojduartem could you forward-port this to v15, please?

rolandojduartem:
@luisg123v Done!

PR #1175, opened 2022-06-13
[14.0][FIX] account_invoice_line_description: unbalanced invoice error in test

tafaRU:
:exclamation: This PR greenifies branch `14.0`.

PR #1183, opened 2022-06-18
[13.0][IMP] stock_picking_return_refund_option: Make tests resilient

pedrobaeza:
The default selected warehouse in the SO may not be the one we expect, so let's fix it to be sure.

@Tecnativa


pedrobaeza:
I have cherry-picked #1175 here, as this branch suffers the same problem.

pedrobaeza:
Merging as the CI is broken due to the `account_invoice_line_description` problem and as this test refactoring is only to assure resilience:

/ocabot merge patch

PR #1184, opened 2022-06-19
[13.0][IMP] stock_picking_return_refund_option: More resilience in tests

pedrobaeza:
There's still some cases that makes the pickings to not be completed, so we force them and also switch to Form for simplifying code.

@Tecnativa


rvalyi (approved):

pedrobaeza:
And working on the integration tests!

/ocabot merge patch

PR #1196, opened 2022-07-03
[13.0][FIX] *: Change _recompute_tax_lines args according upstream

pedrobaeza:
https://github.com/odoo/odoo/commit/11e7486157692bbbe0c605698ecfb4961f68 has changed the number of arguments on the method `_recompute_tax_lines`, so we need to adapt the overrides to such change.

@Tecnativa

PR #1197, opened 2022-07-03
[14.0][FIX] *: Change _recompute_tax_lines args according upstream

pedrobaeza:
https://github.com/odoo/odoo/commit/c9d5f069739d897145f843ee93b7eefd1d39 has changed the number of arguments on the method `_recompute_tax_lines`, so we need to adapt the overrides to such change.

Twin for 14.0 of #1196

@Tecnativa


pedrobaeza:
Fast-tracking this one as well:

/ocabot merge patch

PR #1203, opened 2022-07-13
[13.0][FIX] account_global_discount: leave amount_currency empty

ernestotejeda:
Cc @Tecnativa TT37931

### DIFF
--- a/account_global_discount/models/account_move.py
+++ b/account_global_discount/models/account_move.py
@@ -182,6 +182,9 @@ class AccountMove(models.Model):
         for discount in self.invoice_global_discount_ids.filtered("discount"):
             sign = -1 if self.type in {"in_invoice", "out_refund"} else 1
             disc_amount = sign * discount.discount_amount
+            amount_currency = 0.0
+            if self.currency_id != self.company_id.currency_id:
+                amount_currency = disc_amount
             create_method(
                 {
                     "global_discount_item": True,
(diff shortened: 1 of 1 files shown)


pedrobaeza:
@omar7r your previous patch is causing problems when there's no currency involved. Can you please check for a solution along my colleague?

omar7r:
Hi,

I think in 14.0 all lines have amount_currency but in 13.0 not, only when currency is set. You must check if currency is set and write amount_currency only in this case for 13.0.

ernestotejeda:
ping @pedrobaeza @omar7r

omar7r (approved):
LGTM

pedrobaeza (approved):

pedrobaeza:
@ernestotejeda please fw-port it to v14

ernestotejeda:
> @ernestotejeda please fw-port it to v14

In v14 it is not necessary because as @omar7r said in v14 all lines have amount_currency

pedrobaeza:
OK, right, I didn't remember.

pedrobaeza:
But the sign part is correct?

https://github.com/OCA/account-invoicing/blob/044998afabdfc5816fb76e03050fa156f5c63a82/account_global_discount/models/account_move.py#L197

ernestotejeda:
> But the sign part is correct?
> 
> https://github.com/OCA/account-invoicing/blob/044998afabdfc5816fb76e03050fa156f5c63a82/account_global_discount/models/account_move.py#L197

It's correct

pedrobaeza:
Why then such difference here in 13 (no sign correction) against v14 (sign correction)?

ernestotejeda:
What happens is that in that line of code in v14 there is not exactly a sign correction:
`"amount_currency": (disc_amount > 0.0 and disc_amount or 0.0) - (disc_amount < 0.0 and -disc_amount or 0.0),`
It is equivalent to:
`"amount_currency": disc_amount`

I said before, that it is correct because the code is not wrong even though it can be simplified, but there is really no sign correction.

the sign correction is done a few lines above:
https://github.com/OCA/account-invoicing/blob/044998afabdfc5816fb76e03050fa156f5c63a82/account_global_discount/models/account_move.py#L186

pedrobaeza:
OK, it seems a bit awkward, so we can maybe propose the simplification for not being confusing.

PR #1208, opened 2022-07-26
[13.0][ADD] sale_invoicing_date_selection: New module to set invoice_date when user creates invoices from wizard

sergio-teruel:
cc @Tecnativa

ping @carlosdauden @CarlosRoca13


carlosdauden (approved):

CarlosRoca13 (approved):
Functional and code review

sale_invoicing_date_selection/tests/test_sale_invoicing_date_selection.py
```diff
@@ -0,0 +1,41 @@
...
+            self.env["sale.advance.payment.inv"].with_context(
+                active_model="sale.order", active_ids=orders.ids, open_invoices=True
+            )
+        ) as wiz_invoice_form:
+            wiz_invoice_form.invoice_date = fields.Date.today()
+        wiz = wiz_invoice_form.save()
+        action = wiz.create_invoices()
+        invoices = self.env["account.move"].search(action["domain"])
+        self.assertTrue(invoices[0].invoice_date)
+        self.assertTrue(invoices[1].invoice_date)
```
pedrobaeza: Put here a fixed date and check such date on the test:

```suggestion
            wiz_invoice_form.invoice_date = "2022-07-01"
        wiz = wiz_invoice_form.save()
        action = wiz.create_invoices()
        invoices = self.env["account.move"].search(action["domain"])
        self.assertEqual(invoices[0].invoice_date, fields.Date.from_string("2022-07-01"))
        self.assertEqual(invoices[1].invoice_date, fields.Date.from_string("2022-07-01"))
```

sale_invoicing_date_selection/wizard/sale_advance_payment_inv.py
```diff
@@ -0,0 +1,19 @@
...
+
+    invoice_date = fields.Date(string="Invoice Date")
+
+    def create_invoices(self):
+        """
+        Add date to crate all invoices into user context
+        """
+        return super(
+            SaleAdvancePaymentInv,
+            self.with_context(default_invoice_date=self.invoice_date),
```
pedrobaeza: Isn't there a problem with the accounting date if we force this?

pedrobaeza (changes requested):

sale_invoicing_date_selection/tests/test_sale_invoicing_date_selection.py
sergio-teruel: This changes are not needed!!

pedrobaeza: You are not really testing the real thing. Asserting if there is a date is not the same as asserting that the date you have put is the one that is in the invoice.

sale_invoicing_date_selection/wizard/sale_advance_payment_inv.py
sergio-teruel: Chage done!! Now the same behavior is maintained in the same way that I set the manual date in the invoice

sale_invoicing_date_selection/tests/test_sale_invoicing_date_selection.py
sergio-teruel: Yes.. I know!!
But only need that is not empty

pedrobaeza: But in next versions or with any patch a default date is filled and this mechanism won't work, and tests won't detect the problem being so relaxed in this condition. Why not doing what I'm talking about?

sergio-teruel: Changes done

pedrobaeza:
You need to use `fields.Date.from_string`, or tests won't be correct.

sergio-teruel:
@pedrobaeza Done!

pedrobaeza (approved):

PR #1211, opened 2022-08-09
[13.0][FIX] account_move_tier_validation: fix _get_to_validate_message_name

JordiBForgeFlow:
cc @LoisRForgeFlow

PR #1134, opened 2022-04-10
[13.0][IMP][account_move_tier_validation] Use field messages for review, validation, rejection

JordiBForgeFlow:
Dependencies:
* [x] Review and merge https://github.com/OCA/server-ux/pull/484

PR #1212, opened 2022-08-10
[14.0][IMP] account_move_tier_validation: Use field messages for review, validation, rejection

LoisRForgeFlow:
Forward port of #1134 and #1211.

@ForgeFlow 

cc @JordiBForgeFlow

### DIFF
--- a/account_move_tier_validation/models/account_move.py
+++ b/account_move_tier_validation/models/account_move.py
@@ -1,7 +1,7 @@
 # Copyright <2020> PESOL <info@pesol.es>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
 
-from odoo import models
+from odoo import _, models
 
 
 class AccountMove(models.Model):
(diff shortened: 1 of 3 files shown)


AaronHForgeFlow (approved):

account_move_tier_validation/models/account_move.py
```diff
@@ -1,11 +1,23 @@
...
+        name = super(AccountMove, self)._get_to_validate_message_name()
+        if self.move_type == "in_invoice":
+            name = _("Bill")
+        elif self.move_type == "in_refund":
+            name = _("Refund")
+        elif self.move_type == "out_invoice":
+            name = _("Invoice")
+        elif self.move_type == "out_refund":
+            name = _("Credit Note")
+        return name
```
MiquelRForgeFlow, line 23: what if `self.move_type == "entry"`?

JordiBForgeFlow (approved):

account_move_tier_validation/models/account_move.py
LoisRForgeFlow: default behavior: _description = "Journal Entry"

PR #1219, opened 2022-08-18
[14.0] [IMP] account_invoice_date_due: enable multi-line edit on list view

gdgellatly:
Before this commit, this module would only work on a single record.

We enable the Odoo multi edit functionality to update the due date
of multiple invoices at once.

### DIFF
--- a/account_invoice_date_due/models/account_move.py
+++ b/account_invoice_date_due/models/account_move.py
@@ -37,10 +37,13 @@ class AccountMove(models.Model):
         res = super().write(vals)
         # Propagate due date to move lines
         # that correspont to the receivable/payable account
-        if "invoice_date_due" in vals and self.state == "posted":
-            payment_term_lines = self.line_ids.filtered(
+        if "invoice_date_due" in vals:
+            payment_term_lines = self.filtered(
+                lambda move: move.state == "posted"
+            ).line_ids.filtered(
                 lambda line: line.account_id.user_type_id.type
                 in ("receivable", "payable")
             )
-            payment_term_lines.write({"date_maturity": vals["invoice_date_due"]})
+            if payment_term_lines:
+                payment_term_lines.write({"date_maturity": vals["invoice_date_due"]})
         return res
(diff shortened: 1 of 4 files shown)


account_invoice_date_due/models/account_move.py
```diff
@@ -14,8 +14,9 @@ class AccountMove(models.Model):
     @api.onchange("invoice_date_due_payment_term")
     def _onchange_invoice_date_due_payment_term(self):
         """Propagate from Payment term due date to original field"""
-        if self.invoice_date_due_payment_term:
-            self.invoice_date_due = self.invoice_date_due_payment_term
+        for move in self:
```
pedrobaeza: onchange is by design single-record, so this change is not needed.

pedrobaeza (approved):

account_invoice_date_due/models/account_move.py
gdgellatly: @pedrobaeza it is only explicitly api.onchange. it is also implicitly api.multi and callable outside of an onchange context. 

EDIT: I should add, I don't care much either way for this one. It is just habit for me, onchanges should always iterate records because they work on recordsets. I actually think the extra field is not needed to achieve the modules goals and all this onchange does is get around some view restrictions.

pedrobaeza: All methods are by default `api.multi`, but as convention, all the onchanges are considered single-record. See all of them in across Odoo core. Anyways, this is not blocking.

luisg123v: I agree iterating is not required, it could give the wrong impression it's actually necessary and adds an extra indentation level

luisg123v (changes requested):
Please reword commit's title to follow naming guidelines, e.g.

`[IMP] account_invoice_date_due: enable multi-line edit on list view`

This to fix:
- Missing change type, in this case an improvement
- Version number is not included
- After module name it should be a ":"
- And improved a little short description to be more descriptive

PR #1225, opened 2022-08-24
[FIX][14.0][account_invoice_fixed_discount]: bad tax computation

### LINKED ISSUE #1107: [13.0][14.0] account_invoice_fixed_discount wrong tax
Hi,
I discovered a problem when creating invoices out of a sales order with a fixed discount. The tax amount on the created invoice doesn't get calculated correctly. This behavior occurs when also using the module `sale-workflow/sale_fixed_discount`.

Reproduce:

1. Create a sales order with:
Product with €390 unit price, €100 discount = €290 line total. Tax with 19% results in €55.10
2. Confirm the sales order and create a regular invoice by clicking the "Create Invoice" Button in the sales order
3. Invoice results in €390 unit price, €100 discount, €290 net total but tax 19% results in €74.10 (€390 * 0.19) so the fixed discount is disregarded.

I fixed the behavior by adding the following code to `account.move` but I don't know if its the best way.

```
@api.model
def create(self, vals):
    res = super(SaleOrder, self).create(vals)
    res.with_context(check_move_validity=False)._recompute_tax_lines()
    res.with_context(check_move_validity=False)._onchange_invoice_line_ids()
    return res
```

flotho:
FIX #1107 for v14.0

Possibly same issue for triple discount

### DIFF
--- a/account_invoice_fixed_discount/models/account_move.py
+++ b/account_invoice_fixed_discount/models/account_move.py
@@ -23,7 +23,7 @@ class AccountMove(models.Model):
         for line in vals.keys():
             line.update(vals[line])
         return res
-    
+
     @api.model
     def create(self, vals):
         res = super().create(vals)
--- a/account_invoice_fixed_discount/models/account_move.py
+++ b/account_invoice_fixed_discount/models/account_move.py
@@ -1,8 +1,13 @@
 # Copyright 2017 ForgeFlow S.L.
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
 
+import logging
+
 from odoo import _, api, fields, models
 from odoo.exceptions import ValidationError
+from odoo.tools import float_compare, float_is_zero
+
+_logger = logging.getLogger(__name__)
 
 
 class AccountMove(models.Model):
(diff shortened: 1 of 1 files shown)


flotho:
ping @iggups , any chance to have a review here ?

iggups:
@flotho looks good to me 👍

gaelTorrecillas (approved):

flotho:
hi @iggups you anly make a comment, any chance to make a positive review

iggups (approved):
Looks good!

flotho:
ping @OCA/accounting-maintainers ansy chance to be merged ?

adrienpeiffer:
Could you include a regression test showing the problem that this patch is solving and squash your commits ?

account_invoice_fixed_discount/models/account_move.py
```diff
@@ -71,7 +79,11 @@ def _get_price_total_and_subtotal_model(
         move_type,
     ):
         if self.discount_fixed != 0:
-            discount = ((self.discount_fixed) / price_unit) * 100 or 0.00
+            if price_unit >0.00:
```
grindtildeath: You should use float tools functions such as `float_compare` or `float_is_zero` whenever comparing float values.

flotho: thanks @grindtildeath 
@gaelTorrecillas could you please correct this.
thanks

flotho:
@grindtildeath , thanks for your comments, I've added the proper syntax
@adrienpeiffer I would love to add tests. What I need to test is when it comes from other part of the module Odoo, I mean I need test for data coming from sale.order and pos.order . How is it possible to make the module depnds on other only in test /demo mode ?

rafaelbn:
Quite interesting, don't miss the v16 migration https://github.com/OCA/account-invoicing/pull/1488 

#1488

PR #1226, opened 2022-08-26
[10.0][BACKPORT] account_invoice_tax_note from 11.0 version

fd-oncodna:
Works "as is"


fd-oncodna:
Is it possible to merge this even if Travis fails? I have no way to fix this issue on my end.

```
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-install-0XIsYv/lazy-object-proxy/
The command "travis_install_nightly" failed and exited with 1 during .
```

PR #1231, opened 2022-09-06
[15.0][MIG] account_portal_hide_invoice

Freni-OSI:


adrienpeiffer (approved):
LGTM (Code review)

rafaelbn:
Thank you! 😄 . Just for you curiosity, we get same effect with https://github.com/OCA/sale-workflow/tree/14.0/portal_sale_personal_data_only (from v10 5 years ago) because if the portal user haven't received the invoice they cannot see them but for example, accountants can see the invoices in the portal because the received. Just in the case you don't want to hide the invoices always for everyone!

Invoices: https://github.com/OCA/sale-workflow/blob/14.0/portal_sale_personal_data_only/security/security.xml#L19

This is due the Odoo decision of showing all invoices to all users of the company ...

This module is great to block totally for this use case and I only would like to give you visibility in case of 🤟 

Thank you for this module and migration!

Best regard!

dreispt (approved):

PR #1238, opened 2022-09-14
[13.0][FIX] account_invoice_triple_discount: Calculate taxes correcly consi…

ao-landoo:
…dering all discounts

When creating an invoice from a sale order and using different discounts (sale_triple_discount, still not merged but working on PR https://github.com/OCA/sale-workflow/pull/2028) invoice's taxes were not calculated correctly.

Example:

> Created a sale order where the product unit price is 60. Added discount of 20% and discount 2 of 10%. Taxes are 15%.
> 60 * 0.8 * 0.1 * 0.15 = 6.48 -> This is OK
> 
> Created an invoice from the sale order. The taxes on the invoice are 7.2 instead of 6.48.
> This is because it's not considering discount2 and discount3 field values for the tax calculation
> 60 * 0.8 * 0.15 = 7.2 -> This is WRONG

After I saw it's working correctly on version 14.0, I looked at it's code and applied the differences. Now it's working fine.

@igallart Could you check it?

### DIFF
--- a/account_invoice_triple_discount/models/account_move.py
+++ b/account_invoice_triple_discount/models/account_move.py
@@ -13,23 +13,26 @@ class AccountMove(models.Model):
 
     _inherit = "account.move"
 
-    def _recompute_tax_lines(
-        self, recompute_tax_base_amount=False, tax_rep_lines_to_recompute=None
-    ):
-        vals = {}
-        for line in self.invoice_line_ids.filtered(
-            lambda l: l.discount2 or l.discount3
-        ):
-            vals[line] = {"price_unit": line.price_unit, "discount": line.discount}
+    def _recompute_tax_lines(self, **kwargs):
+        """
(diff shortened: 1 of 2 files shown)


ao-landoo:
@pedrobaeza Could you review this PR? The module is not working OK

pedrobaeza:
Please include a regression test showcasing the problem that this patch is solving.

bi-landoo (approved):

ao-landoo:
> Please include a regression test showcasing the problem that this patch is solving.

Added a test which is not passed without the changes in this commit. There also is a PR in sale-workflow repo (https://github.com/OCA/sale-workflow/pull/2028) which is not passing tests because of this bug.

adrienpeiffer (approved):
So it's seems the issue comme from the usage of invoice_line_ids instead of line_ids

PR #1249, opened 2022-09-29
[14.0][ADD] account_credit_note_keep_invoiced_quantity

StefanRijnhart:
### Keep invoiced quantities while creating a credit note

When creating credit notes for sale or purchase invoices, Odoo will reset the invoiced quantities on the order lines and make the orders become eligible for invoicing again. This module adds a checkbox in the credit note (or 'reversal') wizard to keep the invoiced quantities on the related order lines.

image


celm1990:
Hi, I see wiith this module, relations wiith sale order and new credit note is missing.

I use module https://github.com/OCA/sale-workflow/tree/14.0/sale_force_invoiced for this purpose, from credit note wizard mark sale order as force_invoiced, this not solve your problem?

StefanRijnhart:
@celm1990 no, as our primary use case is for purchase orders. Also in the case of sales orders, I'd consider the approach from this PR more precise as it can be perfectly clear what the invoiced qty needs to be and how it can be achieved. In that way it is safer than sale_force_invoiced (which I was happy to use in the past to deal with orders that came out of a migration in an inconsistent invoice status) as it can be the case that we will want to create an additional invoice in more complex invoicing scenarios.

StefanRijnhart:
But feel free to give this PR a thumbs-down :-1: . If nobody finds this useful, then I'll gladly move it back to my custom addons for this project.

account_credit_note_keep_invoiced_quantity/models/account_move.py
```diff
@@ -0,0 +1,26 @@
...
+        This is used to allow the creation of credit notes without resetting the
+        invoiced quantities on sales or purchases.
+        """
+        avoid_include_business_fields = self.env.context.get(
+            "avoid_include_business_fields"
+        ) and self.env.context.get("include_business_fields")
+        if avoid_include_business_fields:
+            self = self.with_context(include_business_fields=False)
+        res = super().copy_data(default=default)
+        if avoid_include_business_fields and "purchase_line_id" in self._fields:
```
legalsylvain: Hi. maybe not a valid question.
you write "when creating credit notes for sale or purchase invoices" in the doc of the module, but I don't see something for the ``sale_line_id``.
the sale refund is handled by your module ? 
(or maybe I missed something).

StefanRijnhart: Good question! We don't have to worry about `sale_line_ids` because it has copy=False: https://github.com/odoo/odoo/blob/14.0/addons/sale/models/account_move.py#L34-L38. That is why they have to be copied in explicitely in the method that my PR avoids: https://github.com/odoo/odoo/blob/14.0/addons/sale/models/account_move.py#L40-L43

But `purchase_line_id` does not have copy=False and is therefore copied by default. There is also an override of the method to copy it explicitely if requested (https://github.com/odoo/odoo/blob/14.0/addons/purchase/models/account_invoice.py#L141-L144), but it is redundant as I found out when testing this. It also means that avoiding the method is not enough, and the value has to be purged explicitely in these lines of code here.

I consider it a bug in Odoo that `purchase_line_id' is not set to copy=False.

PieterPaulussen (approved):
LGTM

StefanRijnhart:
The module is now improved to include these 'dangling' credit notes in the sale and purchase's invoice_ids as accessible through the smart buttons on the order forms.

StefanRijnhart:
Closing due to obvious lack of interest.

PR #1251, opened 2022-10-04
[12.0][MIG] account_move_tier_validation: Migration from 14.0

marcelsavegnago:


marcelsavegnago:
ping @douglascstd @rvalyi

douglascstd (changes requested):
The model selection allows to select any model, it may cause an error. The suggestion is add a domain to restrict the selection to the **account.move** **model** only.
### Follow example from this PR:
image

This is the domain applied on v14 (source):
image

### The process tested:
- Then the account.move is created by an invoice, the tier validation is bypassed. **OK**
image

- Single Journal Entry: **OK**
image

image
image
image
image

sergiocorato:
@marcelsavegnago This branch is now upgraded to copier (see #1285 ), so please rebase to re-launch tests.

marcelsavegnago:
all green :)

marcelsavegnago:
> The model selection allows to select any model, it may cause an error. The suggestion is add a domain to restrict the selection to the **account.move** **model** only.

@douglascstd 

image

marcelsavegnago:
cc @kaynnan @WesleyOliveira98

PR #1265, opened 2022-10-17
[15.0][mig] account_invoice_blocking

JordiBForgeFlow:
Standard migration


account_invoice_blocking/models/account_move.py
```diff
@@ -0,0 +1,58 @@
+# Copyright 2016 Acsone SA/NV
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class AccountInvoice(models.Model):
```
PieterPaulussen: Perhaps a more suitable name would be 
```suggestion
class AccountMove(models.Model):
```

account_invoice_blocking/models/account_move.py
```diff
@@ -0,0 +1,58 @@
...
+        This method set the value of the field 'blocked' to True
+        If every line of the move is actually blocked
+        """
+        for move in self:
+            move_lines = move._get_move_line()
+            move.blocked = (
+                all(line.blocked for line in move_lines) if move_lines else False
+            )
+
+    blocked = fields.Boolean(
```
PieterPaulussen: I believe this field is best defined right below the models' private attributes like `_inherit` instead of at the bottom of the file.

account_invoice_blocking/views/account_move.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2016 Acsone SA/NV
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<odoo>
+    <record id="account_invoice_form_view" model="ir.ui.view">
```
PieterPaulussen:
```suggestion
    <record id="view_move_form" model="ir.ui.view">
```

PieterPaulussen (changes requested):
A few conventions that can be easily applied, otherwise this looks good to me.

adrienpeiffer (approved):

JordiBForgeFlow:
@PieterPaulussen thanks for your feedback. I have applied your suggestions.

PieterPaulussen (approved):

PR #1266, opened 2022-10-17
[14.0][MIG] purchase_stock_picking_return_invoicing_force_invoiced

eantones:
Standard migration


pedrobaeza:
In theory this one shouldn't be needed in newer versions.

eantones:
In theory, but my customer and I I tested it and without it the module purchase_force_invoiced does not work when the module purchase_stock_picking_return_invoicing is installed.

pedrobaeza:
Please expose them the failing use case and it would be interesting to have it as test in the module. The changes you have done in the tests doesn't seem very correct, changing fixed amounts...

eantones:
Why the tests doesn't seem very correct? They are green
Regarding changing amounts, the only amount I changed is the 50 and I've only changed the sign.

pedrobaeza:
Yeah, I mean that one, but I think this change was done by Odoo. And the steps to reproduce the problem without the module?

eantones:
The code is simple and it's easy to see that the code of "purchase_stock_picking_return_invoicing" is overwritting the values updated by "purchase_force_invoiced". I've attached a quick video to see the behavior on all cases.


https://user-images.githubusercontent.com/10927087/196281421-1b60e3da-c059-41a2-a179-8bb5ca5ffefa.mp4

eantones:
Superseded by https://github.com/nuobit/account-invoicing/pull/1

Issue #1250, opened 2022-10-03
Migration to version 16.0


sbidoul:
# Todo

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0

# Modules to migrate

- [x] account_global_discount - By @pedrobaeza - #1724
- [x] account_invoice_alternate_payer - By @ChrisOForgeFlow - #1492
- [x] account_invoice_blocking - By @AnizR - #1258
- [x] account_invoice_change_currency - By @ljsalvatierra-factorlibre - #1261
- [x] account_invoice_check_picking_date - By @Shide - #1947
- [x] account_invoice_check_total - By @AnizR - #1253
- [x] account_invoice_crm_tag - By @ernesto-garcia-tecnativa - #1572
- [x] account_invoice_customer_no_autofollow - By @sbidoul - #1781
- [x] account_invoice_date_due - By @smussie - #1447
- [x] account_invoice_default_code_column - By @legalsylvain - #1654
- [x] account_invoice_discount_date - By @Shide - #1886
- [x] account_invoice_discount_display_amount - By @samoros1 - #1623
- [x] account_invoice_fiscal_position_update - By @ljsalvatierra-factorlibre - #1263
- [x] account_invoice_fixed_discount - By @PieterPaulussen - #1488
- [x] account_invoice_force_number - By @heliaktiv - #1441
- [ ] account_invoice_mass_sending - By @BT-anieto - #1529
- [x] account_invoice_merge - By @nguyenminhchien - #1400
- [ ] account_invoice_mode_weekly - By @rousseldenis - #1514
- [x] account_invoice_payment_retention - By @Saran440 - #1597
- [x] account_invoice_payment_term_date_due - By @BT-anieto - #1607
- [ ] account_invoice_pricelist - By @hildickethan-S73 - #1337
- [x] account_invoice_pricelist_sale - By @nguyenminhchien - #1415
- [x] account_invoice_refund_line_selection - By @RodrigoBM - #1466
- [x] account_invoice_refund_link - By @ljsalvatierra-factorlibre - #1264
- [x] account_invoice_refund_reason - By @BT-anieto - #1612
- [ ] account_invoice_search_by_reference - By @dsolanki-initos - #1455
- [x] account_invoice_section_sale_order - By @rven - #1451
- [x] account_invoice_show_currency_rate - By @ACheung-FactorLibre - #1522
- [ ] account_invoice_supplier_ref_unique
[...]

Issue #1240, opened 2022-09-17
Account Invoice Triple Discount does not compute correctly price_subtotal


cscocca:
The module does not compute correctly the `price_subtotal` field because the price rounding feature.

## Module
account_invoice_triple_discount

## Describe the bug
When the invoice has a price and a discount (just one discount can reproduce the issue) that produce a discounted price with more decimals than price rounding setting so the `price_subtotal` is wrong.

## To Reproduce
An example is:
qty           price            discount    discount2      discount3     subtotal
9950        0.139              15%            0                   0               1176.09

but the right subtotal is **1175.59** because 0.139 minus 15% is 0.11815 * 9950 = 1175.5925 -> Rounded = 1175.59.

**Affected versions**:

- [x] 12.0

legalsylvain:
Could you add in a PR a test that is failing ?
Thanks !

cscocca:
Just added
https://github.com/OCA/account-invoicing/pull/1242

ramiadavid:
@legalsylvain @c-amara 

I have the same problem, even using only Odoo's default discount field and the problem is here
```
        for line in self.line_ids:
            aggregated_discount = line._compute_aggregated_discount(line.discount)
            old_values_by_line_id[line.id] = {
                "price_unit": line.price_unit,
                "discount": line.discount,
            }
            price_unit = line.price_unit * (1 - aggregated_discount / 100)
            line.update({"price_unit": price_unit, "discount": 0})**
```
To calculate the taxes here the unit price with the aggregated discounts is computed, but when saving the value in the line Odoo rounds to the decimals by default and this produces an inaccuracy when calculating the taxes and the base especially when the line contains many units.

PR #1269, opened 2022-10-19
[16.0][MIG] account_invoice_triple_discount: Migration to 16.0

ramiadavid:
Migration to v16

#1250


legalsylvain:
Hi. Thanks for porting this module. 
I have a functional question. In my company I receive many supplier invoices with Discount 1 & 2. (Odoo 12.)
Some computation are good (I mean Odoo computation = supplier computation), but some computation are bad. That is due to the fact that the supplier makes intermediate rounding.
Do you think, it could be possible to introduce some computation option (depending on the supplier ?)

If yes, I can try to implement that extra feature.

Note : I think that this module is very complex. We should try to talk to the Odoo SA dev team in the next odoo experience to try to have some hook to avoid all this code. Don't you think ?

Kind regard.

ramiadavid:
> Hi. Thanks for porting this module. I have a functional question. In my company I receive many supplier invoices with Discount 1 & 2. (Odoo 12.) Some computation are good (I mean Odoo computation = supplier computation), but some computation are bad. That is due to the fact that the supplier makes intermediate rounding. Do you think, it could be possible to introduce some computation option (depending on the supplier ?)
> 
> If yes, I can try to implement that extra feature.
> 
> Note : I think that this module is very complex. We should try to talk to the Odoo SA dev team in the next odoo experience to try to have some hook to avoid all this code. Don't you think ?
> 
> Kind regard.

Hi @legalsylvain
In this module there is a bug in the calculation of discounts, it is discussed here #1240 , and I have uploaded a PR with the fix, I am waiting for it to be merged

Does this solve your problem?

legalsylvain:
Hi @ramiadavid . thanks for your answer. it's différent topic. I talk about the fact that 2 results are valid depending on the supplier computation method.

FernandoRomera (approved):
Review, OK.

MRomeera (approved):
ok

AntoniRomera (approved):

sbejaoui (changes requested):
Hi, 

Functional test

The invoice total don't match the lines subtotal.

image

ramiadavid:
@sbejaoui @AntoniRomera @FernandoRomera @MRomeera 

This is fixed, but I see that the `discount_2` and `discount_3` fields are only shown if we add the 'Technical / Discount on lines' group, while the `discount` field is always available, likewise the discount field is set to optional="hide" while than the other two as optional="show" and in the `discount` field the name is abbreviated "Disc." while in the others it says "Discount"

Wouldn't it be better to unify these things to follow the same criteria as the original field?

sbejaoui:
> Wouldn't it be better to unify these things to follow the same criteria as the original field?

+1, 
Thank you

FernandoRomera:
@ramiadavid 
In account_invoice_triple_discount/report/invoice.xml, the names are:
                <span>Disc.2 %</span>
                <span>Disc.3 %</span>
without so many blanks and parentheses

>

ramiadavid:
> @ramiadavid In account_invoice_triple_discount/report/invoice.xml, the names are: Disc.2 % Disc.3 % without so many blanks and parentheses
> 
> >

You're right, I've already changed it.

account_invoice_triple_discount/views/account_move.xml
```diff
@@ -0,0 +1,24 @@
...
+    <record id="invoice_triple_discount_form_view" model="ir.ui.view">
+        <field name="name">account.invoice.triple.discount.form</field>
+        <field name="model">account.move</field>
+        <field name="inherit_id" ref="account.view_move_form" />
+        <field name="arch" type="xml">
+            <xpath
+                expr="//field[@name='invoice_line_ids']//tree//field[@name='discount']"
+                position="after"
+            >
+                <field name="discount2" optional="hide" string="Disc.2%" />
```
lmignon, line 12: @ramiadavid If you change the label, you need to update the translation files ..

lmignon (changes requested):
LGTM (Code review + functional) but the source term into the translation files must be updated with the new labels for the discount columns....

account_invoice_triple_discount/views/account_move.xml
rousseldenis: @lmignon This should be done once this is merged through weblate.

account_invoice_triple_discount/models/account_move_line.py
```diff
@@ -0,0 +1,150 @@
...
+            tmp_values = {}
+            discount_changed = (
+                float_compare(old_discount, new_discount, precision_digits=dp_discount)
+                != 0
+            )
+            if discount_changed:
+                values["discount"] = new_discount
+                tmp_values["discount"] = old_discount
+            old_values.append(tmp_values)
+        records = super(AccountMoveLine, self).create(values_list)
```
rousseldenis:
```suggestion
        records = super().create(values_list)
```

account_invoice_triple_discount/models/account_move_line.py
```diff
@@ -0,0 +1,150 @@
...
+        restored after the original process is done
+        """
+        old_values_by_line_id = {}
+        digits = self._fields["discount"]._digits
+        self._fields["discount"]._digits = (16, 16)
+        for line in self:
+            aggregated_discount = line._compute_aggregated_discount(line.discount)
+            old_values_by_line_id[line.id] = {"discount": line.discount}
+            line.update({"discount": aggregated_discount})
+        res = super(AccountMoveLine, self)._compute_totals()
```
rousseldenis:
```suggestion
        res = super()._compute_totals()
```

account_invoice_triple_discount/models/account_move_line.py
```diff
@@ -0,0 +1,150 @@
...
+        restored after the original process is done
+        """
+        digits = self._fields["discount"]._digits
+        self._fields["discount"]._digits = (16, 16)
+        old_values_by_line_id = {}
+        for line in self:
+            aggregated_discount = line._compute_aggregated_discount(line.discount)
+            old_values_by_line_id[line.id] = {"discount": line.discount}
+            line.update({"discount": aggregated_discount})
+        res = super(AccountMoveLine, self)._compute_all_tax()
```
rousseldenis:
```suggestion
        res = super()._compute_all_tax()
```

account_invoice_triple_discount/views/account_move.xml
lmignon: @rousseldenis In this case, the source term is changed. That doesn't mean that we want to loose all the translations of the previous source term. That's why the translation files must be updated also replace the previous label by the new label for the source term.

ramiadavid: @rousseldenis Updated translations for translated langs

lmignon (approved):
Thank you @ramiadavid for all this work. LGTM (Code review + functional)

rousseldenis (approved):

account_invoice_triple_discount/models/account_move_line.py
ramiadavid: done

account_invoice_triple_discount/models/account_move_line.py
ramiadavid: done

account_invoice_triple_discount/models/account_move_line.py
ramiadavid: done

lmignon:
@OCA/accounting-maintainers Someone to merge this one?

lmignon:
Thank you @adrienpeiffer for the merge!

PR #1277, opened 2022-11-04
[14.0][ADD] account_refund_payment_term:

petrus-v:
In Odoo when account user create refund directly (not from an invoice), sales payment terms (those used for invoicing) are used.

This module add new refund payment term field on partner to manage payment terms for refunds different to those used for invoicing.

This avoid to wrongly split generated account move lines with different due dates for refunds and properly configure payment terms messages on refund document (PDF).


petrus-v:
As far I understand pre-commit is running on python 3.11 env it seems that the pylint version we are using rely on [inspect.formatargspec](https://docs.python.org/3.10/library/inspect.html#inspect.formatargspec) which has beend deprecated from python version 3.5 and removed in v3.11 !

kemar (approved):

petrus-v:
To fix pre-commit I'll rebase once #1279 is merged

hparfr (approved):

etobella:
When I created a refunf from an invoice it was not setted (tested on runboat). Can you review that?

If I created a refund, it worked properly.

petrus-v:
> When I created a refunf from an invoice it was not setted (tested on runboat). Can you review that?

It's a fact that I haven't change this behaviour at the moment by this module.

In such case I'm not sure what's the expected behaviour. I've the feeling that it depends if the former invoice has been paid or not. The case I've in mind is that, if as odoo company I want to pay refunds at two different due date (not sure if such business exists) when creating refunds manually with this module I would set payment terms that manage properly my due date for refunds, but I probably don't want to create two account.move.line in my refund when I "cancel" an invoice by creating a refund from such invoice (this is at least the way in France to cancel invoice to keep invoice sequence without gap).

As today I can change behaviour by this module or at least be more explicit in the description, what do you prefer ?

etobella:
Well, right now, it is confusing, so at least I would set on the description that it only works on manually created refunds (not coming from an usual invoice)

petrus-v:
@etobella I've reworded a bit the module description !

etobella (approved):
/ocabot merge nobump

PR #1280, opened 2022-11-08
[14.0][ADD] new module account_invoice_mode_multi_monthday

zamberjo:
This module `account_invoice_mode_multi_monthday` creates monthly invoices for customer on a indicated (one or several) days of the month.  (configuration is in Accounting Settings)

image

This is useful for biweekly billing, for example.


pedrobaeza:
This is redundant with `contract` module, isn't it? And please don't use plurals in module name.

pedrobaeza:
OK, I see it follows the `account_invoice_base_invoicing_mode` module that is for sales, so it doesn't overlap, but please attend the module name thing. You can call it `account_invoice_mode_multi_monthday` or similar.

zamberjo:
`account_invoice_mode_multi_monthday` looks good

zamberjo:
@OCA/accounting-maintainers would appreciate someone checking this PR.

rafaelbn:
@zamberjo I canno test it because after creating several sale order and all ready to invoice I have to way at leat a week... Any idea?

imagen

anmarmo1:
Hi @rafaelbn 
I believe it is not necessary to wait for a week to test it; it just needs to match the billing date with the stock picking "date_done". The process is executed through a scheduled task that creates a "queue job" which is not running in the runboat. I'm not familiar with how the queue_job works in the runboat, whether it's a configuration issue or how the job is created.
@zamberjo Can you please check it?

PR #1283, opened 2022-11-11
[15.0][MIG] account_invoice_check_picking_date: Migration to 15.0

victoralmau:
Migration to 15.0

Please @carlosdauden and @sergio-teruel can you review it?

@Tecnativa TT36418


victoralmau:
I think it would be very interesting to add tests in this module, but it would be necessary to add `purchase_stock` dependency, what do you think about it (description module explains how to use with `purchase` module).

carlosdauden (approved):

sergio-teruel (approved):

PR #1290, opened 2022-11-21
[15.0] [MIG] account_invoice_supplier_self_invoice

Shide:
Migration to 15.0:

What I've done:
- All the steps in [Migration guidelines](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-15.0#technical-method-to-migrate-a-module-from-140-to-150-branch)
- New field `is_self_invoice_number_different` stored=False on `account.move` to check if the ref number is different than the self_invoice_number
- New Report based on the copied and modified from `external_layout_striped` template to maintain a "almost" the same external report structure than Odoo.
- Modified the `report_invoice_document` template to add the case of the self invoice.
- Removed the buttons for self invoice (you can print it vía **Print > Invoice or Invoice without Payments**) to maintain the same workflow than Odoo.
- A new email template (super similiar to `account.email_template_edi_supplier` to send better standard emails to suppliers.
- Added the **Send & Print** buttons to be able to send the self invoices vía email directly with 2 clicks (if all the selected invoices are self invoices, the email template is chosen correctly)
- Filter on Invoices to show only self invoices. (if you want to check what invoices has been sent, you need to create an additional filter with _is_move_sent = True_)
- Fix: Prevent when back to draft an self invoice generates a new number.
- Banner that shows when it's going to generate a new self invoice number.
- Banner that shows when the invoice ref doesn't match the self invoice number.
- Remove unnecesary test of `test_action_view_account_invoice_self`

MT-906 @moduon

@etobella @pedrobaeza @rafaelbn @yajo please feel free to review and suggest anyting.


rafaelbn:
Amazing work @Shide ! ❤️ Start review!

account_invoice_supplier_self_invoice/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+# © 2017 Creu Blanca
+# © 2022 Moduon
```
yajo: Please use this instead. The © character has no legal value.

```suggestion
# Copyright 2022 Moduon
```

account_invoice_supplier_self_invoice/models/res_partner.py
```diff
@@ -0,0 +1,46 @@
...
+from odoo import fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    self_invoice = fields.Boolean(
+        string="Approves self invoice",
+        default=False,
+        help="When checked, all invoices will generate a self-invoice " "on validation",
```
yajo: If I'm not wrong, this is not for all invoices. It's for vendor bills only, right?

```suggestion
        help="When checked, all vendor bills will generate by default a self-invoice on confirmation.",
```

account_invoice_supplier_self_invoice/models/res_partner.py
```diff
@@ -0,0 +1,46 @@
...
+    )
```
yajo: Don't you think it would be good to have this field tracked? So we can know since when the supplier is using self-invoices.

account_invoice_supplier_self_invoice/models/account_move.py
```diff
@@ -0,0 +1,87 @@
...
+
+    self_invoice_number = fields.Char(readonly=True, copy=False)
+    is_self_invoice_number_different = fields.Boolean(
+        string="Different invoice reference and self invoice number",
+        compute="_compute_is_self_invoice_number_different",
+        help="Self invoice number is different than invoice reference",
+    )
+    set_self_invoice = fields.Boolean(
+        help="Set self invoice for this invoice.",
+    )
```
yajo: I think the field name and help aren't the best. One idea:

```suggestion
    self_invoice = fields.Boolean(
        help="If enabled, create a self-invoice when validating.",
    )
```

account_invoice_supplier_self_invoice/models/account_move.py
```diff
@@ -0,0 +1,87 @@
...
+    can_self_invoice = fields.Boolean(related="partner_id.self_invoice")
+
+    def _compute_is_self_invoice_number_different(self):
+        self.update({"is_self_invoice_number_different": False})
```
yajo: This is Odoo 15, so you can do this instead:

```suggestion
        self.is_self_invoice_number_different = False
```

It will act as expected in multi-records.

account_invoice_supplier_self_invoice/models/account_move.py
```diff
@@ -0,0 +1,87 @@
...
+        for move in self:
+            if move.move_type not in ("in_invoice", "in_refund") or not (
```
yajo: Remember to always use sets for `in` or `not in` operators, if possible. This is a 2-element tuple, so the performance benefit if you accept the suggestion will be ridiculous. But my inner nerd forces me to add the suggestion, sorry 😆

```suggestion
            if move.move_type not in {"in_invoice", "in_refund"} or not (
```

account_invoice_supplier_self_invoice/models/account_move.py
```diff
@@ -0,0 +1,87 @@
...
+                move.is_self_invoice_number_different = True
+
+    @api.onchange("partner_id", "company_id")
+    def _onchange_partner_id(self):
+        res = super()._onchange_partner_id()
+        self.set_self_invoice = self.partner_id.self_invoice
+        return res
+
+    def action_post(self):
+        in_move_types = ("in_invoice", "in_refund")
```
yajo: 🤓 yet again

```suggestion
        in_move_types = {"in_invoice", "in_refund"}
```

account_invoice_supplier_self_invoice/models/res_partner.py
```diff
@@ -0,0 +1,46 @@
...
+    self_invoice_sequence_id = fields.Many2one(
+        comodel_name="ir.sequence", string="Self invoice sequence", ondelete="restrict"
+    )
+    self_invoice_report_footer = fields.Text(
+        string="Self invoice footer",
+        help="Footer text displayed at the bottom of the self invoice reports.",
+    )
```
yajo: Shouldn't all these fields be company-dependant?

account_invoice_supplier_self_invoice/models/account_move.py
```diff
@@ -0,0 +1,87 @@
...
+            for onchange_method, changed_fields in onchanges.items():
+                if any(f not in vals for f in changed_fields):
+                    invoice = self.new(vals)
+                    getattr(invoice, onchange_method)()
+                    for field in changed_fields:
+                        if field not in vals and invoice[field]:
+                            vals[field] = invoice._fields[field].convert_to_write(
+                                invoice[field], invoice
+                            )
+        return super().create(vals)
```
yajo: This code is... weird. Also I'd say it won't perform very well. You can replace it with a simple compute:

```suggestion
    @api.depends("partner_id", "company_id")
    def _compute_self_invoice(self):
        """Inherit default for self-invoice from partner."""
        for move in self:
            self.set_self_invoice = self.partner_id.self_invoice
```

Make sure you make the field computed, stored and writable. Also remove the `_onchange_partner_id()` override above.

account_invoice_supplier_self_invoice/models/account_move.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def _get_mail_template(self):
+        in_move_types = ("in_invoice", "in_refund")
```
yajo: 🤓 🤓 

```suggestion
        in_move_types = {"in_invoice", "in_refund"}
```

account_invoice_supplier_self_invoice/models/res_partner.py
```diff
@@ -0,0 +1,46 @@
...
+                            "padding": 4,
+                            "prefix": record._self_invoice_sequence_prefix(),
+                            "use_date_range": True,
+                            "number_next": 1,
+                        }
+                    )
+                )
+
+    def _self_invoice_sequence_prefix(self):
+        return "SI/INV/%(range_year)s/"
```
yajo: Would it make sense to make this prefix translatable?

```suggestion
        return _("SI/INV/%(range_year)s/")
```

yajo (changes requested):
For code review, see comments below. Some functional review here too.

Self-invoice features in partner form have a broken layout:

imagen

When printing the invoice, the vendor appears nowhere:

imagen

account_invoice_supplier_self_invoice/models/res_partner.py
etobella: Not really...

If we keep several sequences according to each company, we need to show different numbers and that might be hard to manage (the creation of the sequence might be a problem....). IMO it is easier to make it only once to simplifie the process.

etobella (commented):
Amazing job @Shide The changes are really good. Just one thing. Might be possible to split the commit in two? The migration and the improvements? This way we can port the commit to 14. 

Also found the same issue with the report that @yajo reported.

Shide:
I'll try to do my best (it will take a bit long time).

Printing the report is failing because external report layout is not configured on company (i'll fix it too).

Shide:
@etobella It's very hard to split the PR in two commits. 

The main reason of the improvement is to remove the old report.
So in the end, the old report will be gone. Making a separate commit fixing the old report and then remove in the next commit it's not viable for me because it's an extra work to fix something that will be out in the same PR.

I'll review and fix all the things that  @yajo said, but in this case, I'm not going to split he PR in 2 commits.

I'll take your suggestions seriously for the next migrations that I do, I promise :)

account_invoice_supplier_self_invoice/models/res_partner.py
Shide: No, because it's a sequence, no need to translate.

account_invoice_supplier_self_invoice/README.rst
```diff
@@ -0,0 +1,93 @@
...
+* Creu Blanca
+* Moduon
+
+Contributors
+~~~~~~~~~~~~
+
+* Enric Tobella <etobella@creublanca.es>
+* Nikul Chaudhary <nikulchaudhary2112@gmail.com>
+* Alba Riera <alba.riera@creublanca.es>
+* Eduardo de Miguel (edu@moduon.team)
```
rafaelbn: The time spent a pre-testing, specification with improvement, validations plus additional testting now, I think make me a contributor 😄

account_invoice_supplier_self_invoice/models/res_partner.py
```diff
@@ -0,0 +1,46 @@
+# © 2017 Creu Blanca
+# License AGPL-3.0 or later (https://www.gnuorg/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
```
rafaelbn, line 8: When duplicating a res.partner all fields added should be empty always (not copy)

- self_invoice
- self_invoice_sequence_id
- self_invoice_report_footer

imagen

Shide:
Weird things happening requesting reviews... @yajo @etobella

account_invoice_supplier_self_invoice/data/mail_template_data.xml
```diff
@@ -0,0 +1,59 @@
...
+    <record id="email_template_edi_supplier_self_invoice" model="mail.template">
+        <field name="name">Purchase Self Invoice: Send by email</field>
+        <field name="model_id" ref="account.model_account_move" />
+        <field
+            name="email_from"
+        >{{ (object.invoice_user_id.email_formatted or user.email_formatted) }}</field>
+        <field name="partner_to">{{ object.partner_id.id }}</field>
+        <field
+            name="subject"
+        >{{ object.company_id.name }} Invoice (Ref {{ object.name or 'n/a' }})</field>
```
rafaelbn: In this template subject should print the `ref` and not the name. The same than in the body

imagen

account_invoice_supplier_self_invoice/views/report_self_invoice.xml
```diff
@@ -0,0 +1,190 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <!-- Self invoice layout view -->
+    <template
+        id="external_layout_striped_self_invoice"
```
rafaelbn, line 5: Somethings missing in this report. Yajo commented one of them. Here you are all:

imagen

account_invoice_supplier_self_invoice/models/res_partner.py
```diff
@@ -0,0 +1,48 @@
...
+    def set_self_invoice(self):
+        for record in self:
+            record.self_invoice = not record.self_invoice
+            if record.self_invoice_sequence_id:
+                continue
+            if record.self_invoice:
+                record.self_invoice_sequence_id = (
+                    self.env["ir.sequence"]
+                    .sudo()
+                    .create(
```
rafaelbn: This is not and error but and improvement. Auto creation sequences with diferent by vendor as having all the same is not a BUG or incorrect but could be a problematic for UX, looking for references, **bank reconcile**... etc.

imagen

rafaelbn (commented):
Some review done!

rafaelbn:
Hello @etobella @Shide !

Pressing the button "self invoice" for a vendor is not something that a billing user should make. This is a legal staff so I would like to change all this configuration from "Sales & Purchase" tab to the "invoicing" Tab, or at least control that alll the fields of this module are read only for everyone and only Billing administrator can create o edit them.

What do you think? Which is you opinion about it? 😄 

Thank you!

rafaelbn (changes requested):
Hello @Shide ! I found a bug while trying to send invoices in bulk

- [x] Missing send/print button only for self-invoices
- [x] BUG while trying to send invoices in bulk

```
`Traceback (most recent call last):
  File "/opt/odoo/odoo/tools/cache.py", line 85, in lookup
    r = d[key]
  File "/opt/odoo/odoo/tools/func.py", line 71, in wrapper
    return func(self, *args, **kwargs)
  File "/opt/odoo/odoo/tools/lru.py", line 34, in __getitem__
    a = self.d[obj]
KeyError: ('ir.model.data', <function IrModelData._xmlid_lookup at 0x7f847b2c3ca0>, 'account_invoice_supplier_self_invoice.                email_template_edi_supplier_self_invoice')

During handling of the above exception, another exception occurred:

  File "/opt/odoo/odoo/tools/safe_eval.py", line 332, in safe_eval
    return unsafe_eval(c, globals_dict, locals_dict)
  File "", line 2, in <module>
  File "/opt/odoo/addons/account/models/account_move.py", line 3217, in action_send_and_print
    'default_template_id': self.env.ref(self._get_mail_template()).id,
  File "/opt/odoo/odoo/api.py", line 578, in ref
    res_model, res_id = self['ir.model.data']._xmlid_to_res_model_res_id(
  File "/opt/odoo/odoo/addons/base/models/ir_model.py", line 1943, in _xmlid_to_res_model_res_id
    return self._xmlid_lookup(xmlid)[1:3]
  File "<decorator-gen-35>", line 2, in _xmlid_lookup
  File "/opt/odoo/odoo/tools/cache.py", line 90, in lookup
    value = d[key] = self.method(*args, **kwargs)
  File "/opt/odoo/odoo/addons/base/models/ir_model.py", line 1936, in _xmlid_lookup
    raise ValueError('External ID not found in the system: %s' % xmlid)
ValueError: External ID not found in the system: account_invoice_supplier_self_invoice.                email_template_edi_supplier_self_invoice

The above exception was the direct cause of the following exception:

  File "/opt/odoo/odoo/addons/base/models/ir_http.py", line 237, in _dispatch
    result = request.dispatch()
  File "/opt/odoo/odoo/http.py", line 687, in dispatch
    result = self._call_function(**self.params)
  File "/opt/odoo/odoo/http.py", line 359, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/opt/odoo/odoo/service/model.py", line 94, in wrapper
    return f(dbname, *args, **kwargs)
  File "/opt/odoo/odoo/http.py", line 348, in checked_call
    result = self.endpoint(*a, **kw)
  File "/opt/odoo/odoo/http.py", line 916, in __call__
    return self.method(*args, **kw)
  File "/opt/odoo/odoo/http.py", line 535, in response_wrap
    response = f(*args, **kw)
  File "/opt/odoo/addons/web/controllers/main.py", line 1628, in run
    result = action.run()
  File "/opt/odoo/odoo/addons/base/models/ir_actions.py", line 643, in run
    res = runner(run_self, eval_context=eval_context)
  File "/opt/odoo/odoo/addons/base/models/ir_actions.py", line 512, in _run_action_code_multi
    safe_eval(self.code.strip(), eval_context, mode="exec", nocopy=True)  # nocopy allows to return 'action'
  File "/opt/odoo/odoo/tools/safe_eval.py", line 348, in safe_eval
    raise ValueError('%s: "%s" while evaluating\n%r' % (ustr(type(e)), ustr(e), expr))
Exception

The above exception was the direct cause of the following exception:

  File "/opt/odoo/odoo/http.py", line 643, in _handle_exception
  File "/opt/odoo/odoo/http.py", line 301, in _handle_exception
ValueError: <class 'ValueError'>: "External ID not found in the system: account_invoice_supplier_self_invoice.                email_template_edi_supplier_self_invoice" while evaluating
'if records:\n                    action = records.action_send_and_print()'

```

account_invoice_supplier_self_invoice/models/res_partner.py
Shide: I will add the vendor reference if the field is not empty. (partner.ref  on Sales and Purchases > Misc group)

Example: 
partner ref = "  IKEA  "
Resulting sequence: "IKEA/INV/2022/00001"

partner ref = None
Resulting sequence: "SI/INV/2022/00001"

Are you agreed with that?

Shide:
Do you have any legal text to put in self invoice footer by default?

I'm going to add "Invoiced by the recipient" in the default parameter of `self_invoice_report_footer` in order to match https://getquipu.com/blog/que-es-una-autofactura/#como-hace-autofactura

Shide:
> Hello @etobella @Shide !
> 
> Pressing the button "self invoice" for a vendor is not something that a billing user should make. This is a legal staff so I would like to change all this configuration from "Sales & Purchase" tab to the "invoicing" Tab, or at least control that alll the fields of this module are read only for everyone and only Billing administrator can create o edit them.
> 
> What do you think? Which is you opinion about it? smile
> 
> Thank you!

I'm ok with that

Shide:
All things done @rafaelbn  @etobella

rafaelbn (changes requested):
Reviewed 😄 ❤️ Still pending somethings but it looks great

- [x] Send button is missing now from customer invoices. In my test, now it's show only when res.partner is "self"
- [x] 3 terms to change in the email template for sefl
- [x] 1 term to change in the PDF for the self-bill 

After this FIX/IMP It looks ready. I'm going to record a video for other contributors 😄

account_invoice_supplier_self_invoice/models/res_partner.py
yajo: So let's say I have companies A and B, and I have supplier C which will self-invoice me.

C would have the same self-invoice sequence for A and B? I don't think so.

What if C self-invoices only B? Odoo would be suggesting self-invoices for A purchase orders. Not good IMHO.

account_invoice_supplier_self_invoice/models/res_partner.py
yajo: Partners have refs? Or are you talking about adding a new field for that? If so, maybe it would be better to use partner ID? Example:

    SI/P%(partner_id)s/%(year)s/ -> SI/P123/2022/00001

account_invoice_supplier_self_invoice/models/res_partner.py
Shide: I think Yajo is in the correct position.

Sequence has the `company_id` field, so the creation I think will not be a problem.

rafaelbn: Great point @yajo 

I think this is obvious, all the sequences should be by res.company . Also in de case:

- Company A
- Company B
- Vendor Self invoice 1

Case 1:

- Company A make self-biiling to "Vendor Self invoice 1" so invoice is SI/INV/2022/11/0001
- Company B make self-biiling to "Vendor Self invoice 1" so invoice is SI/INV/2022/11/0001, absolutely other invoice sequence. Company A & B are different legal entities.

Case 2: 

- Company A make self-biiling to "Vendor Self invoice 1" so invoice is SI/INV/2022/11/0001
- Company B DON'T make self-biiling to "Vendor Self invoice 1" so invoice is the number the vendor send us.


Any other use case is not correct. Multi-company resilience please!

Thank you!

rafaelbn:
- [x] Self-invoice Sequence per res.company in vendors (res.partners) @Shide Thanks

Shide:
@rafaelbn @etobella Ready to review with multicompany.

Shide:
Ready to review with multi company.

You must go to Accounting/Invoicing settings and set the **Default Self Billing prefix** on each company. (It's in USAGE.rst)
Then you can do the normal test.

@rafaelbn @etobella

etobella:
I think it is not a good idea to make company_dependent... I must do a different kind of sequences in order to not cross invoices for my supplier. this is an extra work that probably is not necessary. Also, previous versions were not company_dependent, so you should think how to keep the previous functionality (not company-dependent and company-shared sequence for a supplier)

Also, IMO, company_dependent fields are hard to manage in multi-company environments, because that might give you wrong sensations if you are working with several companies at the same time.

account_invoice_supplier_self_invoice/models/res_partner.py
etobella: This is broking previous installations of this module, also, both systems have sense, so, IMO, or another glue module is needed to make it company_dependent or find a way to keep the current behavior please.

I have been working with this approach for 3 years with Odoo and 20 with other kind of softwares, and it is a valid options.

etobella (changes requested):
Don't make it company_dependent. It is breaking previous installations.

account_invoice_supplier_self_invoice/models/res_partner.py
rafaelbn: - Sequences should be same length (Best practices) so "partner_id will change length
- Partners have ref, yes.

Could it be a magic to have always for example 6 digit in the id?

`SI/P%(partner_id)s/%(year)s/ -> SI/P123/2022/00001`

This we should think more about that in an internacnional enviroment with multi-company

yajo:
> Don't make it company_dependent. 

This is a drastic solution for the backwards incompatibility problem... But doesn't fix that this module not supporting multi-company case is a serious design problem.

Quoting (and translating, for wider audience) from the link posted above in https://github.com/OCA/account-invoicing/pull/1290#issuecomment-1323779213:

> **Each self-bill receiver** must be assigned a different serial number, but within this, it **must follow its own sequential numbering**.

I'm sad to say this @etobella, but if you've been really doing self-invoicing with multiple companies using the same sequence, you've been probably doing it wrong 😅 (AFAICS not only technically, but also legally, at least in Spain).

So this is something that has to be fixed.

> It is breaking previous installations.

Now this is the real problem. But don't worry! In this case, we just need a migration script. In that script, we should keep the sequences working as before the migration. If the same sequence is applied for multiple companies before the migration, then the same must be true after it. Can you please program that @Shide?

account_invoice_supplier_self_invoice/models/res_partner.py
yajo: > Sequences should be same length (Best practices) so "partner_id will change length

In this case, it should not affect, because the partner ID is not part of the sequence, but it is part of the prefix.

In any case, I'm seeing that it wouldn't help, because Partner 123 is the same for company A and B. It should be the company prefix instead.

> Partners have ref, yes.

Where?

etobella:
Don't worry, I already made small proposal (also splitting the migration commit) https://github.com/tegin/account-invoicing/tree/15.0-mig-account_invoice_supplier_self_invoice WDYT?

etobella:
Anyway, the current way to manage (creating and pressing the button for each company) might be hard when you have a lot of suppliers with self-invoice or companies... now, the sequence is generated on the fly when needed (and company dependent). We might discuss if the flag should be company_dependent or not, but not blocking from my side (easy to change at the end...)

rafaelbn (commented):
### Review
- [ ] Self invoice footer is empty after click the button "supplier approves self billing". See: https://www.loom.com/share/325edbf452554a7e957ba0d930536a85?t=180
- [x] In Vendor bill list view if you filter by "self invoice" you see only self invoice bill but we cannot select all and "post entries" (validate) because it says we should set a invoice date, which is only correct for not self invoices bills: https://www.loom.com/share/325edbf452554a7e957ba0d930536a85?t=625
- [ ] I still think we should take care and improve that all sequence of vendors for self invoice are diferent per vendor, see this: https://www.loom.com/share/325edbf452554a7e957ba0d930536a85?t=915

### Next step
- [ ] We should forget to create a purchase module to filter easily purchases that are going to be self invoices

### Not related to this module
- [ ] We cannot confirm purchases massively
- [ ] If you select one single vender a create invoices from purchases Odoo group then in a single invoice BUT if you select all purchases an create invoices from "Action -> Invoice" Odoo create a single invoice per purchase

Shide:
> ### Review
> * [ ]  Self invoice footer is empty after click the button "supplier approves self billing". See: https://www.loom.com/share/325edbf452554a7e957ba0d930536a85?t=180
> * [x]  In Vendor bill list view if you filter by "self invoice" you see only self invoice bill but we cannot select all and "post entries" (validate) because it says we should set a invoice date, which is only correct for not self invoices bills: https://www.loom.com/share/325edbf452554a7e957ba0d930536a85?t=625
> * [ ]  I still think we should take care and improve that all sequence of vendors for self invoice are diferent per vendor, see this: https://www.loom.com/share/325edbf452554a7e957ba0d930536a85?t=915
> 
> ### Next step
> * [ ]  We should forget to create a purchase module to filter easily purchases that are going to be self invoices
> 
> ### Not related to this module
> * [x]  We cannot confirm purchases massively
> * [ ]  If you select one single vender a create invoices from purchases Odoo group then in a single invoice BUT if you select all purchases an create invoices from "Action -> Invoice" Odoo create a single invoice per purchase

Now you can confirm invoices massively. I've changed the method who perform all the actions when an invoice is confirmed.

Shide:
@etobella mira a ver si ves bien el cambio de la función (cambiado el action_post por el _post), que es lo único que he cambiado.

etobella (approved):

rafaelbn (approved):
Let's go @etobella 👍

PR #1294, opened 2022-11-30
[ADD] contract_invoicing_grouping_criteria

paradoxxxzero:
This module allows to group invoicing of several contracts and to use custom criteria for grouping contracts to be invoiced.

This module mimics the sale.order grouping behavior of the sale module and adds the sale_order_invoicing_grouping_criteria's custom criteria mechanism.


paradoxxxzero:
@mourad-ehm @hparfr

contract_invoicing_grouping_criteria/models/contract.py
```diff
@@ -0,0 +1,188 @@
...
+        # if x is a record, return its id to get a consistent sorting
+        return x.id if isinstance(x, models.BaseModel) else x
+
+    def get_with_or(x, key):
+        """Get key from x. Handles keys containing ||"""
+        for k in key.split("||"):
+            val = getattr(x, k)
+            if val:
+                return val
+        return val
```
mourad-ehm, line 19: val must be defined before the for loop if key is empty we can have a bug.

paradoxxxzero: Split with an argument always returns a non empty list
(cf: https://docs.python.org/3.3/library/stdtypes.html#str.split)

mourad-ehm: in this case the last return is useless ?

paradoxxxzero: Not very useful indeed but this does not matter much

PR #1297, opened 2022-11-30
[15.0][FIX] account_invoice_date_due: Fix tests. Compare fields with sudo user

sergio-teruel:
cc @Tecnativa TT40606

ping @pilarvargas-tecnativa @chienandalu 

As the tests use a Form if other module adds a field in account.move model with an extra security group this field in not in Form fields, so crash the tests in a full private CI


sergio-teruel:
Unrelated runboat error..

chienandalu (approved):

chienandalu:
We could remove João from the mantainers as he's no longer in the Odoo world.

pedrobaeza (approved):

PR #1298, opened 2022-12-01
[Add] account_invoice_policy_order_subtotal

santostelmo:
### Context
On sale order lines we want to have the possibility to add a line where the product of type
consumable (product for which stock is not managed) represents a financial amount not linked to any
delivered product or service.

image

For this type of lines, the invoicing policy is focused on the subtotal amount.
This means that the line can be invoiced as long as the amount already invoiced does not reach
the subtotal amount.


account_invoice_policy_order_subtotal/models/sale_order.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    def _get_invoiceable_lines(self, final=False):
+        precision = self.env["decimal.precision"].precision_get("Product Price")
+        lines = super()._get_invoiceable_lines(final=final)
+        subtot_policy_lns = self.order_line._get_policy_order_subtotal_lines()
+        # Filter subtot_policy_lns from invoiveable lines got in super call
```
yankinmax, line 15:
```suggestion
        # Filter subtot_policy_lns from invoiceable lines got in super call
```

yankinmax (approved):

account_invoice_policy_order_subtotal/models/sale_order_line.py
```diff
@@ -0,0 +1,116 @@
...
+                    precision_digits=precision,
+                )
+                <= 0
+            ):
+                line.invoice_status = "invoiced"
+            else:
+                line.invoice_status = "no"
+        super(SaleOrderLine, self - subtot_policy_lns)._compute_invoice_status()
+
+    amount_invoiced = fields.Float(
```
grindtildeath: Can you please move the fields definition at the top of the class to follow conventions?

account_invoice_policy_order_subtotal/models/sale_order_line.py
```diff
@@ -0,0 +1,116 @@
...
+        compute="_compute_amount_invoiced",
+        string="Invoiced Amount",
+        store=True,
+        readonly=True,
+        compute_sudo=True,
+        digits="Product Unit of Measure",
```
grindtildeath: I don't think we want to use UOM precision if we're managing amount there. I would expect to use the currency precision in that case.

grindtildeath (changes requested):
I have a really hard time understanding the functional scope of this module. :confused: 
Could you explain a bit more what's the goal with this?

To me there's one important issue though: you're adding a selection value in `invoice_policy` but this value is supposed to be used only for `consumable` products. However, someone can still use this value with a product of another type, and we don't know what will happen...

IMO, would be better to use another field for this, but at the minimum you should add a constraint to make sure we cannot define this invoice policy for another type of product.

account_invoice_policy_order_subtotal/models/sale_order_line.py
santostelmo: > I have a really hard time understanding the functional scope of this module. confused Could you explain a bit more what's the goal with this?

Thanks @grindtildeath  for the review!  
When using invoice policy focused on order subtotal amount, we consider a line fully invoiced when the total invoiced amount equal subtotal amount. 
The goal of that development is that: often you don't invoice a sale order by quantity, you invoice a certain amount until you reach a threshold that in this case is the price_subtotal.

Note that we still calculate invoiced quantities based on the proportion of the amount invoiced. Like that modules that calculate accrued revenue are still compatible with this development.

santostelmo:
> IMO, would be better to use another field for this, but at the minimum you should add a constraint to make sure we cannot define this invoice policy for another type of product.

I agree, that policy is supposed to be used only when the product type is `consumable`

santostelmo:
@grindtildeath I made the requested changes

PR #1301, opened 2022-12-12
[14.0][ADD] accounting_partner_category: utility module helping to filter invoices

petrus-v:


accounting_partner_category/i18n/fr.po
```diff
@@ -0,0 +1,172 @@
...
+msgid "Last Updated on"
+msgstr "Dernière mise à jour le"
+
+#. module: accounting_partner_category
+#: model_terms:ir.actions.act_window,help:accounting_partner_category.action_accounting_partner_category
+msgid ""
+"Manage accounting contact tags to better classify them for tracking, filter "
+"and analysis purposes."
+msgstr ""
+"Gère les tags de contact pour la comptabilité pour facilité le suivi de "
```
kemar, line 111:
```suggestion
"Gère les tags de contact pour la comptabilité pour faciliter le suivi de "
```

accounting_partner_category/i18n/fr.po
```diff
@@ -0,0 +1,172 @@
...
+#: model:ir.model.fields,field_description:accounting_partner_category.field_account_payment__category_ids
+msgid "Tags"
+msgstr "Tags"
+
+#. module: accounting_partner_category
+#: model:ir.model.fields,help:accounting_partner_category.field_account_bank_statement_line__category_ids
+#: model:ir.model.fields,help:accounting_partner_category.field_account_move__category_ids
+#: model:ir.model.fields,help:accounting_partner_category.field_account_payment__category_ids
+msgid "Tags when partner was added to the invoice."
+msgstr "Tags utilisé pour filter les pièces comptable."
```
kemar, line 156:
```suggestion
msgstr "Tags utilisé pour filtrer les pièces comptable."
```

The translation does not seem to match `msgid`?

kemar (approved):

accounting_partner_category/models/account_move.py
```diff
@@ -0,0 +1,22 @@
...
+        string="Tags",
+        help="Tags when partner was added to the invoice.",
+    )
+
+    @api.onchange("partner_id")
+    def _onchange_partner_id(self):
+        super()._onchange_partner_id()
+        self.category_ids = (
+            self.partner_id.commercial_partner_id.accounting_category_ids
+        )
```
jguenat: Hi, I think you should use a computed field. Onchange won't be triggered when importing invoices from other sources (or just if created anywhere else than in the form view).
You can take a look how it was done there
https://github.com/OCA/account-invoicing/blob/15.0/account_invoice_transmit_method/models/account_move.py

accounting_partner_category/i18n/foodles_accounting_partner_category.pot
```diff
@@ -0,0 +1,169 @@
+# Translation of Odoo Server.
```
hparfr: rename this file

accounting_partner_category/models/account_move.py
petrus-v: the problem with such api.depends implementation, when creating new invoice and you want to remove all tags they will appear while saving as api.depends is called after create where partner_id will be considered changed so the user input will be lost given some frustrations

petrus-v: @jguenat I'm not sure how to avoid this side effect ?

So I've also implement #1302 when creating invoice from sale order at the moment...

Also if invoice are imported from other source there is a chance that they already call [upstream `_onchange_partner_id` method](https://github.com/OCA/OCB/blob/14.0/addons/account/models/account_move.py#L436) to get proper informations on invoice.

jguenat: If you have the compute="_compute_your field" and readonly=False defined in conjunction with the @api.depends I'm pretty sure the behaviour is similar to an onchange on the form view with the benefit of the compute called on creation in any cases. You can check the functionality of account_invoice_transmit_method :)

petrus-v: My bad, I probably missed something during my initial implementation/test... I'll close #1302 then :)

jguenat (approved):

petrus-v:
Is there any @OCA/accounting-maintainers could merge this approved PR ?

damdam-s:
@pedrobaeza @grindtildeath can one you you guys can merge this one? Double-approved and all 🟢 
thanks a lot

PR #1303, opened 2022-12-13
[FIX] account_invoice_supplier_self_invoice: Creation must be set with sudo

etobella:
otherwise, the following error is raised:

image

TEsted on runboat

@rafaelbn @Shide


rafaelbn (approved):
👍 

@Shide

Shide (approved):
LGTM

PR #1306, opened 2022-12-15
[IMP] faster SQL hook, avoids memory limit reached for large datasets

gfcapalbo:
On a customer implementation : 90K invoices. 1.2Million invoice lines. 2 GB of system memory, the module was uninstallable due to memory limits reached.
This version inits the sequence fields with a quick SQL operation, reducing install time drastically and avoiding the memory problem.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -34,26 +34,21 @@ repos:
           - requirements.txt
           - --header
           - "# generated from manifests external_dependencies"
-  - repo: https://github.com/OCA/mirrors-flake8
-    rev: v3.4.1
+  - repo: https://github.com/PyCQA/flake8
+    rev: 3.8.3
     hooks:
       - id: flake8
-        name: flake8 excluding __init__.py
-        exclude: __init__\.py
-  - repo: https://github.com/pre-commit/mirrors-pylint
-    rev: v2.5.3
+        name: flake8
+        additional_dependencies: ["flake8-bugbear==20.1.4"]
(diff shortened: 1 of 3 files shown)


legalsylvain (approved):
Thanks. Makes sense. Code review / no test.

AaronHForgeFlow:
Makes sense! Can you check pre-commit?

gfcapalbo:
seems like pre-commit cannot find "python3.6" i have same problem on local , made some changes/updates in precommit,   without decreasing checks, and then added some small lints.
On local CI is successful, still failing here: it seems that we need python 3.7 or pin libraries to an older version:
we have a similar problem here:https://github.com/csuhan/ReDet/issues/14

gfcapalbo:
@legalsylvain @AaronHForgeFlow I have tried updating a bit of things in my pre-commit , and locally it now works, all green. 
It seems that we require python 3.7 as default python 3 in our CI to avoid an environment setup error, but i don;t know how to do that.  on my machine "python3" is python 3.10, if i set python 3.6 as global python i get the same error on local.

I see only two solutions:
     Downgrade all libraries to work with python 3.6 (set pylint  and pylint-odoo to a even lower version)  or set python 3.10 as our default "python3" to solve initialization problem.

gfcapalbo:
CI fixed, by replicating the setup  on other branches, the checks are effectively the same . please double check. 

All work concerning the hook fix in 1 commit.
All work involving CI in another commit.

PR #1309, opened 2022-12-19
[FIX] [15.0] account_invoice_supplier_self_invoice: New partner billing prefix

Shide:
Use a new partner self billing prefix to not use the `partner.ref` field.
Added a button in the partner to create both sequences before the first self bill invoice (maybe you want to modify the sequence)

Please review if you want @etobella @rafaelbn 

MT-906 @moduon

### DIFF
--- a/account_invoice_supplier_self_invoice/views/res_partner_views.xml
+++ b/account_invoice_supplier_self_invoice/views/res_partner_views.xml
@@ -7,16 +7,28 @@
         <field name="arch" type="xml">
             <xpath expr="//page[@name='accounting']/group" position="inside">
                 <group name="group_self_invoice" string="Self Billing">
-                    <field name="self_invoice" />
+                    <field name="self_invoice" widget="boolean_toggle" />
+                    <field
+                        name="self_invoice_partner_prefix"
+                        attrs="{'invisible': [('self_invoice', '=', False)], 'readonly': [('self_invoice_sequence_id', '!=', False), ('self_invoice_refund_sequence_id', '!=', False)]}"
+                    />
+                    <button
+                        name="action_set_self_invoice"
+                        string="Create missing Self Billing sequences"
+                        type="object"
+                        class="oe_highlight mb-2"
+                        attrs="{'invisible': ['|', ('self_invoice', '=', False), '&amp;', ('self_invoice_sequence_id', '!=', False), ('self_invoice_refund_sequence_id', '!=', False)]}"
+                        colspan="2"
+                    />
                     <field
                         name="self_invoice_sequence_id"
                         readonly="1"
-                        attrs="{'invisible': [('self_invoice', '=', False)]}"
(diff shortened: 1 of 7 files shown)


rafaelbn (approved):
Tested in runboat 👍

rafaelbn:
Could I recommend, please:

- [x] Make visible the button for sending the invoice in draft to the vendor

Why? Because for our general case, we send it in draft, the vendor confirm it by replying and then we validate.

A [requirement](https://www.sumup.com/es-es/facturas/glosario/facturacion-por-el-destinatario/) is

> Por cada factura emitida, el empresario o profesional ha de mostrar su aceptación, mediante un procedimiento que acordará previamente con su cliente. 

Thanks!

account_invoice_supplier_self_invoice/views/res_partner_views.xml
```diff
@@ -7,16 +7,28 @@
...
+                        string="Create missing Self Billing sequences"
+                        type="object"
+                        class="oe_highlight mb-2"
+                        attrs="{'invisible': ['|', ('self_invoice', '=', False), '&amp;', ('self_invoice_sequence_id', '!=', False), ('self_invoice_refund_sequence_id', '!=', False)]}"
+                        colspan="2"
+                    />
                     <field
                         name="self_invoice_sequence_id"
                         readonly="1"
-                        attrs="{'invisible': [('self_invoice', '=', False)]}"
```
etobella, line 14: You should hide it only if it is null, no need to check on the other. WDYT?

etobella (approved):
Technical review.

Just a small comment.

:+1:

account_invoice_supplier_self_invoice/views/res_partner_views.xml
etobella: Te same for the refund sequence :wink:

Shide: Because when both fields (`self_invoice_sequence_id` and `self_invoice_refund_sequence_id`) are not set, only the button must be displayed. 
This is why is the same attrs on both fields.

Shide:
> Could I recommend, please:
> 
> * [ ]  Make visible the button for sending the invoice in draft to the vendor
> 
> Why? Because for our general case, we send it in draft, the vendor confirm it by replying and then we validate.
> 
> A [requirement](https://www.sumup.com/es-es/facturas/glosario/facturacion-por-el-destinatario/) is
> 
> > Por cada factura emitida, el empresario o profesional ha de mostrar su aceptación, mediante un procedimiento que acordará previamente con su cliente.
> 
> Thanks!

Done

Recap:
Added to show button when invoice is not cancelled
Added "This Self-bill is not valid" when the invoice is in draft state on the footer 
Added ES missing translations

etobella:
@Shide @rafaelbn translation should be managed with weblate not directly here. Touching it outside from it can give us some errors :dizzy_face:

PR #1315, opened 2022-12-21
[FIX] account_invoice_supplier_self_invoice: change the groups to view in order to use them without debug mode

etobella:
@rafaelbn


rafaelbn (approved):
Tested in runboat! 👍 Thank you @etobella !

/ocabot merge patch

PR #1316, opened 2022-12-22
[FIX] Pre-commit flake8 repo gitlab->github

joshkreud:
According to https://flake8.pycqa.org/en/latest/user/using-hooks.html the Flake8 pre-commit hook now lives on github. 

The gitlab repo runs into a 404.


pedrobaeza:
For achieving the same result (and other update to the template), you can do `pipx run copier -f update`, but let's merge it to fix at least the 404 error:

/ocabot merge nobump

PR #1319, opened 2022-12-22
[IMP] account_invoice_supplier_self_invoice: Improve footer management

etobella:
Allow to use a translated value
Write it only on if self_invoice is checked

### DIFF
--- a/account_invoice_supplier_self_invoice/models/res_partner.py
+++ b/account_invoice_supplier_self_invoice/models/res_partner.py
@@ -43,12 +43,15 @@ class ResPartner(models.Model):
     )
     self_invoice_report_footer = fields.Text(
         string="Self Billing footer",
-        default="Invoiced by the recipient",
         help="Footer text displayed at the bottom of the self invoice reports.",
         copy=False,
         tracking=True,
         company_dependent=True,
     )
+
+    @api.model
+    def _default_self_invoice_report_footer(self):
+        return _("Invoiced by the recipent")
 
     def _get_self_invoice_number(self, invoice):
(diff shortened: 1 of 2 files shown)


account_invoice_supplier_self_invoice/models/res_partner.py
```diff
@@ -115,6 +118,4 @@ def action_set_self_invoice(self):
     @api.onchange("self_invoice")
     def onchange_self_invoice(self):
         if self.self_invoice and not self.self_invoice_report_footer:
-            self.self_invoice_report_footer = self._fields[
-                "self_invoice_report_footer"
-            ].args["default"]
+            self.self_invoice_report_footer = self._default_self_invoice_report_footer()
```
Shide: I think it's better to use the partner language instead of user language to generate the translated value.

```suggestion
            self.self_invoice_report_footer = self.with_context(lang=self.lang)._default_self_invoice_report_footer()
```

`_` translation method uses lang in context

Shide (changes requested):
Just a small suggestion

account_invoice_supplier_self_invoice/models/res_partner.py
rafaelbn: Hello @etobella , i think this is a good point. The language should be from partner, not the user. I guess you agree too. Let us know, thanks!

etobella: :+1:

Shide (approved):
LGTM

rafaelbn (approved):
/ocabot merge minor

PR #1320, opened 2022-12-23
[12.0][FIX] view created payment

sergiocorato:
When creating a payment and clicking on `Validate and view payment` payment is created but the view opened is empty.
With this PR the payment is opened correctly.


legalsylvain (approved):

sergiocorato:
To pass tests #1321 is needed

PR #1327, opened 2022-12-27
[13.0][FIX] sale_line_refund_to_invoice_qty: do not change invoiced qty

LoisRForgeFlow:
when you mark a refund to not be reinvoiced, the invoice qty should not be reduced, because it is already invoiced and you won't invoice it again.

@ForgeFlow

PR #1339, opened 2023-01-05
[15.0][FIX] sale_line_refund_to_invoice_qty: do not change invoiced qty

LoisRForgeFlow:
when you mark a refund to not be reinvoiced, the invoice qty should not be reduced, because it is already invoiced and you won't invoice it again.

Forward port of #1327

### DIFF
--- a/sale_line_refund_to_invoice_qty/models/sale.py
+++ b/sale_line_refund_to_invoice_qty/models/sale.py
@@ -13,28 +13,27 @@ class SaleOrderLine(models.Model):
     )
 
     @api.depends(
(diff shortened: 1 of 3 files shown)


LoisRForgeFlow:
cc @juanydforgeflow @AaronHForgeFlow

JuanyDForgeflow:
@LoisRForgeFlow seems like the test is not working

LoisRForgeFlow:
@JuanyDForgeflow Thanks for the headsup, it is fixed now. Could you check?

AaronHForgeFlow (approved):

PR #1344, opened 2023-01-11
[OU] account_invoice_supplier_self_invoice: Create field  by script

etobella:
In order to make it faster, we will create the field and compute it with SQL.


rafaelbn (approved):
I don't see any problem. Please @Shide could you make a technical review? Thank you 👍

Shide (approved):

rafaelbn (approved):
/ocabot merge patch

PR #1352, opened 2023-01-22
[14.0][FIX] account_invoice_pricelist: Make tests resilient

pedrobaeza:
- Depending on the installed set of modules, the company currency may be USD or EUR. If the second case, these tests will fail, so we make sure that the company currency is USD for our tests, doing the change by SQL, as there's a Python constraint that prevents it.
- On contrary, the main pricelist doesn't change its currency, so we can have an incompatible currency.
- The onchange partner test should assert fixed data.

@Tecnativa

### DIFF
--- a/account_invoice_pricelist/tests/test_account_move_pricelist.py
+++ b/account_invoice_pricelist/tests/test_account_move_pricelist.py
@@ -27,6 +27,13 @@ class TestAccountMovePricelist(common.SavepointCase):
         cls.journal_sale = cls.env["account.journal"].create(
             {"name": "Test sale journal", "type": "sale", "code": "TEST_SJ"}
         )
+        # Make sure the currency of the company is USD, as this not always happens
+        # To be removed in V17: https://github.com/odoo/odoo/pull/107113
+        cls.company = cls.env.company
+        cls.env.cr.execute(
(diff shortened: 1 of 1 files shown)


pedrobaeza:
FYI, this weird company's currency switch has been finally solved in odoo/odoo#107113 for v17.

account_invoice_pricelist/tests/test_account_move_pricelist.py
```diff
@@ -27,6 +27,12 @@ def setUpClass(cls):
         cls.journal_sale = cls.env["account.journal"].create(
             {"name": "Test sale journal", "type": "sale", "code": "TEST_SJ"}
         )
+        # Make sure the currency of the company is USD, as this not always happens
```
legalsylvain, line 30: If I understand correctly your point, regarding V17.

```suggestion
        # Make sure the currency of the company is USD, as this not always happens
        # Remove this SQL request when porting this module in V17.
        # See : https://github.com/odoo/odoo/pull/107113 
```

pedrobaeza: Added, @legalsylvain

ernestotejeda (approved):

carlosdauden (approved):

PR #1353, opened 2023-01-23
IMP 15.0 sale_order_invoicing_grouping_criteria: extract _prepare_so_grouping

leemannd:
Extract the so-grouping to allow pre or post processing.

### DIFF
--- a/sale_order_invoicing_grouping_criteria/models/sale_order.py
+++ b/sale_order_invoicing_grouping_criteria/models/sale_order.py
@@ -34,8 +34,9 @@ class SaleOrder(models.Model):
             group_key.append(self[field.name])
         return tuple(group_key)
 
-    def _create_invoices(self, grouped=False, final=False, date=None):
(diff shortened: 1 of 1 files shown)


pedrobaeza:
You already have `_get_sale_invoicing_group_key` for changing the way the groups are done. Why another hook? Just add an extra key for the grouping criteria you want.

leemannd:
Hello, given the result of the grouping result I want to trigger distinct methods if the invoice is the result of a grouping or not.
The previous hook prepares the data for the grouping, but we don't know at this moment if the invoice that will be created will be from grouped SO.

pedrobaeza:
I don't get exactly your explanation, as you get the link SO line <> invoice line for doing any kind of traceability between both without having to call the grouping method IMO, but well, I don't want to oppose to this. Pre-commit is not OK though.

leemannd:
Closing as old and outdated

PR #1359, opened 2023-01-29
[FIX] account_invoice_supplier_self_invoice: Use country

etobella:
The country used does not exist on 14.0

@rafaelbn


Shide (approved):

rafaelbn (approved):
/ocabot merge patch

PR #1361, opened 2023-01-31
[15.0][ADD] account_move_tax_adjust

TheerayutEncoder:
This module allow user's ability to edit the tax amounts to fix rounding errors on customer invoice.


account_move_tax_edition/models/account_move.py
```diff
@@ -0,0 +1,49 @@
...
+    @api.depends(
+        "line_ids.amount_currency",
+        "line_ids.tax_base_amount",
+        "line_ids.tax_line_id",
+        "partner_id",
+        "currency_id",
+        "amount_total",
+        "amount_untaxed",
+    )
+    def _compute_tax_totals_json(self):
```
Saran440: you shouldn't overwrite function core odoo if not necessary.

May be you can use this

```suggestion
    def _compute_tax_totals_json(self):
        res = super()._compute_tax_totals_json()
        for move in self:
            tax_totals = json.loads(move.tax_totals_json)
            tax_totals["allow_tax_edition"] = (
                move.is_purchase_document(include_receipts=False)
                or move.is_sale_document(include_receipts=False)
            ) and move.state == "draft"
```

it clear and not affect with other module.

Saran440 (changes requested):
@TheerayutEncoder I think this module should be name "account_move_tax_adjust"
and please add test script in this module.

account_move_tax_edition/models/account_move.py
TheerayutEncoder: Thanks, I will improve my code.

DucTruongKomit (approved):
Look good to me now!

Saran440 (approved):
Code Review and Functional Test :+1:

PR #1364, opened 2023-02-01
[16.0][ADD] account_tax_one_vat

glitchov:
Limits the selection of taxes of  type VAT to one. This is done at different levels: so line, po line, product or move line.


account_tax_one_vat/data/account_tax_group.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" ?>
+<odoo noupdate="1">
+  <record id="vat_tax_group" model="account.tax.group">
```
legalsylvain: Hi. I think you should handle that point (know if it is a vat or not), by adding a is_vat on `account.tax.`(template)
It can occures that there are a vat group 5,5% and a vat group 10% and a vat group 20%. All that configuration are valid.

account_tax_one_vat/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+{
+    "name": "Account Tax One VAT",
+    "summary": "Allow only the selection of one VAT Tax.",
+    "version": "16.0.1.0.0",
+    "author": "Camptocamp, Acsone, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "category": "account",
+    "depends": ["account", "sale", "purchase"],
```
legalsylvain: I think that you could make a generic module `account_tax_one_vat` and some glue modules `account_tax_one_vat_purchase` `account_tax_one_vat_sale`.
That should not ask a lot of work, and allow better modularity.

account_tax_one_vat/models/account_move_line.py
```diff
@@ -0,0 +1,13 @@
...
+
+from odoo import api, models
+
+
+class AccountMoveLine(models.Model):
+    _name = "account.move.line"
+    _inherit = ["account.move.line", "one.vat.mixin"]
+
+    @api.onchange("tax_ids")
+    def _onchange_tax_ids(self):
```
legalsylvain: I don't understand how you handle Intra community vat. When you are in a intra com fiscal position, there are 2 vat.

legalsylvain (commented):
Hi. 

- Some functional reviews and questions inline.

- do you know the existence of the OCA module `account_product_fiscal_classification`. Basically, it allow to have product correctly "vat" configured,(especially regarding consistency between purchase and sales tax.) but with more modularity.
(see for V16 : https://github.com/OCA/account-fiscal-rule/pull/327)

- This module should maybe go in account-fiscal-rule repo. don't you think ? CC : @OCA/accounting-maintainers.

account_tax_one_vat/models/one_vat_mixin.py
```diff
@@ -0,0 +1,38 @@
...
+class OneVatMixin(models.AbstractModel):
+    _name = "one.vat.mixin"
+    _description = (
+        "Abstract the methods needed to ensure only one VAT tax is " "selected."
+    )
+
+    def _check_only_one_vat_tax_field(self, field_name):
+        vat_group = self.env.ref("account_tax_one_vat.vat_tax_group")
+        for rec in self:
+            if getattr(rec.company_id, "account_tax_one_vat", False):
```
lmignon: `getattr` is useless... you can simply write `if rec.company_id.account_tax_one_vat`

account_tax_one_vat/models/one_vat_mixin.py
```diff
@@ -0,0 +1,38 @@
...
+                )
+                if len(vat_taxes) > 1:
+                    msg = _(
+                        "Multiple customer tax of type VAT are selected. Only one is allowed."
+                    )
+                    raise ValidationError(msg)
+
+    def _onchange_one_vat_tax_field(self, field_name):
+        """Warning if multiple VAT taxes are selected."""
+        if getattr(self.company_id, "account_tax_one_vat", False):
```
lmignon: see previous comment

lmignon:
@jbaudoux your opinion would be welcome here. I don't remember why we've this kind of restriction in our project. The 'why' this module has been developed is missing and would help to understand the 'how'.

jbaudoux:
@lmignon In Belgium, we need one and only one vat tax on each invoice line. There is the oca module  `account_invoice_tax_required` that ensures the line has a tax but it doesn't ensure it is a vat and that only one is set.
For example, it prevents invoice line with taxes like:
* recupel only
* 6% and 21%

Initially I had proposed to enhance the `account_invoice_tax_required` module but the additional complexity was refused, so this module was created.

`account_product_fiscal_classification` is helpful in the product configuration but it's not putting a hard constraint on the accounting.

account_tax_one_vat/data/account_tax_group.xml
jbaudoux: Indeed, looks cleaner than relying on an xmlid

account_tax_one_vat/models/account_move_line.py
jbaudoux: The fiscal position maps a national tax to an intracom tax. It is always a 1-1 mapping

account_tax_one_vat/models/one_vat_mixin.py
```diff
@@ -0,0 +1,38 @@
...
+    def _check_only_one_vat_tax_field(self, field_name):
+        vat_group = self.env.ref("account_tax_one_vat.vat_tax_group")
+        for rec in self:
+            if getattr(rec.company_id, "account_tax_one_vat", False):
+                vat_taxes = rec[field_name].filtered(
+                    lambda r: r.tax_group_id == vat_group
+                )
+                if len(vat_taxes) > 1:
+                    msg = _(
+                        "Multiple customer tax of type VAT are selected. Only one is allowed."
```
jbaudoux:
```suggestion
                        "Multiple taxes of type VAT are selected. Only one is allowed."
```

lmignon:
Thank you @jbaudoux for the explanation.

account_tax_one_vat/models/account_move_line.py
legalsylvain: Well, not was always the case. 

In fact : 
- in Odoo V12, the mapping is 1 -> 2. [see](https://github.com/odoo/odoo/blob/12.0/addons/l10n_fr/data/account_fiscal_position_template_data.xml#L133) tva_acq_normale -> tva_intra_normale + tva_acq_intra_normale.
- Since, V13, the mapping is 1 -> 1. [See](https://github.com/odoo/odoo/blob/16.0/addons/l10n_fr/data/account_fiscal_position_template_data.xml#L151) tva_acq_normale -> tva_intra_normale_biens. 

so your design is now OK.

lmignon:
ping @glitchov

account_tax_one_vat/models/product_template.py
```diff
@@ -0,0 +1,27 @@
...
+    @api.constrains("taxes_id")
+    def _check_only_one_vat_customer_tax(self):
+        self._check_only_one_vat_tax_field("taxes_id")
+
+    @api.constrains("supplier_taxes_id")
+    def _check_only_one_vat_supplier_tax(self):
+        self._check_only_one_vat_tax_field("supplier_taxes_id")
+
+    @api.depends("taxes_id")
+    def _compute_product_vat(self):
```
rousseldenis, line 36: You should add a depends_context("company")

account_tax_one_vat/models/product_template.py
```diff
@@ -0,0 +1,27 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class ProductTemplate(models.Model):
+    _name = "product.template"
+    _inherit = ["product.template", "one.vat.mixin"]
+
+    vat_id = fields.Many2one("account.tax", compute="_compute_product_vat")
```
rousseldenis:
```suggestion
    vat_id = fields.Many2one("account.tax", compute="_compute_product_vat", string="VAT")
```

account_tax_one_vat/models/one_vat_mixin.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+class OneVatMixin(models.AbstractModel):
+    _name = "one.vat.mixin"
+    _description = "Abstract the methods needed to ensure only one VAT tax is selected."
+
+    def _check_only_one_vat_tax_field(self, field_name):
+        for rec in self:
+            if rec.company_id.account_tax_one_vat:
+                vat_taxes = rec[field_name].filtered(lambda r: r.is_vat)
```
rousseldenis:
```suggestion
                vat_taxes = rec[field_name].filtered("is_vat")
```

rousseldenis (changes requested):
Some comments.

lmignon (approved):
LGTM (Code review)

account_tax_one_vat/models/one_vat_mixin.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+class OneVatMixin(models.AbstractModel):
+    _name = "one.vat.mixin"
+    _description = "Abstract the methods needed to ensure only one VAT tax is selected."
+
+    def _check_only_one_vat_tax_field(self, field_name):
+        for rec in self:
+            if rec.company_id.account_tax_one_vat:
+                vat_taxes = rec[field_name].filtered("is_vat")
```
jbaudoux: To support multi-company, and model like product where company_id is not required, I would do:
```
company = self.company_id or env.user.company_id
if company.account_tax_one_vat:
    vat_taxes = rec[field_name].filtered(lambda r: r.is_vat and r.company_id == company)
```

account_tax_one_vat/models/product_template.py
```diff
@@ -0,0 +1,29 @@
...
+
+from odoo import api, fields, models
+
+
+class ProductTemplate(models.Model):
+    _name = "product.template"
+    _inherit = ["product.template", "one.vat.mixin"]
+
+    vat_id = fields.Many2one(
+        "account.tax", compute="_compute_product_vat", string="VAT"
```
jbaudoux: this is company dependent

account_tax_one_vat/models/product_template.py
```diff
@@ -0,0 +1,29 @@
...
+        self._check_only_one_vat_tax_field("taxes_id")
+
+    @api.constrains("supplier_taxes_id")
+    def _check_only_one_vat_supplier_tax(self):
+        self._check_only_one_vat_tax_field("supplier_taxes_id")
+
+    @api.depends("taxes_id", "company_id")
+    def _compute_product_vat(self):
+        for rec in self:
+            vat = rec.taxes_id.filtered(lambda r: r.is_vat)
```
jbaudoux: add `env.user.company_id` in the filter

account_tax_one_vat/models/product_template.py
```diff
@@ -0,0 +1,29 @@
...
+            rec.vat_id = vat if rec.company_id.account_tax_one_vat else False
+            rec.vat = vat.name if vat and rec.company_id.account_tax_one_vat else False
```
jbaudoux: should evaluate the env company, not the record company

account_tax_one_vat/models/one_vat_mixin.py
```diff
@@ -0,0 +1,33 @@
...
+                if len(vat_taxes) > 1:
+                    msg = _(
+                        "Multiple taxes of type VAT are selected. Only one is allowed."
+                    )
+                    raise ValidationError(msg)
+
+    def _onchange_one_vat_tax_field(self, field_name):
+        """Warning if multiple VAT taxes are selected."""
+        if self.company_id.account_tax_one_vat:
```
jbaudoux: In case company_id is not set
`company = self.company_id or env.user.company_id`
or drop this if statement as it is already part of the called check

jbaudoux (commented):
Some remarks for multi-companies

glitchov:
@jbaudoux Thank you for the help on multi-companies aspects

account_tax_one_vat/models/product_template.py
rousseldenis: This is not true. As the field is purely computed, you should add a `@api.depends_context("company")`on the field.

account_tax_one_vat/models/product_template.py
rousseldenis: @jbaudoux Careful on that. company_id on the user is the default company. If we want to have the current company, we should use `self.env.company`

jbaudoux: yes, thanks for this clarification

SkiBY (approved):

account_tax_one_vat/models/product_template.py
```diff
@@ -0,0 +1,37 @@
...
+    vat = fields.Char(compute="_compute_product_vat", string="VAT name")
+
+    @api.constrains("taxes_id")
+    def _check_only_one_vat_customer_tax(self):
+        self._check_only_one_vat_tax_field("taxes_id")
+
+    @api.constrains("supplier_taxes_id")
+    def _check_only_one_vat_supplier_tax(self):
+        self._check_only_one_vat_tax_field("supplier_taxes_id")
+
```
jbaudoux, line 29: Add onchange when filling the field (like you have on account move line)

```suggestion
    @api.onchange("taxes_id")
    def _onchange_only_one_vat_customer_tax(self):
      self._onchange_one_vat_tax_field("taxes_id")
    
    @api.onchange("supplier_taxes_id")
    def _onchange_only_one_vat_supplier_tax(self):
      self._onchange_one_vat_tax_field("supplier_taxes_id")
```

account_tax_one_vat/models/account_move_line.py
jbaudoux: To avoid conflicts, name the method with what it is doing, not what is triggering it
```suggestion
    def _onchange_only_one_vat_tax(self):
```

jbaudoux (commented):
Would be nice to have a call to self._check_only_one_vat_tax_field on account move action_post like done in account_invoice_tax_required

Then we would have the warning & the constraint on both product & invoice.
Currently, there is the constraint on product and warning on invoice.

Otherwise, tested and works fine.

account_tax_one_vat/models/account_move.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    def _check_invoice_line_only_one_vat_tax(self):
+        errors = []
+        error_template = _("Invoice has a line %s with more than one vat tax")
+        loggedin_company = self.env.company
+        for invoice_line in self.mapped("invoice_line_ids").filtered(
```
jbaudoux: No need for mapped

jbaudoux (approved):
Thanks for the improvements.
Looks good, not retested

rousseldenis (approved):
Code review

rousseldenis:
@legalsylvain Could you update your review ?

account_tax_one_vat/models/account_tax.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2023 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class AccountTaxTemplate(models.Model):
```
legalsylvain:
```suggestion
class AccountTax(models.Model):
```

legalsylvain (approved):
a typo suggestion. otherwise :+1: 
thanks for the explanation.

account_tax_one_vat/models/account_tax.py
rousseldenis: @glitchov Could you fix this ? Thanks

glitchov:
@OCA/accounting-maintainers

PR #1365, opened 2023-02-01
[FIX] account_invoice_supplier_self_invoice: Use a smaller logo

etobella:
Right now, it uses a 1024 logo, and it is too large. With this change, we use a smaller logo

@rafaelbn @Shide


Shide (approved):

rafaelbn (approved):
/ocabot migration patch

PR #1369, opened 2023-02-03
16.0 mig account invoice supplier self invoice

vince-dynapps:


rafaelbn:
@moduon MT-2255

SkiBY (approved):

Shide (approved):

Shide (changes requested):
Cannot generate the PDF, so if there is an error because other oca models are installed it's ok.
Button Send & Print doesn't work too (it's generating the email, so it's logic)

Can you provide an example of the resulting pdfs?

SkiBY:
@Shide the problem is with account_invoice_triple_discount addon. There is an method in report template, is not in code.
If you uninstall it on runboat, print works
Issue is created [here](https://github.com/OCA/account-invoicing/issues/1375)

Shide (changes requested):
@vince-dynapps The report shows the same partner (the vendor) in both positions.
In the bottom position must show the company partner.

Here is the printed report: 
[self_bill_inv_v16.pdf](https://github.com/OCA/account-invoicing/files/10821329/self_bill_inv_v16.pdf)

rven:
> @vince-dynapps The report shows the same partner (the vendor) in both positions. In the bottom position must show the company partner.
> 
> Here is the printed report: [self_bill_inv_v16.pdf](https://github.com/OCA/account-invoicing/files/10821329/self_bill_inv_v16.pdf)

@Shide should be fixed now!

Shide:
@rven great!
One thing more. The partner image is not in the correct position. See:

image

account_invoice_supplier_self_invoice/views/report_self_invoice.xml
```diff
@@ -0,0 +1,207 @@
...
+        >
+            <div class="o_background_header container-fluid">
+                <t
+                    t-set="image_offset"
+                    t-value="o.partner_id.image_1024 and 'offset-2' or ''"
+                />
+                <img
+                    t-if="o.partner_id.image_1024"
+                    t-att-src="image_data_uri(o.partner_id.image_1024)"
+                    class="float-left p-0"
```
Shide: I think this need to be changed

Shide (changes requested):

account_invoice_supplier_self_invoice/views/report_self_invoice.xml
rven: Fixed, I think ;-)

Shide: Perfect! Thanks!

Shide (approved):
LGTM, all suggestions attended!
Good job!

rafaelbn:
@etobella would you like to review? 😄

rafaelbn (approved):
Tested un runboat functionally 👍

account_invoice_supplier_self_invoice/views/report_self_invoice.xml
```diff
@@ -92,14 +92,18 @@
         priority="99"
     >
         <!-- Address switcher -->
-        <xpath expr="//t[@t-set='address']" position="replace">
+        <xpath
```
etobella, line 95: It was improved on 14.0 in order to avoid the replacement, as it could give as problems with other modules, like sale

https://github.com/OCA/account-invoicing/pull/1418/files

It was not ported to 15. My error. Sorry, however, it might be necessary

etobella (changes requested):

account_invoice_supplier_self_invoice/views/report_self_invoice.xml
rven: I don't think it was really needed, because in this new version the report is split up.
The `information_block` is only shown when the delivery address is different, but this is not possible on a vendor invoice.
But of course it is better to not replace the standard logic :-)

rven: @etobella Can you verify this one?

rafaelbn:
Ping @etobella  😄

etobella (approved):
I think we can proceed with merge on this one 

/ocabot merge nobump

PR #1282, opened 2022-11-10
[14.0] [MIG] account_invoice_mass_sending

jguenat:
Standard migration

I want to propose an improvement :
Instead of a new action "Mass Send & Print" I want to inherit the standard Send&Print wizard and add a new button "Enqueue Send". This would allow to select a mail template and would be more intuitive to use (inspired by account_invoice_validation_queued)
What do you think @RobinetDenisAcsone @ThomasBinsfeld ?

If yes what would be the procedure, add a commit for the improvement in this PR or open a new one?

Thanks

PR #1382, opened 2023-02-16
[14.0][ADD] account_invoice_auto_send_by_email

TDu:


jguenat:
Hello @TDu ,

I am migrating the account_invoice_mass_sending #1282 

I want to propose an improvement there to use the wizard in a similar way you are doing here.

I'll review your PR soon.

account_invoice_auto_send_by_email/models/account_move.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    def cron_send_email_invoice(self):
+        invoices = self.search(self._email_invoice_to_send_domain())
+        for invoice in invoices:
+            description = "Send invoice {} by email".format(invoice.name)
+            invoice.with_delay(description=description)._execute_invoice_sent_wizard()
```
jguenat, line 14: Maybe it would be interesting to check if the invoice is already being processed with a boolean field "sending_in_progress"? (like its done in account_invoice_mass_sending)

TDu: Good point.
I have just added a check at the beginning of the function executed by the job (cf. fixup commit)

TDu:
Some tests are red. 
In this repository the module `account_invoice_line_sequence` has been configured as a rebel module 
https://github.com/OCA/account-invoicing/blob/0c957efd3212f3bedeb9d942c683d38d31f1a852/.copier-answers.yml#L18
But for the tests when only that particular module should be tested, every module are tested !?
This behavior does not seem to happen for other PRs of the repository.
This PR is rebased on the 14.0 branch.

jguenat (approved):

account_invoice_auto_send_by_email/models/account_move.py
sebalix: And to avoid multiple jobs to be spawn for the same invoice, you can use `identity_key` parameter with `identity_exact` helper:
```python
from odoo.addons.queue_job.job import identity_exact
[...]
invoice.with_delay(identity_key=identity_exact)._execute_invoice_sent_wizard()
```

jguenat: Good to know thanks, I'll look into it.
(though last message was from TDu^^)

TDu: Adding an identity key will prevent creating a new job if some other job with the same key have NOT RUN yet. Which in this case is not so important because of the daily schedule.
The change I made was at the beginning of the job to check that the invoice has not yet been sent, to prevent an invoice being sent multiple times (by the same job, manually or multiple jobs) to the customer.

simahawk:
> Some tests are red. In this repository the module `account_invoice_line_sequence` has been configured as a rebel module
> 
> https://github.com/OCA/account-invoicing/blob/0c957efd3212f3bedeb9d942c683d38d31f1a852/.copier-answers.yml#L18
> 
> But for the tests when only that particular module should be tested, every module are tested !?
> This behavior does not seem to happen for other PRs of the repository.
> This PR is rebased on the 14.0 branch.

Is it causing this error?
```
 2023-02-23 14:05:15,962 251 ERROR odoo odoo.addons.account_invoice_section_sale_order.tests.test_invoice_group_by_sale_order: ERROR: TestInvoiceGroupBySaleOrder.test_create_invoice
  File "/__w/account-invoicing/account-invoicing/account_invoice_section_sale_order/tests/test_invoice_group_by_sale_order.py", line 109, in test_create_invoice
    self.assertEqual(line.name, result[line.sequence][0])
KeyError: 1
```
https://github.com/OCA/account-invoicing/actions/runs/4252960824/jobs/7398120263#step:8:569

IMO we should fix that module. 
This kind of trick is fragile and ties behaviors to test cases https://github.com/OCA/account-invoicing/blob/14.0/account_invoice_line_sequence/models/account_move.py#L58
I would add a flag on the company to be set via config settings to turn on and off this behavior.
It should be off by default and its tests can enable it before running.
This way you won't have any impact on the rest of the modules -> you can remove it from rebel modules.
WDYT?

ivantodorovich (approved):

jguenat:
> IMO we should fix that module.
> This kind of trick is fragile and ties behaviors to test cases https://github.com/OCA/account-invoicing/blob/14.0/account_invoice_line_sequence/models/account_move.py#L58
> I would add a flag on the company to be set via config settings to turn on and off this behavior.
> It should be off by default and its tests can enable it before running.
> This way you won't have any impact on the rest of the modules -> you can remove it from rebel modules.
> WDYT?

Hello @simahawk ,
I tried to fix it in #1431 can you take a look? thanks

rousseldenis:
@TDu I will do the migration for v16

rousseldenis:
@TDu My first question is : shouldn't we split the code in order to have one module per transmit method ??? @jbaudoux

simahawk:
IMO this should be refactored on top of edi framework (just think of "transmit method" as "exchange type").

rousseldenis:
> IMO this should be refactored on top of edi framework (just think of "transmit method" as "exchange type").

Yes, thought about it. But it seems not yet mature to do it. @jbaudoux 

Or, EDI could implement specialized transmit method with this module as base

jbaudoux:
We should not depend on a shit load of modules for printing papers or send emails.
So I'm against having the transmit method depend on the edi framework.

We have a base module for defining the transmit method.
This module implements the sending of emails.
You are free to make another module to implement another transmit method (that could or not depend on edi framework)

According to me, this is perfect as it is and does not require refactoring. We should merge https://github.com/OCA/account-invoicing/pull/1431https://github.com/OCA/account-invoicing/pull/1431 and this PR

jbaudoux (approved):

simahawk:
> We should not depend on a shit load of modules for printing papers or send emails. So I'm against having the transmit method depend on the edi framework.
> We have a base module for defining the transmit method. This module implements the sending of emails. You are free to make another module to implement another transmit method (that could or not depend on edi framework)
> According to me, this is perfect as it is and does not require refactoring. We should merge #1431 and this PR

You are not looking at it from the right side IMO. You are already maintaining many modules that do a piece of what the framework does. It implements its own UI, it implements its own models and workflows. If you apply the same concept to all the similar use cases you'll end up with w/ tons of modules anyway. And all of them specific to their own use cases, own models, own workflows, etc, etc. 

So, I buy the fact that you want to reuse something that is already here, it works, no refactoring, bla bla bla, but I don't by at all the "shit load of modules" argument.

That being said, I missed to specify above, that I was talking about the migration to 16.0, not about this PR so... splitting the module - as @rousseldenis said - here in 14.0 is not worth. Again, in my opinion :wink:

simahawk:
@TDu squash is needed and build is red: could you have a look pls?

TDu:
> @TDu My first question is : shouldn't we split the code in order to have one module per transmit method ???

Well do you need this feature for other transmit method in a real case scenario ? If not I would not refactor.
In the Odoo implementation I know, other transmit method are handled differently (click of a button, change of state,...)

mt-software-de:
Thanks. 
What is the state of the PR is there any change to merge this in the near future?

TDu:
In my view this is ready to be merged. This module is being used on a live instance.

jguenat:
I rebased #1431 in case you still need it

mt-software-de (approved):
LGTM

jbaudoux:
@OCA/accounting-maintainers Can we get this simple module merged?

account_invoice_auto_send_by_email/tests/test_account_invoice_auto_send_by_email.py
```diff
@@ -0,0 +1,112 @@
+# Copyright 2023 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+import mock
+
+from odoo.tests.common import SavepointCase
+
+
+class TestAccountInvoiceAutoSendByEmail(SavepointCase):
```
grindtildeath, line 8: Would be nice to use an accounting test case instead of having to spend to much code on creating accounts and journals in the setup.

account_invoice_auto_send_by_email/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+{
+    "name": "Account Invoice Auto Send By Email",
+    "summary": "Invoice with the email transmit method are send automatically.",
+    "version": "14.0.1.0.0",
+    "category": "Accounting",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "depends": ["account", "account_invoice_transmit_method", "queue_job"],
```
grindtildeath, line 11: I'm wondering, couldn't this module avoid depending on queue job and send the emails with the cron? Then could be extended by another module to generate the jobs?

simahawk:
Here on v14 I'd keep the module as it is as it's used in prod since quite some time.
Further improvements can be considered for later.
Likely EDI framework will be used starting from v18.

/ocabot merge nobump

edescalona:
Hi @rousseldenis did you start migrating to v16 or any other higher version?

PR #1383, opened 2023-02-16
[15.0][FIX] sale_line_refund_to_invoice_qty, fix default value on aml

ChrisOForgeFlow:
@ForgeFlow

### DIFF
--- a/sale_line_refund_to_invoice_qty/models/account.py
+++ b/sale_line_refund_to_invoice_qty/models/account.py
@@ -12,9 +12,9 @@ class AccountMove(models.Model):
         move_vals = super(AccountMove, self)._reverse_move_vals(
             default_values, cancel=cancel
         )
-        if self.env.context.get("sale_qty_to_reinvoice", False):
-            for vals in move_vals["line_ids"]:
-                vals[2].update({"sale_qty_to_reinvoice": True})
+        sale_qty_to_reinvoice = self.env.context.get("sale_qty_to_reinvoice", False)
+        for vals in move_vals["line_ids"]:
+            vals[2].update({"sale_qty_to_reinvoice": sale_qty_to_reinvoice})
         return move_vals
 
 
@@ -22,5 +22,6 @@ class AccountMoveLine(models.Model):
     _inherit = "account.move.line"
 
     sale_qty_to_reinvoice = fields.Boolean(
+        default=True,
         help="Leave it marked if you will reinvoice the same sale order line",
     )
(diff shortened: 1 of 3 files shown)


sale_line_refund_to_invoice_qty/models/account.py
```diff
@@ -12,15 +12,17 @@ def _reverse_move_vals(self, default_values, cancel=True):
...
+        for vals in move_vals["line_ids"]:
+            vals[2].update({"sale_qty_to_reinvoice": sale_qty_to_reinvoice})
         return move_vals
 
 
 class AccountMoveLine(models.Model):
     _inherit = "account.move.line"
 
     sale_qty_to_reinvoice = fields.Boolean(
+        string="Sale qty to reinvoice",
```
RodrigoBM: this is not necessary.
```suggestion
```
https://github.com/OCA/OCB/blob/16.0/odoo/fields.py#L507

RodrigoBM (changes requested):

sale_line_refund_to_invoice_qty/models/account.py
ChrisOForgeFlow: Done!

RodrigoBM (changes requested):
Hi @ChrisOForgeFlow ,

If you launch a wizard credit note and uncheck "This credit note will be reinvoiced" and  check credit method "Full refund and new draft invoice"

Selección_434

A wrong calculation of the billed lines is obtained.

Selección_435

if option "Full refund and new draft invoice" is selected, the value of the field "This credit note will be reinvoiced must always be True, and in my opinion it should be hidden.

RodrigoBM:
> Hi @ChrisOForgeFlow ,
> 
> If you launch a wizard credit note and uncheck "This credit note will be reinvoiced" and check credit method "Full refund and new draft invoice"
> 
> Selección_434
> 
> A wrong calculation of the billed lines is obtained.
> 
> Selección_435
> 
> if option "Full refund and new draft invoice" is selected, the value of the field "This credit note will be reinvoiced must always be True, and in my opinion it should be hidden.

```
def reverse_moves(self):
    sale_qty_to_reinvoice = True if self.refund_method == "modify" else self.sale_qty_to_reinvoice
    return super(
        AccountMoveReversal,
        self.with_context(sale_qty_to_reinvoice=sale_qty_to_reinvoice),
    ).reverse_moves()
```

RodrigoBM:
ping @ChrisOForgeFlow

PR #1384, opened 2023-02-17
[14.0][ADD] sale_order_invoicing_qty_percentage

pedrobaeza:
**Sales order invoicing by percentage of the quantity**

This module allows to invoice a percentage of the sales order, modifying invoiced quantities according such percentage.

Example: You have a service line to invoice 10 hours. You choose to invoice a 50% of the quantity, so the generated invoice line has 5 hours.

@Tecnativa TT41746


sale_order_invoicing_qty_percentage/models/sale_order.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order.line"
+
+    def _prepare_invoice_line(self, **optional_values):
+        """If invoicing by quantity percentage, modify quantities."""
+        res = super()._prepare_invoice_line(**optional_values)
+        if self.env.context.get("qty_percentage"):
+            res["quantity"] *= self.env.context["qty_percentage"]
```
carlosdauden, line 14: Shouldn't we use float_round to avoid potential rounding issues?

carlosdauden (changes requested):

sale_order_invoicing_qty_percentage/models/sale_order.py
pedrobaeza: AFAIK, on create, values are rounded automatically by the precision unit, so no need to do that.

carlosdauden (approved):

sergio-teruel (approved):

PR #1399, opened 2023-02-27
[16.0][FIX] account_invoice_refund_link - correct link creation

pedrocs-exo:
The current implementation creates incorrect links between refund lines and their origin invoice lines. This patch fixes it. It's a very small and obvious change so it should be easy to review and merge.

### DIFF
--- a/account_invoice_refund_link/models/account_move.py
+++ b/account_invoice_refund_link/models/account_move.py
@@ -23,7 +23,11 @@ class AccountMove(models.Model):
                     refund_lines = move.line_ids.filtered(
                         lambda x: x.display_type == "product"
                     )
-                    for i, line in enumerate(self.invoice_line_ids):
+                    for i, line in enumerate(
+                        self.invoice_line_ids.filtered(
+                            lambda x: x.display_type == "product"
+                        )
+                    ):
                         if i < len(refund_lines):
                             refund_lines[i].origin_line_id = line.id
         return reverse_moves
--- a/account_invoice_refund_link/tests/test_invoice_refund_link.py
+++ b/account_invoice_refund_link/tests/test_invoice_refund_link.py
@@ -36,6 +36,14 @@ class TestInvoiceRefundLinkBase(TransactionCase):
                 0,
                 False,
                 {
+                    "name": "Test section",
+                    "display_type": "line_section",
+                },
+            ),
+            (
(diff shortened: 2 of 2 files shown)


pedrobaeza:
Please squash in one commit and change the commit message to:
```
[FIX] account_invoice_refund_link: Exclude section/notes origin lines

On the side by side comparison, exclude the lines that are not product, the 
same we do for the refund lines.
```

rafaelbn:
Hello @pedrocasi , thank you very much.

rafaelbn:
@ljsalvatierra-factorlibre could you please review? As you migrate it this could be interenting for you 😄  thank you!

ljsalvatierra-factorlibre:
> @ljsalvatierra-factorlibre could you please review? As you migrate it this could be interenting for you smile thank you!

Yes, of course.

Thank you @pedrocasi, could you please add unittests? Sorry for the trouble.

pedrocs-exo:
Added a line section on the unittests reference invoice and adapted the assertions to match.

ljsalvatierra-factorlibre:
> Added a line section on the unittests reference invoice and adapted the assertions to match.

Thank you, i'll take a look tomorrow morning

ljsalvatierra-factorlibre (approved):
Thank you for the fix!

pedrobaeza (approved):
/ocabot merge patch

I suppose this problem is the same in lower versions. Can you please backport it?

pedrocs-exo:
@pedrobaeza I could backport it but it's not that simple. Before 16.0, there are 3 types of invoices lines: sections, notes or regular lines. All of them link with refund lines and it's working fine. On 16.0 you have other line options, like payment terms, taxes, rounding, etc. The person that migrated the module (@ljsalvatierra-factorlibre ) decided to establish the link only on regular lines (display_type == 'product') and I tend to agree (to be honest I'm not familiar with the implications of the other line types so I prefer to stick with the safe behaviours).

With that being said, in my opinion we shouldn't backport. What do you think?

pedrobaeza:
OK then. Let's keep it for now.

pedrobaeza:
On #1526, I'm refactoring the way of matching the refund lines against the origin ones, and I see no reason for not linking section and the rest of the lines as well. Please check.

PR #1400, opened 2023-02-28
[16.0][MIG] account_invoice_merge

nguyenminhchien:


account_invoice_merge/models/account_move.py
```diff
@@ -107,7 +108,11 @@ def make_key(br, fields):
                     or field == "tax_ids"
                     or field == "sale_line_ids"
                 ):
-                    field_val = ((6, 0, tuple(v.id for v in field_val)),)
+                    # field_val = ((6, 0, tuple(v.id for v in field_val)))
+                    field_val = tuple(v.id for v in field_val)
```
mmequignon:
```python
                    field_val = fields.Commands.set(field_val.ids)
```
according to [odoo documentation](https://www.odoo.com/documentation/16.0/developer/reference/backend/orm.html#odoo.models.Model.write)

account_invoice_merge/tests/test_account_invoice_merge.py
```diff
@@ -1,15 +1,62 @@
...
+        )
+        cls.invoice6 = cls.init_invoice(
+            "out_invoice",
+            partner=cls.partner_a,
+            products=cls.product_a,
+            invoice_date=invoice_date,
+            company=cls.company,
+        )
+
     def setUp(self):
```
mmequignon: I think this should go in `setUpClass`

account_invoice_merge/models/account_move.py
```diff
@@ -107,7 +108,11 @@ def make_key(br, fields):
                     or field == "tax_ids"
```
mmequignon: I don't like this `make_key` implementation.
Instead of checking value types with getattr, I would use smth like `type(br._fields[field])`.
There, depending on the type (Selection, Many2many,…) you can decide what to do.
```python
if type(br._fields[field]) == fields.Many2many:
    records = br.get(field)
    value = set(records.ids) if records else False
```
You can do that for the special cases, but for most of them, you only need to use:
```python
value = br.get(field, False)
```

mmequignon (commented):
We need to split the `do_merge` method.

account_invoice_merge/models/account_move.py
mmequignon: Also, I would say that this `make_key` func should be extracted, and tested.

mmequignon:
quickly checked from runboat, looks like its working :blush:

account_invoice_merge/models/account_move.py
nguyenminhchien: @mmequignon  your suggestion will cause this error :
```
  File "/account-invoicing/16.0/account_invoice_merge/models/account_move.py", line 171, in do_merge
    o_line = invoice_infos["invoice_line_ids"].setdefault(line_key, {})
TypeError: unhashable type: 'list'
```

mmequignon: yep, my bad, this should be a list of field.Commands
```suggestion
                    field_val = [fields.Commands.set(field_val.ids)]
```

account_invoice_merge/tests/test_account_invoice_merge.py
nguyenminhchien: Updated

account_invoice_merge/models/account_move.py
nguyenminhchien: @mmequignon i have refactored that function

StephaneMangin (approved):

account_invoice_merge/models/account_move.py
```diff
@@ -64,10 +66,39 @@ def _get_first_invoice_fields(self, invoice):
...
+        """
+        Return a hashable key
+        """
+        list_key = []
+        for field in fields:
+            field_val = getattr(br, field)
+            if isinstance(field_val, dict):
+                field_val = str(field_val)
+            elif isinstance(field_val, models.Model):
+                field_val = tuple(sorted(field_val.ids))
```
mmequignon, line 97: I don't think this `sorted()` should be used, here. You probably break the `_order` defined on the model itself.

account_invoice_merge/models/account_move.py
```diff
@@ -64,10 +66,39 @@ def _get_first_invoice_fields(self, invoice):
...
+        for field_name, val in origin_vals.items():
+            if field_name in field_names:
+                vals[field_name] = val
+        return vals
+
     def _get_draft_invoices(self):
         """Overridable function to return draft invoices to merge"""
         return self.filtered(lambda x: x.state == "draft")
 
+    def make_key(self, br, fields):
```
mmequignon, line 87: this method should probably be tested.
For a given list of fields:
`["m2o_field", "m2m_field", "o2m_field", "char_field", ...]`
Should return something like that
```python
{
"m2o_field": ID,
"m2m_field": [fields.Commands.set(IDS)],
"o2m_field": [fields.Commands.set(IDS)],
"char_field": VALUE,
…
}
```
Thanks :pray:

account_invoice_merge/models/account_move.py
nguyenminhchien: "_order" is useful when reading values, but here i just want to make a unique key.

account_invoice_merge/models/account_move.py
nguyenminhchien: No: it does not care the type of fields but the type of values.
Yes: i will add the test for 3 types: Dict, Model, Other

mmequignon: I had to check… Looks like odoo became very permissive about how to write on relational fields!
```python
>>> so
sale.order(1,)
>>> so2
sale.order(2,)
>>> so.order_line
sale.order.line(1, 2)
>>> lines = so2.order_line
>>> lines
sale.order.line(3, 4, 5)
>>> so.write({"order_line": [fields.Command.set(lines.ids)]})
True
>>> so.write({"order_line": tuple(lines.ids)})
True
>>> so.write({"partner_id": so2.partner_id.id})
True
>>> so.write({"partner_id": tuple(so2.partner_id.ids)})
True
```
My bad, I wasn't aware about this… :pray:

mmequignon: And I was also confused about the usage of this make key method. Sorry about that

mmequignon (approved):

baimont (approved):
Code review, LGTM

PR #1403, opened 2023-03-01
[15.0][mig]-account_invoice_line_description

JuanyDForgeflow:


MiquelRForgeFlow (approved):

AaronHForgeFlow (commented):
According to the code the description in the product is the only thing that should appear in the invoice line as label. However it is appearing a compound name, for example, for a sales order of a product with name "Widget" and description for sales "this is a widget" the label in the invoice line is "Widget this is a widget". Is that intended?

rafaelbn:
Please @JuanyDForgeflow could you answer @AaronHForgeFlow ?

AaronHForgeFlow (changes requested):
@rafaelbn I checked and in v14 the behaviour is not the same, the product label is not in the description. So that must be fixed.

AaronHForgeFlow (approved):
Last commit makes this to behave same as v14. I will wait for others to review before merging it.

PR #1405, opened 2023-03-02
[IMP] account_invoice_supplierinfo_update_*_discount : Do not display discounts column if all the values are null

legalsylvain:
**Rational :**
purchase_discount and purchase_triple_discount exists to handle invoices of some suppliers that apply up to 3 levels of discounts.
But in most cases, suppliers doesn't provide discounts level, so, the display of column with null value is useless.

This trivial PR hide discount column if no discount is applied.


legalsylvain:
patch deployed since 6 monthes without problem.

/ocabot merge patch

quentinDupont (approved):
Code review and works on prod :heavy_check_mark:

legalsylvain:
- Used in production since 1 year.
- 1 approval
- accepted in V14

/ocabot merge minor

PR #1417, opened 2023-03-14
[FIX]account_invoice_refund_link: Fix index overflow.

javierjcf:
Fixed error when we use the module **account_invoice_refund_link** along with **account_invocie_refund_line_selection**.


pedrobaeza:
The problem is that the result is not correct although you don't have an error. They are simply incompatible.

pedrobaeza:
Closing in favor of #1523

PR #1418, opened 2023-03-14
[FIX] account_invoice_supplier_self_invoice: Fix printing when using delivery address

etobella:
We cannot override the address parameter as it is modified in other reports. The only way to keep the logic is the following change

@rafaelbn @Shide

### DIFF
--- a/account_invoice_supplier_self_invoice/views/report_self_invoice.xml
+++ b/account_invoice_supplier_self_invoice/views/report_self_invoice.xml
@@ -98,8 +98,9 @@
         priority="99"
     >
         <!-- Address switcher -->
-        <xpath expr="//t[@t-set='address']" position="replace">
+        <xpath expr="//t[@t-set='address']" position="after">
             <t t-if="o.set_self_invoice">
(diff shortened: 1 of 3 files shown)


account_invoice_supplier_self_invoice/views/report_self_invoice.xml
```diff
@@ -98,8 +98,9 @@
         priority="99"
     >
         <!-- Address switcher -->
-        <xpath expr="//t[@t-set='address']" position="replace">
+        <xpath expr="//t[@t-set='address']" position="after">
             <t t-if="o.set_self_invoice">
+                <t t-set="information_block" />
```
etobella, line 103: We need to override information_block too as it is modified by sale

etobella:
The third commits tries to solve the following error:

If you select a partner with self_invoice, the boolean to uncheck if you don't want self invoice appear, but it disappears when we refresh

kluna1998 (approved):

sanchonuria (approved):

PR #1419, opened 2023-03-14
[MIG] timesheet_invoice_description migration to 16.0

jasperjumelet:


azoellner:
(@jasperjumelet) This PR became a duplicate of the (more up-to-date) #1475, and should thus probably be closed.

PR #1420, opened 2023-03-15
[ADD] account_invoice_line_complimentary

SirTakobi:


account_invoice_change_currency/models/account_change_currency.py
sergiocorato, line 1: This change is needed by the new module?

SirTakobi: Thanks for having a look!
Yes it is needed by the unit test contained in the new module.
Without this change, the unit test fails: you can see how in the test run of the previous commit (https://github.com/OCA/account-invoicing/actions/runs/4426084114/jobs/7761937518#step:8:236) and in the message of the commit containing this change.

sergiocorato (approved):
LGTM
Thanks!

account_invoice_line_complimentary/models/res_company.py
```diff
@@ -0,0 +1,13 @@
...
+
+from odoo import fields, models
+
+
+class ResCompany (models.Model):
+    _inherit = 'res.company'
+
+    complimentary_account_id = fields.Many2one(
+        comodel_name='account.account',
+        string="Complimentary Account",
```
DucTruongKomit: I think we don't need the string here

DucTruongKomit (commented):
Hi! Here just 1 small comment!

account_invoice_line_complimentary/models/res_company.py
SirTakobi: :wave:
Thanks for having a look!
You are right, I have removed it.

mrcast (approved):
Functional test, ok!

eLBati (approved):
Thanks

DucTruongKomit (approved):
It should be okay now!

PR #1422, opened 2023-03-17
[13.0][FIX] account_invoice_supplierinfo_update: migration script

chienandalu:
We have to drop the foreign key so its recreated with the right
references to move_id

cc @Tecnativa TT38570

please review @sergio-teruel @pedrobaeza

### DIFF
--- a/account_invoice_supplierinfo_update/migrations/13.0.1.0.0/pre-migration.py
+++ b/account_invoice_supplierinfo_update/migrations/13.0.1.0.0/pre-migration.py
@@ -1,0 +1,19 @@
+# Copyright 2023 Tecnativa - David Vidal
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        ALTER TABLE wizard_update_invoice_supplierinfo
+        DROP CONSTRAINT IF EXISTS wizard_update_invoice_supplierinfo_invoice_id_fkey
+        """,
+    )
+    # Prevent having records that weren't vacuumed as they could provoke a key error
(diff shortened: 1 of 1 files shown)


legalsylvain (approved):
Thanks.

chienandalu:
Still working on it, please don't merge yet :)

chienandalu:
Ready for review :)

account_invoice_supplierinfo_update/migrations/13.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,20 @@
...
+        ALTER TABLE wizard_update_invoice_supplierinfo
+        DROP CONSTRAINT IF EXISTS wizard_update_invoice_supplierinfo_invoice_id_fkey
+        """,
+    )
+    # Prevent having records that weren't vacuumed as they could provoke a key error
+    # for the new constraint
+    openupgrade.logged_query(
+        env.cr, "DELETE FROM wizard_update_invoice_supplierinfo_line"
+    )
+    openupgrade.logged_query(env.cr, "DELETE FROM wizard_update_invoice_supplierinfo")
```
pedrobaeza: Do `TRUNCATE wizard_update_invoice_supplierinfo CASCADE`, and prevent any problem with extensions.

pedrobaeza (commented):
This would be `[OU-ADD]` instead of `[FIX]`.

There's a debate about doing an automatic cleaning of all the transient tables before doing other things in OpenUpgrade.

account_invoice_supplierinfo_update/migrations/13.0.1.0.0/pre-migration.py
chienandalu: Done

pedrobaeza (approved):

pedrobaeza:
Please rename the tag next time as indicated.

PR #1267, opened 2022-10-18
[14.0] ADD account_receipt_sale

eLBati:
Depends on https://github.com/OCA/account-invoicing/pull/1260 and https://github.com/OCA/account-invoicing/pull/1427

PR #1427, opened 2023-03-28
[ADD] account_receipt_base

SirTakobi:


rafaelbn:
Please @SirTakobi could you explain a little bit the purpose of this module, really interested 😄 cc @Shide

SirTakobi:
> Please @SirTakobi could you explain a little bit the purpose of this module, really interested :smile: cc @Shide

Thanks for having a look!
The purpose of this module is to be the base for other modules that improve Receipts Management in `14.0`, for instance:

- refund Receipts: https://github.com/OCA/account-invoicing/pull/1428
- Sale Orders generating Receipts instead of Invoices: https://github.com/OCA/account-invoicing/pull/1267
  the linked module currently does not need this module, but it probably might use delegating generic stuff to this base module

Here we are also planning to add minor improvements that might not deserve a dedicated module, like:

- the Receipts button in the partner (similar to Invoices button),
- showing Receipts in the Invoices Report

rafaelbn:
OK! Great and thank you very much @SirTakobi for explanation.

- [ ] We need that receipt can be created in a Account Journal Type | Miscellaneous

Do you think this is something we can improve in account_receipt_journal or you would prefer we create a new module?

The need is that in receipts, in the receipt lines, we need use more accounts type than the ones are now allowed by Odoo in Journes type "Sale"

cc @Shide

SirTakobi:
> OK! Great and thank you very much @SirTakobi for explanation.
> 
> * [ ]  We need that receipt can be created in a Account Journal Type | Miscellaneous
> 
> Do you think this is something we can improve in account_receipt_journal or you would prefer we create a new module?
> 
> The need is that in receipts, in the receipt lines, we need use more accounts type than the ones are now allowed by Odoo in Journes type "Sale"
> 
> cc @Shide

@eLBati what do you think about this?

eLBati:
Ciao @rafaelbn 
having `account_receipt_journal` a specific name, I think the modification of domain in receipt lines could go in another place, like another module or even an option in `account_receipt_base`

rafaelbn:
Hello Lorenzo! 

> Ciao @rafaelbn
> having account_receipt_journal a specific name, I think the modification of domain in receipt lines could go in another place, like another module or even an option in account_receipt_base

OK, I will think about this one

another need @eLBati @SirTakobi that we can add to this module is that we can define

- [ ] If receipts by default have empty tax or
- [ ] Receipt journal by default set an specific tax to all lines of all receipts

This is because receipts are not invoices therefore fiscally shouldn't be reported to administrations in the same way and Mainly could don't apply any tax

Thank you!

eLBati:
Hello @rafaelbn 
I think these needs could be covered later and we could merge this module as is.
What do you think?

tafaRU:
@rafaelbn please see https://github.com/OCA/account-invoicing/pull/1427#issuecomment-1692975338

rafaelbn (approved):
👍🏼

primes2h:
@SirTakobi @eLBati 

image

Could you please force a rebuild?

eLBati:
cc @siraiontech

SirAionTech:
> @SirTakobi @eLBati
> 
> image
> 
> Could you please force a rebuild?

> cc @SirAionTech

@SirTakobi is no more used, so I superseded with #1551; this can be closed.

PR #1431, opened 2023-03-30
[14.0][FIX] account _invoice_line_sequence : use a flag on company to avoid breaking tests

jguenat:
- false by default to avoid breaking tests of other modules #1382

### DIFF
--- a/account_invoice_line_sequence/__manifest__.py
+++ b/account_invoice_line_sequence/__manifest__.py
@@ -5,7 +5,7 @@
 
 {
     "name": "Account Invoice line with sequence number",
-    "version": "14.0.1.0.1",
+    "version": "14.0.2.0.1",
     "summary": "Adds sequence field on invoice lines to manage its order.",
     "category": "Accounting",
     "author": "Camptocamp, "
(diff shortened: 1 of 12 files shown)


TDu (approved):
LG, thank you for doing that.

account_invoice_line_sequence/__manifest__.py
```diff
@@ -5,15 +5,19 @@
 
 {
     "name": "Account Invoice line with sequence number",
-    "version": "14.0.1.0.1",
+    "version": "14.0.2.0.1",
```
jbaudoux, line 8: The migration script should not be in a directory including the last digit ?
"14.0.2.0.1" vs "14.0.2.0"

jguenat: indeed, corrected
It was working but it's nicer like that

jbaudoux (approved):

mt-software-de (approved):
LGTM

bosd:
Why Introduce a new setting while the goal is to fix test compatibility?

Why not use a conditional context on the default test value?

jguenat:
> Why Introduce a new setting while the goal is to fix test compatibility?
> 
> Why not use a conditional context on the default test value?

it's discussed here https://github.com/OCA/account-invoicing/pull/1382#issuecomment-1455664425

bosd:
> it's discussed here [#1382 (comment)](https://github.com/OCA/account-invoicing/pull/1382#issuecomment-1455664425)

Thanks fro the clarification.

AaronHForgeFlow (approved):
Code LG

hbrunn:
ocabot is merging for quite a while already, so I take the opportunity to ask: Did you try to just not write the sequence field? I've disentangled sequence and sequence2 in the v15 (#1565) migration, and I think this is actually a better fix

AaronHForgeFlow:
The merge is aborted. Better to backport the refactor from https://github.com/OCA/account-invoicing/pull/1565 at this point. Otherwise 13.0, 14.0 and 15.0 would have very different code.

dreispt:
Superseded by https://github.com/OCA/account-invoicing/pull/1565

PR #1433, opened 2023-04-04
[16.0][ADD] account_tax_change

sebalix:
Declare your tax changes starting from a date, and apply these changes on your existing invoices.

The following glue module has also been created to be compatible with `account_invoice_start_end_dates`:
- https://github.com/OCA/account-closing/pull/243

PR #1435, opened 2023-04-07
[10.0][ADD] account_tax_change

sebalix:
Backport of:
- https://github.com/OCA/account-invoicing/pull/1433


account_tax_change/models/account_tax_change.py
```diff
@@ -0,0 +1,62 @@
...
+        comodel_name="account.tax",
+        ondelete="cascade",
+        string="From tax",
+        required=True,
+        index=True,
+    )
+    type_tax_use = fields.Selection(
+        related="tax_src_id.type_tax_use",
+    )
+    tax_dest_id = fields.Many2one(
```
grindtildeath, line 59: I guess we need a constraint to somehow make sure `tax_src_id` and `tax_dest_id` have the same `type_tax_use`.

account_tax_change/security/ir.model.access.csv
```diff
@@ -0,0 +1,9 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_account_tax_change_internal_user,account.tax.change internal user,model_account_tax_change,base.group_user,1,0,0,0
+access_account_tax_change_line_internal_user,account.tax.change.line internal user,model_account_tax_change_line,base.group_user,1,0,0,0
+access_account_tax_change_invoice,account.tax.change,model_account_tax_change,account.group_account_invoice,1,0,0,0
+access_account_tax_change_line_invoice,account.tax.change.line,model_account_tax_change_line,account.group_account_invoice,1,0,0,0
+access_account_tax_change_manager,account.tax.change,model_account_tax_change,account.group_account_manager,1,1,1,1
+access_account_tax_change_line_manager,account.tax.change.line,model_account_tax_change_line,account.group_account_manager,1,1,1,1
+access_account_move_apply_tax_change_invoice,account.move.apply.tax.change,model_account_move_apply_tax_change,account.group_account_invoice,1,1,1,1
+access_account_move_apply_tax_change_manager,account.move.apply.tax.change.manager,model_account_move_apply_tax_change,account.group_account_manager,1,1,1,1
```
grindtildeath: IIRC there are no access rights for wizards in v10.0

account_tax_change/wizards/account_move_apply_tax_change.py
```diff
@@ -0,0 +1,95 @@
...
+            " ('company_id', '=', company_id),"
+            "]"
+        ),
+    )
+
+    @api.multi
+    def validate(self):
+        """Apply the tax changes on the selected invoices."""
+        self.ensure_one()
+        src_taxes = self.tax_change_id.change_line_ids.tax_src_id
```
grindtildeath: I guess we need a `mapped` here in v10

account_tax_change/wizards/account_move_apply_tax_change.py
```diff
@@ -0,0 +1,95 @@
...
+                    continue
+                # - invoice date doesn't match
+                if self._skip_line(line):
+                    continue
+                self._change_taxes(line)
+        return True
+
+    def _skip_invoice(self, invoice):
+        """No tax change on invoice already posted or paid."""
+        return invoice.state == "posted"
```
grindtildeath:
```suggestion
        return invoice.state in ("open", "paid", "cancel")
```

account_tax_change/wizards/account_move_apply_tax_change.py
```diff
@@ -0,0 +1,95 @@
...
+        required=True,
+        string="Tax Change",
+        domain="[('company_id', '=', company_id)]",
+    )
+    invoice_ids = fields.Many2many(
+        comodel_name="account.invoice",
+        required=True,
+        string="Invoices",
+        domain=(
+            "[('state', 'in', ('draft', 'open')),"
```
grindtildeath: I guess we could allow `proforma` and `proforma2` here (https://github.com/odoo/odoo/blob/10.0/addons/account/models/account_invoice.py#L228) :thinking:

account_tax_change/wizards/account_move_apply_tax_change.py
```diff
@@ -0,0 +1,95 @@
...
+        active_model = self.env.context.get("active_model")
+        if active_model == "account.invoice":
+            invoice_ids = self.env.context.get("active_ids")
+            invoices = self.env[active_model].search(
+                [
+                    ("id", "in", invoice_ids),
+                    ("state", "in", ("draft", "open")),
+                ]
+            )
+            res["invoice_ids"] = [(6, 0, invoices.ids)]
```
grindtildeath: IMO we should raise an error if one of the selected invoices is not in an "allowed" state.

account_tax_change/wizards/account_move_apply_tax_change.py
```diff
@@ -0,0 +1,95 @@
...
+        return invoice.state == "posted"
+
+    def _skip_line(self, line):
+        """No tax change on line if the invoice date doesn't match.
+
+        Other modules could override this method.
+        """
+        return (
+            not line.invoice_id.date_invoice
+            or line.invoice_id.date_invoice < self.tax_change_id.date
```
grindtildeath: IMO this should also go in `_skip_invoice` as the date is not defined on the lines. (but it's good to keep the hook)

grindtildeath (commented):
Few things to improve, and missing utf-8 headers for python2

account_tax_change/wizards/account_move_apply_tax_change.py
grindtildeath: Actually thinking through this one is tricky because the application of tax change will depend if the product that is sold is supplied after the tax change date, and we cannot know this.
A same product could be:
 - sold in 2022 and supplied in 2022 (in such case no tax change is required) 
 - sold in 2022 and supplied in 2023 (in such case tax change is required)

IMO let's remove the condition on the invoice date, as the user has to know what he's doing.

sebalix: I have some doubts about this, the tax to apply isn't based on the invoice date, whatever the shipping date is?

sebalix: It depends... reading this for France: https://www.demarches.interieur.gouv.fr/professionnels/quelles-dates-exigibilite-matiere-tva
so you are right, better to remove this constraint and let the user decide.

account_tax_change/wizards/account_move_apply_tax_change.py
sebalix: We should skip `open` invoices in your opinion? In the `invoice_ids` M2M above I added a domain that allows them.
EDIT: And I added a `_check_invoices` used by `default_get` as well to check the invoices state, so this `skip_invoice` will do nothing now, but we could keep it as a hook.

sebalix:
@grindtildeath I addressed all your remarks, what was relevant for v16 has been done there first (https://github.com/OCA/account-invoicing/pull/1433) as fixups and backported here + adapted if needed in the last commit.

account_tax_change/wizards/account_move_apply_tax_change.py
grindtildeath: Open invoices are the "posted" ones which did already generate an account.move, so in case we modify such an invoice, we'd have to either modify or regenerate the move. I guess it's then easier to restrict application to invoices that did not generate a move.

account_tax_change/wizards/account_move_apply_tax_change.py
```diff
@@ -0,0 +1,107 @@
...
+                self._change_taxes(line)
+        return True
+
+    def _skip_invoice(self, invoice):
+        """No tax change on invoice already posted or paid."""
+        # FIXME: this method does nothing now we have '_check_invoices'
+        return invoice.state in ("open", "paid", "cancel")
+
+    def _skip_line(self, line):
+        """Return `True` if the tax change should be applied on the line.
```
grindtildeath: docstring to fix. Shouldn't we return false as default instead of an implicit None?

grindtildeath (commented):
Just wondering, shouldn't we change the module name to be a bit more specific?
i.e. it's changing taxes on invoices and not on changing the tax themselves, so maybe something like account_invoice_tax_change?

account_tax_change/wizards/account_move_apply_tax_change.py
sebalix: We could return False yes

account_tax_change/wizards/account_move_apply_tax_change.py
sebalix: Yes I agree, it's easier for me ;) But was wondering if it was correct as it probably depends also on the localization.

account_tax_change/wizards/account_move_apply_tax_change.py
```diff
@@ -0,0 +1,108 @@
...
+        """
+        return False
+
+    def _change_taxes(self, line):
+        """Apply the tax change in the invoice line.
+
+        Other modules could override this method.
+        """
+        line.invoice_line_tax_ids -= self.tax_change_id.change_line_ids.tax_src_id
+        line.invoice_line_tax_ids |= self.tax_change_id.change_line_ids.tax_dest_id
```
grindtildeath:
```suggestion
        for tax_change_line in self.tax_change_id.change_line_ids:
            if tax_change_line.tax_src_id in line.invoice_line_tax_ids:
                line.invoice_line_tax_ids -= tax_change_line.tax_src_id
                line.invoice_line_tax_ids |= tax_change_line.tax_dest_id
```

sebalix: fixed

PR #1442, opened 2023-04-17
[14.0][IMP]Account_invoice_triple_discount: For in invoice_line_ids and not in line_ids

odooNextev:
We need this fix to avoid issues with `l10n_it_fatturapa_in` (https://github.com/OCA/l10n-italy/tree/14.0/l10n_it_fatturapa_in).
In this module there's a test that trying to import a refund with negative amounts and without this fix it fails cause tax sign it's not reversed unlike price unit.
So we analyzed how taxes are recomputed in 16.0 version of `account_invoice_triple_discount` and we found something different that adapted and applied to 14.0 version should no longer get issues with negative amounts.

### DIFF
--- a/account_invoice_triple_discount/models/account_move.py
+++ b/account_invoice_triple_discount/models/account_move.py
@@ -16,9 +16,9 @@ class AccountMove(models.Model):
         restored after the original process is done
         """
         old_values_by_line_id = {}
-        digits = self.line_ids._fields["price_unit"]._digits
-        self.line_ids._fields["price_unit"]._digits = (16, 16)
-        for line in self.line_ids:
+        digits = self.invoice_line_ids._fields["price_unit"]._digits
+        self.invoice_line_ids._fields["price_unit"]._digits = (16, 16)
+        for line in self.invoice_line_ids:
             aggregated_discount = line._compute_aggregated_discount(line.discount)
             old_values_by_line_id[line.id] = {
                 "price_unit": line.price_unit,
@@ -26,9 +26,9 @@ class AccountMove(models.Model):
             }
             price_unit = line.price_unit * (1 - aggregated_discount / 100)
             line.update({"price_unit": price_unit, "discount": 0})
-        self.line_ids._fields["price_unit"]._digits = digits
+        self.invoice_line_ids._fields["price_unit"]._digits = digits
         res = super(AccountMove, self)._recompute_tax_lines(**kwargs)
-        for line in self.line_ids:
+        for line in self.invoice_line_ids:
             if line.id not in old_values_by_line_id:
                 continue
             line.update(old_values_by_line_id[line.id])
--- a/account_invoice_triple_discount/models/account_move.py
+++ b/account_invoice_triple_discount/models/account_move.py
@@ -12,12 +12,13 @@ class AccountMove(models.Model):
     def _recompute_tax_lines(self, **kwargs):
         """
         As the taxes are recalculated based on a single discount, we need to
-        simulate a multiple discount by changing the unit price. Values are
-        restored after the original process is done
+        simulate a multiple discount by changing temporarily the first discount
+        with the aggregated discount.
+        Values are restored after the original process is done
         """
+        old_values_by_line_id = {}
         digits = self.line_ids._fields["discount"]._digits
         self.line_ids._fields["discount"]._digits = (16, 16)
-        old_values_by_line_id = {}
         for line in self.line_ids:
             aggregated_discount = line._compute_aggregated_discount(line.discount)
             old_values_by_line_id[line.id] = {
--- a/account_invoice_triple_discount/models/account_move.py
+++ b/account_invoice_triple_discount/models/account_move.py
@@ -15,20 +15,18 @@ class AccountMove(models.Model):
         simulate a multiple discount by changing the unit price. Values are
         restored after the original process is done
         """
+        digits = self.line_ids._fields["discount"]._digits
+        self.line_ids._fields["discount"]._digits = (16, 16)
         old_values_by_line_id = {}
-        digits = self.invoice_line_ids._fields["price_unit"]._digits
-        self.invoice_line_ids._fields["price_unit"]._digits = (16, 16)
-        for line in self.invoice_line_ids:
+        for line in self.line_ids:
             aggregated_discount = line._compute_aggregated_discount(line.discount)
             old_values_by_line_id[line.id] = {
-                "price_unit": line.price_unit,
                 "discount": line.discount,
             }
-            price_unit = line.price_unit * (1 - aggregated_discount / 100)
-            line.update({"price_unit": price_unit, "discount": 0})
-        self.invoice_line_ids._fields["price_unit"]._digits = digits
+            line.update({"discount": aggregated_discount})
         res = super(AccountMove, self)._recompute_tax_lines(**kwargs)
-        for line in self.invoice_line_ids:
+        self.line_ids._fields["discount"]._digits = digits
+        for line in self.line_ids:
             if line.id not in old_values_by_line_id:
                 continue
             line.update(old_values_by_line_id[line.id])
@@ -39,6 +37,6 @@ class AccountMove(models.Model):
         return any(
             [
                 line._compute_aggregated_discount(line.discount) > 0
-                for line in self.invoice_line_ids
+                for line in self.line_ids
             ]
         )
--- a/account_invoice_triple_discount/tests/test_invoice_triple_discount.py
+++ b/account_invoice_triple_discount/tests/test_invoice_triple_discount.py
@@ -196,12 +196,13 @@ class TestInvoiceTripleDiscount(SavepointCase):
             line_form.quantity = 10
             line_form.price_unit = -2.00
         invoice_form.save()
-        for line in invoice.invoice_line_ids:
+        for line in invoice.line_ids:
             line.with_context(check_move_validity=False).update(
                 {"price_unit": -line.price_unit}
             )
         invoice.with_context(check_move_validity=False)._recompute_dynamic_lines(
-            recompute_all_taxes=True
+            recompute_all_taxes=True,
         )
+        invoice._check_balanced()
         self.assertEqual(invoice.move_type, "in_refund")
         self.assertEqual(round(invoice.amount_total, 2), 23.0)
--- a/account_invoice_triple_discount/tests/test_invoice_triple_discount.py
+++ b/account_invoice_triple_discount/tests/test_invoice_triple_discount.py
@@ -41,9 +41,7 @@ class TestInvoiceTripleDiscount(SavepointCase):
     @classmethod
     def _create_refund(cls, tax=False, date=False, in_type=False):
         refund_form = Form(
-            cls.env["account.move"].with_context(
-                default_move_type="in_refund"
-            )
+            cls.env["account.move"].with_context(default_move_type="in_refund")
         )
         # refund_form.partner_id = partner
         refund_form.name = "Test Refund for Triple Discount"
@@ -188,7 +186,7 @@ class TestInvoiceTripleDiscount(SavepointCase):
 
     def test_05_refund_negative_taxes(self):
         """
-        Tests refund negative taxes 
+        Tests refund negative taxes
         """
         invoice = self._create_refund(0)
         invoice_form = Form(invoice)
--- a/account_invoice_triple_discount/tests/test_invoice_triple_discount.py
+++ b/account_invoice_triple_discount/tests/test_invoice_triple_discount.py
@@ -38,19 +38,18 @@ class TestInvoiceTripleDiscount(SavepointCase):
         )
         cls.sale_journal = cls.Journal.search([("type", "=", "sale")], limit=1)
 
-    @classmethod
-    def _create_refund(cls):
+    def _create_refund(self):
         refund_form = Form(
-            cls.env["account.move"].with_context(default_move_type="in_refund")
+            self.env["account.move"].with_context(default_move_type="in_refund")
         )
         refund_form.name = "Test Refund for Triple Discount"
 
         with refund_form.invoice_line_ids.new() as refund_line:
-            refund_line.quantity = 1.00
-            refund_line.name = "test refund line"
-            refund_line.price_unit = 100.00
+            refund_line.quantity = 10
+            refund_line.name = "Negative amounts"
+            refund_line.price_unit = -2.00
             refund_line.tax_ids.clear()
-            refund_line.tax_ids.add(cls.tax)
+            refund_line.tax_ids.add(self.tax)
 
         refund = refund_form.save()
         return refund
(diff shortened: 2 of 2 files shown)


odooNextev:
@SirTakobi I opened this PR try to solving https://github.com/OCA/l10n-italy/pull/3269 tests.
What do you think? Thanks

SirTakobi:
> @SirTakobi I opened this PR try to solving [OCA/l10n-italy#3269](https://github.com/OCA/l10n-italy/pull/3269) tests.
What do you think? Thanks

I'm sorry but I can't see how this change is affecting the behavior of this module, could you please add a test here that shows which use-case is being fixed? Or at least explain it a bit?

Also, I don't understand from your message if this change is actually fixing the tests of https://github.com/OCA/l10n-italy/pull/3269, could you please clarify that?

odooNextev:
Test failed because of negatives price unit and tax amount on invoice line(in a refund invoice). Using this module it retrieve a wrong total amount because tax amount always had wrong sign. This change fixed this problem according to our tests.

SirTakobi:
So it has to do with Tax lines changing their sign, that's why now you are only considering `invoice_line_ids` instead of `line_ids`.
Why are the Tax lines changing their sign, before this change? I don't see anything related to changing signs, only copy/pasting the same values.

I'm sorry but this fix is still making little sense to me as a standalone change, is it possible to add a test showing what has been fixed?

odooNextev:
@SirTakobi I added a test trying to repoduce what l10n_it_fatturapa_in does with refunds with negative total. It look that it works properly

francesco-ooops:
@SirTakobi what do you think?

account_invoice_triple_discount/models/account_move.py
```diff
@@ -16,19 +16,19 @@ def _recompute_tax_lines(self, **kwargs):
         restored after the original process is done
         """
         old_values_by_line_id = {}
-        digits = self.line_ids._fields["price_unit"]._digits
-        self.line_ids._fields["price_unit"]._digits = (16, 16)
-        for line in self.line_ids:
+        digits = self.invoice_line_ids._fields["price_unit"]._digits
+        self.invoice_line_ids._fields["price_unit"]._digits = (16, 16)
+        for line in self.invoice_line_ids:
```
SirTakobi: Thanks for the test, now I can check what is going wrong.

As you suggested, for some reason the sign of `price_unit` of the tax line is changed when calling https://github.com/odoo/odoo/blob/5a3a8d0b1fcc982f813ca77f744a0612abbe7b94/addons/account/models/account_move.py#L602; the method itself is not very clear to me, but the actual switch might be happening in https://github.com/odoo/odoo/blob/5a3a8d0b1fcc982f813ca77f744a0612abbe7b94/addons/account/models/account_move.py#L628.

Apparently the module is trying to manage adding multiple discounts on the tax lines but we can skip them, and we probably should also because multiple discounts can only be assigned on invoice lines (not tax lines) from the UI.

Another fix might be to only edit the lines that we have to: for instance skip the lines that have no `aggregated_discount` because for those lines the module is only writing the same `price_unit` and `discount` after calling `super`, that is not very useful

account_invoice_triple_discount/tests/test_invoice_triple_discount.py
```diff
@@ -164,3 +183,25 @@ def test_04_discounts_decimals_tax(self):
...
+            line_form.price_unit = -2.00
+        invoice_form.save()
+        for line in invoice.invoice_line_ids:
+            line.with_context(check_move_validity=False).update(
+                {"price_unit": -line.price_unit}
+            )
+        invoice.with_context(check_move_validity=False)._recompute_dynamic_lines(
+            recompute_all_taxes=True
+        )
+        self.assertEqual(invoice.move_type, "in_refund")
```
SirTakobi: In my opinion, editing the invoice with `check_move_validity=False` is risky, and should be avoided as much as possible.
If really needed, the invoice should be balanced when you are done editing it.

I tried this locally and here the `invoice` is no more balanced: calling `invoice._check_balanced()` raises:
```
odoo.sql_db: bad query: UPDATE "account_move_line" SET "amount_currency"='0.00',"amount_residual"='23.00',"amount_residual_currency"='0.00',"balance"='23.00',"credit"='0.00',"debit"='23.00',"price_subtotal"='0.00',"price_total"='0.00',"price_unit"='0.00',"reconciled"=false,"tax_audit"='',"write_uid"=1,"write_date"=(now() at time zone 'UTC') WHERE id IN (242)
ERROR: new row for relation "account_move_line" violates check constraint "account_move_line_check_amount_currency_balance_sign"
DETAIL:  Failing row contains (242, 83, Test Refund for Triple Discount, 2023-06-12, null, draft, 2, 1, 2, 14, 50049, 10, , 1.00, 0.00, 0.00, 23.00, 0.00, 23.00, 0.00, 0.00, 0.00, f, f, 2023-06-12, 2, null, null, null, null, null, null, null, null, null, 0.00, t, null, , 23.00, 0.00, null, null, null, null, f, t, 1, 2023-06-12 10:52:10.539377, 1, 2023-06-12 10:52:10.539377, 0.00, 0.00).
```

I know this is only the test and you are not using `check_move_validity=False` in the module's code but looks to me like you are checking assertions on something that is _not possible_ (because usually you can't have an invoice that is not balanced), thus the assertions have little value.

odooNextev: @SirTakobi Thanks for your review. I tried to add a test to replace behavior of l10n_it_fatturapa if there are some negative lines https://github.com/OCA/l10n-italy/blob/14.0/l10n_it_fatturapa_in/models/account.py#L321.

francesco-ooops:
@SirTakobi is it good now or is some other change needed?

SirTakobi:
> @SirTakobi is it good now or is some other change needed?

https://github.com/OCA/account-invoicing/pull/1442#discussion_r1226492079

account_invoice_triple_discount/tests/test_invoice_triple_discount.py
odooNextev: @SirTakobi I made some changes and now the invoice is balanced. Could you please check if everything look good to you? Thanks

account_invoice_triple_discount/tests/test_invoice_triple_discount.py
```diff
@@ -38,6 +38,25 @@ def setUpClass(cls):
         )
         cls.sale_journal = cls.Journal.search([("type", "=", "sale")], limit=1)
 
+    @classmethod
+    def _create_refund(cls, tax=False, date=False, in_type=False):
+        refund_form = Form(
+            cls.env["account.move"].with_context(default_move_type="in_refund")
+        )
+        # refund_form.partner_id = partner
```
SirTakobi: What is this comment explaining? If not useful, should be removed

account_invoice_triple_discount/tests/test_invoice_triple_discount.py
```diff
@@ -38,6 +38,25 @@ def setUpClass(cls):
         )
         cls.sale_journal = cls.Journal.search([("type", "=", "sale")], limit=1)
 
+    @classmethod
+    def _create_refund(cls, tax=False, date=False, in_type=False):
```
SirTakobi: `date` and `in_type` are not used in the method, should be removed.

Maybe the whole method might be substituted by calling `init_invoice`? See https://github.com/odoo/odoo/blob/f71626fa6ba413625d7242f5e13d05b8b95be48c/addons/account/tests/common.py#L366

account_invoice_triple_discount/tests/test_invoice_triple_discount.py
```diff
@@ -164,3 +183,26 @@ def test_04_discounts_decimals_tax(self):
         invoice_form.save()
 
         self.assertEqual(invoice.amount_tax, 177.61)
+
+    def test_05_refund_negative_taxes(self):
```
SirTakobi, line 184: I tried this locally and this test is passing even without the changes to `account_move`.
When adding a fix, the included test should pass with the fix and break without it.

SirTakobi (changes requested):
Please follow https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#71commit-message for the commits messages.

For instance:
> Always put meaningful commit messages: commit messages should be self explanatory (long enough) including the name of the module that has been changed and the reason behind that change. Do not use single words like "bugfix" or "improvements".

account_invoice_triple_discount/tests/test_invoice_triple_discount.py
odooNextev: @SirTakobi now I fixed this test and without our fix in _recompute_tax_lines it fails

account_invoice_triple_discount/models/account_move.py
```diff
@@ -15,19 +15,17 @@ def _recompute_tax_lines(self, **kwargs):
         simulate a multiple discount by changing the unit price. Values are
         restored after the original process is done
         """
+        digits = self.line_ids._fields["discount"]._digits
+        self.line_ids._fields["discount"]._digits = (16, 16)
         old_values_by_line_id = {}
-        digits = self.line_ids._fields["price_unit"]._digits
-        self.line_ids._fields["price_unit"]._digits = (16, 16)
```
SirTakobi: In order to have a minimal diff, you might keep the `old_values_by_line_id = {}` line as first

account_invoice_triple_discount/models/account_move.py
```diff
@@ -15,19 +15,17 @@ def _recompute_tax_lines(self, **kwargs):
         simulate a multiple discount by changing the unit price. Values are
```
SirTakobi: The method documentation should be updated because the `unit price` is no more changed.

account_invoice_triple_discount/tests/test_invoice_triple_discount.py
SirTakobi: 👍🏻

account_invoice_triple_discount/tests/test_invoice_triple_discount.py
SirTakobi: 👍🏻

account_invoice_triple_discount/tests/test_invoice_triple_discount.py
SirTakobi: 👍🏻

account_invoice_triple_discount/tests/test_invoice_triple_discount.py
```diff
@@ -164,3 +180,18 @@ def test_04_discounts_decimals_tax(self):
...
+        """
+        refund = self._create_refund()
+        for line in refund.invoice_line_ids:
+            line.with_context(check_move_validity=False).update(
+                {"price_unit": -line.price_unit}
+            )
+        refund.with_context(check_move_validity=False)._recompute_dynamic_lines(
+            recompute_all_taxes=True,
+        )
+        self.assertEqual(refund.move_type, "in_refund")
```
SirTakobi, line 196: I tried this locally and here the `refund` is no more balanced: calling `refund._check_balanced()` raises:
```
odoo.sql_db: bad query: UPDATE "account_move_line" SET "amount_currency"='0.00',"amount_residual"='23.00',"amount_residual_currency"='0.00',"balance"='23.00',"credit"='0.00',"debit"='23.00',"price_subtotal"='0.00',"price_total"='0.00',"price_unit"='0.00',"reconciled"=false,"tax_audit"='',"write_uid"=1,"write_date"=(now() at time zone 'UTC') WHERE id IN (125)
ERROR: new row for relation "account_move_line" violates check constraint "account_move_line_check_amount_currency_balance_sign"
DETAIL:  Failing row contains (125, 44, Test Refund for Triple Discount, 2023-08-07, null, draft, 2, 1, 2, 14, 50049, 10, , 1.00, 0.00, 0.00, 23.00, 0.00, 23.00, 0.00, 0.00, 0.00, f, f, 2023-08-07, 2, null, null, null, null, null, null, null, null, null, 0.00, t, null, , 23.00, 0.00, null, null, null, null, f, t, 1, 2023-08-07 08:48:13.137036, 1, 2023-08-07 08:48:13.137036, 0.00, 0.00).
```
so I have the same concerns of 2 months ago: https://github.com/OCA/account-invoicing/pull/1442#discussion_r1226492079

account_invoice_triple_discount/models/account_move.py
```diff
@@ -39,6 +37,6 @@ def _has_discount(self):
         return any(
             [
                 line._compute_aggregated_discount(line.discount) > 0
-                for line in self.invoice_line_ids
+                for line in self.line_ids
```
SirTakobi, line 41: This check was on `line_ids`, why is it changed? Does not seem to affect the added test result

SirTakobi (changes requested):

rafaelbn:
Please @eantones could you review this PR?

matteoopenf:
@SirTakobi  @odooNextev  have you news?

odooNextev:
> @SirTakobi @odooNextev have you news?

I just solved a couple of issues and I tried to relaunch tests cause in my local environment I wasn't able to reproduce tests fails in account_invoice_mass_sending.

account_invoice_triple_discount/tests/test_invoice_triple_discount.py
odooNextev: > I tried this locally and here the `refund` is no more balanced: calling `refund._check_balanced()` raises:
> 
> ```
> odoo.sql_db: bad query: UPDATE "account_move_line" SET "amount_currency"='0.00',"amount_residual"='23.00',"amount_residual_currency"='0.00',"balance"='23.00',"credit"='0.00',"debit"='23.00',"price_subtotal"='0.00',"price_total"='0.00',"price_unit"='0.00',"reconciled"=false,"tax_audit"='',"write_uid"=1,"write_date"=(now() at time zone 'UTC') WHERE id IN (125)
> ERROR: new row for relation "account_move_line" violates check constraint "account_move_line_check_amount_currency_balance_sign"
> DETAIL:  Failing row contains (125, 44, Test Refund for Triple Discount, 2023-08-07, null, draft, 2, 1, 2, 14, 50049, 10, , 1.00, 0.00, 0.00, 23.00, 0.00, 23.00, 0.00, 0.00, 0.00, f, f, 2023-08-07, 2, null, null, null, null, null, null, null, null, null, 0.00, t, null, , 23.00, 0.00, null, null, null, null, f, t, 1, 2023-08-07 08:48:13.137036, 1, 2023-08-07 08:48:13.137036, 0.00, 0.00).
> ```
> 
> so I have the same concerns of 2 months ago: [#1442 (comment)](https://github.com/OCA/account-invoicing/pull/1442#discussion_r1226492079)

@SirAionTech I don't understand what's wrong in lines recomputation at line 193...
If you duplicate lines 193-195 or manually launch double `_recompute_dynamic_lines()` from debug console, lines will be balanced and it will pass the test...
I don't know why we need 2 recomputations to balance taxes lines and I guess there are no issues in this method cause it's from Odoo standard accounting

eLBati:
I propose to move here https://github.com/OCA/account-invoicing/pull/1592

sergiocorato:
superseeded by #1592

PR #1448, opened 2023-04-26
[14.0][IMP] account_global_discount, added init hook to avoid memory issues on big database installation

ChrisOForgeFlow:
@ForgeFlow

### DIFF
--- a/account_global_discount/hooks.py
+++ b/account_global_discount/hooks.py
@@ -1,0 +1,28 @@
+from odoo.tools.sql import column_exists
+
+
+def _pre_init_global_discount_fields(cr):
+    if not column_exists(cr, "account_move", "amount_global_discount"):
+        cr.execute(
+            """
+                ALTER TABLE "account_move"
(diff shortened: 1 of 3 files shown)


account_global_discount/hooks.py
```diff
@@ -0,0 +1,23 @@
+from odoo.tools.sql import column_exists
+
+
+def _pre_init_global_discount_fields(cr):
+    if not column_exists(cr, "account_move", "amount_global_discount"):
+        cr.execute(
+            """
+                ALTER TABLE "account_move"
+                ADD COLUMN "amount_global_discount" double precision
```
pedrobaeza: No 0 initialization?

ChrisOForgeFlow: Sorry, I added default and drop after, thank you

pedrobaeza (approved):

rvalyi (approved):

hildickethan (approved):

PR #1455, opened 2023-05-09
[16.0][mig] account_invoice_search_by_reference: Migration to 16.0

dsolanki-initos:


account_invoice_search_by_reference/models/account_move.py
```diff
@@ -0,0 +1,29 @@
...
+    _inherit = "account.move"
+
+    @api.model
+    def name_search(self, name, args=None, operator="ilike", limit=100):
+        args = args or []
+        domain = []
+        if name:
+            domain = ["|", ("ref", operator, name), ("name", operator, name)]
+        invoices = self.search(domain + args, limit=limit)
+        return invoices.name_get()
```
hildickethan, line 17: this should already be covered by core `_rec_names_search = ['name', 'partner_id.name', 'ref']`

SoniaViciana:
This module maybe it's not necessary as long as you can configure [base_name_search_improved](https://github.com/OCA/server-tools/pull/2477)

AaronHForgeFlow:
That other module seems very interesting. But searching invoices by reference is so common that I think this one is still useful. If you only want to improve this specific search this modules is much light than the other one, that is more useful for more complex cases.

legalsylvain:
Hi @dsolanki-initos. Your branch is conflicting. could you take a look ?

legalsylvain (changes requested):
AFAIK, the feature is native in odoo ``account`` module in V16.
https://github.com/OCA/OpenUpgrade/pull/4288
https://github.com/odoo/odoo/pull/103400/files

CC : original author : @AaronHForgeFlow, @nikul-serpentcs (v10), @bjeficent (v12) , @newtratip (v13), @chusamo (v14) , @koenloodts (v15)

AaronHForgeFlow:
@legalsylvain thanks for the clarification, I think we can close this

PR #1458, opened 2023-05-12
[15.0][FIX] sale_timesheet_invoice_description: Fix tests according to qty_delivered compute change since

victoralmau:
Fix tests according to `qty_delivered` compute change since https://github.com/odoo/odoo/pull/115907/commits/12355a80042c70d4c4c23f6e22145358ccc747f8

Before, `qty_delivered` was calculated by summing up the analytical lines and converting with `uom_id`
Now, `qty_delivered` is computed by iterating each analytical lines and converting with `uom_id` each line.

This change of behavior, causes in some use cases that the rounding makes it differ by 0.01 (or more in some cases).

Please @pedrobaeza can you review it?

@Tecnativa


pedrobaeza (approved):
Fast-tracking for solving CI error:

/ocabot merge patch

PR #1460, opened 2023-05-18
[16.0][ADD] stock_picking_accounting_date

AungKoKoLin1997:
This module adds accounting date in picking and pass the value to accounting date of SVL's journal entry.

When a user makes the transfer for the products with "real time" inventory valuation, effective date in transfer couldn't be selected.
And Accounting date for stock valuation journal entry will be the same with effective date.
So, for the users who want to specify each transfer's actual accounting date needs to reset to draftmanually and 
change acocunting date. This module intends to get rid of this inconvenience by adding accounting date in transfer and pass to journal entry.

@qrtl


yostashiro:
@AungKoKoLin1997 I don't think this module should belong here.  Maybe https://github.com/OCA/stock-logistics-warehouse?

AungKoKoLin1997:
> I don't think this module should belong here. Maybe https://github.com/OCA/stock-logistics-warehouse?

@yostashiro I found modules like [stock_picking_invoicing_incoterm](https://github.com/OCA/account-invoicing/tree/15.0/stock_picking_invoicing_incoterm) is in this repo and I propose this PR. I also know module like account_move_line_stock_info exist in `stock-logistics-warehouse` repo. That's why I thought this module can host in this repo.

AungKoKoLin1997:
This module was proposed to stock-logistics-warehouse.
https://github.com/OCA/stock-logistics-warehouse/pull/1711

PR #1463, opened 2023-05-23
[16.0] [MIG] account_invoice_supplier_ref_unique: Migration to 16.0

carlosluissb:


antoniocanovas:
I think this module is no need anymore.
Standard v16 send a warning if duplicated supplier ref.

Captura de pantalla 2023-05-24 a las 11 59 26

carlosluissb:
Thank you for your comment.

I think the module needs to be migrated because in Odoo 16 although there is a warning that the invoice reference number is not repeated, it can still be confirmed, while this module avoids that invoices with the same reference number can exist.

BT-anieto:
Hello @carlosluissb! Hve you followed https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0#how-to to migrate the module? I only see two commits, while you should have all the commits from previous versions. Could you take a look at that? I will review it then.

Gelo-fl:
@carlosluissb If you try to validate a vendor invoice with the same reference and **the same invoice date**, a warning will appear and you will not be able to validate it. 
image

If it is the same invoice, these two fields should be identical in both invoices. As @antoniocanovas said I don't think this is necessary either.

rousseldenis:
@carlosluissb I think this module do several things other than unique supplier reference:

- Add a supplementary invoice reference (because you can have a different supplier reference than the 'ref' field which is used for payment for instance)
- Set the default 'ref' value with that new field
- Check the suppler number unicity
- Avoid to copy the reference when duplicating the invoice
- Voids the reference when move reversing

For all those features, I propose to rename the module to 'account_invoice_supplier_reference' because the unicity is not the only feature.

Moreover, I would have added (after renaming) a configuration parameter to enable or not the unicity constraint.

@ThomasBinsfeld @adrienpeiffer @alexis-via As you have worked on this and as this is an historical module when Odoo decided to remove the field to keep only the 'ref' one.

max3903 (approved):

sbidoul:
See also #1484 which did not get a conclusion yet.

BT-anieto:
@carlosluissb hi carlos. This module is already in v16. Could you please close thois pr?

carlosluissb:
> @carlosluissb hi carlos. This module is already in v16. Could you please close thois pr?

Ok thanks

PR #1467, opened 2023-05-29
[15.0][IMP] account_invoice_payment_term_date_due: view invoice_date_due in…

polqubiq:
Improvement consisting in being able to see the `invoice_date_due` field in edit mode.

### DIFF
--- a/account_invoice_payment_term_date_due/views/account_move.xml
+++ b/account_invoice_payment_term_date_due/views/account_move.xml
@@ -5,23 +5,23 @@
     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 -->
 <odoo>
-
     <record id="view_move_form" model="ir.ui.view">
         <field name="model">account.move</field>
         <field name="inherit_id" ref="account.view_move_form" />
         <field name="arch" type="xml">
-            <field name="invoice_payment_term_id" position="after">
-                <span
-                    name="invoice_payment_term_date_due"
-                    class="text-muted oe_read_only"
+            <xpath expr="//field[@name='invoice_payment_term_id']" position="after">
+                <i
+                    class="fa fa-arrow-right mx-3 text-muted pl-4 pt-1"
                     attrs="{'invisible': ['|', ('invoice_payment_term_id', '=', False), ('invoice_date_due', '=', False)]}"
-                    t-translation="off"
-                >
-                    <i class="fa fa-arrow-right mx-3" />
-                    <field name="invoice_date_due" />
-                </span>
-            </field>
+                    title="Arrow icon"
+                />
+                <field
+                    class="text-muted"
+                    name="invoice_date_due"
+                    force_save="1"
+                    attrs="{'readonly': [('invoice_payment_term_id', '!=', False)], 'invisible': ['|', ('invoice_payment_term_id', '=', False), ('invoice_date_due', '=', False)]}"
+                />
+            </xpath>
         </field>
     </record>
-
 </odoo>


account_invoice_payment_term_date_due/views/account_move.xml
```diff
@@ -5,23 +5,23 @@
...
-            </field>
+                    style="padding-left: 30px;padding-top: 2px;"
+                />
+                <field
+                    class="text-muted"
+                    name="invoice_date_due"
+                    force_save="1"
+                    attrs="{'readonly': [('invoice_payment_term_id', '!=', False)], 'invisible': ['|', ('invoice_payment_term_id', '=', False), ('invoice_date_due', '=', False)]}"
+                />
+            </xpath>
```
ivantodorovich, line 24: I found it easier to read before, with the `invisible` attribute being set on the parent span directly. 🤔 Same for the `text-muted` class..
Is there a reason for this change?

account_invoice_payment_term_date_due/views/account_move.xml
```diff
@@ -5,23 +5,23 @@
...
         <field name="model">account.move</field>
         <field name="inherit_id" ref="account.view_move_form" />
         <field name="arch" type="xml">
-            <field name="invoice_payment_term_id" position="after">
-                <span
-                    name="invoice_payment_term_date_due"
-                    class="text-muted oe_read_only"
+            <xpath expr="//field[@name='invoice_payment_term_id']" position="after">
+                <i
+                    class="fa fa-arrow-right mx-3 text-muted"
```
ivantodorovich:
```
2023-05-29 15:48:21,047 254 WARNING odoo odoo.addons.base.models.ir_ui_view: A <i> with fa class (fa fa-arrow-right mx-3 text-muted) must have title in its tag, parents, descendants or have text
View error context:
{'file': '/__w/account-invoicing/account-invoicing/account_invoice_payment_term_date_due/views/account_move.xml',
 'line': 3,
 'name': 'account.move form',
 'view': ir.ui.view(1445,),
 'view.model': 'account.move',
 'view.parent': ir.ui.view(668,),
 'xmlid': 'view_move_form'} 
```

While you're at this, could you solve this warning?
it's not hurting anyone, but still..

account_invoice_payment_term_date_due/views/account_move.xml
```diff
@@ -5,23 +5,23 @@
...
                     attrs="{'invisible': ['|', ('invoice_payment_term_id', '=', False), ('invoice_date_due', '=', False)]}"
-                    t-translation="off"
-                >
-                    <i class="fa fa-arrow-right mx-3" />
-                    <field name="invoice_date_due" />
-                </span>
-            </field>
+                    style="padding-left: 30px;padding-top: 2px;"
```
ivantodorovich: Generally is better to avoid inline styles like this one.
Can't it be solved with some boostrap classes instead? (ie: `pl-2 pt-1` or something like that)

account_invoice_payment_term_date_due/views/account_move.xml
polqubiq: Yes, before it was easier to read, but the idea of the change is that even though it is being edited, the `invoice_date_due` field can be seen.

The `force_save` is important for it to work correctly.

You can test it if you wish. It should work great!

account_invoice_payment_term_date_due/views/account_move.xml
polqubiq: Done! I added the line `title="Arrow icon"`

account_invoice_payment_term_date_due/views/account_move.xml
polqubiq: Perfect! I added the `pl-4 pt-1` like you suggested. It is displayed correctly.

polqubiq:
@ivantodorovich changes have been applied successfully! Thanks!

chusamo (approved):
LGTM

account_invoice_payment_term_date_due/views/account_move.xml
ivantodorovich: What I mean is that you can keep the `span` while achieving that, which also results in a minimal diff
For example, something like this:

```diff
             <field name="invoice_payment_term_id" position="after">
                 <span
                     name="invoice_payment_term_date_due"
-                    class="text-muted oe_read_only"
+                    class="text-muted"
                     attrs="{'invisible': ['|', ('invoice_payment_term_id', '=', False), ('invoice_date_due', '=', False)]}"
                     t-translation="off"
                 >
                     <i class="fa fa-arrow-right mx-3" />
-                    <field name="invoice_date_due" />
+                    <field
+                        name="invoice_date_due"
+                        force_save="1"
+                        attrs="{'readonly': [('invoice_payment_term_id', '!=', False)]}"
+                    />
                 </span>
             </field>
```

polqubiq: @ivantodorovich I tried to do it like you suggest:

view


And in **non edit mode** it's okay:

image

But in **edit mode** it is not displayed correctly:

image


With my suggestion, it is displayed correctly with **both modes**!

Thanks.

ivantodorovich (approved):
Alright, then.
Thanks for the improvement!

PR #1470, opened 2023-06-05
[15.0] FIX CI

HviorForgeFlow:
CC @AaronHForgeFlow


AaronHForgeFlow:
Thank you @HviorForgeFlow , I spoke with @victoralmau . That test was added in order to take into account Odoo changes here: https://github.com/odoo/odoo/pull/115907 However, Odoo reverted the changes here: https://github.com/odoo/odoo/pull/122431

We agreed to comment the check, because it makes no sense now, but we do not delete it, just to keep track of Odoo changes.

MiquelRForgeFlow (approved):

victoralmau (approved):

PR #1475, opened 2023-06-08
[16.0][mig] sale_timesheet_invoice_description: Migration to 16.0

dsolanki-initos:


victor-champonnois (approved):
Functionnal review OK

sale_timesheet_invoice_description/models/account_move.py
```diff
@@ -26,19 +28,19 @@ def _link_timesheets_to_invoice_line(self, start_date=None, end_date=None):
...
-                if end_date:
-                    domain = expression.AND([domain, [("date", "<=", end_date)]])
+                if date:
+                    domain = expression.AND([domain, [("date", ">=", date)]])
 
                 timesheets = self.env["account.analytic.line"].sudo().search(domain)
                 timesheets.write({"timesheet_invoice_line_id": aml.id})
 
-    def _check_balanced(self):
+    @contextmanager
```
pedrobaeza: Why adding this? And why we need to override this check in first place? Maybe the reason has gone or it should be implemented in other way.

sale_timesheet_invoice_description/readme/CONTRIBUTORS.rst
```diff
@@ -6,6 +6,8 @@
   * Manuel Calero
   * Luis D. Lafaurie
 
+
 * `Akretion <https://www.akretion.com>`_:
 
   * Clément Mombereau <clement.mombereau@akretion.com.br>
+  * Dhara Solanki <dhara.solanki@initos.com>
```
pedrobaeza, line 17: Incorrect indentation. This way, it's inside Akretion section.

tbaden:
@dsolanki-initos can you have a look into the requested changes

sale_timesheet_invoice_description/models/account_move.py
robinkeunen: It comes from this commit https://github.com/OCA/account-invoicing/commit/98c253e32b508207b31171a8bce6be96537cf7a1 . It looks like splitting the account move line by timesheet breaks the debit/credit balance by line.

robinkeunen: I'm afraid I don't know account.move well enough to suggest a better way.

azoellner: The new (in comparison to v15) `@contextmanager` here is needed, because the [super method in the Odoo v16 core](https://github.com/odoo/odoo/blob/3b24a13ee217b2d3846ad0953ccf8885d7b1f15a/addons/account/models/account_move.py#L1714-L1738) got one in the reworked accounting code, in commit https://github.com/odoo/odoo/commit/d8d47f9ff8554f4b39487fd2f13c153c7d6f958d.

However, the super call is implemented incorrectly here (a couple of lines below). It should read something line `with super()._check_balanced(container): yield`. The [flake8-bugbear warning B901](https://github.com/PyCQA/flake8-bugbear#opinionated-warnings), suppressed here with `# noqa`, points to that issue.

👉 ~~I am fixing this.~~ [Update:] Fixed in 4e2d0ce518c7294e751153fcc8ee60750953ef02.

sale_timesheet_invoice_description/readme/CONTRIBUTORS.rst
azoellner: 👉 ~~I am fixing this.~~ [Update:] Fixed in 4e2d0ce518c7294e751153fcc8ee60750953ef02.

Thank you for your advice.

azoellner:
@pedrobaeza (@robinkeunen @tbaden) I now fixed the issues that you mentioned above, did run `pre-commit` afterwards, and finally rebased this feature branch onto the current head of its target "16.0".

pedrobaeza:
Great, now it's available for reviewers. Check https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#review for knowing the process. You can review other PRs and ask in exchange that they review yours.

/ocabot migration sale_timesheet_invoice_description

azoellner:
@pedrobaeza While working on migrating the feature "split-by-task" (see PR #1501 for version 15.0, which is still on my agenda), I found and fixed some issues related to version 16.0 core changes (to timesheets and accounting), pushed in fafcaefd7d70b3fd347966d00e9ce4a0b40f0343.

P.S. Note that the aforementioned PR #1501 does contain further fixes, unrelated to the migration itself. I did not include them here, as imho this migration branch should only be concerned with migrating the version 15.0 code, not fixing it.

robinkeunen (approved):
👍 Last 3 commits sould be squashed IMO.

Thx !

azoellner:
@robinkeunen I squashed the last three commits into one, as you suggested. Thx.

PR #1478, opened 2023-06-09
[15.0][FIX] account_invoice_clearing: sudo when reading action

Shide:
Perform a sudo when reading the action to relaunch wizard.

MT-3101 @moduon @yajo @rafaelbn


rafaelbn (approved):
👍🏼

PR #1485, opened 2023-06-12
[16.0][FIX] account_invoice_pricelist: Avoid division by zero

ferran-S73:
When calculating the discount, pricelists discount_policy not "with_discount", the percentage of the discount is calculated. If base_price is 0.0 we receive an error.

Cherry-pick from https://github.com/OCA/account-invoicing/pull/1076

### DIFF
--- a/account_invoice_pricelist/models/account_move.py
+++ b/account_invoice_pricelist/models/account_move.py
@@ -77,6 +77,8 @@ class AccountMoveLine(models.Model):
         return res
 
     def _calculate_discount(self, base_price, final_price):
+        if base_price == 0.0:
+            return 0.0
         discount = (base_price - final_price) / base_price * 100
         if (discount < 0 and base_price > 0) or (discount > 0 and base_price < 0):
             discount = 0.0


RodrigoBM (approved):
it would be interesting to add the test

hildickethan (approved):
https://github.com/odoo/odoo/blob/1e96c21eed9556f87bb8dae2384580dd784f12c0/addons/sale/models/sale_order_line.py#L577
This is just replicating the same behaviour as Odoo's pricelist in sales (that this module attempts to replicate for invoices), should be an easy merge

ljsalvatierra-factorlibre (approved):

RodrigoBM:
can @rafaelbn merge it? thanks!!!

rafaelbn:
Could you add a test for that please? Is it reasonable? 😄

AaronHForgeFlow:
@rafaelbn I did not see you comment and now the code is merged in versions 13.0 to 15.0. I am merging this just to keep branches even. Then another PR can be created to cover the new test case.

/ocabot merge patch

PR #1481, opened 2023-06-10
[16.0][MIG] account invoice fixed discount: Migration to 16.0

TITUS6304658:
16.0 mig account invoice fixed discount

Migration to version 16.0

PR #1488, opened 2023-06-19
[16.0][MIG] account_invoice_fixed_discount: Migration to 16.0

PieterPaulussen:


CRogos:
@PieterPaulussen are you still working on this migration?

PieterPaulussen:
@CRogos Hi, sorry, left this PR a bit hanging at the moment. There are a few fixes that still need to be applied to the code. If you want, you can fork the repo and apply the fixes. I'll see if I can do it later this evening myself otherwise.

CRogos:
Thank you for the quick response. If you can do it in the near future, I'll happy to wait. Contact me if you need some help or when you are ready for review.

account_invoice_fixed_discount/models/account_move_line.py
```diff
@@ -0,0 +1,55 @@
+# Copyright 2017 ForgeFlow S.L.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+
+from odoo import api, fields, models
+
+
+class AccountMove(models.Model):
```
FrancoMaxime: I think it would be better to split the file into two. The first for account_move and the second for account_move_line

account_invoice_fixed_discount/views/account_move_view.xml
```diff
@@ -22,12 +22,6 @@
             >
                 <field name="discount_fixed" groups="base.group_no_one" />
             </xpath>
-            <xpath
```
FrancoMaxime: Why do you remove this?

account_invoice_fixed_discount/tests/test_account_fixed_discount.py
```diff
@@ -2,30 +2,21 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
 
 from odoo.exceptions import ValidationError
-from odoo.tests import TransactionCase
+from odoo.tests import SavepointCase
 
 
-class TestInvoiceFixedDiscount(TransactionCase):
+class TestInvoiceFixedDiscount(SavepointCase):
```
FrancoMaxime: SavepointCase class is now deprecated, you should use TransactionCase instead.

FrancoMaxime (changes requested):

FrancoMaxime:
Thanks for contributing @PieterPaulussen

rousseldenis:
@PieterPaulussen This is a duplicate as there is already #1481

CRogos:
Thx for the link to the other PR. Didn't find it,  too. But the test of the other PR are only green, because the were all deleted. Nevertheless there is some good changes which can be added here, too.

PieterPaulussen:
@rousseldenis @CRogos I believe the other PR is insufficient. I'm testing the module locally for the migration and noticed that most of the upstream methods do not exist at all. So all the `super()` calls there seem faulty. They are also not tested using a `Form` which is suboptimal but only visible during a manual test.

The implementation of the price fields has completely changed in V16.0 compared to the last version of this module by refactoring all the fields to computed fields with inverse onchanges.

My interpretation of the module is that this serves as a 'convenience' module to the user to give explicit discounts. Therefor I'm going to suggest to change the implementation. By reverse calculating the discount percentage per line when setting the `fixed_discount` and then relying on the core to apply the appropriate downstream logic to update the `discount` and `price_unit` fields.

Any thoughts on that?

rousseldenis:
> @rousseldenis @CRogos I believe the other PR is insufficient. I'm testing the module locally for the migration and noticed that most of the upstream methods do not exist at all. So all the `super()` calls there seem faulty. They are also not tested using a `Form` which is suboptimal but only visible during a manual test.
> 
> The implementation of the price fields has completely changed in V16.0 compared to the last version of this module by refactoring all the fields to computed fields with inverse onchanges.
> 
> My interpretation of the module is that this serves as a 'convenience' module to the user to give explicit discounts. Therefor I'm going to suggest to change the implementation. By reverse calculating the discount percentage per line when setting the `fixed_discount` and then relying on the core to apply the appropriate downstream logic to update the `discount` and `price_unit` fields.
> 
> Any thoughts on that?

Indeed, I've just saw that.

A refactoring is needed. You take care of that ?

PieterPaulussen:
@rousseldenis working on it now 👍

rafaelbn:
Did you follow up the FIX in 14? 

https://github.com/OCA/account-invoicing/pull/1225

CRogos (changes requested):
The merge of discount_fixed with discount has another side effect when resetting a discount to 0.
[video (12).webm](https://github.com/OCA/account-invoicing/assets/1799080/cad8b9c3-3624-4ac9-ae6e-da8dc5772ce2)

PieterPaulussen:
> Did you follow up the FIX in 14?
> 
> #1225

@CRogos I've checked the case described in the [PR](https://github.com/OCA/account-invoicing/pull/1225) and the original case from the [ticket](https://github.com/OCA/account-invoicing/issues/1107). I believe my changes to the implementation do not require tax recalculation, but I have implemented the required `float_round` and `float_compare`. 
Because the implementation "steers" the value of the discount field making ensuring the tax is always updated accordingly. It is always the core `discount` field value that is used instead of having a separate field which should be taken into account in other calculations.


> The merge of discount_fixed with discount has another side effect when resetting a discount to 0. [video (12).webm](https://github.com/OCA/account-invoicing/assets/1799080/cad8b9c3-3624-4ac9-ae6e-da8dc5772ce2)

I've also checked this and it was the last if block of the onchange which circumvented the update of the discount field. That should be fixed now.

CRogos (approved):
Code + functional test LGTM.

CRogos (commented):
Sorry, found one more thing which is not 100% perfect.

Can you update discount_fixed, when it is != 0 and the discount field is updated?

[video (13).webm](https://github.com/OCA/account-invoicing/assets/1799080/c6d6c88e-91dc-490e-ab49-f92ccba47bf8)

PieterPaulussen:
> Sorry, found one more thing which is not 100% perfect.
> 
> Can you update discount_fixed, when it is != 0 and the discount field is updated?
> 
>  video.13.webm

Fixed in latest update.

CRogos:
It is now updated when you set the dicount field to 0, but if you change it from 10% to 20% it stays on 75€
image

PieterPaulussen:
@CRogos Same issue here as with the `sale_fixed_discount` regarding the onchanges. I've applied the same onchange context values here in order to support the workflow.

PieterPaulussen:
Great the tests are botched now because of the onchanges. Apparently the O2MForm class does not support inserting context values. I'll have to think of something else.

CRogos:
oh no... I really appreciate all the effort you've already put here. 

Have you tried to make discount_fixed a computed attribute (non stored) with an inverse function setting discount?

account_invoice_fixed_discount/reports/report_account_invoice.xml
```diff
@@ -0,0 +1,28 @@
...
+            </th>
+        </xpath>
+        <xpath expr="//td[@t-if='display_discount']" position="before">
+            <td
+                t-if="display_discount_fixed"
+                t-attf-class="text-end {{ 'd-none d-md-table-cell' if report_type == 'html' else '' }}"
+            >
+                <span t-field="line.discount_fixed" />
+            </td>
+        </xpath>
```
CRogos: We already talked about to combine the value in the PDF. I would prefer this.
image

PieterPaulussen:
@CRogos Another update for this module. I've refactored it to improve the usability and now the expected workflow should be supported. Had to dig deep into the tax calculation of Odoo to account for the rounding differences (similar to the behavior of the `sale_fixed_discount` module. Feel free to try it again.

CRogos:
@PieterPaulussen here are still the issues like in the sale https://github.com/OCA/sale-workflow/pull/2614

- Setting the discount to fixed 32€ changes the discount to 10%, but changing it back to 0, leaves the discount to 10%
- In the PDF layout Fixed discount and % are not merged into 1 column as it is now in sale

PieterPaulussen:
> @PieterPaulussen here are still the issues like in the sale [OCA/sale-workflow#2614](https://github.com/OCA/sale-workflow/pull/2614)
> 
> * Setting the discount to fixed 32€ changes the discount to 10%, but changing it back to 0, leaves the discount to 10%
> * In the PDF layout Fixed discount and % are not merged into 1 column as it is now in sale

@CRogos Sorry, I made the changes to the code but never pushed it. You can re-test it now.

CRogos:
Yes can confirm it is working. 

I discovered one small difference between invoice and quotation. In one PDF we have the currency, in the other it is missing. Not sure which is the better solution, but I think it should be implemented the same.

image
image

rousseldenis (approved):

CRogos:
@pedrobaeza I think this can be merged.

pedrobaeza:
OK, merging blindly due to reviews:

/ocabot merge nobump

PR #1490, opened 2023-06-20
[15.0][ADD] account_multi_branch_company

Saran440:
This module add branch on `account.move` (invoices/bills/refund/credit note) and `account.payment` (payment)

Depend on
- [x] base_multi_branch_company - https://github.com/OCA/server-ux/pull/670


account_multi_branch_company/models/account_journal.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2023 Ecosoft Co., Ltd (https://ecosoft.co.th)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from odoo import fields, models
+
+
+class AccountJournal(models.Model):
+    _inherit = "account.journal"
+
+    branch_id = fields.Many2one(comodel_name="res.branch")
```
kittiu, line 10: I think account.journal has no need for branch_id, it add more complication.

account_multi_branch_company/models/account_payment.py
```diff
@@ -0,0 +1,21 @@
...
+        compute="_compute_journal_branch",
+        store=True,
+        readonly=False,
+        tracking=True,
+    )
+
+    @api.depends("journal_id")
+    def _compute_journal_branch(self):
+        for rec in self:
+            rec.branch_id = rec.journal_id.branch_id.id
```
kittiu, line 21: Since account.payment already inherit accout.move. So, I think this is not needed.

kittiu, line 21:
```suggestion
```

kittiu:
@Saran440 this module serve both invoice, paymnet and journal entry. I think it should move to `account-financial-tools` WDYT?

Saran440:
@kittiu Ok, I fixed it. Thank you for advise.
Closed and move to https://github.com/OCA/account-financial-tools/pull/1783

PR #1500, opened 2023-06-30
[15.0][FIX] account_invoice_fixed_discount: Avoid division by zero

ao-landoo:
It avoids cases of divisions by zero in case of `price_unit` being zero

### DIFF
--- a/account_invoice_fixed_discount/models/account_move.py
+++ b/account_invoice_fixed_discount/models/account_move.py
@@ -70,7 +70,7 @@ class AccountMoveLine(models.Model):
         taxes,
         move_type,
     ):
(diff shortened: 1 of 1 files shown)


rafaelbn:
@LoisRForgeFlow @MiquelRForgeFlow What do you think?

LoisRForgeFlow (commented):
I tested it on runbot and seems that it is still possible to do a division by zero:

image

What I did was to create a new invoice for an existing product, write the fixed discount, then change the unit price to zero -> error raises.

rafaelbn:
Thank you @LoisRForgeFlow , @ao-landoo I set label need fixing! Thank you!

PR #1503, opened 2023-07-04
 [15.0][FIX] account_invoice_fixed_discount: _recompute_tax_lines

ao-landoo:
Fixes function _recompute_tax_lines, which is iterating invoice_line_ids instead of line_ids

This issue was also found on account_invoice_triple_discount module and fixed on this PR: https://github.com/OCA/account-invoicing/pull/1238

### DIFF
--- a/account_invoice_fixed_discount/models/account_move.py
+++ b/account_invoice_fixed_discount/models/account_move.py
@@ -9,12 +9,20 @@ class AccountMove(models.Model):
     _inherit = "account.move"
 
     def _recompute_tax_lines(self, **kwargs):
+        # If another module of this repo inherited this function do not execute
+        # following code in order to avoid conflict
+        if self.env.context.get("avoid_inherit", False) or not any(
+            line.discount_fixed for line in self.line_ids
+        ):
+            return super(AccountMove, self)._recompute_tax_lines(**kwargs)
         old_values_by_line_id = {}
         for line in self.line_ids:
             old_values_by_line_id[line.id] = {"price_unit": line.price_unit}
             price_unit = line.price_unit - line.discount_fixed
             line.update({"price_unit": price_unit})
-        res = super(AccountMove, self)._recompute_tax_lines(**kwargs)
+        res = super(
+            AccountMove, self.with_context(avoid_inherit=True)
+        )._recompute_tax_lines(**kwargs)
         for line in self.line_ids:
             if line.id not in old_values_by_line_id:
                 continue
--- a/account_invoice_fixed_discount/models/account_move.py
+++ b/account_invoice_fixed_discount/models/account_move.py
@@ -8,20 +8,17 @@ from odoo.exceptions import ValidationError
 class AccountMove(models.Model):
     _inherit = "account.move"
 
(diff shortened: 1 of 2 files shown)


ao-landoo:
@adrienpeiffer @pedrobaeza account_invoice_triple_invoice and account_invoice_fixed_discount are in conflict on tests as both modules are installed. They are calling the same function (_recompute_tax_lines), and modifying lines' price_unit field both at the same time, so tests are failing.

Any idea what could be done?

ao-landoo:
@adrienpeiffer @pedrobaeza could you check if you like the solution I applied? I think it's the best solution possible for the issue

pedrobaeza:
About the solution, is inhibiting one of the 2 methods, so it doesn't seem correct. They shouldn't act and interfere if no global or fixed discount AFAIK.

Please don't add empty lines inside the method.

ao-landoo:
Both modules are incompatible, it does not work properly if you install both of them. As runboat installs all modules on the repo, tests were not passing. This is why I applied this solution.

rafaelbn:
@LoisRForgeFlow @MiquelRForgeFlow any help here to go ahead or close it?

account_invoice_fixed_discount/models/account_move.py
```diff
@@ -8,20 +8,25 @@
 class AccountMove(models.Model):
     _inherit = "account.move"
 
-    def _recompute_tax_lines(
-        self, recompute_tax_base_amount=False, tax_rep_lines_to_recompute=None
-    ):
-        vals = {}
-        for line in self.invoice_line_ids.filtered("discount_fixed"):
-            vals[line] = {"price_unit": line.price_unit}
+    def _recompute_tax_lines(self, **kwargs):
```
MiquelRForgeFlow, line 11:
```suggestion
    def _recompute_tax_lines(self, recompute_tax_base_amount=False, tax_rep_lines_to_recompute=None, **kwargs):
```

account_invoice_triple_discount/models/account_move.py
```diff
@@ -15,6 +15,12 @@ def _recompute_tax_lines(self, **kwargs):
         simulate a multiple discount by changing the unit price. Values are
         restored after the original process is done
         """
+        # If another module of this repo inherited this function do not execute
+        # following code in order to avoid conflict
+        if self.env.context.get("avoid_inherit", False) or not any(
+            line._compute_aggregated_discount(line.discount) for line in self.line_ids
+        ):
+            return super(AccountMove, self)._recompute_tax_lines(**kwargs)
```
MiquelRForgeFlow, line 23:
```suggestion
            return super(AccountMove, self)._recompute_tax_lines(recompute_tax_base_amount, tax_rep_lines_to_recompute, **kwargs)
```

rafaelbn:
Please @eantones could you review this PR?

MiquelRForgeFlow:
FYI (related to my request): An example of how passing explicitly the needed arguments besides the kwargs is important: https://github.com/odoo/odoo/commit/49b91e47c6c24e0023110365e7f9084a2c13ca5a

PR #1505, opened 2023-07-10
[FIX] account_invoice_refund_link: Inverted account_invoice_line_refu…

TheClaud99:
…nds_rel in hook

* [FIX] account_invoice_refund_link: Inverted account_invoice_line_refunds_rel in hook

The relation account_invoice_line_refunds_rel between original and refund invoice lines is inverted in the hook

Steps to reproduce:

* Create a refund in a db without the module installed
* Install the module
* The ids in the columns refund_line_id and original_line_id in account_invoice_line_refunds_rel are inverted for the already existing records

All Odoo versions of this module are affected

### DIFF
--- a/account_invoice_refund_link/hooks.py
+++ b/account_invoice_refund_link/hooks.py
@@ -30,7 +30,7 @@ def match_origin_lines(refund, invoice):
             )
             if match:
                 invoice_lines -= invoice_line
-                invoice_line.origin_line_ids = [(6, 0, refund_line.ids)]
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Only 10.0 version is affected. The rest are correct:

https://github.com/OCA/account-invoicing/blob/11.0/account_invoice_refund_link/hooks.py#L24

Unfortunately, this version is too old and the CI is red.

PR #1509, opened 2023-07-14
[15.0][ADD] Add module account_invoice_crm_tag

stefan-tecnativa:
TT44457 cc @Tecnativa

This module adds crm tags for invoices and refunds coming from sale_orders, and links them to its respective order.

@pedrobaeza @CarlosRoca13 please review!


account_invoice_crm_tag/models/sale_order.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    @api.depends("order_line.invoice_lines")
+    def _get_invoiced(self):
+        res = super()._get_invoiced()
+        self.invoice_ids.tags_ids = self.tag_ids
+        return res
```
CarlosRoca13: This is not correct, you have to search the method used to create the invoices from the sale.order and see how it feels the info.

CarlosRoca13 (changes requested):

account_invoice_crm_tag/models/sale_order.py
pedrobaeza: It's a `_prepare...` method in `sale.order.line` and you have to use `[(4, id)]` syntax.

stefan-tecnativa: I saw both methods from `sale.order` and `sale.order.line`, but I wasn't sure which to use. Finally i decided to use the `sale.order` one, but still wondering what is the best decision.

account_invoice_crm_tag/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2023 Tecnativa - Stefan Ungureanu
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Account Invoice CRM Tag",
+    "version": "15.0.1.0.1",
```
pedrobaeza: Why this version?

```suggestion
    "version": "15.0.1.0.0",
```

account_invoice_crm_tag/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+    "category": "Accounting",
+    "author": "Tecnativa, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/account-invoicing",
+    "license": "AGPL-3",
+    "depends": ["account", "crm", "sale"],
```
pedrobaeza: No need of such dependencies, as the tags in the sale.order belongs to `sale`, and `sale` already depends on `account`:

```suggestion
    "depends": ["sale"],
```

account_invoice_crm_tag/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,2 @@
+This module allows you to add crm tags to invoices coming from sales orders
```
pedrobaeza: You can put manually the tags. The propagation from sales order is an extra feature that you should mention also, but not only.

account_invoice_crm_tag/readme/USAGE.rst
```diff
@@ -0,0 +1,5 @@
+#. Configure a sale_order, confirm it, create an invoice and open it.
```
pedrobaeza: This usage should be oriented to only see and handle the tags on the invoice.

account_invoice_crm_tag/report/account_invoice_report.py
```diff
@@ -0,0 +1,28 @@
...
+        relation="account_move_crm_tag_rel",
+        column1="account_move_id",
+        column2="crm_tag_id",
+        string="Invoice CRM Tags",
+        readonly=True,
+    )
+
+    def _select(self):
+        select_str = super()._select()
+        return "%s, account_move_crm_tag_rel.crm_tag_id as tag_ids" % select_str
```
pedrobaeza: Seeing https://github.com/odoo/odoo/blob/15.0/addons/project/report/project_report.py, there's no need of adding it to the select.

stefan-tecnativa: Changes done, I updated the from insertion more accordingly with your reference.

chienandalu: Isn't just declaring the orm m2m field enough?

stefan-tecnativa: @chienandalu You are right, thank you

chienandalu (approved):
:+1:

pedrobaeza (approved):
/ocabot merge nobump

PR #1514, opened 2023-07-25
[MIG][16.0] partner_invoicing_mode_weekly (rename account_invoice_mode_weekly)

rousseldenis:
- Migrated to 16.0
- Remove dependency on stock (as introduced just for tests...)
- Added dependency on : #1515 

- [x] Dependency on : #1515


rousseldenis:
@StefanRijnhart IMHO, stock dependency should be removed in monthly module too (see second commit).

StefanRijnhart (approved):
Nice one!

account_invoice_mode_weekly/models/sale_order.py
```diff
@@ -0,0 +1,76 @@
...
+            self.with_delay()._generate_invoices_by_partner(saleorder_ids)
+        companies.write({"invoicing_mode_weekly_last_execution": datetime.now()})
+        return saleorder_groups
+
+    @api.model
+    def _get_groupby_fields_for_weekly_invoicing(self):
+        """Returns the sale order fields used to group them into jobs."""
+        return ["partner_invoice_id", "payment_term_id"]
+
+    def _generate_invoices_by_partner(self, saleorder_ids, invoicing_mode="weekly"):
```
rousseldenis: @StefanRijnhart @jcoux After having read this, it's impossible to have several modes on the same instance as:

- This is should be implemented in base module
- This is always overridden without super in each specific module...

I'll try to fix that

javierjcf:
Hi, 
I'm testing this PR.
In my system i got an error when install due to:
https://github.com/acsone/account-invoicing/blob/b17ec89d6a1f1798010e90e0ad4e94707260fe85/partner_invoicing_mode_weekly/views/res_config_settings_views.xml#L7

I see in the history that was changed, but it does not exist in v16, i think it should be:
`<div id="invoicing_settings" position="after">`
like before

I also see that the function ` def generate_weekly_invoices(self, companies=None):` was removed and now we have
```py 
return self.generate_invoices(
            companies,
            invoicing_mode="weekly",
            last_execution_field="invoicing_mode_weekly_last_execution",
        )
```
But `generate_invoices` does not exist, and i can not find it.
Is this PR Work in progress or i'm missing something? It seems a different structure compared to partner_invoicing_mode_monthly that is already merged.

javierjcf:
> Hi, I'm testing this PR. In my system i got an error when install due to: https://github.com/acsone/account-invoicing/blob/b17ec89d6a1f1798010e90e0ad4e94707260fe85/partner_invoicing_mode_weekly/views/res_config_settings_views.xml#L7
> 
> I see in the history that was changed, but it does not exist in v16, i think it should be: `<div id="invoicing_settings" position="after">` like before
> 
> I also see that the function ` def generate_weekly_invoices(self, companies=None):` was removed and now we have
> 
> ```python
> return self.generate_invoices(
>             companies,
>             invoicing_mode="weekly",
>             last_execution_field="invoicing_mode_weekly_last_execution",
>         )
> ```
> 
> But `generate_invoices` does not exist, and i can not find it. Is this PR Work in progress or i'm missing something? It seems a different structure compared to partner_invoicing_mode_monthly that is already merged.

I response myself: It seems we ned this other PR: https://github.com/OCA/account-invoicing/pull/1515 in order to test this one, so function `generate_invoices` is on that PR.

i think is a nice refactor, but what about montlhy module that is already in 16.0 branch? It seems we need another PR if this one is merged

rafaelbn (approved):
/ocabot merge patch

rafaelbn:
@rousseldenis it looks like a dependency is missing 🔴 

```
Run for reqfile in requirements.txt test-requirements.txt ; do
odoo-addon-partner-invoicing-mode @ git+https://github.com/OCA/account-invoicing@refs/pull/1515/head#subdirectory=setup/partner_invoicing_mode
Unreleased dependencies found in test-requirements.txt.
Error: Process completed with exit code 1.
```

rousseldenis:
> @rousseldenis it looks like a dependency is missing red_circle
> 
> ```
> Run for reqfile in requirements.txt test-requirements.txt ; do
> odoo-addon-partner-invoicing-mode @ git+https://github.com/OCA/account-invoicing@refs/pull/1515/head#subdirectory=setup/partner_invoicing_mode
> Unreleased dependencies found in test-requirements.txt.
> Error: Process completed with exit code 1.
> ```

@rafaelbn This one is normal.

It's a warning to say we need to merge the dependency PR before this one. :-)

rafaelbn:
OK @rousseldenis ! I see! I miss it because I knew that one was migrated some time ago!

https://github.com/OCA/account-invoicing/tree/16.0/partner_invoicing_mode

So I closed your PR, could you change this one using the actual migrated? 😄 

Thank you!

rousseldenis:
> could you change this one using the actual migrated? 😄

Already done...

rafaelbn (commented):
Tested, I'm unable to make this works, is it possible to test this module in runboat? See:


https://github.com/OCA/account-invoicing/assets/8736623/9a05c3b2-3e4d-42db-8854-54e986a3a118

acsonefho:
@rousseldenis Is it possible to merge this one as it's validated?

StefanRijnhart:
@rousseldenis I think test dependency can be removed.

acsonefho:
@rousseldenis Is it possible to update this PR and merge please?

PR #1515, opened 2023-07-27
[16.0][IMP] partner_invoicing_mode: Allow better inheritance + use correctly grouping option

rousseldenis:
Improvements:

- Move the invoicing code to base module from weekly module (and test code).
- Fix the grouping functionality as it does not work if several sale orders have mixed values (for grouping).
- Add a field that compute the next invoice date (e.g.: for the 14 days module, see #1534)
- Allow to change the grouping option per sale order.

UPDATE:

- Added some hooks to allow changing behavior when validating invoices.
- Uses the grouping function from sale module


rousseldenis:
@StefanRijnhart @jcoux @TDu

rafaelbn:
@rousseldenis this módulo is already merged!

https://github.com/OCA/account-invoicing/tree/16.0/partner_invoicing_mode

😄

rousseldenis:
> @rousseldenis this módulo is already merged!
> 
> https://github.com/OCA/account-invoicing/tree/16.0/partner_invoicing_mode
> 
> 😄

Maaaaah, this is an improvement... So, please reopen it

rafaelbn:
Hahaha! Sorry! After-holidays stuf!!

@StefanRijnhart , please could you review this improvements? 

Thank you ! And sorry for the noise! 🙏🏼

partner_invoicing_mode/__manifest__.py
```diff
@@ -8,9 +8,11 @@
     "website": "https://github.com/OCA/account-invoicing",
     "license": "AGPL-3",
     "category": "Accounting & Finance",
-    "depends": ["account", "queue_job", "sale"],
+    "depends": ["account", "base_partition", "queue_job", "sale"],
```
rafaelbn, line 11: So the main big change is using a new dependency. We are not using in this moment this module. I didn't know about it

https://github.com/OCA/server-tools/tree/16.0/base_partition

rafaelbn (commented):
I usualy don't like more dependencies , but I leave the review to @TDu and @StefanRijnhart .

rafaelbn:
@rousseldenis in case this PR is merged with new dependencies, then a new PR is needed to https://github.com/OCA/account-invoicing/tree/16.0/partner_invoicing_mode_monthly 

Is it?

Runbots don't looks like building 🟠

partner_invoicing_mode/__manifest__.py
rousseldenis: It's not a big one. Technical one.

rousseldenis:
> Runbots don't looks like building 🟠

It's not a failed build (not red)

It happens from time to time, relaunching the build is sufficient in most cases.

rousseldenis:
IMHO @OCA/accounting-maintainers automatic generation of invoices should deserves its own repository

rafaelbn:
> IMHO https://github.com/orgs/OCA/teams/accounting-maintainers automatic generation of invoices should deserves its own repository

I agree, this module prepare to automate but don't automate. We use this one and the others, month, week, etc to classify not to automate 

This module just add the periodicity to invoicing, monthly, weekley and so no like terms (due date) or methods (how the pay)

1. Payment term (dates)
2. Payment mode (how)
3. Invoicing period (every xx time or in a concret moment like shipping) <== **Invoicing Mode**

That's why I don't like to depend in `base_partition` because then for this so simple function we should leave this module and create a new one simpler for the point 3

Let's see what @TDu and @StefanRijnhart says 😄 

> It happens from time to time, relaunching the build is sufficient in most cases.

Thank you @rousseldenis !

rafaelbn (commented):
Tested, I'm unable to make this works see:


https://github.com/OCA/account-invoicing/assets/8736623/5844174d-7884-4de3-902a-b3f1190e3a4a

rousseldenis:
> That's why I don't like to depend in `base_partition` because then for this so simple function we should leave this module and create a new one simpler for the point 3

I don't understand the 'why'... This is a helper module that implement a partition on recordsets in order to split them on criteria.

To avoid writing too much code, that one is very useful in many modules. It does not introduces database nor models changes. So not harmful at all. So, don't block for that.

rousseldenis:
@TDu I have two questions on this as it seems to have not been integrated in OCA module:

- The grouping mode and the invoicing mode on payment mode level
- The ability to define the grouping mode on sale order level separately from configuration on partner or payment mode (as one may want to exclude some sale orders individually)

Thanks for your answers

@jbaudoux

rousseldenis:
I have answered the first one (not used).

I will implement the second one here too. So, we'll be able to set the grouping option on sale order individually (e.g.: grouping is set on partner but he wants to have a separate invoice for a particular sale order).

partner_invoicing_mode/models/sale_order.py
```diff
@@ -70,12 +70,16 @@ def _generate_invoices_by_partner(self, saleorder_ids):
         )
         if not sales:
             return "No sale order found to invoice ?"
-        invoices = sales._create_invoices(
-            grouped=sales[0].partner_invoice_id.one_invoice_per_order, final=True
```
rousseldenis: @rafaelbn FYI This line is not correct and fixed by using the partition function. The current code takes the first sale order property (and don't distinguish each one).

rousseldenis:
> Tested, I'm unable to make this works see:
> 
>  test1-16-IMP-partner_invoicing_mode-1515.mp4

That's because queue_job is not loaded at this time on runboat instances. So, not possible to test ir functionally apart in a local environment.

rafaelbn:
Well any case here is missing the 2 more importants reviews Stepahn and Thierry 😄

partner_invoicing_mode/__manifest__.py
TDu: I agree, code reuse is always good.

TDu:
> @TDu I have two questions on this as it seems to have not been integrated in OCA module:
> 
>     * The grouping mode and the invoicing mode on payment mode level
> 
>     * The ability to define the grouping mode on sale order level separately from configuration on partner or payment mode (as one may want to exclude some sale orders individually)
> 
> 
> Thanks for your answers
> 
> @jbaudoux

The grouping and invoicing mode are on partners because it was required like that for the first instance using this. Having the same possibilities at payment mode level is a good idea (if needed) but raises the question of which settings takes priority. Anyway that would need to be tackle in an other module IMO.

For the 2nd point, adding the option to disable/enable grouping by sales order (like is done here) gives more flexibility, so why not.

partner_invoicing_mode/models/res_partner.py
```diff
@@ -12,10 +12,21 @@ class ResPartner(models.Model):
         default=False,
         help="Do not group sale order into one invoice.",
     )
+    next_invoice_date = fields.Date(
```
TDu, line 15: Is this date recomputed when the invoicing mode is changed on the partner ?

partner_invoicing_mode/models/res_partner.py
```diff
@@ -12,10 +12,21 @@ class ResPartner(models.Model):
...
             "one_invoice_per_order",
+            "next_invoice_date",
         ]
+
+    def _update_next_invoice_date(self):
+        """
+        This will update the next invoice date from the configuration set on
+        the partner if needed (not for standard invoicing_mode).
+        """
+        return
```
TDu, line 31: So this is not really used yet by  depending modules ?

TDu (commented):
Looks good to me. Except for storing the next invoicing date and how it is being used at the moment

partner_invoicing_mode/models/res_partner.py
rousseldenis: The idea here was to have a simple field that is updated after invoice generation.

Like here : https://github.com/OCA/account-invoicing/pull/1534/files#diff-d59687f72a9a4701fe1b9c5796040a8477fbb8a09c4faaa0925a1ce2293aac83R22

For the 14 days module, the next invoice date can be different on each partner (but the gap is 14 days), so I needed to store that next invoice date. As some extensions would maybe need that field too, that's why I put it on the base module (with no update).

If we change the invoicing mode, maybe indeed it should be updated to False. What do you think ?

partner_invoicing_mode/models/res_partner.py
rousseldenis: See https://github.com/OCA/account-invoicing/pull/1534/files#diff-d59687f72a9a4701fe1b9c5796040a8477fbb8a09c4faaa0925a1ce2293aac83R22

partner_invoicing_mode/models/res_partner.py
TDu: Only thinking about the change of behavior of the module.
Before this change we had one invoicing date (by invoicing mode), every monthly invoiced partner were invoiced on the same day (If I recall correctly).
Now we could have the monthly invoiced partner invoiced on different days.

Good, bad ? I don't know...

Should we have an option to allow for both behavior ?

rafaelbn: In my opinion,
- Most common case is the same day every week, month o whatever because in our case clients invoice monthly (for ex.) one day of the month
- Less common but Denis have this use case, every single partner has different day of invoices (which in my management opinion of a company is crazy but there are lots of opinions in the world)

So:
- We shouldn't change the behaviour, we should respect how it was working
- We should add a new behaviour for this new use case

rafaelbn:
@rousseldenis I can review functionally if you ping me, finally we are going for other more simpler way: https://github.com/OCA/sale-workflow/issues/2669

partner_invoicing_mode/models/res_partner.py
rousseldenis: > Only thinking about the change of behavior of the module. Before this change we had one invoicing date (by invoicing mode), every monthly invoiced partner were invoiced on the same day (If I recall correctly). Now we could have the monthly invoiced partner invoiced on different days.
> 
> Good, bad ? I don't know...
> 
> Should we have an option to allow for both behavior ?

@TDu That behavior hasn't been changed as this function remains the same:

https://github.com/OCA/account-invoicing/blob/16.0/partner_invoicing_mode_monthly/models/sale_order.py#L69

rousseldenis: > * Less common but Denis have this use case, every single partner has different day of invoices (which in my management opinion of a company is crazy but there are lots of opinions in the world)

@rafaelbn Indeed, that's in customer requirements as they have different customer profiles and they want to manage it like that.

That's why I have extended the existing module in order to lower the necessary code if flow is not needed (and that's why I don't like putting everything in the same module)

mt-software-de:
Fyi: @rousseldenis i have done a refactoring for Version 14 https://github.com/OCA/account-invoicing/pull/1598 it is already merged.

partner_invoicing_mode/models/res_company.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    invoicing_mode_standard_last_execution = fields.Datetime(
+        string="Last execution (standard)",
+        help="Last execution of standard invoicing.",
+        readonly=True,
+    )
```
jbaudoux, line 14: I don't understand what this serves?

You can configure the cron job interval.
If you want a weekly or fortnightly invoicing on a fixed day, you can configure the cron to run on that specific day with an interval of respectively 7 or 14 days.
Otherwise you can configure the cron to run on a daily basis and the partner `next_invoice_date` will ensure you don't invoice more often than the interval.

So what brings the `invoicing_mode_*_last_execution` company value ?

jbaudoux:
> Fyi: @rousseldenis i have done a refactoring for Version 14 #1598 it is already merged.

Would be good to include the forward port of #1598 in this PR

rousseldenis:
@jbaudoux @mt-software-de You'll be interested in the last commit : https://github.com/OCA/account-invoicing/pull/1515/commits/c0c5d8c061594a39fd5f8c36e8401030e02ef8b8

When doing thousands of sales validations, this avoid as much query as customers count.

rousseldenis:
> > Fyi: @rousseldenis i have done a refactoring for Version 14 #1598 it is already merged.
> 
> Would be good to include the forward port of #1598 in this PR

@jbaudoux The problem is: It seems that my changes are quite the same as the ones done in the base module, moreover, the module name has changed. So, it will be difficult.

partner_invoicing_mode/models/res_company.py
rousseldenis: I think it's done at purpose to avoid duplicate runs. @TDu You can confirm this ?

rousseldenis:
@TDu @rafaelbn @jbaudoux I think this is ready to be merged. Some other PR's depend on it.

rousseldenis:
I think a refinement could be launching the invoice generation/validation for a particular invoicing_mode with a particular date manually could be interesting if something went wrong in the automatic process.

partner_invoicing_mode/models/res_partner.py
```diff
@@ -12,10 +12,21 @@ class ResPartner(models.Model):
...
+    next_invoice_date = fields.Date(
+        help="This is the date at which the next invoice will be generated."
+    )
 
     @api.model
     def _commercial_fields(self):
         return super()._commercial_fields() + [
             "invoicing_mode",
             "one_invoice_per_order",
+            "next_invoice_date",
```
jbaudoux: This change requires to modify the `_update_next_invoice_date` method in the other modules to update only the commercial partner

rousseldenis: Why ? This will sync the value.

jbaudoux: Because you update each partner of a company one by one and you trigger the sync on each contact of the company

rousseldenis: Yes; but that's the Odoo core behavior. You want to get wrong data values across partners of a same company ?

rousseldenis: Moreover, it is done on the whole recordset and not one by one :

https://github.com/odoo/odoo/blob/16.0/odoo/addons/base/models/res_partner.py#L619

jbaudoux: I was referring to this where you process each contact of a company one by one: https://github.com/OCA/account-invoicing/pull/1534#discussion_r1433953276

partner_invoicing_mode/models/sale_order.py
```diff
@@ -1,10 +1,118 @@
...
+            self.with_delay()._generate_invoices_by_partner(saleorder_group["sale_ids"])
+        companies.write({last_execution_field: Datetime.now()})
+        return saleorder_groups
+
+    @api.model
+    def _get_groupby_fields_for_invoicing(self):
+        """Returns the sale order fields used to group them into jobs."""
+        return ["partner_invoice_id", "payment_term_id"]
+
+    def _get_generated_invoices(self, partition):
```
jbaudoux, line 102: it's more a make than a get
```suggestion
    def _generate_invoices(self, partition):
```

partner_invoicing_mode/models/sale_order.py
```diff
@@ -1,10 +1,118 @@
...
+        if not sales:
+            return "No sale order found to invoice ?"
+        # Create invoices using grouping when needed, so partition sales
+        invoice_ids = set()
+        for partition, sales in sales.partition(
+            lambda sale: sale.one_invoice_per_order
+        ).items():
+            invoices = sales._get_generated_invoices(partition=partition)
+            # Update each partner next invoice date
+            sales.partner_invoice_id._update_next_invoice_date()
```
jbaudoux, line 129: move this outside loop ?

jbaudoux: you may have concurrent update on this call as you have one job per `_get_groupby_fields_for_invoicing` and this can concern the same commercial entity

partner_invoicing_mode/models/sale_order.py
jbaudoux: Ah you named it like that because of this https://github.com/OCA/account-invoicing/pull/1628/files#diff-d6d818e70aecbfc330e04d2e0d2da2486f430bea0968f96952e191abe245704aR26
Can you add a comment ?

partner_invoicing_mode/models/sale_order.py
```diff
@@ -1,10 +1,118 @@
...
+            invoice_ids.update(invoices.ids)
+            for invoice in invoices:
+                invoice.with_delay()._validate_invoice()
```
jbaudoux, line 132: @rousseldenis Can you add a hook here?

I have a use case where I need one_invoice_per_shipping but the invoice should not be validated before the goods reach the customer (delay is between 0 and 2 days) because I have to credit on the same invoice the goods returned by the customer.

rousseldenis: You can maybe look at : https://github.com/OCA/account-invoicing/pull/1628/files

This one will create only the draft invoice, then will respect the invoicing_mode (14 days, monthly, ...)

jbaudoux: Yes but I cannot validate the invoice before it reaches the customer. I get the info from the transporter. If there is an invoicing mode, I would like to prevent any early automatic validation.
Example: Goods are shipped on 30th Nov with one invoice per shipping. During the night of the 30th (end of month), I don't want the invoice to be validated. Goods reach the customer on the 2nd Dec and invoice gets validated.

rousseldenis: No problem for that

partner_invoicing_mode/models/sale_order.py
rousseldenis: In that case, the job is automatically postponed

partner_invoicing_mode/models/sale_order.py
rousseldenis: @jbaudoux Is the _get_generated_invoices() could be the one you want (as you maybe don't want to update the next invioce date ? ) ?

partner_invoicing_mode/models/sale_order.py
rousseldenis: See : https://github.com/OCA/queue/blob/16.0/queue_job/controllers/main.py#L127

partner_invoicing_mode/models/sale_order.py
jbaudoux: ah yes, sounds the way to go :)
Thanks

partner_invoicing_mode/models/sale_order.py
jbaudoux: but still worth moving the `_update_next_invoice_date()` outside the loop as it is independent from the partition and the `_get_generated_invoices`

rousseldenis:
@jbaudoux @TDu @rafaelbn Could we move forward on this ?

rafaelbn:
Hello @rousseldenis , I'm not using this module. After analyse it we discard it.

I can help you with pleasure to move it forward but non contributor @bosd @jbaudoux @StefanRijnhart or @TDu have validated it 😢 .

We need at leat 2 approvals 👍🏼

partner_invoicing_mode/models/res_company.py
jbaudoux: ping @TDu @sebalix

partner_invoicing_mode/models/sale_order.py
jbaudoux: @rousseldenis what about moving `_update_next_invoice_date` outside partition to prevent useless jobs?

jbaudoux:
Overall looking good. I still have a few comments open. I'll try your PR

rousseldenis:
> Overall looking good. I still have a few comments open. I'll try your PR

I'll make one refinement as:

- The invoice date should be set before creating validation jobs as real execution could be done the next day (if the job stack is huge)

partner_invoicing_mode/models/sale_order.py
jbaudoux: not that sure anymore. Anyway, the next invoice date is not relevant I guess in such case as you don't want auto-validation

jbaudoux:
> I'll make one refinement as:
> 
> * The invoice date should be set before creating validation jobs as real execution could be done the next day (if the job stack is huge)

and move outside the loop as it is independent from the partition

jbaudoux (approved):

jbaudoux:
@TDu @sebalix Wanna review this one ?

rousseldenis:
@jbaudoux Done

partner_invoicing_mode/models/sale_order.py
jbaudoux: I see you changed it, so solved :)

jbaudoux (approved):
All goods for me.

Would be nice to merge this one as there are many PR depending on it

rousseldenis:
@OCA/accounting-maintainers

cyrilmanuel:
hi @sbidoul, @pedrobaeza  can you merge this PR please ?

bosd (approved):
Code review LGTM

bosd:
@OCA/accounting-maintainers
Or @AaronHForgeFlow Do you have super powers here?

StefanRijnhart (approved):

StefanRijnhart:
I'm afraid merges are stalled by https://github.com/OCA/oca-addons-repo-template/issues/250 too.

PR #1516, opened 2023-07-28
[FIX] account_tax_change: update chart_template_ref in test

ajaniszewska-dev:
without that tests were skipped

### DIFF
--- a/account_tax_change/tests/common.py
+++ b/account_tax_change/tests/common.py
@@ -8,8 +8,8 @@ from odoo.addons.account.tests.common import AccountTestInvoicingCommon
 
 class AccountTaxChangeCommon(AccountTestInvoicingCommon):
     @classmethod
-    def setUpClass(cls):
-        super().setUpClass()
+    def setUpClass(cls, chart_template_ref=None):
+        super().setUpClass(chart_template_ref=chart_template_ref)
         cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
         # Prepare the taxes and products
         cls.product_b.taxes_id -= cls.product_a.taxes_id


account_tax_change/tests/common.py
```diff
@@ -8,8 +8,10 @@
 
 class AccountTaxChangeCommon(AccountTestInvoicingCommon):
     @classmethod
-    def setUpClass(cls):
-        super().setUpClass()
+    def setUpClass(cls, chart_template_ref=None):
+        super().setUpClass(
+            chart_template_ref="l10n_generic_coa.configurable_chart_template"
+        )
```
simahawk: Unless you need a specific COA here you should not override what comes from the method arg.

```suggestion
        chart_template_ref = chart_template_ref or "l10n_generic_coa.configurable_chart_template"
        super().setUpClass(chart_template_ref=chart_template_ref)
```
IF you need it here, otherwise just pass over the argument and drop the dependency you added.

simahawk (changes requested):

account_tax_change/tests/common.py
```diff
@@ -8,8 +8,11 @@
 
 class AccountTaxChangeCommon(AccountTestInvoicingCommon):
     @classmethod
-    def setUpClass(cls):
-        super().setUpClass()
+    def setUpClass(cls, chart_template_ref=None):
+        chart_template_ref = (
+            chart_template_ref or "l10n_generic_coa.configurable_chart_template"
```
simahawk: are you sure we need this default?

Seems to be the default in core https://github.com/odoo/odoo/blob/c0dce31bd509de3b966887f739fa565f3c737720/addons/account/tests/common.py#L37

ajaniszewska-dev: @simahawk  i seen it but somehow even tests for this addon are skipped because template not provided. I think it doesn't harm to have it here if it helps to run tests correctly.

simahawk (approved):

yankinmax (approved):

simahawk (changes requested):
Looking at this once again: we should drop the dependency and the default chart ref.
IMO is not needed. Look:

1. tests are running on 16.0

image

2. when tests do not run you get this [log line](https://github.com/odoo/odoo/blob/16.0/addons/account/tests/common.py#L41) and is not the case here
3. that module in installed automatically WHEN needed https://github.com/odoo/odoo/blob/16.0/addons/account/__init__.py#L51
4. this module has no functional dep on that module

Here we should only fix the propagation of the param.

simahawk (approved):
Can you squash pls?

ajaniszewska-dev:
@simahawk commits squashed

sebalix (approved):

PR #1523, opened 2023-08-07
[13.0][IMP] account_invoice_refund_link: Make more reliable the linking

pedrobaeza:
Hooking on a high level method like `_reverse_move_vals` (although still being private), makes that other modules hooking into it may alter the number of returned lines (like for example, `account_invoice_refund_line_selection`).

We choose the low-level `copy_data` method that is used in such method for linking the refund lines with their origin ones, avoiding the problem.

@Tecnativa


pedrobaeza:
@javierjcf this should solve your problem in #1417. Please try.

stefan-tecnativa (approved):

CarlosRoca13 (approved):

PR #1526, opened 2023-08-07
[16.0][IMP] account_invoice_refund_link: Make more reliable the linking

pedrobaeza:
Forward-port of #1523

Hooking on a high level method like `_reverse_moves` (although still being private), makes that other modules hooking into it may alter the number of returned lines (like for example, account_invoice_refund_line_selection).

We choose the low-level `copy_data` method that is used in such method for linking the refund lines with their origin ones, avoiding the problem.

@Tecnativa

### DIFF
--- a/account_invoice_refund_link/__manifest__.py
+++ b/account_invoice_refund_link/__manifest__.py
@@ -1,7 +1,7 @@
 # Copyright 2004-2011 Pexego Sistemas Informáticos. (http://pexego.es)
 # Copyright 2016 Antonio Espinosa <antonio.espinosa@tecnativa.com>
-# Copyright 2014-2017 Pedro M. Baeza <pedro.baeza@tecnativa.com>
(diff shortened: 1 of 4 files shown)


pedrobaeza:
@pedrocasi @ljsalvatierra-factorlibre here I refactor the linking code, and I see no reason for excluding of the link of other lines like sections, notes, and so on. Do you see any problem?

stefan-tecnativa (approved):

CarlosRoca13 (approved):

pedrocs-exo:
Hi,

A can't think of any inconvenient of linking also the sections and notes. Except for the fact that it's probably useless but we can live with that.

pedrobaeza:
OK, I don't think it's totally useless, as this still keeps trace that a section in the origin invoice generates a section in the refund, on contrary of a manually created section. Let's keep this way and see if there's any cons:

/ocabot merge patch

PR #1529, opened 2023-08-11
[16.0][MIG] account_invoice_mass_sending: Migration to 16.0

BT-anieto:
Migration from 14.0


BT-anieto:
@jguenat  As promised, here the migration to v16. Could you take a look?

BT-mrobles (approved):
LGTM!

BT-dmoreno (approved):

account_invoice_mass_sending/tests/test_account_invoice_mass_sending.py
```diff
@@ -131,8 +129,8 @@ def test_invoice_mass_sending_3(self):
         # test one invoice to send, one with no mail and one already in progress
         self.invoices = (
             self.first_eligible_invoice
-            | self.invoice_without_mail
-            | self.invoice_in_progress
+            + self.invoice_without_mail
+            + self.invoice_in_progress
```
jguenat, line 133: I usually prefer the union over the addition because with the addition you could end up with a recordset containing twice the same id which makes no sense in Odoo.

jguenat (approved):
LGTM
Thanks for the migration

PR #1531, opened 2023-08-16
[15.0][FIX] account_global_discount: Make tests resilient

victoralmau:
Depending on the installed set of modules, the company currency may be USD or EUR. If the second case, these tests will fail, so we make sure that the company currency is USD for our tests, doing the change by SQL, as there's a Python constraint that prevents it.

Not needed in v17 due to: odoo/odoo#107113.

@Tecnativa


pedrobaeza (approved):
Fast-tracking this for fixing CIs:

/ocabot merge patch

PR #1537, opened 2023-08-26
[14.0][FIX] account_invoice_triple_discount: recomputed taxes overwritten by old move line values

eantones:
We only need to save and restore the invoice lines, not all the move lines as is currently happening.

The restoration of the old lines is overwriting the changes made in the tax lines by the tax recomputation in line https://github.com/OCA/account-invoicing/blob/9b64ecc2b8e9b5146a12994b5c13f9a603b0db17/account_invoice_triple_discount/models/account_move.py#L30

This is a problem in a certain situations like in [repair_discount](https://github.com/OCA/manufacture/tree/14.0/repair_discount) module where the old values are kept due to this bug.
https://github.com/OCA/manufacture/blob/035a4de34a0de7e4f0db1a9c0ab3354acf235e99/repair_discount/models/mrp_repair.py#L115-L119

### DIFF
--- a/account_invoice_triple_discount/models/account_move.py
+++ b/account_invoice_triple_discount/models/account_move.py
@@ -16,9 +16,9 @@ class AccountMove(models.Model):
         restored after the original process is done
         """
         old_values_by_line_id = {}
-        digits = self.line_ids._fields["price_unit"]._digits
-        self.line_ids._fields["price_unit"]._digits = (16, 16)
-        for line in self.line_ids:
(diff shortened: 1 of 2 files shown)


rafaelbn:
If all of you, in this great OCA maka a cross review of the same module! I'm sure the force will be with us and you can fix all BUGs in all versions

@eantones @ao-landoo @Segui-S79 @odooNextev @ramiadavid 

imagen


❤️ OCA ❤️

eantones:
Fixed in https://github.com/OCA/account-invoicing/pull/1605

PR #1539, opened 2023-08-29
[16.0][FIX] account_invoice_currency_taxes: get the company's currency through the record environment

ghost:
Fix added so that in case `record` does not have `company_id` field, it can get the value from its own `env`.


JavierIniesta (commented):
Please add a description to the PR.

JavierIniesta (approved):
Code review: LGTM

gurneyalex (approved):

PR #1535, opened 2023-08-24
[13.0][IMP] account_product_move

ntsirintanis:
@NL66278

PR #1544, opened 2023-09-06
[ADD] [13.0] account_product_move enhanced

NL66278:
This module is already used successfully in a live environment at one of our customers for quite some time. However, the PR offering the module was actually never merged. Now the customer needs have expanded, and again we think more organizations could benefit from this.


NL66278:
Builds on: https://github.com/OCA/account-invoicing/pull/1535

NL66278:
Can anyone explain to me while with all files 100% covered by tests, the codecov/project still complains about decreased coverage?

NL66278:
@ntsirintanis @lfreeke Please review.

simahawk:
> Can anyone explain to me while with all files 100% covered by tests, the codecov/project still complains about decreased coverage?

That's normal for new modules. Just look at the diff stats in such cases :wink:

simahawk:
Could you please add a bit of description here? By reading the title I think of a new module due to "ADD" but  "enhanced" makes me think about an improving of something existing... and when I look at the version, it says "...4.0.0" which means it's something that was already there.... but where? Or.. the version is wrong!

And what about https://github.com/OCA/account-invoicing/pull/1535? Has to be closed? :dizzy_face: :smile:

NL66278:
@simahawk I expanded the description. Basically this module is already in use at one of our customers for some time, That's why the version numbers, supporting migration of older data. Other organizations might also be using the module already from the PR. As for the normal red cross for coverage for new modules: the problem is lots of reviewers will not even look at PR's that do not pass all tests. let alone merge them. So if something could be done about that, it would be great, no only for this module, but for all.

simahawk:
@NL66278 tnx for the update.
> Other organizations might also be using the module already from the PR. 

Fine, just explaining this in the description is enough for the rest of the world to understand and review based on this :wink: 

> As for the normal red cross for coverage for new modules: the problem is lots of reviewers will not even look at PR's that do not pass all tests.

That's not true, at least from my experience. In fact, not everybody care - or care enough - about test cov not to mention full test cov. It's unlikely that someone will block such PRs just because there's a not perfect score on codecov validation :wink: 

In any case, thanks for providing full test cov: we need more PRs with this approach :)

NL66278:
After a suggestion from @yajo I rebased the branch and all coverage is green now. Apparently the project coverage compares the submitted branch with the current (13.0) branch, instead of the current branch as it is now, with how it would change after merging the PR. Anyway I have now a way to fix these problems. Thanks @yajo and @simahawk for your comments!

PR #1546, opened 2023-09-07
[16.0][IMP] account_invoice_mass_sending: missing commits from 15

sergio-teruel:
Recover missing commits from v15

https://github.com/OCA/account-invoicing/commit/faabd09b330eabf4b177d2d2a791789806fd99bf
https://github.com/OCA/account-invoicing/commit/6f1b6740a37069597b8c1afe565848900d3d1a5f

cc @Tecnativa

ping @chienandalu @CarlosRoca13


pedrobaeza (approved):

account_invoice_mass_sending/README.rst
```diff
@@ -38,7 +38,7 @@ This module adds a mass sending feature on invoices, using asynchronous jobs.
 Configuration
 =============
 
-queue_job module needs to be configured, see the module `documentation <https://github.com/OCA/queue/tree/14.0/queue_job>`_
+queue_job module needs to be configured, see the module `documentation <https://github.com/OCA/queue/tree/15.0/queue_job>`_
```
pedrobaeza: This README seems outdated.

sergio-teruel: Changes Done!!

jguenat (approved):
I'm not against the change but could you explain why the definition of the job channel and function is necessary?
It is my understanding that it has no effect on its own (when only one channel) and also depends on how you define your channels capacity in odoo.conf.

chienandalu (approved):

pedrobaeza:
AFAIK, the idea is to be able to categorize jobs for being able to dedicate numbers for each channel.

/ocabot merge minor

PR #1562, opened 2023-09-25
[ADD] account_invoice_purchase_quantity

rousseldenis:


account_invoice_purchase_quantity/tests/test_invoice_purchase_quantity.py
```diff
@@ -0,0 +1,52 @@
...
+            {
+                "name": "Product Test",
+                "type": "product",
+            }
+        )
+        cls._create_order()
+
+    @classmethod
+    def _create_order(cls):
+        with Form(cls.env["purchase.order"]) as order_form:
```
IT-Ideas, line 23: I guess that `with Form()` is used to trigger some `onchange`. Otherwise I would advise not to use `with Form()`.

IT-Ideas (approved):
Thanks!

account_invoice_purchase_quantity/tests/test_invoice_purchase_quantity.py
rousseldenis: > I guess that `with Form()` is used to trigger some `onchange`. Otherwise I would advise not to use `with Form()`.

Why ? It is designed for tests.(but for tests only)

IT-Ideas: Just because they are explicitly calling `onchanges` which modifies the behavior that you would have if the use case could be called by a piece of code.
So if you test is not relying on `onchanges` it is safer not to use `Form`. Because in that case we presume that it can be called server side and that you will get the desired behavior. The idea being that if someone modifies your code and add an onchange to make it work, the test will fail... preserving the potentials server side calls.
Lastly to my understanding, `onchanges` should be changed into `computes`, except in some cases, so better not relying on them if not explicitly needed.

IT-Ideas: This is not blocking though 🙂 I approved the PR and wrote a comment. Just to share it with you.

rousseldenis:
@sebalix @jbaudoux @sbidoul @lmignon

jbaudoux (commented):
LG

jbaudoux:
Can you reopen, I'll approve

sbejaoui (approved):

lmignon (approved):

jbaudoux (approved):

jbaudoux:
@OCA/procurement-purchase-maintainers Could you merge this approved PR? Thanks :)

PR #1563, opened 2023-09-25
[16.0][IMP] account_invoice_pricelist: Move discount to onchange

hildickethan:
- Refactor some methods to reuse them elsewhere
- Move the `discount` calc to a dedicated `onchange` method (as discount doesn't have a `compute` here), as being in another fields compute isn't really correct

In the 16.0 migration the onchange methods were refactored to computes, but the editing of discount remained in the price_unit compute, which causes problems when editing the quantity and discount at once.

My case is that I have an invoice created from a sale order with 2 units at 200€ with a 50% discount (manually input, the pricelist calcs to 0%). A refund is created for this invoice that only wants to refund 1 of the 2 units. When creating the invoice everything is OK, but when editing the quantity to 1 the discount gets set to 0% due to the pricelist calc, this is fine but when you edit the discount back to 50% the subtotal compute ignores your 50% input and uses the 0% from the pricelist calc despite visually showing on the form 50% discount.

<details><summary>This occurs on runboat too</summary>
<p>

Invoice creation:
image

Manually editing the lines:
image

After saving:
image

With this PR the subtotal in the last screenshot becomes $100 like it should

</p>
</details>

### DIFF
--- a/account_invoice_pricelist/models/account_move.py
+++ b/account_invoice_pricelist/models/account_move.py
@@ -58,9 +58,10 @@ class AccountMove(models.Model):
         return res
 
     def button_update_prices_from_pricelist(self):
-        self.filtered(
-            lambda r: r.state == "draft"
-        ).invoice_line_ids._compute_price_unit()
+        moves = self.filtered(lambda r: r.state == "draft")
+        moves.invoice_line_ids._compute_price_unit()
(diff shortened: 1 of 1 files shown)


aritzolea (approved):
Code review, looks good to me

PR #1498, opened 2023-06-27
[ADD][16.0] account_invoice_validation: init

AnizR:
[New addon]

This module extends the functionality of vendor bills and refunds and adds an approbation mechanism. The approbation status is purely informative and doesn't block any other functionality's flow.

It includes a new entry point with a new 'approbation view' of account moves

A default validator can be set in the settings and this setting is company dependent.
An email is set when assigning approbation to a new user and a cron sends an email every week to remind assigned user to check every 'waiting approval' .

- [x] Possibility to add an validator on vendors. When this functionality is enabled, it adds a second level of approval that is applied before the validation of validator set on company.
- [ ]  _To improve: add icon?_

PR #1573, opened 2023-10-11
[16.0][ADD] account_invoice_validation_purchase_type

FrancoMaxime:
This PR needs https://github.com/OCA/account-invoicing/pull/1498 
It allows to define a product as a purchase type product. When a account move is approbated with a purchase type product, it updates all taxes and the account of all invoice lines with those defined on the purchase type product.


account_invoice_validation_purchase_type/models/account_move.py
```diff
@@ -0,0 +1,40 @@
...
+    def update_move_line_with_product_type(self):
+        for move in self:
+            product = move.product_type_id
+            taxes = product.taxes_id
+            expense_account = product.property_account_expense_id
+            lines = move.invoice_line_ids.filtered(
+                lambda l: l.display_type == "product"
+            )
+            for line in lines:
+                if line.move_type in ("in_invoice", "in_refund"):
```
acsonefho: Please use `move.get_concerned_types()` and you could do this `if` into the `filtered(...)`.

https://github.com/OCA/account-invoicing/pull/1498/files#diff-8a0f1c858462292484dca32e824b5f60d1ee759c3b2aa6ed1047b27b3ee534b3R180

account_invoice_validation_purchase_type/models/account_move.py
```diff
@@ -0,0 +1,40 @@
...
+                    name = line.name
+                    line.write(
+                        {
+                            "product_id": product.id,
```
acsonefho: Can you create a function to build this `dict`. :pray: 
As there are some changes compared to the original module, we should keep a way to add more values into this `dict`.

account_invoice_validation_purchase_type/models/account_move.py
```diff
@@ -0,0 +1,40 @@
...
+                            "name": name,
+                            "tax_ids": taxes.ids,
+                            "account_id": expense_account.id,
+                        }
+                    )
+
+    def write(self, vals):
+        res = super().write(vals)
+        if "product_type_id" in vals:
```
acsonefho, line 40: I think this should be only in the `create(...)`. The `product_type_id` shouldn't be updated.

account_invoice_validation_purchase_type/tests/test_account_invoice_validation_purchase_type.py
```diff
@@ -0,0 +1,201 @@
...
+                "amount": 10.0,
+            }
+        )
+        self.product1 = Product.create(
+            {
+                "name": "Product purchase type - Unit test 1",
+                "purchase_type": True,
+                "list_price": 500,
+                "property_account_expense_id": account_payable.id,
+                "taxes_id": tax_p1.ids,
```
acsonefho: You should use `Command.xxx(...)`. Giving a list of id works only because it's a test.

account_invoice_validation_purchase_type/tests/test_account_invoice_validation_purchase_type.py
```diff
@@ -0,0 +1,201 @@
...
+            }
+        )
+        self.product2 = Product.create(
+            {
+                "name": "Product purchase type - Unit test 2",
+                "purchase_type": True,
+                "list_price": 200,
+                "property_account_expense_id": account_payable.id,
+                "taxes_id": tax_p2.ids,
```
acsonefho: Same here

account_invoice_validation_purchase_type/tests/test_account_invoice_validation_purchase_type.py
```diff
@@ -0,0 +1,201 @@
...
+            }
+        )
+        self.product3 = Product.create(
+            {
+                "name": "Product purchase type - Unit test 3",
+                "purchase_type": True,
+                "list_price": 736.25,
+                "property_account_expense_id": account_payable.id,
+                "taxes_id": tax_p1.ids,
```
acsonefho: Same here

acsonefho (changes requested):

account_invoice_validation_purchase_type/models/account_move.py
FrancoMaxime: @acsonefho First somebody create the account.move, then another employee must validate the account.move and attach a purchase type to the account.move, so that can't be done in a create function as the account.move is already created when we attach a purchase type

acsonefho (approved):

PR #1576, opened 2023-10-13
 [15.0] [IMP] improve tests performance

josep-tecnativa:
CC @Tecnativa TT45458


sergio-teruel:
@josep-tecnativa with tracking_disable=True in context is enough.

pedrobaeza (approved):
Sergio, other places use that context to perform certain operations.

Example: https://github.com/odoo/odoo/blob/015a4ee8603106db744a2016034445980e602e05/addons/mail/models/mail_thread.py#L1894

so better to avoid the maximum. If not, the payback is only few memory bytes to store the expanded context.

/ocabot merge patch

PR #1583, opened 2023-10-18
[FIX] account_invoice_triple_discount: avoid updating lines w/o disc

ferran-S73:
I've encountered some cases in which the base amount of a tax line is recalculated in the super and then it's price_unit is set to it's old value in here. Seeing the original PR that included this I don't really see the point in doing this process for every aml since only the ones in the invoice lines tab can haves discounts.


legalsylvain (approved):
make sense

ferran-S73:
Close as it has already been merged at https://github.com/OCA/account-invoicing/commit/747083ae652e35b48d9030f5b5b346cf610a6031

PR #1584, opened 2023-10-19
[IMP] account_invoice_refund_line_selection: propagate 'invoice_payment_term_id' on account move reversal

ghost:
Propagate `invoice_payment_term_id` from the original `account.move` to the `account.move.reversal`.


account_invoice_refund_line_selection/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 
 {
     "name": "Account invoice refund line",
-    "version": "16.0.1.0.0",
+    "version": "16.0.1.0.1",
```
ljsalvatierra-factorlibre: There is no need to bump version inside the PR.

account_invoice_refund_line_selection/tests/test_refund_line.py
```diff
@@ -60,6 +60,9 @@ def test_partial_refund_in_invoice(self):
             refund.invoice_line_ids.product_id,
         )
         self.assertNotEqual(refund.amount_total, self.out_invoice.amount_total)
+        self.assertEqual(
+            refund.invoice_payment_term_id, self.out_invoice.invoice_payment_term_id
+        )
```
ljsalvatierra-factorlibre, line 66: We should check first if the invoice has indeed payment term.

ljsalvatierra-factorlibre (changes requested):

PR #1586, opened 2023-10-24
[FIX] 14.0 account_invoice_fixed_discount: fixing precompute tax line…

jumeldi74:
…s fixing in function _recompute_tax_lines when retrive from sale order

### DIFF
--- a/account_invoice_fixed_discount/models/account_move.py
+++ b/account_invoice_fixed_discount/models/account_move.py
@@ -12,7 +12,7 @@ class AccountMove(models.Model):
         self, recompute_tax_base_amount=False, tax_rep_lines_to_recompute=None
     ):
         vals = {}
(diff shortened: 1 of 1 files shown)


hitrosol:
LGTM, the test is also green...

maybe you are interested on this @flotho @Saran440

hitrosol (approved):

damdam-s (approved):
works like a charm. thanks

petrus-v:
Dear @OCA/accounting-maintainers,

This receive approved label, do you mind to merge it?

Thanks a lot for your time.

sergiocorato (changes requested):
Sorry but I don't understand the logic of this change, why use line_ids from an account.move?

What bug does this PR fix?

Tests are obviously green as the lines are the same plus the created lines to manage VAT and partner's debit/credit, which do not have the fixed discount as they are not visible on the main page of the invoice.

PR #1593, opened 2023-11-03
[16.0][IMP] account_invoice_fixed_discount: avoid replace in xml

CRogos:
I am currently facing a problem with to many decimal numbers in a standard Odoo installation on the invoice PDF. 
image

I assume this causes by this code section: 
image

I also evaluated that adding this options will solve the issue in standard Odoo:
t-options='{"widget": "float", "decimal_precision": "Discount"}'

https://github.com/odoo/odoo/blob/5911f056f9ed45a82a694782f82aa3170d0b74ac/addons/account/views/report_invoice.xml#L126C49-L126C100

But this module is using replace, which would bring back the problem. Therefore I would like to change the xml not using "replace".

### DIFF
--- a/account_invoice_fixed_discount/reports/report_account_invoice.xml
+++ b/account_invoice_fixed_discount/reports/report_account_invoice.xml
@@ -1,35 +1,32 @@
 <!-- Copyright 2017 ForgeFlow S.L.
      License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
 <odoo>
-    <template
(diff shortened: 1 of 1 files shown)


CRogos:
@PieterPaulussen could you review?

lucuswolfius (approved):
Changes LGTM

MohamedOsman7 (approved):
Changes LGTM

CRogos:
@JordiBForgeFlow could you review and merge this?

PR #1598, opened 2023-11-07
[14.0][IMP] account_invoice_base_invoicing_mode: Adding shared methods and tests

mt-software-de:
Refactoring of account_invoice_base_invoicing_mode, account_invoice_mode_* modules.
Move duplicated methods into account_invoice_base_invoicing_mode.
Creating a commen test


mt-software-de:
@jbaudoux @etobella @kluna1998 do you have time for a review?

kluna1998 (approved):
Great job! The refactoring makes sense to me 👍🏽

account_invoice_base_invoicing_mode/__manifest__.py
```diff
@@ -8,7 +8,7 @@
     "website": "https://github.com/OCA/account-invoicing",
     "license": "AGPL-3",
     "category": "Accounting & Finance",
-    "depends": ["account", "queue_job", "sale"],
```
etobella: Why do you need sale_stock?

I think you are using at the test, but you could do it manually by edting the qty_delivered, isn't it?

pedrobaeza: You shouldn't change dependencies of a stable module. If required, do a glue module.

pedrobaeza (changes requested):

account_invoice_base_invoicing_mode/__manifest__.py
mt-software-de: I is needed because of the refactoring, there was the dependency stock in account_invoice_mode_weekly/__manifest__.py declared.
But in fact it would also possible to remove this change and there for refactor also the unittests.

mt-software-de:
@pedrobaeza and @etobella could you review it please again?

pedrobaeza (approved):

etobella (approved):
LGTM

/ocabot merge minor

PR #1601, opened 2023-11-10
[12.0] ADD account_invoice_default_code_column

legalsylvain:
In odoo, user has the possibility to rename the "description" field of an invoice line, and at the same time delete / alter the product code.
But in some situation, removing the product code is a problem, because it should be present on the invoice report.

This module avoids this problem.

For that purpose, this module adds a new dedicated column "Default code"
in the account invoice report.
As a side effect, it improves readability of the invoices.

**Without this module**

1_without_module

**With this module**

2_with_module


account_invoice_default_code_column/models/account_invoice_line.py
```diff
@@ -0,0 +1,22 @@
...
+    name_without_default_code = fields.Char(
+        compute="_compute_name_without_default_code"
+    )
+
+    @api.depends("name", "product_id.default_code")
+    def _compute_name_without_default_code(self):
+        for line in self:
+            line.name_without_default_code = line.name
+            if line.product_id and line.product_id.default_code:
+                line.name_without_default_code = line.name_without_default_code.replace(
```
quentinDupont, line 20: very "habile"

quentinDupont (approved):
Code review and works in production ! Clarifies reading and managing products in PDF

rvalyi (approved):

PR #1592, opened 2023-10-31
[14.0] FIX account_invoice_triple_discount - _recompute_tax_lines involving only relevant lines: invoice_line_ids

eLBati:
Ref https://github.com/OCA/l10n-italy/pull/3681

PR #1605, opened 2023-11-14
[14.0] [FIX] account_invoice_triple_discount: for in invoice_line_ids if exists, else on line_ids

odooNextev:
…ts, else on line_ids
fix https://github.com/OCA/account-invoicing/pull/1592#issuecomment-1807762328

### DIFF
--- a/account_invoice_triple_discount/models/account_move.py
+++ b/account_invoice_triple_discount/models/account_move.py
@@ -16,9 +16,10 @@ class AccountMove(models.Model):
         restored after the original process is done
         """
         old_values_by_line_id = {}
-        digits = self.invoice_line_ids._fields["price_unit"]._digits
-        self.invoice_line_ids._fields["price_unit"]._digits = (16, 16)
-        for line in self.invoice_line_ids:
+        lines = self.invoice_line_ids or self.line_ids
+        digits = lines._fields["price_unit"]._digits
+        lines._fields["price_unit"]._digits = (16, 16)
+        for line in lines:
             aggregated_discount = line._compute_aggregated_discount(line.discount)
             old_values_by_line_id[line.id] = {
                 "price_unit": line.price_unit,
@@ -26,9 +27,9 @@ class AccountMove(models.Model):
             }
             price_unit = line.price_unit * (1 - aggregated_discount / 100)
             line.update({"price_unit": price_unit, "discount": 0})
-        self.invoice_line_ids._fields["price_unit"]._digits = digits
+        lines._fields["price_unit"]._digits = digits
         res = super(AccountMove, self)._recompute_tax_lines(**kwargs)
-        for line in self.invoice_line_ids:
+        for line in lines:
             if line.id not in old_values_by_line_id:
                 continue
             line.update(old_values_by_line_id[line.id])
@@ -36,9 +37,7 @@ class AccountMove(models.Model):
 
     def _has_discount(self):
         self.ensure_one()
+        lines = self.invoice_line_ids or self.line_ids
         return any(
-            [
-                line._compute_aggregated_discount(line.discount) > 0
-                for line in self.invoice_line_ids
-            ]
+            [line._compute_aggregated_discount(line.discount) > 0 for line in lines]
         )


francesco-ooops:
cc @eLBati @sergiocorato

francesco-ooops:
@OCA/accounting-maintainers this looks urgent to me

francesco-ooops:
@SirAionTech what do you think of this one?

HekkiMelody (approved):
Code review LGTM

**chore**: please adjust the commit message following the guidelines: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message 

**suggestion (non-blocking)**: since this is a rather intricate and delicate matter (had to read multiple issues in multiple languages to understand the context), I think adding a test would be helpful in avoiding regressions and it would implicitly also explain the reason behind the change. If not a test, at least a comment.

**chore**: once approved, please FW to v15 and 16!

francesco-ooops:
@ferran-S73 fyi

ferran-S73 (approved):

account_invoice_triple_discount/models/account_move.py
```diff
@@ -16,29 +16,28 @@ def _recompute_tax_lines(self, **kwargs):
         restored after the original process is done
         """
         old_values_by_line_id = {}
-        digits = self.invoice_line_ids._fields["price_unit"]._digits
-        self.invoice_line_ids._fields["price_unit"]._digits = (16, 16)
-        for line in self.invoice_line_ids:
+        lines = self.invoice_line_ids or self.line_ids
```
sergiocorato, line 19: I think it would be better to check if it `is_invoice` instead.

odooNextev: Working on the broken test, is_invoice returns True but without invoice_line_ids. So in this case it will not fix the test

SirAionTech:
> @SirAionTech what do you think of this one?

I think there should be a test in this module that explains what this change is fixing and ensures that there are no regressions in the future.
Basically I agree with https://github.com/OCA/account-invoicing/pull/1605#pullrequestreview-1733844809, but I'm more strict :angel:.
Changing this module to fix tests of another module, _in another repo_ too, makes little sense in my opinion.

If this module ain't broken, don't fix it.

almumu (approved):

anmarmo1 (approved):
Functional review is ok.

eLBati (approved):
Thanks
(rif https://github.com/OCA/l10n-italy/pull/3681#discussion_r1401858443)

eLBati:
> Changing this module to fix tests of another module, _in another repo_ too, makes little sense in my opinion

Why?

SirAionTech:
> > Changing this module to fix tests of another module, _in another repo_ too, makes little sense in my opinion
> 
> Why?

Because in my opinion, a module should test its own behavior.
This avoids regressions: otherwise nothing would stop someone else from changing back this code: CI stays :green_circle:, so why not?

If triple discount works good for invoices, why is it breaking sales? I see a few options:
- Is **account_invoice_triple_discount** not working good? It should be reproducible with this module alone, so there should be a test in here.
- Is **sale_triple_discount** expecting something different from this module? Why? Maybe because it was relying on some wrong behavior of **account_invoice_triple_discount**? There might be something to fix in there.

francesco-ooops:
@pedrobaeza can I ask your input here? 

the urgency is due to sale-workflow 14 tests currently failing

eLBati:
> This avoids regressions: otherwise nothing would stop someone else from changing back this code: CI stays 🟢, so why not?

This makes sense.
@odooNextev did you already try to add a test that makes `account_invoice_triple_discount` fail without the changes of this PR?
Otherwise I can work on this next week.

odooNextev:
> This makes sense. @odooNextev did you already try to add a test that makes `account_invoice_triple_discount` fail without the changes of this PR? Otherwise I can work on this next week.

We tried adding a test from `l10n_it_fatturapa_in` like this: https://github.com/OCA/l10n-italy/blob/14.0/l10n_it_fatturapa_in/tests/test_import_fatturapa_xml.py#L658

You can found it in the old PR: https://github.com/OCA/account-invoicing/pull/1442/files#diff-3c2a0c335b3658f70e6cc646d2f82097e866dbcd7267a0b3497df142d9cd2392R184

pedrobaeza:
This patch sounds weird, but let's continue for unblocking the CI:

/ocabot merge patch

PR #1613, opened 2023-11-21
[FIX] account_invoice_refund_reason: Filter refunds

BT-anieto:
Before this fix, when a second refund was created for an invoice
with a different reason_id, the reason_id field on the first invoice
was switched.

### DIFF
--- a/account_invoice_refund_reason/wizard/account_move_reversal.py
+++ b/account_invoice_refund_reason/wizard/account_move_reversal.py
@@ -19,6 +19,9 @@ class AccountMoveReversal(models.TransientModel):
                 record.reason = record.reason_id.name
 
     def reverse_moves(self):
+        """Overriden to set the reason_id fields in the new created refunds"""
         res = super().reverse_moves()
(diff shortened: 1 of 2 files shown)


account_invoice_refund_reason/wizard/account_move_reversal.py
```diff
@@ -19,6 +19,22 @@ def _compute_reason(self):
                 record.reason = record.reason_id.name
 
     def reverse_moves(self):
+        """Overriden to set the reason_id fields in the new created refunds"""
         res = super().reverse_moves()
-        self.move_ids.mapped("reversal_move_id").write({"reason_id": self.reason_id.id})
```
pedrobaeza, line 23: I think it's simpler to do:

```suggestion
        self.move_ids.mapped("reversal_move_id").filtered(lambda x: not x.reason_id).write({"reason_id": self.reason_id.id})
```

pedrobaeza (approved):

BT-nfrancisco (approved):
Looks good

AaronHForgeFlow (approved):
Code LGTM :+1:

PR #1619, opened 2023-11-27
[17.0][MIG] account_invoice_pricelist: Migration to 17.0

peluko00:
Module migrated to version 17.0

cc https://github.com/APSL 148829

@miquelalzanillas @lbarry-apsl @javierobcn @mpascuall please review

- [Issue] https://github.com/OCA/account-invoicing/issues/1600


mpascuall (approved):
Working correctly, checked in runboat

lbarry-apsl (approved):
Works, tested on Runboat

hildickethan:
I've been having some issues installing this PR alongside `l10n_es` in an environment with demo data, though i think any non-USD accounting would have the same issue. Creating a db with `odoo -i l10n_es,account_invoice_pricelist` with `without_demo=False` will error when installing the latter.
image

From my digging the problem seems to be that in this commit https://github.com/odoo/odoo/commit/83c52575d0ce3d7deb1737c622c6ce366db8c31b the way pricelists work in 17.0 was altered to not be required and to use the company's currency if it isn't set.
image

This is causing the demo sales/invoices to have EUR currency and then error in the `_check_currency` due to `_compute_pricelist_id` computing the default USD pricelist for the partners, which in turn changes the pricelist but the `_check_currency` happens before the `_compute_currency_id` fixes the currency :melting_face:

But the reason this only happens in 17.0 is due to aforementioned commit, it removes the basic "Public Pricelist", which is what would be the default computed EUR pricelist in previous versions (I checked on 16.0).
image

Since this no longer exists, it can only default to USD.
The easiest way to handle this is to avoid computing `pricelist_id` on existing invoices altogether. It also has the benefit of speeding up installation times which in big databases could be a few minutes. I don't think it's unreasonable to not touch existing invoices either.

This is the simple pre_init_hook I added locally that fixes the issue
<details><summary>Patch</summary>
<p>

```patch
diff --git a/account_invoice_pricelist/__init__.py b/account_invoice_pricelist/__init__.py
index 0650744..548c73e 100644
--- a/account_invoice_pricelist/__init__.py
+++ b/account_invoice_pricelist/__init__.py
@@ -1 +1,2 @@
+from .hooks import pre_init_hook
 from . import models
diff --git a/account_invoice_pricelist/__manifest__.py b/account_invoice_pricelist/__manifest__.py
index 7e365ba..b30a7cf 100644
--- a/account_invoice_pricelist/__manifest__.py
+++ b/account_invoice_pricelist/__manifest__.py
@@ -11,4 +11,5 @@
     "depends": ["account", "sale_management"],
     "data": ["views/account_invoice_view.xml"],
     "installable": True,
+    "pre_init_hook": "pre_init_hook",
 }
diff --git a/account_invoice_pricelist/hooks.py b/account_invoice_pricelist/hooks.py
new file mode 100644
index 0000000..ce5fe6a
--- /dev/null
+++ b/account_invoice_pricelist/hooks.py
@@ -0,0 +1,15 @@
+# Copyright 2024 Studio73 - Ethan Hildick <ethan@studio73.es>
+
+from odoo.tools.sql import column_exists, create_column
+
+
+def pre_init_hook(env):
+    cr = env.cr
+    # Speed up the installation of the module on an existing Odoo instance
+    # by not computing the pricelist for every invoice. Also avoids
+    # a possible computation error of
+    # computing all pricelists -> _check_currency constraint error before the
+    # currency can recompute itself
+    if not column_exists(cr, "account_move", "pricelist_id"):
+        create_column(cr, "account_move", "pricelist_id", "int4")
```

</p>
</details>

hildickethan:
> This is the simple pre_init_hook I added locally that fixes the issue
> 
> Patch

@peluko00 sorry for the long message but could you add this hook to the migration?

peluko00:
> > This is the simple pre_init_hook I added locally that fixes the issue
> > Patch
> 
> @peluko00 sorry for the long message but could you add this hook to the migration?

Thanks for the fix! Can you review now please

ecino (approved):
LGTM :+1:

grindtildeath:
Can you please cherry-pick https://github.com/OCA/account-invoicing/pull/1751 ? :pray:

yvaucher:
I'll merge this and cherry-pick #1751 in a new PR

yvaucher:
@grindtildeath I created the PR here: https://github.com/OCA/account-invoicing/pull/1806

PR #1625, opened 2023-12-12
[ADD] [15.0] account_invoice_discount_date: Early Payment Discount Date

Shide:
Add early discount date to move lines and invoices as Odoo 17 does.

This module must be deprecated in Odoo 17.

https://www.loom.com/share/13eca4ab1e3d4790938b518eee39a654?sid=eaf99617-a10b-4f36-9d29-036f43aa7e2b

MT-4074 @moduon @rafaelbn @yajo @EmilioPascual please review if you want :)


rafaelbn (approved):
Designed, worked and reviewed 👍🏼 

This PR is the conclusion of a big work and deeply analysed around the new functionality en Odoo 17 for this Stuff (that is great).

https://www.odoo.com/documentation/17.0/es/applications/finance/accounting/customer_invoices/cash_discounts.html

README @Shide is wonderful! Thank you for this lovely work! ❤️

fcvalgar (approved):
Functional review

LGTM, thank you @Shide

Gelojr (approved):
LGTM Nice job @Shide 
https://www.loom.com/share/104f120a2540477a9d36f06d6940d48c?sid=d454f450-e8ed-4d72-9329-ec285b72b9fc

account_invoice_discount_date/models/account_move.py
```diff
@@ -0,0 +1,31 @@
...
+        "for the Early Payment Discount to be granted",
+    )
+
+    @api.depends("line_ids.discount_date")
+    def _compute_discount_date(self):
+        for record in self:
+            d_dates = sorted(
+                record.line_ids.filtered("discount_date").mapped("discount_date")
+            )
+            record.discount_date = d_dates and sorted(d_dates)[0] or None
```
yajo: I think you don't need to sort them twice:

```suggestion
            record.discount_date = d_dates and d_dates[0] or None
```

account_invoice_discount_date/models/account_move.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def _inverse_discount_date(self):
+        """When set discount_date set on all move lines with date maturity"""
+        for record in self:
+            for line in record.line_ids:
+                line.discount_date = line.date_maturity and record.discount_date
```
yajo, line 32: If this method were right, it'd have a bug:

```suggestion
                line.discount_date = line.date_maturity and line.discount_date
```

However, `_compute_discount_date()` depends on the value of `line_ids.discount_date`, so the normal thing would be that an inverse method on `discount_date` would write in those fields instead.

But... an inverse field writing on itself? It's the 1st time I see this. 🤔  I think we could approach this better.

It seems to me that what you're trying to achieve is to forbid users from setting a `discount_date` if `date_maturity` is empty, right? If so, the standard way to do that would be:
1. Make `_compute_discount_date` depend on `date_maturity`, and set `discount_date = False` if `date_maturity` is `False`.
2. In the view, make `discount_date` readonly or invisible when `date_maturity == False`.

account_invoice_discount_date/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+{
+    "name": "Account Invoice Discount Date",
+    "summary": "Set the early discount date on invoices",
+    "version": "15.0.1.0.0",
+    "development_status": "Alpha",
+    "category": "Accounting",
+    "website": "https://github.com/OCA/account-invoicing",
+    "author": "Moduon, Odoo Community Association (OCA)",
+    "maintainers": ["Shide", "rafaelbn"],
+    "license": "AGPL-3",
```
yajo: According to our internal contribution guidelines, this should be LGPL-3

yajo (changes requested):

account_invoice_discount_date/models/account_move.py
Shide: I'll explan what I want to do:

1. If you set `discount_date` on the invoice, the **lines** with `date_maturity` will write its `discount_date` with the `discount_date` you put on the **invoice**. (This is where `_inverse_discount_date` triggers)
2. If you change the `discount_date` of a **line** that has `date_maturity` (found a bug here), get the lower `discount_date` of its **lines** and if it's different than the `discount_date` of the invoice, update it (and then trigger `_inverse_discount_date` again)

Regarding the bug if found, this works as I expected.

account_invoice_discount_date/__manifest__.py
Shide: All modules on account-invoicing have AGPL-3

account_invoice_discount_date/models/account_move.py
yajo: Sorry I completely misunderstood the code! It's OK indeed.

yajo (approved):
/ocabot merge patch

PR #1614, opened 2023-11-22
[FIX] account_invoice_fixed_discount: Check base_line type

tsliepen-ws:
hr_expense calls _convert_to_tax_base_line_dict with base_line set to None. Comparison of base_line._name in account_invoice_fixed_discount causes an attribute error.

Add extra check to ensure there is an account.move.line to check against in base_line.

PR #1636, opened 2024-01-05
[16.0][IMP] account_invoice_fixed_discount: Error on create expense

CRogos:
When hr_expense is installed, base_line can be None:

https://github.com/odoo/odoo/blob/fa0a6b077b171473da950c969de9e25c3577058e/addons/hr_expense/models/hr_expense.py#L187

image


CRogos:
Add unit test for #1614

@pedrobaeza

pedrobaeza (approved):
Well, this unit test is not very valid, as it's calling a direct internal method that doesn't reflect a real flow, but I understand that this only happens with `hr_expense` installed, so let's continue this way.

/ocabot merge patch

PR #1638, opened 2024-01-05
[16.0][REF] account_invoice_triple_discount: Consolidate discount in std field

grindtildeath:
Until now, the triple discount feature did use the standard Odoo field as the first discount field, adding only extra fields for the second and the third discount. This implied we had to override any function using discount to consider the other discounts properly.

By adding an extra field discount1 to store the first discount, it allows to redefine the standard discount field to a computed field that will consolidate the triple discount from the other fields, and avoid the need to redefine anything relying on the discount field as it will already consider the triple discounts.


grindtildeath:
ping @chienandalu @pedrobaeza @Nikul-Chaudhary @SirAionTech @ramiadavid 

Comments welcome

grindtildeath:
FTR: This change originates from this discussion https://github.com/odoo/odoo/pull/145513#pullrequestreview-1774865804

SirAionTech:
> ping @chienandalu @pedrobaeza @Nikul-Chaudhary @SirAionTech @ramiadavid
> 
> Comments welcome

Thanks for your contribution!
This is a cleaner approach indeed and I have talked about it a few times; I never got to an implementation because I was always told that both approaches had been discussed and the current one had been chosen in the end.
I think there was even another module with the same name implementing this approach in some fork.

Before looking at the code, I'd like to know from the authors if this is ok

chienandalu (commented):
We don't currently mantain this module. I like the approach though :smile: :+1:

leemannd (approved):
I like the approach.

The only point missing is an update of the reamde, especially the USAGE section

account_invoice_triple_discount/migrations/16.0.2.0.0/post-migrate.py
```diff
@@ -0,0 +1,23 @@
...
+    env["account.move.line"].search(
+        [
+            "|",
+            "|",
+            ("discount1", "!=", 0),
+            ("discount2", "!=", 0),
+            ("discount3", "!=", 0),
+        ]
+    )
+    env["account.move.line"]._compute_discount()
```
lmignon: IMO this will raise an exception for all the lines with a date prior to the fiscal year lock date...

account_invoice_triple_discount/migrations/16.0.2.0.0/pre-migrate.py
```diff
@@ -0,0 +1,29 @@
+# Copyright 2024 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+import openupgrade
```
legalsylvain: Hi. It should be openupgradelib, don't you think ?

account_invoice_triple_discount/migrations/16.0.2.0.0/post-migrate.py
legalsylvain: I guess it should be done by sql. Otherwise, it will raise recomputation for total od account move line. This is not desired (and forbidden)

legalsylvain (commented):
Nice refactoring and simplification ! 
Some remarks inline. 

I guess purchase module should be changed in the same time, don't you think ?
https://github.com/OCA/purchase-workflow/blob/16.0/purchase_triple_discount/__manifest__.py#L13

ramiadavid:
I am migrating purchase_triple_discount to v17 and I am using this method, I have added a post_ini_hook to move all the values ​​from "discount" to discount1 when the module is installed in a database that already has discount lines, I should not add it here too ?

In my case I have used:
`UPDATE purchase_order_line upd
SET discount1=pol.discount
FROM (SELECT *
      FROM purchase_order_line
      WHERE discount IS NOT NULL
        AND discount1 IS NULL
        AND discount2 IS NULL
        AND discount3 IS NULL) as pol
WHERE upd.id = pol.id`

grindtildeath:
Sorry for the delay in answering here, I had some holidays and then lot of catch up to do.

@ramiadavid Thanks for working on the migration. IMO the migration script from 16.0 to 17.0 should check if a `discount1` column exists before copying the value, so that it allows "migrating" to this new computation from v16.0, if we manage to complete this PR.

@lmignon @legalsylvain Thanks for the warning on the fiscal year lock date. Indeed the compute will not work with the field set. Do you think it's an acceptable solution to remove the fiscal year lock date before calling the compute and resetting it afterwards?

legalsylvain:
> @lmignon @legalsylvain Thanks for the warning on the fiscal year lock date. Indeed the compute will not work with the field set. Do you think it's an acceptable solution to remove the fiscal year lock date before calling the compute and resetting it afterwards?

Hi. SQL is is totally adequate to : 
- populate quickly big tables
- avoid warning / constrains / check like @lmignon mentioned. (but maybe other things (from odoo / OCA or custom modules) can generates errors.
- avoid to alter undesired things. (like write_date, or extra fields that depends on changed field)

Could you elaborate, why use ORM instead of SQL requests ?

grindtildeath:
@legalsylvain I have nothing against SQL, just wanted to avoid reimplementing the calculation logic that is already defined at python level :grimacing:

Anyway, seems I found my way to avoid the rounding issue, so would be nice if anyone could test this on a large DB.

ramiadavid:
I was just seeing some performance problems when confirming or creating invoices with many lines (500). I know it is not very common, but in this case we need to create them and it takes up to 10 minutes to process it.  And the problem has brought me exactly here, the method that exists until now makes many updates on all the lines when the taxes are recalculated (even if they do not have discount2 or discount3) because it does not check it.  I was thinking about uploading a patch, but with this change you are making here it would be solved since we stopped writing the discount field twice for each line.  If I can collaborate on something, you just have to tell me.  To finish as soon as possible.  Thank you!

lmignon:
@ramiadavid In a first time you could try #1664. It should improve performances.

ramiadavid:
I'll try it, I hadn't seen it since the title is not well written...
[16.0][FIX] **account_invoice _riple_discount**: Fix tax recompute on closed fiscal year

lmignon:
The title is right; it fixes a bug when taxes are recomputed on a closed
fiscal year. As a side effect, it improves the compute method and avoid
useless recompute...

account_invoice_triple_discount/migrations/16.0.2.0.0/pre-migrate.py
```diff
@@ -0,0 +1,32 @@
+# Copyright 2024 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+from openupgradelib import openupgrade
+
+
+def migrate_discount_to_discount1(env):
```
legalsylvain, line 6: This function should be called in the post-init hook, once the module is installed.
Otherwise, database will be corrupted, because discount1 = 0, even if discount != 0. Don't you think ? 

Edit : done here : https://github.com/grindtildeath/account-invoicing/pull/7

legalsylvain:
hi @grindtildeath 
- I made a simplification + fix here : https://github.com/grindtildeath/account-invoicing/pull/7 could you take a look ? 
- A very big thanks for this PR. I've rarely seen such a complex OCA module made so simple (and so easy to maintain).

Kind regards.

legalsylvain:
Note : I adapted purchase_discount module to this refactoring. see : https://github.com/OCA/purchase-workflow/pull/2296

ramiadavid (approved):
Just as a suggestion, why change to optional="show" by default when odoo has the discount field hidden, wouldn't it be more consistent to also leave it hidden by default?

legalsylvain (changes requested):
wait for merge of https://github.com/grindtildeath/account-invoicing/pull/7

legalsylvain:
> Just as a suggestion, why change to optional="show" by default when odoo has the discount field hidden, wouldn't it be more consistent to also leave it hidden by default?

I'm not sure. Odoo set optional="hide" because it considers that maybe the feature is not very used, to make the UI cleaner. 

In our context, if we install account_invoice_triple_discount, we can guess that users want to access this field, so optional="show" is not inconsistent for me.

Just a point of view. Matter of taste for me, not a blocking point.

grindtildeath:
@legalsylvain Thank you so much for your help and your PR :star_struck: 

@ramiadavid One one hand, you're right (setting it hidden would be closer to std odoo). On the other hand, those having the module installed are more likely to use these fields. I'd be ok to set it as hidden if that's what other want.

legalsylvain (approved):
This one is ready to merge with https://github.com/OCA/purchase-workflow/pull/2296

account_global_discount/__manifest__.py
```diff
@@ -9,6 +9,9 @@
     "website": "https://github.com/OCA/account-invoicing",
     "license": "AGPL-3",
     "depends": ["account", "base_global_discount"],
+    "excludes": [
```
sergiocorato, line 12: As I will use nobump, in this manifest the version has to be bumped manually.

sergiocorato (changes requested):

account_global_discount/__manifest__.py
legalsylvain: Nice catch !

account_invoice_triple_discount/report/invoice.xml
```diff
@@ -5,14 +5,26 @@
         inherit_id="account.report_invoice_document"
         priority="100"
     >
-        <xpath expr="//t[@t-set='display_discount']" position="replace">
-            <t t-set="display_discount" t-value="o._has_discount()" />
+        <xpath expr="//t[@t-set='display_discount']" position="after">
```
Feyensv:
```suggestion
        <xpath expr="//t[@t-set='display_discount']" position="attributes">
```

grindtildeath: Good catch :+1:  (and thx for the review)

legalsylvain:
@grindtildeath : could you : 
- rebase (there is conflict)
- bump manually the version of account_global_discount. (See [comment](https://github.com/OCA/account-invoicing/pull/1638/files#diff-fb8b000cd9b77a2aa896ca6638f67b253d8e1487f4989b48f36a9be01370c724))

So we can merge it after.
Thanks !

grindtildeath:
@legalsylvain 
Actually there's an issue when `account_invoice_pricelist` is computed as the module updates the `discount` field when computing `price_unit`: https://github.com/OCA/account-invoicing/blob/14.0/account_invoice_pricelist/models/account_move.py#L191-L223  :weary: 

I need to find more time to refactor that module to have this working 100% but I don't know when I'll be able to.

legalsylvain:
Hi @grindtildeath Thanks for your answer. 
Could you at least rebase ? I want to use it with gitaggregate, but for the time being, there are conflict.

For the write of discount in account_invoice_pricelist, indeed, there is some problem overthere. i'll try to make a PoC.

legalsylvain:
@grindtildeath  could you take a look here : https://github.com/grindtildeath/account-invoicing/pull/8

I set an inverse function on ``discount`` field.
- It works (see tests).
- it doesn't generate recursive computation when we write on ``discount`` field.

what do you think ? 

thanks !

legalsylvain:
@sergiocorato is it now OK for you ?

grindtildeath:
@legalsylvain the issue as I see it is the inverse function is only executed on write, whereas the update of discount field is executed through an onchange by the compute function. Which means on the UI, when the product is selected, discount is updated but discount1 stay 0 if we don't save. :disappointed:

legalsylvain:
Hi @grindtildeath. I don't get the use case. Do you mean : with or without account_invoice_pricelist ?

ramiadavid:
I don't know if you have already seen this, but when an invoice is created from a sales order the discount1 field is not filled with the discount field of the order

ramiadavid:
I think this should be added:

```
@api.model_create_multi
def create(self, vals_list):
    for vals in vals_list:
        if vals.get("discount") and not any(vals.get(field) for field in self._get_multiple_discount_field_names()):
            vals["discount1"] = vals.pop("discount")
    return super().create(vals_list)
```

And this way this would no longer be necessary:
https://github.com/OCA/purchase-workflow/blob/2104054c4322916a202a47c6caab692ce582f5fc/purchase_discount/models/purchase_order.py#L112-L119

legalsylvain:
@ramiadavid : Done here https://github.com/grindtildeath/account-invoicing/pull/9

could you review ? 

CC : @grindtildeath

legalsylvain:
>  Which means on the UI, when the product is selected, discount is updated but discount1 stay 0 if we don't save. 😞

if Odoo/account module is installed (without any OCA module) : 
- on purchase invoice, when we select a product, the price is set based on the 'standard_price' of the product. There is no discount set. (even if there is a discount on the "product.supplierinfo" model.
- on sale invoice, when we select a product, the price is set, based on the 'list_price' of the product. There is no discount set.

could you elaborate ? for the time being Odoo doesn't populate 'discount' field. (unlike the sale and purchase module.)

ramiadavid:
If you create an invoice from a purchase order (with three discounts) the invoice is created with the three discounts added in the discount1 field.

The problem seems to occur in the `_inverse_discount` function

http://oca-account-invoicing-16-0-pr1638-64605a2af360.runboat.odoo-community.org/web#id=11&cids=1&menu_id=204&action=319&model=purchase.order&view_type=form

legalsylvain:
Hi @ramiadavid. Sorry I don't get it. the link you provide does'nt seems to be valid. Could you elaborate the step to reproduce ? Which module installed ? The action to reproduce the bug. (text or printscreen...)
thanks !

ramiadavid:
Yes, install purchase_triple_discount, create purchase order with three discounts, then create invoice from purchase order and discount in invoice its not correct

legalsylvain:
> Yes, install purchase_triple_discount, create purchase order with three discounts, then create invoice from purchase order and discount in invoice its not correct

but AFAIK, it should be fixed in purchase_triple_discount not here. Or did I missed something ?

ramiadavid:
I think not, because the only thing the invoice_triple_discount module does is add the fields discount1, discount2 and discount3 to the values ​​to create the invoice line, the invoice creation process is standard odoo

The problem is that after creating the standard "discount" field when calculating, the inverse_discount function deletes the other discounts

legalsylvain:
> The problem is that after creating the standard "discount" field when calculating, the inverse_discount function deletes the other discounts

Hi. Thanks for reporting. I dont reproduce this bug. Could you provide a test that is failing with the current implémentation ?

Thanks !

ramiadavid:
@legalsylvain 
https://github.com/grindtildeath/account-invoicing/pull/10

grindtildeath:
@legalsylvain @ramiadavid 
Sorry for the delay, and thank you very much for your contributions.

I just merged both of your PRs, reverted the inverse function, and squashed a bit the commits.

Moreover I opened #1751 to fix the existing issue with account_invoice_pricelist.

I will test a bit more but it's starting to look quite good :sunglasses:

legalsylvain:
> reverted the inverse function

Why ? this make the CI red. the test ``test_08_write_on_main_discount`` is failing. (https://github.com/OCA/account-invoicing/actions/runs/9714603374/job/26814030864?pr=1638#step:8:47)

Write on main ``discount`` field should work. Otherwise, this module will not work with extra module that write an discount field.

grindtildeath:
@legalsylvain 
The issue with the inverse function is it's triggered only on save/write. That means, when a user is creating an invoice manually on the UI, here is a description of what happens (with for example account_invoice_pricelist installed) and what I'm trying to avoid:
1. After the product is selected, the onchange function will set the `discount1` field.
2. The user enters another discount in `discount2`. (which will recompute `discount`)
3. When saving, `discount` is written, and the inverse function will write its value in `discount1` and reset `discount2` to 0.

I don't think we need this inverse, as IMO when this module is installed, `discount` is a computed field and should not be written into.

account_invoice_triple_discount/tests/test_invoice_triple_discount.py
```diff
@@ -153,56 +154,100 @@ def test_04_discounts_decimals_tax(self):
...
+        invoice_line1 = invoice.invoice_line_ids[0]
+        # 1000 * 0.9 * 0.8 * 0.7
+        self.assertEqual(invoice_line1.price_subtotal, 504.0)
+
+        invoice_line1.discount = 50
+        self.assertEqual(invoice_line1.discount1, 50.0)
+        self.assertEqual(invoice_line1.discount2, 0.0)
+        self.assertEqual(invoice_line1.discount3, 0.0)
+        self.assertEqual(invoice_line1.price_subtotal, 500.0)
+
```
legalsylvain:
```suggestion
```

legalsylvain (commented):
> I don't think we need this inverse, as IMO when this module is installed, discount is a computed field and should not be written into.

Ok ! let's go with that design ! The interest to add an inverse function was to avoid to patch some extra modules (OCA or custom modules) to write a "if triple_discount installed,  write on discount1 else, write on discount field)
like : https://github.com/OCA/purchase-workflow/pull/2296

but once that point handled, I think it's OK.

Indeed, there is a design problem with the inverse function. Better to remove this section.

could you fix CI, and so we can merge this one ? (see remarks inline + pre-commit)

thanks for your work !

legalsylvain:
@grindtildeath 
Could you take a look on red precommit CI ?

grindtildeath:
@legalsylvain Sorry for the delay.

IMO it looks like it must be good to merge from.

grindtildeath:
@sergiocorato can you please update your review?

legalsylvain:
@sergiocorato question has been answered.

3 approvals, we can merge. Thanks all for your review, and thanks @grindtildeath for your great simplification ! 

/ocabot merge nobump

legalsylvain:
an @OCA/accounting-maintainers is required here ! 
Thanks ! 
(please, use nobump)

grindtildeath:
@legalsylvain I'm part of the accouting maintainers, however, IMO we should make sure we can merge https://github.com/OCA/sale-workflow/pull/2885 at the same time we merge this one.

I'm also wondering, if we shouldn't add a warning in the logs in case both modules are not updated together...

legalsylvain:
> however, IMO we should make sure we can merge https://github.com/OCA/sale-workflow/pull/2885 at the same time we merge this one.

Could you take a look on red CI in the other PR ?

> I'm also wondering, if we shouldn't add a warning in the logs in case both modules are not updated together...

I dont understand in which case it can occure...

grindtildeath:
@legalsylvain 
I fixed the PR in sale-workflow and added a POC for versions dependencies warning, but I'll work on a better solution next week.

There are so many ways a single module could have its code updated and not another, it all depends how each project is managed.

legalsylvain:
> I fixed the PR in sale-workflow and added a POC for versions dependencies warning, but I'll work on a better solution next week.

The CI looks still red.

pedrobaeza:
Can you avoid the modules exclusion in the CI? And what about if you want to use both modules together? They are not excluding features.

rousseldenis:
@grindtildeath

legalsylvain:
finished here : https://github.com/OCA/account-invoicing/pull/1840

legalsylvain:
hi @grindtildeath : 
- this PR can be closed, as #1840 is merged.
- Could you review #1938 ? There is a regression in the invoice report. (discount columns are not displayed for the time being).

Note: there is some remaining work to do in sale-workflow https://github.com/OCA/sale-workflow/pull/2885.

PR #1642, opened 2024-01-11
[15.0][FIX] account_invoice_tax_required: Don't bypass restriction on mass validation

pedrobaeza:
If you use the action in the invoice list for doing mass validation, the restriction about taxes required is bypassed. That's because the mass validation wizard is not calling `action_post`, but directly `_post`.

@Tecnativa TT47059


pedrobaeza:
Added a second commit for improving the message, indicating the affected invoice.

carlosdauden (approved):

carolinafernandez-tecnativa (approved):

Issue #1637, opened 2024-01-05
account_invoice_warn_message: not only for customer invoices


dalonsod:
Hello @HviorForgeFlow @manuelregidor ,

I've noticed that this addon only takes in account customer invoices, and doesn't work for vendor bills, it's intetionally coded for `out_` moves. Is this behavior desired for any reason? If we make a PR extending it for `in_` moves, will you consider it?

Thanks in advance!!

cc @ChristianSantamaria @Anxo82

dalonsod:
Developed in #1645

PR #1645, opened 2024-01-12
[13.0][IMP] account_invoice_warn_message: added full invoice types support

dalonsod:
This is achieved with a hook method that returns move types enabled for warning messages, so it caold be extended.

Linked issue: #1637

### DIFF
--- a/account_invoice_warn_message/__manifest__.py
+++ b/account_invoice_warn_message/__manifest__.py
@@ -10,5 +10,5 @@
     "author": "ForgeFlow, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/account-invoicing",
     "depends": ["account"],
-    "data": ["views/account_move_views.xml"],
+    "data": ["views/account_move_views.xml", "views/res_config_settings_views.xml"],
 }
--- a/account_invoice_warn_message/models/__init__.py
+++ b/account_invoice_warn_message/models/__init__.py
@@ -1,1 +1,3 @@
+from . import res_company
+from . import res_config_settings
 from . import account_move
--- a/account_invoice_warn_message/models/account_move.py
+++ b/account_invoice_warn_message/models/account_move.py
@@ -3,6 +3,19 @@
 
 from odoo import api, fields, models
 
+from .res_company import INVOICE_WARN_MESSAGE_SELECTION
+
+INVOICE_WARN_MESSAGE_SEL_MAPPING = {
+    INVOICE_WARN_MESSAGE_SELECTION[0][0]: ["out_invoice", "out_refund"],
+    INVOICE_WARN_MESSAGE_SELECTION[1][0]: ["in_invoice", "in_refund"],
+    INVOICE_WARN_MESSAGE_SELECTION[2][0]: [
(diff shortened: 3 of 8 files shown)


dalonsod:
@HviorForgeFlow @manuelregidor could you review? If merged, I'll port it to v15

manuelregidor (approved):
LGTM. However, with your change you are already including purchase invoices. @dalonsod Don't you think we should return just a list composed just by "out_invoice" and "out refund"?

dalonsod:
Hello, I don't understand... the goal of this change is actually adding `in_*` invoices support, and I've only made an extra refactoring to help customization at this point. What do you exactly propose?

HviorForgeFlow:
Hello @dalonsod the module was created to show warning messages when invoices are created. At the first point no bill warn was considered. I'm just wondering what would happen if an specific partner has warns on invoicing and billing. Could it be considered to split this warns for both or it is better to keep simplicity and allow only one message for billing and invoice.

Thanks to arise the issue!

manuelregidor:
@dalonsod What I mean is that before your improvement, the warn message would only be shown in customer invoices. However, with your improvement, by default, it would be shown in customer and provider invoices. What I mean is that, because the  _get_move_types_warn method allows customization, it might return only "out_invoice" and "out_refund" and, in case we also need to show warn messages in provider invoices, we could inherit the funcion and add "in_invoice" and "in_refund". This way, people who are already using the module won't notice any difference and people who need to show the warn messages in provider invoices can inherit the method and add new types. Does it make sense now? It is not that I don't agree with what you propose,  the module is easily cosutomizable now, and I think is good. Thank you!

dalonsod:
Thanks for your explanations @manuelregidor @HviorForgeFlow that's why I've opened #1637 before and asked for the better option.

I'll modify this PR and simply add hook method, so the next steps could be developed later.

dalonsod:
Changes done

dalonsod:
@HviorForgeFlow could you review again and approve it, if it's OK?

HviorForgeFlow (approved):
Code review :+1:

I'm not blocking adding bill warns feature here allow it with company or config setting and keeping backward compatibility

dalonsod:
Don't worry @HviorForgeFlow I could make in another PR final changes as you suggested

dalonsod:
@pedrobaeza could you merge this? Thanks!

pedrobaeza:
I don't think this should go with a hook, but directly configuring what do you want to consider, as this module should include both.

dalonsod:
Thanks, Pedro... I don't understand, what do you mean when you're saying "as this module should include both"?

pedrobaeza:
IMO, this module should allow to show the warning for:

- Customers
- Vendors
- Both

being configurable for example at system level.

dalonsod:
Got it, thanks! This also fits @HviorForgeFlow requirements... I'll adapt it

dalonsod:
@pedrobaeza comments attended

pedrobaeza (approved):

dalonsod:
I'll plan port this to v15, thanks for all

PR #1647, opened 2024-01-17
[15.0][ADD]account_invoice_custom_rounding: Custom taxes rounding method in invoices

manuelregidor:
New module to select a custom taxes ronding method in customer invoices.


manuelregidor:

account_invoice_custom_rounding/models/account_move.py
```diff
@@ -0,0 +1,66 @@
...
+        compute="_compute_tax_calculation_rounding_method",
+        store=True,
+        readonly=False,
+        help="How total tax amount is computed. If no value selected, "
+        "the method defined in the company is used.",
+    )
+
+    @api.depends("partner_id")
+    def _compute_tax_calculation_rounding_method(self):
+        for sel in self:
```
ValentinVinagre: change 'sel' to 'move' as in 'def _recompute_dynamic_lines'

ValentinVinagre (approved):
LGTM 👍🏻

account_invoice_custom_rounding/models/account_move.py
```diff
@@ -0,0 +1,66 @@
...
+        help="How total tax amount is computed. If no value selected, "
+        "the method defined in the company is used.",
+    )
+
+    @api.depends("partner_id")
+    def _compute_tax_calculation_rounding_method(self):
+        for move in self:
+            tax_calculation_rounding_method = False
+            if (
+                move.move_type in ["out_invoice", "out_refund"]
```
etobella: Why only out_invoices? Could be applied to in_invoices, isn't it?

I would recomend to do `move.is_invoice()`

etobella (changes requested):
It has a bug

You cannot edit anymore the tax amount on supplier invoices. probably related to the write modification of account.move

etobella:
https://github.com/OCA/account-invoicing/assets/28590170/b624af25-8a4a-444c-b576-e07ec0e56314

account_invoice_custom_rounding/models/account_move.py
manuelregidor: @etobella Thank you, we'll take a look.

manuelregidor:
@etobella Now the module also works with vendor invoices and the issue you address in the video should be sorted out now.

account_invoice_custom_rounding/models/account_move.py
```diff
@@ -0,0 +1,63 @@
...
+            )._recompute_dynamic_lines(recompute_all_taxes, recompute_tax_base_amount)
+        return super(
+            AccountMove,
+            self.filtered(
+                lambda a: not (a.is_invoice() and a.tax_calculation_rounding_method)
+            ),
+        )._recompute_dynamic_lines(recompute_all_taxes, recompute_tax_base_amount)
+
+    def write(self, vals):
+        res = super().write(vals)
```
etobella, line 56: Why do you need to add this change in write?

manuelregidor: So taxes are recomputed in case the value in the field tax_calculation_rounding_method changes.

etobella: And then why you need the context?

manuelregidor: Sorry, you're right. Because of the last changes, part of the context is not needed anymore. However, check_move_validity=False is still needed. Otherwise, it's not possible to save changes when the rounding method is changed.

etobella: I see... Can you add a comment on the function then?

manuelregidor: Done

ValentinVinagre:
@HaraldPanten could u review please 😄  ?

Jesarregui (approved):
LGTM

etobella (approved):
Thanks for the effort!

PR #1648, opened 2024-01-17
[15.0][IMP] account_invoice_triple_discount: only manage lines with discount2 or discount3

manuelregidor:
Before this improvement, all the account.move.lines in an invoice were treated as if they had the second or third discount applied.

With this improvement, only lines affected by the second or third discount are managed in the inherited method _recompute_tax_lines, so the price_unit field is only written in those lines that need it.

### DIFF
--- a/account_invoice_triple_discount/models/account_move.py
+++ b/account_invoice_triple_discount/models/account_move.py
@@ -3,6 +3,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from odoo import models
+from odoo.tools import float_compare
 
 
 class AccountMove(models.Model):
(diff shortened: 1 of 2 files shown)


legalsylvain (approved):
that makes sense. nice improvement. Could you port this patch in V16, once merged ? 

thanks !

manuelregidor:
@legalsylvain Yes, sure, I'll port it to v16 once merged.

manuelregidor:

ValentinVinagre (approved):
LGTM 👍🏻

HaraldPanten (approved):

manuelregidor:
@legalsylvain I said that I could port this patch to v16. However, I've seen this PR https://github.com/OCA/account-invoicing/pull/1638, that changes the entire logic of the module, so I will not port it to v16 until we know if this PR is finally merged if that's OK.

PR #1651, opened 2024-01-22
[15.0][IMP] account_invoice_warn_message: added full invoice types support

ChristianSantamaria:
With this addition, warn messages are set only for customers by default, but can be set for vendors or both customers and vendors.

Port from v13 #1645


ChristianSantamaria:
@HviorForgeFlow @manuelregidor, could you review?

PR #1654, opened 2024-01-28
[16.0] mig account_invoice_default_code_column

legalsylvain:
Trivial port from 12.0

### Description 
In odoo, user has the possibility to rename the "description" field of an invoice line, and at the same time delete / alter the product code.
But in some situation, removing the product code is a problem, because it should be present on the invoice report.

This module avoids this problem.

For that purpose, this module adds a new dedicated column "Default code"
in the account invoice report.
As a side effect, it improves readability of the invoices.

**Without this module**

1_without_module

**With this module**

2_with_module


SandieFavre (approved):
Functionnal review

AungKoKoLin1997 (approved):
Code Review: LGTM

legalsylvain:
/ocabot merge nobump
thanks for your reviews

rvalyi (approved):

PR #1659, opened 2024-02-01
[15.0][IMP] account_invoice_tax_required: Add customer information to error message for missing taxes in invoice lines

pilarvargas-tecnativa:
The customer is included in the error message of the line containing a product without tax and an action is added to go directly to the list of invoices containing those lines.

cc@Tecnativa TT47521

@carlosdauden @pedrobaeza  please review

### DIFF
--- a/account_invoice_tax_required/models/account_move.py
+++ b/account_invoice_tax_required/models/account_move.py
@@ -5,7 +5,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from odoo import SUPERUSER_ID, _, models
-from odoo.exceptions import UserError
+from odoo.exceptions import RedirectWarning
 from odoo.tools import config
 
 
@@ -14,8 +14,10 @@ class AccountMove(models.Model):
 
     def _test_invoice_line_tax(self):
         errors = []
+        invoice_error_ids = []
         error_template = _(
-            "Invoice %(invoice)s has a line with product %(product)s with no taxes"
+            "Invoice %(invoice)s for customer %(customer)s "
+            "has a line with product %(product)s with no taxes"
         )
         for invoice_line in self.mapped("invoice_line_ids").filtered(
             lambda x: x.display_type is False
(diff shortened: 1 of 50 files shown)


account_invoice_tax_required/models/account_move.py
```diff
@@ -15,14 +15,16 @@ class AccountMove(models.Model):
     def _test_invoice_line_tax(self):
         errors = []
         error_template = _(
-            "Invoice %(invoice)s has a line with product %(product)s with no taxes"
+            "Invoice %(invoice)s  for customer %(customer)s "
```
pedrobaeza: Extra space:

```suggestion
            "Invoice %(invoice)s for customer %(customer)s "
```

pedrobaeza (commented):
Adapt also the POT and PO files.

Why not putting the ID?

carlosdauden:
Why do not show invoices instead of lines?

pilarvargas-tecnativa:
> Why do not show invoices instead of lines?

You mean the message?
Right now the message does save the line error, i.e. a text string for each line/product but the action leads to the list of invoices containing the error.

pilarvargas-tecnativa:
> Why do not show invoices instead of lines?

I had changed it because it didn't fit to show the lines, but I had saved the changes in the wrong commit...

Now it is correct

account_invoice_tax_required/models/account_move.py
```diff
@@ -14,22 +14,44 @@ class AccountMove(models.Model):
...
+                "domain": [("id", "in", invoice_error_ids)],
+                "search_view_id": [
+                    self.env.ref("account.view_account_move_filter").id,
+                    "search",
+                ],
+                "views": [
+                    [self.env.ref("account.view_move_tree").id, "list"],
+                    [self.env.ref("account.view_move_form").id, "form"],
+                ],
+            }
```
carlosdauden: Can you try this?

```suggestion
            invoice_error_ids = list(set(invoice_error_ids))
            action_error = {
                "view_mode": "form",
                "name": _("Invoices with Missing Taxes"),
                "res_model": "account.move",
                "type": "ir.actions.act_window",
                "search_view_id": [
                    self.env.ref("account.view_account_move_filter").id,
                    "search",
                ],
                "views": [
                    [self.env.ref("account.view_move_form").id, "form"],
                ],
            }
            if len(invoice_error_ids) > 1:
                action_error["view_mode"] = "tree,form"
                action_error["domain"] = [("id", "in", invoice_error_ids)]
                action_error["views"].insert(
                    0, [self.env.ref("account.view_move_tree").id, "list"])
            else:
                action_error["res_id"] = invoice_error_ids[0]
```

carlosdauden (changes requested):

account_invoice_tax_required/models/account_move.py
pilarvargas-tecnativa: done! :)

carlosdauden:
Message and action only contains one invoice errors, change this to check all invoices:
https://github.com/OCA/account-invoicing/blob/9b1ac9c831daac428acb03431223f6e359a5ba85/account_invoice_tax_required/models/account_move.py#L83-L85

pilarvargas-tecnativa:
> Message and action only contains one invoice errors, change this to check all invoices:

done!

account_invoice_tax_required/models/account_move.py
```diff
@@ -50,7 +80,7 @@ def _post(self, soft=True):
                 config["test_enable"],
             )
         )
-        for move in self:
-            if move.move_type != "entry" and (force_test or not skip_test):
-                move._test_invoice_line_tax()
+        self.filtered(
+            lambda m: m.move_type != "entry" and (force_test or not skip_test)
+        )._test_invoice_line_tax()
```
carlosdauden:
```suggestion
        if force_test or not skip_test:
            self.filtered(lambda m: m.is_invoice())._test_invoice_line_tax()
```

carlosdauden (changes requested):

account_invoice_tax_required/models/account_move.py
pilarvargas-tecnativa: thanks!

pedrobaeza (approved):
Thanks for caring about having all translations synced.

/ocabot merge minor

pedrobaeza:
Please fw-port it to v16

PR #1660, opened 2024-02-01
[FIX] account_invoicing: compute price recalcule

adrip-s73:
This change is to prevent the unit price from being recalculated since when making a refund invoice the price is recalculated and the price of the product is taken.


legalsylvain (commented):
Could you improve message commit ? Correct module name and description ?
Otherwise, lgtm.

legalsylvain (approved):
Thanks !

PR #1663, opened 2024-02-05
[16.0][IMP] account_invoice_tax_required: Improve error message for lack of tax

pilarvargas-tecnativa:
Port from v16 #1659 

The customer is included in the error message of the line containing a product without tax and an action is added to go directly to the list of invoices containing those lines.

In addition, the method call is changed so that it receives and checks all the invoices instead of doing it one at a time.

cc @Tecnativa TT47521

@pedrobaeza @carlosdauden please review


chienandalu (approved):

pedrobaeza (approved):
/ocabot merge minor

PR #1664, opened 2024-02-09
[16.0][FIX] account_invoice_triple_discount: Fix tax recompute on closed fiscal year

lmignon:
When computing the taxes and the totals on account.move.line it's required to prevent calls to the write method on account.move or triggering recompute of unexpected fields while we assign a temporary value to the discount field to use the aggregated one bedore calling the base implementation. Otherwise the update of the discount field will invalidate totals and ... on related accoun.move and trigger calls to the write method of account.move. Any call to the write method will trigger a set of check to see if the balance is still right or check if you try to write on a record prior to the fiscal year close date or .... At the end, every time you try to access to the taxes on the records, some errors could be unexpeced raised due to the trick a temporarily changing the discount value of the field while computing others fields.

In the future we should use dedicated fields for the 3 discount fields and make the native discount field a computed one with readonly=False. This approach would simplify a lot the code. See https://github.com/OCA/account-invoicing/pull/1638

### DIFF
--- a/account_invoice_triple_discount/README.rst
+++ b/account_invoice_triple_discount/README.rst
@@ -84,6 +84,7 @@ Contributors
 * `Aion Tech <https://aiontech.company/>`_:
 
   * Simone Rubino <simone.rubino@aion-tech.it>
+* Laurent Mignon <laurent.mignon@acsone.eu>
 
 Maintainers
 ~~~~~~~~~~~
--- a/account_invoice_triple_discount/models/account_move_line.py
+++ b/account_invoice_triple_discount/models/account_move_line.py
@@ -3,6 +3,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 import functools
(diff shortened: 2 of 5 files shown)


lmignon:
cc @sbejaoui @rousseldenis

rousseldenis (approved):
Code review.

Seems fair to use protecting() in this case.

rousseldenis:
@grindtildeath

lmignon:
@grindtildeath Can you take a look at this one?

grindtildeath (approved):
That looks good, but do we want to merge this or move forward with #1638 ?

lmignon:
@grindtildeath This one is used in production for months. IT would be nice to have it merged even if we move forward with #1638 It's an important fix for those using the current version.

ramiadavid:
@lmignon 
Tested and works correctly. The performance on invoices with many lines the performance improvement is amazing.

PR #1671, opened 2024-02-29
[14.0][MIG] account_portal_hide_invoice

mtelahun:
Migrate account_portal_hide_invoice to 14.0


hbrunn (approved):
well, you could just remove the whole [t-call](https://github.com/OCA/OCB/blob/14.0/addons/account/views/account_portal_templates.xml#L17-L21), but I actually like what you do better. Let's see what other reviewers think

PR #1672, opened 2024-02-29
[14.0][FIX] stock_picking_invoicing qty_invoiced value compute

christian-ramos-tecnativa:
This PR fix the invoice qty compute on the sale when the invoice is created from the picking.

### DIFF
--- a/stock_picking_invoicing/models/__init__.py
+++ b/stock_picking_invoicing/models/__init__.py
@@ -6,3 +6,4 @@ from . import account_move
 from . import stock_move
 from . import stock_picking
 from . import stock_picking_type
+from . import sale
--- a/stock_picking_invoicing/models/sale.py
+++ b/stock_picking_invoicing/models/sale.py
@@ -1,0 +1,57 @@
+# Copyright (C) 2019-Today: Odoo Community Association
+# @ 2019-Today: Binhex - www.binhex.cloud -
+#   Christian-RB <c.ramos@binhex.es>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import api, models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    @api.depends("order_line.invoice_status")
+    def _get_invoiced(self):
+        """
+        Override to get the invoices related to the stock moves of the sale order
+        """
+        super()._get_invoiced()
+        for order in self:
+            linked_invoice_ids = order.invoice_ids
+            new_invoice_ids = (
+                order.order_line.move_ids.invoice_line_ids.move_id.filtered(
+                    lambda r: r.move_type in ("out_invoice", "out_refund")
+                )
+            )
+            linked_invoice_ids |= new_invoice_ids
+            order.invoice_ids = [(6, 0, list(set(linked_invoice_ids.ids)))]
+            order.invoice_count = len(linked_invoice_ids)
+
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    @api.depends(
+        "move_ids.invoice_line_ids.move_id.state",
+        "move_ids.invoice_line_ids.quantity",
+        "invoice_lines.move_id.state",
+        "invoice_lines.quantity",
+        "invoice_status",
+    )
+    def _get_invoice_qty(self):
+        """
+        Override to get the quantity invoiced related to the stock moves of the sale order
+        """
+        res = super()._get_invoice_qty()
+        for line in self:
+            qty_invoiced = line.qty_invoiced
+            for invoice_line in line.move_ids.invoice_line_ids - line.invoice_lines:
+                if invoice_line.move_id.state != "cancel":
+                    if invoice_line.move_id.move_type == "out_invoice":
+                        qty_invoiced += invoice_line.product_uom_id._compute_quantity(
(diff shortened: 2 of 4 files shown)


mbcosta:
@Christian-RB there are a problem about your PR but is not related direct about your code, the problem are the dependecy of Sale module, this was included indirect by the module stock_picking_invoice_link https://github.com/OCA/stock-logistics-workflow/blob/14.0/stock_picking_invoice_link/__manifest__.py#L20 , the main reason are modularization and isolation of code, because the main goal of this module are "create invoices directly from picking, without having to use Sale or Purchase Orders." https://github.com/OCA/account-invoicing/blob/14.0/stock_picking_invoicing/readme/DESCRIPTION.rst,  (I thinking about propusing to remove it to avoid the direct dependency) to integrate this module with Sale or Purchase modules the better approuch should be create new modules sale_stock_picking_invoicing and purchase_stock_picking_invoicing.

There are an example of how it can done in Brazilian Localization, where we have the case of the Invoice creation from Picking without relation with a Sale or Purchase Order ( by the Fiscal rules, even when the company move a product to different address, is not a Sale or Purchase case, the company has the Fiscal obligation to generate an Invoice/Fiscal Document ), but we also are using this module when we want to force the creation of the Invoice from the Picking that was created by a Sale or Purchase Order, you can see here https://github.com/OCA/l10n-brazil/tree/14.0/l10n_br_sale_stock and https://github.com/OCA/l10n-brazil/tree/14.0/l10n_br_purchase_stock .

I have made a PR in this repo based in l10n_br_sale_stock module for v14 https://github.com/OCA/account-invoicing/pull/1025 at that time I was think it just used by Brazilian Localization, so as I wrote in that PR we plan to propuse add this module in v16 and check if there are intest, but if other people has necessity to use this module, we can think and consider re-open the PR. But it's important to hear from you, and other people, about the use of the  module and if we should to add it in v14 or waititng for v16, what do you think?

cc @rvalyi @renatonlima

christian-ramos-tecnativa:
Hello @mbcosta. Thank you for taking the trouble to review the PR. I understand your point of view regarding the unnecessary dependency on the 'sale' module and that a separate module should be created to add this functionality. Despite this, I believe that separating the functionalities of an existing module (which surely has people using it) will lead to more problems than solutions. I think the most appropriate approach would be to separate them at the time of module migration, so that openupgrade can handle the necessary adjustments. Let me know what you think.

mbcosta:
@Christian-RB  thank you for reply, I believe will be better to solve it with the module sale_stock_picking_invoicing, because Brazilian Localization use it for a while so the isolation, modularization and operation seems works as necessary, the only doubt in the PR was if people beyond Brazilian Localization has interest in the module and as you said "which surely has people using it" I'm reopen the PR and appreciate if you can review.

There are a difference from your PR, it's was unnecessary changed the methods _get_invoiced and _get_invoice_qty because during the creation of the Invoice from Picking the sale_line_ids are informed https://github.com/akretion/account-invoicing/blob/14.0-NEW-sale_stock_picking_invoicing/sale_stock_picking_invoicing/wizards/stock_invoice_onshipping.py#L87

The module was made as reference Brazilian Localization in the implementation, so will be important the review of other people and cases to check if something need to be change to have a better compatible, if you find something please let me know.

I'm looking to solve the problem of the tests in the PR.

mbcosta:
PR of sale_stock_picking_invoicing https://github.com/OCA/account-invoicing/pull/1025 is Ready for Review

rvalyi:
Please @OCA/accounting-maintainers , don't merge this too quickly and read @mbcosta remarks and mine in https://github.com/OCA/account-invoicing/pull/1025 . Indeed it would be much better if stock_picking_invoicing wouldn't carry a hard dependency to the sale module as @mbcosta explained and this PR would just make it harder to make it modular later...
cc @pedrobaeza @marcelsavegnago @antoniospneto

rvalyi:
@mbcosta I believe something great could be migrate stock_picking_invoice_link to v17 making it fully independent from sale and create a new `sale_stock_picking_invoice_link` in the same mood as `purchase_stock_picking_invoice_link` this would fix the problem for the future version (at the end of next year we will be implementing v18). As for the present versions, well we have to find the best work around and may be the best work around is not merging this PR and merging #1025 instead, I should still look better myself...

mbcosta:
Hi @Christian-RB , take some time to merged the **sale_stock_picking_invocing** but since 25 of Octuber the module was integrated in the **14.0** branch, the time was necessary and important to make the implementation robust enough to get the Invoice created by **Stock Picking** being very similar to the created by **Sale Order** and with this make unnecessary "glue modules" for cases like [account_payment_sale](https://github.com/OCA/bank-payment/blob/14.0/account_payment_sale/models/sale_order.py#L41) and [sale_commission](https://github.com/OCA/commission/blob/14.0/sale_commission/models/sale_order.py#L64), so please consider check if the module can help you and solve the case of this PR and also, if you can, consider helping in the Review of the migration to v15 https://github.com/OCA/account-invoicing/pull/1826

@rvalyi thank you for take time in this problem, it will be important to solve the indirect depency of **stock_picking_invocing** from **sale** with a new **sale_stock_picking_invoice_link**, I belive after the answer of @Christian-RB you can consider Close the PR

christian-ramos-tecnativa:
Thanks @mbcosta for the heads up! I will check the work done and close this if proceed. In the meantime I will turn it to draft.

PR #1674, opened 2024-03-04
[16.0][FIX] account_invoice_fixed_discount: Monetary to Float

hildickethan:
This field was Float pre-16, it was changed to Monetary [there](https://github.com/OCA/account-invoicing/commit/3384403d138904130496cbbbdfd042dbe5061285) I assume as a natural part of the refactorization

https://github.com/OCA/sale-workflow/blob/627b3bee6fa0cb224c1e073df068019a90e96dce/sale_fixed_discount/models/sale_order_line.py#L12-L16
In `sale_fixed_discount` the field is `Float` and has configurable decimal precision through `Product Price`, but when creating an invoice from this sale it will get cut down to the Monetary precision causing inconsistencies.
Since this field goes hand in hand with `price_unit`, it makes more sense to keep it at the same precision as that field.

Sale:
image

Invoice before changes:
image

### DIFF
--- a/account_invoice_fixed_discount/models/account_move_line.py
+++ b/account_invoice_fixed_discount/models/account_move_line.py
@@ -8,10 +8,10 @@ from odoo.tools.float_utils import float_is_zero
 class AccountMoveLine(models.Model):
     _inherit = "account.move.line"
 
-    discount_fixed = fields.Monetary(
+    discount_fixed = fields.Float(
         string="Discount (Fixed)",
         default=0.0,
-        currency_field="currency_id",
+        digits="Product Price",
         help=(
             "Apply a fixed amount discount to this line. The amount is multiplied by "
             "the quantity of the product."


hildickethan:
ping @PieterPaulussen since the migration was yours, in case you have any insight

PieterPaulussen:
Hi @hildickethan-S73, what about changing the sale order line field to a Monetary one instead of the other way around?
I don't believe the field goes hand in hand with price unit, but rather with the subtotals?
The PR should not be created here, but in https://github.com/OCA/sale-workflow IMHO

PieterPaulussen (changes requested):
PR should be created the other way around: https://github.com/OCA/sale-workflow

hildickethan:
> Hi @hildickethan-S73, what about changing the sale order line field to a Monetary one instead of the other way around? I don't believe the field goes hand in hand with price unit, but rather with the subtotals?

The help text reads `Apply a fixed amount discount to this line. The amount is multiplied by the quantity of the product.` and that is how it works in practice
image
When using this field the real `price_unit` is directly `price_unit - fixed_discount`, rather than subtracting itself from the subtotal. That is why I would say they should share decimal precision.

My end use case is having some cases where the unit prices and fixed discounts look like the PR's screenshots with 3 decimals (this is ironically due to being given a subtotal-based fixed discount by an API that I have to divide between the unit price). I think changing it to be subtotal-based instead of unit price might be the best long-term, but I don't think it's a good idea to refactor that heavily now that it's already merged

CRogos (approved):
My first thought was that price should be linked to a currency and therefore be monetary. But indeed the price field in Odoo is Float in nearly all cases and because the fields.Monetary field is always rounded by the precision of the linked currency, this won't work for price or average_price information.

I also agree that discount_fixed is more linked to unit_price then to sub_total, because it is multiplied by quantity.

To make it consistent with the Odoo base, I would also change it to float, even though I still think every price should be linked with a currency.

PR #1558, opened 2023-09-20
[15.0][FIX] account_invoice_fixed_discount: Fix Division by zero  error

angelmoya:

PR #1680, opened 2024-03-11
[16.0][FIX] account_invoice_fixed_discount: Division by zero

ioans73:


Jortolsa-S73 (approved):

ioans73:
@tien-ld @rvalyi could you review this? It's the same as https://github.com/OCA/sale-workflow/pull/3021

tien-ld (approved):

rvalyi (approved):

account_invoice_fixed_discount/models/account_move_line.py
```diff
@@ -82,4 +82,8 @@ def _get_discount_from_fixed_discount(self):
         currency = self.currency_id or self.company_id.currency_id
         if float_is_zero(self.discount_fixed, precision_rounding=currency.rounding):
```
sergiocorato, line 83: Why not add here the condition `or not self.discount`?

ioans73: You're right, it could also be done that way. I have tried to do it the same as in this PR https://github.com/OCA/account-invoicing/pull/1558

ioans73:
@rvalyi Can it be merged?

ioans73:
@pedrobaeza since it's approved, could you merge? Thanks

PR #1688, opened 2024-03-18
[14.0][IMP] manual invoicing mode

Kev-Roche:
Actually, the invoicing mode only work with weekly, monthly automatic invoicing. But if you select manually some sale orders in a list,the action create invoice doesn't take into account the  "1 invoice by sale" parameter.
Add the invoicing mode when select the sale orders manually.

### DIFF
--- a/account_invoice_base_invoicing_mode/models/sale_order.py
+++ b/account_invoice_base_invoicing_mode/models/sale_order.py
@@ -62,3 +62,14 @@ class SaleOrder(models.Model):
             self.with_delay()._generate_invoices_by_partner(saleorder_ids)
         companies.write({last_execution_field_name: datetime.now()})
         return saleorder_groups
+
+    def _create_invoices(self, grouped=False, final=False, date=None):
+        moves = self.env["account.move"]
+        for partner_invoice in self.partner_invoice_id:
+            sales = self.filtered(lambda r: r.partner_invoice_id == partner_invoice)
+            moves += super(SaleOrder, sales)._create_invoices(
+                grouped=partner_invoice.one_invoice_per_order,
+                final=final,
+                date=date,
+            )
+        return moves


account_invoice_base_invoicing_mode/models/sale_order.py
```diff
@@ -62,3 +62,14 @@ def generate_invoices_by_invoice_mode(
             self.with_delay()._generate_invoices_by_partner(saleorder_ids)
         companies.write({last_execution_field_name: datetime.now()})
         return saleorder_groups
+
+    def _create_invoices(self, grouped=False, final=False, date=None):
+        moves = self.env["account.move"]
+        for partner_invoice in self.partner_invoice_id:
```
hbrunn, line 68: this breaks if you work on multiple sale orders

hbrunn (changes requested):
it's very unclear to me what your goal here is, so please elaborate to allow a sensible review

account_invoice_base_invoicing_mode/models/sale_order.py
Kev-Roche: I tried with multiple sale orders and multiple customers, i had no problem, what is the breaking case ?

Kev-Roche:
> it's very unclear to me what your goal here is, so please elaborate to allow a sensible review

Sorry, I update the PR presentation.

account_invoice_base_invoicing_mode/models/sale_order.py
hbrunn: any code of the form (saleorder1 + saleorder2)._create_invoices() will break here. Core supports this, so should you

Kev-Roche: In the test of this module, I tried `(self.so1 + self.so2)._create_invoices()`without any error.

hbrunn: oh apologies, right, I was confused with scalar fields. So what you do here is group sale orders by partner_invoice_id, and pass this partner's choice of grouping. Must have been tired yesterday.

This will not apply to deposits, but then I also don't see so much point in grouping deposits.

hbrunn (approved):

sebastienbeau (approved):
LGTM (code review and test)

PR #1690, opened 2024-03-23
[14.0][FIX] account_receipt_sale: invoice word occurrences

primes2h:
If using receipts, all texts containing "invoice" (buttons, wizard, etc.) will be changed accordingly.

sale_order_receipt


MaurizioPellegrinet (approved):
LGTM

FrancescoBallerini (approved):
Functional: LGTM

account_receipt_sale/wizard/sale_make_invoice.py
```diff
@@ -1,11 +1,30 @@
-from odoo import models
+from odoo import fields, models
 
 
-class SaleAdvancePaymentInv(models.TransientModel):
+class SaleAdvancePaymentRec(models.TransientModel):
+    _name = "sale.advance.payment.rec"
```
FrancescoBallerini, line 5: Hi Sergio, is it necessary to re-define a new name for the model? If that's the case could you also add `_description` attribute? Otherwise it will give annoying warnings in the log during the module update

FrancescoBallerini (changes requested):

account_receipt_sale/wizard/sale_make_invoice.py
primes2h: > Hi Sergio, is it necessary to re-define a new name for the model? If that's the case could you also add `_description` attribute? Otherwise it will give annoying warnings in the log during the module update

I'd say yes, `_description` added [here](https://github.com/OCA/account-invoicing/commit/7d62986ddb7af13927ba251ef90ed0319d65219c#diff-b8cc7465a8d18dac20baa7aa762f5df2c62b39e53e49f7dd65201071dc7af4e5R6), thanks!

FrancescoBallerini (approved):

account_receipt_sale/wizard/sale_make_invoice.py
FrancescoBallerini: Perfect, thanks!

eLBati:
Hi @primes2h 
did you consider overriding [fields_view_get](https://github.com/odoo/odoo/blob/f2765d2cab5671a010404c36842bf1b4c4d6350b/odoo/models.py#L1602) of the involved models to replace strings where needed?
Thanks

PR #1694, opened 2024-03-28
[16.0][MIG] account_invoice_validation_queued

Saran440:
Standard MIgrated and Update readme


hbrunn (approved):
/ocabot migration account_invoice_validation_queued

francesco-ooops:
@Saran440 (and @pedrobaeza as module author): do you know if there is a specific reason why it's not possible to validate 2 invoices with a different value in field "date"? 

If this is done to preserve invoicing sequence from job queue completion order, seems to me the check could be done on invoice_date field. Is there something I'm missing?

pedrobaeza:
I can't say, as I'm not using this module in newer versions (the speed improvements done on core have been enough for our customers).

Saran440:
@francesco-ooops That's a good question. I'm not sure myself. I migrated from standard.
However, In my project I inherit this module for delete code validation too.

Should we delete it in this module because job queue processes tasks one by one?
@pedrobaeza

francesco-ooops:
@Saran440 I think the only thing we must make sure is not to validate together invoices with different invoice date as to preserve sequence

currently this check is done on field "date" which is not a visible field in the form and AFAIU is = creation date (not sure what is the purpose of this field)

adriapalleja:
@Saran440 @hbrunn How is the migration of this module going? Could we help somehow in this migration? Thanks!

Saran440:
@adriapalleja Could you please review and approve this PR if it works? 
We can merge once 2 reviewers approve.

adriapalleja:
@Saran440 I've just reviewed and tested the module. Everything is working fine on my side, except from this topic in "Usage": 
- If any of the invoices is already enqueued, there will be a message saying so and avoiding to perform the process

The process is working as expected, but I would remove "there will be a message saying so" from the Usage Readme.

adriapalleja (changes requested):
Everything looking good! Only a minor change requested in the "Usage" Readme.

adriapalleja:
Hi @Saran440 ! Could you apply the minor changes suggested in order to move forward with this PR? Thanks!

francesco-ooops:
For the record, we edited this module internally by implementing the error message "You can't enqueue invoices with different dates." (seems to appear only in translation files at the moment) and forcing the user to validate invoices grouped by date (from older to newer in order to preserve sequence)

If date is not set in invoice, then "today" is used

Saran440:
@francesco-ooops I tested and I think this version can use invoice with multiple dates. if you find any bugs, please explain and provide the steps to reproduce the error, and we will fix it.

For this module, it runs 1 queue per document, so the sequence will function correctly.
I have removed the validation error in this module because I couldn't find a reason for check.

@adriapalleja I updated readme, please recheck.

francesco-ooops:
@Saran440 if you're positive that enqueuing invoices with different dates results in invoice numbers following sequence correctly, then it's good for me 👍

adriapalleja (approved):

sergiobstoj:
merge?

Saran440:
@hbrunn Hi, Can you help merge, please?

PR #1628, opened 2023-12-18
[IMP] partner_invoicing_mode_at_shipping: Add grouping option per picking

rousseldenis:
Depends on :

- #1515

PR #1695, opened 2024-03-28
[16.0][IMP] partner_invoicing_mode_at_shipping: Extract _invoicing_at_shipping_validation function and add option auto validate invoice on partner

tuantrantg:
Depends on:
* https://github.com/OCA/account-invoicing/pull/1628


partner_invoicing_mode_at_shipping/models/res_partner.py
```diff
@@ -11,3 +12,44 @@ class ResPartner(models.Model):
...
+
+    one_invoice_per_shipping = fields.Boolean(
+        index=True,
+        help="Check this if you want to create one invoice per shipping using the"
+        " partner invoicing mode that should be different than 'At Shipping'.",
+    )
+
+    auto_validate_invoice = fields.Boolean(
+        help="Check this if you want to validate the invoice automatically"
+    )
```
jbaudoux: Do not add this. There is already that option
* invoicing mode at shipping: generate & validate invoice
* one_invoice_per_shipping: generate invoice & choose another option in invoicing mode if you want auto validation

jbaudoux (changes requested):

jbaudoux:
This should be a PR on https://github.com/OCA/account-invoicing/pull/1628 unless @rousseldenis you rebase and merge your PR, then you can rebase this one...

partner_invoicing_mode_at_shipping/models/res_partner.py
tuantrantg: > invoicing mode at shipping: generate & validate invoice
> one_invoice_per_shipping: generate invoice & choose another option in invoicing mode if you want auto validation

Hi @jbaudoux could you please explain this one, I don't get this point yet

And it's nice to have a summary of this issue, because:
* in here, the validation of invoice bases on the config in `res.partner` [1]
* in https://github.com/OCA/stock-logistics-transport/pull/118, it bases on COD payment term config [2]

It makes me confused:
* In [2], after we get invoices need to validate, should we return them (to validate the invoices) or pass them to the call of `super` to check the condition [1] again?

jbaudoux: The method `_invoicing_at_shipping_validation` https://github.com/OCA/account-invoicing/pull/1695/files#diff-9805ceb19bf3584cef2fb4a911f57df11606716d8af657ebe54680627f90f2dfR25 currently implements [1].
We need to add there a condition in case of COD [2] to validated only if the COD payment term is configure so. So if COD, we rely only on COD config, otherwise we call super. This is because we use this module from COD to make the invoices. So we still need to rely only on COD config when called from there.

tuantrantg:
Hi @jbaudoux https://github.com/acsone/account-invoicing/pull/4 is created to follow this PR. I will close this one

PR #1697, opened 2024-03-28
[16.0] account_invoice_supplierinfo_update: fix updating the right supplierinfo

sebastienbeau:
If you have variant of price per qty the module is not going to update the right supplierinfo

Reuse the native _sellect_seller to pick the right supplierinfo + add test

@legalsylvain

### DIFF
--- a/account_invoice_supplierinfo_update/models/account_move_line.py
+++ b/account_invoice_supplierinfo_update/models/account_move_line.py
@@ -12,8 +12,9 @@ class AccountMoveLine(models.Model):
         """Given an invoice line, return the supplierinfo that matches
         with product and supplier, if exist"""
         self.ensure_one()
-        supplierinfos = self.product_id.seller_ids.filtered(
-            lambda seller: seller.partner_id == self.move_id.supplier_partner_id
+        supplierinfos = self.product_id._select_seller(
+            partner_id=self.move_id.supplier_partner_id,
+            quantity=self.quantity,
         )
         return supplierinfos and supplierinfos[0] or False
 
--- a/account_invoice_supplierinfo_update/tests/test_module.py
+++ b/account_invoice_supplierinfo_update/tests/test_module.py
@@ -42,6 +42,106 @@ class TestModule(AccountTestInvoicingCommon):
         cls.WizardUpdateSupplierinfo = cls.env["wizard.update.invoice.supplierinfo"]
         cls.ProductSupplierinfo = cls.env["product.supplierinfo"]
 
+    def test_get_the_right_variant_supplierinfo(self):
+        # Variant the product A and set a price on variation 1
+        tmpl_a = self.product_a.product_tmpl_id
+        tmpl_a.write(
+            {
+                "attribute_line_ids": [
+                    (
+                        0,
+                        0,
+                        {
+                            "attribute_id": self.env.ref(
+                                "product.product_attribute_2"
+                            ).id,
+                            "value_ids": [
+                                (
+                                    6,
+                                    0,
+                                    [
+                                        self.env.ref(
+                                            "product.product_attribute_value_3"
+                                        ).id,
+                                        self.env.ref(
+                                            "product.product_attribute_value_4"
+                                        ).id,
+                                    ],
+                                )
+                            ],
+                        },
+                    )
+                ]
+            }
+        )
+        product_a_1, product_a_2 = tmpl_a.product_variant_ids
+
+        supplier_product_a_1 = self.env["product.supplierinfo"].create(
+            [
+                {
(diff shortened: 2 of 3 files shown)


account_invoice_supplierinfo_update/tests/test_module.py
```diff
@@ -42,6 +42,106 @@ def setUpClass(cls, chart_template_ref=None):
...
 
+    def test_get_the_right_variant_supplierinfo(self):
+        # Variant the product A and set a price on variation 1
+        tmpl_a = self.product_a.product_tmpl_id
+        tmpl_a.write(
+            {
+                "attribute_line_ids": [
+                    (
+                        0,
+                        0,
```
legalsylvain, line 53: nitpicking, could be replaced by Command.create

account_invoice_supplierinfo_update/tests/test_module.py
```diff
@@ -42,6 +42,106 @@ def setUpClass(cls, chart_template_ref=None):
...
+                        {
+                            "attribute_id": self.env.ref(
+                                "product.product_attribute_2"
+                            ).id,
+                            "value_ids": [
+                                (
+                                    6,
+                                    0,
```
legalsylvain, line 61: same here ( Command.set([...]))

legalsylvain (approved):
Nice catch ! 
Thanks !
(I personnaly don't use many supplier info for the same product and partner, so I never faced such issue).

account_invoice_supplierinfo_update/models/account_move_line.py
```diff
@@ -12,8 +12,9 @@ def _get_supplierinfo(self):
         """Given an invoice line, return the supplierinfo that matches
         with product and supplier, if exist"""
         self.ensure_one()
-        supplierinfos = self.product_id.seller_ids.filtered(
-            lambda seller: seller.partner_id == self.move_id.supplier_partner_id
+        supplierinfos = self.product_id._select_seller(
+            partner_id=self.move_id.supplier_partner_id,
+            quantity=self.quantity,
```
rvalyi, line 17: "Information is a mass noun. It has no plural form"* so infos looks like informations which would be Frenglish and looks TinyERP. Better change that to supplier_info or supplier_data as you want.

* https://preply.com/en/question/plural-of-information
#onestpasamiserieux

rvalyi (changes requested):

account_invoice_supplierinfo_update/models/account_move_line.py
legalsylvain: Hi. 

@rvalyi. We already talk about that in another PR. summary of the previous comments : 

what you mention is not an OCA rule. 
- "infos" is used in many OCA repo. (12 occurences in V16, only in the repos I cloned for GRAP project)
- "datas" is used in many OCA repo. (19 occurences in V16, only in the repos I cloned for GRAP project)

So if you want to introduce that new rule, please create a dedicated PR here https://github.com/OCA/odoo-community.org/tree/master/website/Contribution

In the meantime, please avoid to block valid PR.

Thanks !

rvalyi: I didn't block, only suggest to use better variable names. When native English speakers read that they feel it's cheap while it's easy to improve.

legalsylvain: 1)
> I didn't block

You set "changes requested". Could you remove it ?

2) as said not the place to create new rules. Please create a dedicated issue. 

Thanks !

legalsylvain:
@sebastienbeau : could you take into account the proposal of use new "command" syntax ?
@rvalyi : could you unblock the PR ? 

thanks a lot !

account_invoice_supplierinfo_update/views/account_invoice_view.xml
```diff
@@ -4,7 +4,7 @@
         <field name="model">account.move</field>
         <field name="inherit_id" ref="account.view_move_form" />
         <field name="arch" type="xml">
-            <xpath expr="//button[@name='action_post']" position="after">
+            <xpath expr="//button[@name='button_draft']" position="after">
```
legalsylvain, line 7: why this change ? (not sure to understand).

legalsylvain:
Ping @sebastienbeau !

account_invoice_supplierinfo_update/models/account_move_line.py
legalsylvain: Hi. @rvalyi : blocking OCA PR because it doesn't respect your personal tastes is quite annoying. Please do not waste contributors' time in the future.

<hr />

image

<hr />

image

<hr />
image

<hr />
image

<hr />

Issue #1698, opened 2024-04-04
Cannot confirm invoice/vendor bill with account_move_tier_validation


MarcoDuesuenceliSAMSA-IT:
## Module
account_move_tier_validation version 16

## Describe the bug
Although all validations have been completed and are even listed in the history, the message "The operation is under validation" is always displayed.

## To Reproduce
**Affected versions**: V16

1. Create any definition that causes a request for validation
2. Start and finish all validations so that you can in histoy that are all finished
3. Click on "confirm" in a vendor bill e.g.

**Expected behavior**
I would expect that the vendor bill or invoice get confirmend and posted.

**Additional context**
Add any other context about the problem here. (e.g. OS, Python version, ...)

bosd:
@MarcoDuesuenceliSAMSA-IT  This seems fixed in 16 Can this one be closed?

PR #1703, opened 2024-04-16
[16.0][FIX] account_move_tier_validation: Pass context to allow confirming moves with tier validation

ByteMeAsap:
Fixes https://github.com/OCA/account-invoicing/issues/1698

### DIFF
--- a/account_move_tier_validation/models/account_move.py
+++ b/account_move_tier_validation/models/account_move.py
@@ -33,3 +33,8 @@ class AccountMove(models.Model):
         elif self.move_type == "out_refund":
             name = _("Credit Note")
         return name
+
+    def action_post(self):
+        return super(
+            AccountMove, self.with_context(skip_validation_check=True)
+        ).action_post()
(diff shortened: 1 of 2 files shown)


StefanRijnhart (changes requested):
Thanks for the fix. It's not entirely clear to me what causes the issue. Would it be possible to add a reproducing test?

ByteMeAsap:
> Thanks for the fix. It's not entirely clear to me what causes the issue. Would it be possible to add a reproducing test?

@StefanRijnhart , if you check the issue https://github.com/OCA/account-invoicing/issues/1698 , it details the steps that causes the issue. Is it possible for you to review it?

StefanRijnhart:
@ByteMeAsap If it's that easy, then it should definitely be included as a test in the module.

ByteMeAsap:
> @ByteMeAsap If it's that easy, then it should definitely be included as a test in the module.

@StefanRijnhart , I have added the test

ByteMeAsap:
> @ByteMeAsap If it's that easy, then it should definitely be included as a test in the module.

@StefanRijnhart , I have added the test scripts

StefanRijnhart (approved):
Excellent, thanks for adding the test!

gurneyalex (approved):

ByteMeAsap:
Can we have this merged?

PR #1712, opened 2024-04-23
[15.0][IMP-FIX] account_invoice_transmit_method: Update domain of the second button "Send and print"

pilarvargas-tecnativa:
The "Send and print" button should be displayed when the transmission method is defined as e-mail or False. In case an invoice modifies its transmit method to make a first sending and is changed again to another transmission method than fake or e-mail the second button will be visible. For this reason the attributes of the second button that is displayed when an invoice has already been sent must also be modified.

cc @Tecnativa TT48851

@pedrobaeza @carolinafernandez-tecnativa please review


pedrobaeza (approved):

carolinafernandez-tecnativa (approved):
LGTM Thanks :)

PR #1725, opened 2024-05-20
[16.0][IMP] account_global_discount: exclude products from global disc

ferran-S73:
Depends on https://github.com/OCA/server-backend/pull/221

### DIFF
--- a/account_global_discount/models/account_move.py
+++ b/account_global_discount/models/account_move.py
@@ -89,6 +89,8 @@ class AccountMove(models.Model):
         for inv_line in _self.invoice_line_ids.filtered(
             lambda l: l.display_type not in ["line_section", "line_note"]
         ):
+            if inv_line.product_id.bypass_global_discount:
+                continue
             taxes_keys.setdefault(tuple(inv_line.tax_ids.ids), 0)
             taxes_keys[tuple(inv_line.tax_ids.ids)] += inv_line.price_subtotal
         # Reset previous global discounts
@@ -119,6 +121,8 @@ class AccountMove(models.Model):
         base_total = 0
         zero_taxes = self.env["account.tax"]
         for line in self.line_ids.filtered("tax_ids"):
+            if line.product_id.bypass_global_discount:
+                continue
             key = tuple(line.tax_ids.ids)
             if taxes_keys.get(key):
                 base_total += line.price_subtotal


ferran-S73:
Hi @pedrobaeza, I've added the m2m field in the global discounts as suggested, however after a bit of testing I find it a bit confusing to use now.

Normally this module adds a table on the invoice in which the base before discount field gets decreased so it's easy to see which discount has been applied.
After my modification I've tweaked it so that this field shows the actual amount it's used to calculate the discount and so I've found some instances where this amount increases from one line to another leading to some confusion.

I'll show one example, in my invoice I have 2 global discounts (disc1 and disc2) and 2 product lines (product1 and product2) with the same tax. I put product2 on the excluded products for disc1 so my invoice line table and global discount table look like this:
imagen
imagen
As you can see the base for disc2 is greater than the base from disc1 even though it is applied after disc1 because for the first discount 1 product line is ignored.


I think this is over complicating the module a bit but technically is doable as you can see, so I'd like your opinion on whether this is functionally correct. Thanks in advance

pedrobaeza:
OK, let's keep the flag by product to avoid that situations.

QuocDuong1306:
Hello @ferran-S73 , maybe you need to update [test-requirements.txt](https://github.com/OCA/maintainer-tools/wiki/Use-temporary-reference%28s%29-to-another-pull-request%28s%29) to check the tests

cyrilmanuel (approved):

mariancuadranetkia:
Hello @ferran-S73 ,
I'm testing it in runboat, but it's returning an error about **bypass_global_discount** field (it not exists) when I try to create invoice with global discount... 
I've checked and module _base_global_discount_ is installed.. so, I'm not sure about the reason, it's possible that you need update test-requirements?

imagen


Thanks!

ferran-S73:
> Hello @ferran-S73 , I'm testing it in runboat, but it's returning an error about **bypass_global_discount** field (it not exists) when I try to create invoice with global discount... I've checked and module _base_global_discount_ is installed.. so, I'm not sure about the reason, it's possible that you need update test-requirements?
> 
> imagen
> 
> Thanks!

@mariancuadranetkia the PR that this one depends on is already merged so maybe just the runboat needs to be rebuild or something. I don't think I need to add test-requirements for a merged pr. Maybe a maintainer can help in here I don't really know. If there is something I can do please let me know

pedrobaeza:
Try to rebase + forced push to force runboat rebuild.

ferran-S73:
@mariancuadranetkia runboat should be ready now. @pedrobaeza thanks for the tip!

mariancuadranetkia:
@ferran-S73 perfect! Now I could test it and runs ok!

mariancuadranetkia (approved):
LGTM

PR #1740, opened 2024-06-13
Replace author 'Camptocamp SA' with 'Camptocamp'

vvrossem:
This pull request contains changes to replace 'Camptocamp SA' with 'Camptocamp' in __manifest__.py files for branch 14.0.


etobella (changes requested):
Use the proper commit message

PR #1748, opened 2024-06-26
[14.0][IMP]sale_line_refund_to_invoice_qty: credit note default

PicchiSeba:
add setting for default value in credit note wizard

### DIFF
--- a/sale_line_refund_to_invoice_qty/wizards/account_move_reversal.py
+++ b/sale_line_refund_to_invoice_qty/wizards/account_move_reversal.py
@@ -1,6 +1,6 @@
 # Copyright 2021 ForgeFlow (http://www.forgeflow.com)
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
-from odoo import fields, models
+from odoo import api, fields, models
 
 
 class AccountMoveReversal(models.TransientModel):
@@ -8,10 +8,20 @@ class AccountMoveReversal(models.TransientModel):
 
     sale_qty_to_reinvoice = fields.Boolean(
         string="This credit note will be reinvoiced",
-        default="True",
         help="Leave it marked if you will reinvoice the same sale order line "
         "(standard behaviour)",
     )
+
+    @api.model
+    def default_get(self, fields_list):
+        res = super().default_get(fields_list)
+        company = (
+            self.env["res.company"].browse(res.get("company_id"))
+            if res.get("company_id")
+            else self.env.company
+        )
+        res["sale_qty_to_reinvoice"] = company.reinvoice_credit_note_default
+        return res
 
     def reverse_moves(self):
         return super(
--- a/sale_line_refund_to_invoice_qty/README.rst
+++ b/sale_line_refund_to_invoice_qty/README.rst
@@ -49,6 +49,9 @@ Furthermore, after creating the credit note, you can mark the field "Sale qty
 not to reinvoice" to decide the criteria to be used for each specific invoice
 line.
 
+Enabling the "Reinvoice credit note by default" flag in the Invoice settings, will
+enable the "This credit note will be reinvoiced" flag on Credit Notes by default.
+
 Bug Tracker
 ===========
 
--- a/sale_line_refund_to_invoice_qty/__manifest__.py
+++ b/sale_line_refund_to_invoice_qty/__manifest__.py
@@ -14,6 +14,7 @@
     "depends": ["sale_management"],
     "data": [
         "views/account_move_views.xml",
+        "views/res_config_settings_views.xml",
         "views/sale_order_views.xml",
         "wizards/account_move_reversal_view.xml",
     ],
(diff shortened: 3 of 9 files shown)


francesco-ooops (approved):
Functional ok!

Enabling the "Do not reinvoice credit note by default" flag in the Invoice settings, will
disable the "This credit note will be reinvoiced" flag on Credit Notes by default

ilyasProgrammer (approved):
LGTM

francesco-ooops:
@JordiMForgeFlow @LoisRForgeFlow is this addition ok for you? thanks!

JordiMForgeFlow:
@francesco-ooops I think it is much more understandable to have boolean variables that reflect "enabling" something when they are set to True, as it is more intuitive. I would recommend instead of having `no_default_reinvoice_credit_note` to have a `default_reinvoice_credit_note` that is set to True by default.

francesco-ooops:
@PicchiSeba can you attend comment?

PicchiSeba:
@JordiMForgeFlow I updated this PR with your suggestion

sale_line_refund_to_invoice_qty/wizards/account_move_reversal.py
```diff
@@ -8,7 +8,7 @@ class AccountMoveReversal(models.TransientModel):
 
     sale_qty_to_reinvoice = fields.Boolean(
         string="This credit note will be reinvoiced",
-        default="True",
+        default=lambda self: self.env.company.reinvoice_credit_note_default,
```
JordiMForgeFlow: the account.move.reversal wizard already has a company field. What you could do is to put the default of the sale_qty_to_reinvoice in the `def default_get(self, fields)` method of the wizard, where the company is filled. Check: https://github.com/odoo/odoo/blob/14.0/addons/account/wizard/account_move_reversal.py#L39

PicchiSeba: @JordiMForgeFlow I updated with your suggestion. Let me know what you think about my implementation.

sale_line_refund_to_invoice_qty/wizards/account_move_reversal.py
```diff
@@ -1,18 +1,28 @@
...
+from odoo import api, fields, models
 
 
 class AccountMoveReversal(models.TransientModel):
     _inherit = "account.move.reversal"
 
     sale_qty_to_reinvoice = fields.Boolean(
         string="This credit note will be reinvoiced",
-        default="True",
+        default=lambda self: self.company_id.reinvoice_credit_note_default,
```
JordiMForgeFlow: i don't think this is needed if you are already handling it in the default_get

sale_line_refund_to_invoice_qty/wizards/account_move_reversal.py
```diff
@@ -1,18 +1,28 @@
...
         help="Leave it marked if you will reinvoice the same sale order line "
         "(standard behaviour)",
     )
 
+    @api.model
+    def default_get(self, flds):
```
JordiMForgeFlow:
```suggestion
    def default_get(self, fields):
```

sorry for the nitpicking 🙏🏼

JordiMForgeFlow (commented):
Could you also attend the failed CI? Thank you :)

sale_line_refund_to_invoice_qty/wizards/account_move_reversal.py
PicchiSeba: Oh damn, I totally forgot to remove it

sale_line_refund_to_invoice_qty/wizards/account_move_reversal.py
PicchiSeba: **question**: I renamed the argument to `fields` because it shadows `odoo.fields` that we import at the top of the file. Is it absolutely necessary to leave the name as `fields`?

JordiMForgeFlow: I see, Odoo uses 'fields' anyway, but maybe better to rename it. However I would prefer maybe 'fields_list', what do you think?

PicchiSeba: Yeah, it fits nicely. Thanks.

francesco-ooops:
@JordiMForgeFlow looks good to go :)

sale_line_refund_to_invoice_qty/wizards/account_move_reversal.py
```diff
@@ -1,18 +1,28 @@
...
         "(standard behaviour)",
     )
 
+    @api.model
+    def default_get(self, fields_list):
+        res = super().default_get(fields_list)
+        company = (
+            self.env["res.company"].browse(res.get("company_id"))
+            if res.get("company_id")
+            else self.env.company.id
```
JordiMForgeFlow:
```suggestion
            else self.env.company
```

JordiMForgeFlow (approved):
code review 👍🏼

PicchiSeba:
@JordiMForgeFlow  Thanks for your patience!

francesco-ooops (approved):
Functional is ok as well, thanks again @JordiMForgeFlow 

@pedrobaeza can we merge?

PR #1751, opened 2024-06-28
[16.0] account_invoice_pricelist: Use setter for triple discount

grindtildeath:
Relates to #1638

### DIFF
--- a/account_invoice_pricelist/models/account_move.py
+++ b/account_invoice_pricelist/models/account_move.py
@@ -104,7 +104,7 @@ class AccountMoveLine(models.Model):
                     self.tax_ids,
                     self.company_id,
                 )
-                self.with_context(check_move_validity=False).discount = 0.0
+                self._set_discount(0.0)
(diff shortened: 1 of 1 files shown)


legalsylvain:
This should not be required, with latest refactoring. don't you think ? 

Could you provide a test that is failing without that patch ? Thanks !

grindtildeath:
It is needed if you test in the UI.

Without this, the onchange on product (when creating a new line and selecting the product) will not set the discount in discount1 field. We would need the inverse function for this to happen when the document is saved, which will in turn overwrite any changes done in other discount fields.

Requiring the user to save the invoice before being able to set the fields properly is not nice.

legalsylvain (approved):

legalsylvain:
hi @OCA/accounting-maintainers : could you merge this one ? it is required to merge https://github.com/OCA/account-invoicing/pull/1840

thanks !

mmequignon (approved):

PR #1754, opened 2024-07-09
[14.0][FIX] purchase_stock_picking_return_invoicing: Add digits="Product Unit of Measure" to qty_refunded + qty_returned fields

victoralmau:
Add digits="Product Unit of Measure" to `qty_refunded` + `qty_returned` fields

Please @pedrobaeza can you review it?

@Tecnativa TT50034


pedrobaeza (approved):
Yes, makes totally sense, as the rest of the fields have this digits. Fast-tracking thus:

/ocabot merge patch

PR #1755, opened 2024-07-09
[16.0] [MIG] account_invoice_refund_reinvoice: Migration to 16.0

luisDIXMIT:
Standard migration.

@etobella


account_invoice_refund_reinvoice/models/account_move.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def _prepare_reinvoice_reversal(self):
+        return {
+            "ref": _("Reversal of: %s") % (self.name),
+            "journal_id": self.journal_id.id,
+            "invoice_payment_term_id": None,
+            "invoice_user_id": self.invoice_user_id.id,
+        }
+
+    def _reverse_move_vals(self, default_values, cancel=True):
```
etobella: I think that this function was removed on 16. Also, I think you found another way to do the job.

luisDIXMIT: You're right! I removed the unnecessary overridden method.

etobella (approved):

kluna1998 (approved):

PR #1773, opened 2024-08-08
[13.0][FIX] account_invoice_refund_link: migate check column old_invoice_line_id exists

EmilioPascual:
@Shide @edlopen please reveiw

@moduon MT-6161


Shide (approved):

pedrobaeza (changes requested):
The same comment as https://github.com/OCA/contract/pull/1111#pullrequestreview-2228306604

EmilioPascual:
> The same comment as [OCA/contract#1111 (review)](https://github.com/OCA/contract/pull/1111#pullrequestreview-2228306604)

Code updated putting contrary condition.

pedrobaeza:
But the enterprise question remains.

PR #1774, opened 2024-08-09
[IMP][14.0] account_invoice_section_sale_order: Fix read/write access

mmequignon:
The `sale.order._create_invoices()` method does not only requires `create` access to the `account.move` and `account.move.line` models.

Because we're actually sorting lines after the creating, it requires `read` access to sort, and `write` access to modify their sequence.

This is problematic when interracting with other modules like `invoice_mode_at_shipping` where stock users are the ones to create invoices.

This PR adds a few `sudo()` in the code, in order to avoid granting `read` and `write` access to users that shouldn't be allowed to read and write invoices.

### DIFF
--- a/account_invoice_section_sale_order/models/sale_order.py
+++ b/account_invoice_section_sale_order/models/sale_order.py
@@ -16,14 +16,16 @@ class SaleOrder(models.Model):
         the group name.
         Only do this for invoices targetting multiple groups
         """
-        invoice_ids = super()._create_invoices(grouped=grouped, final=final, date=date)
-        for invoice in invoice_ids:
+        invoices = super()._create_invoices(grouped=grouped, final=final, date=date)
+        for invoice in invoices.sudo():
             if (
                 len(invoice.line_ids.mapped(invoice.line_ids._get_section_grouping()))
                 == 1
             ):
                 continue
             sequence = 10
+            # Because invoices are already created, this would require
+            # an extra read access in order to read order fields.
             move_lines = invoice._get_ordered_invoice_lines()
             # Group move lines according to their sale order
             section_grouping_matrix = OrderedDict()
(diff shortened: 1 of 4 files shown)


henrybackman (approved):

sebalix (approved):

account_invoice_section_sale_order/models/sale_order.py
```diff
@@ -24,7 +24,9 @@ def _create_invoices(self, grouped=False, final=False, date=None):
             ):
                 continue
             sequence = 10
-            move_lines = invoice._get_ordered_invoice_lines()
+            # Because invoices are already created, this would require
+            # an extra read access in order to read order fields.
+            move_lines = invoice.sudo()._get_ordered_invoice_lines()
```
simahawk: can't we avoid calling `sudo` everywhere? It all starts from line 19 where you can call `sudo` at the end of the super call. Plus, can we kindly rename `invoice_ids` to `invoices`? :angel:

simahawk (changes requested):

account_invoice_section_sale_order/models/sale_order.py
mmequignon: Updated, can you please approve and merge please ?

simahawk (approved):

account_invoice_section_sale_order/models/sale_order.py
```diff
@@ -56,10 +58,14 @@ def _create_invoices(self, grouped=False, final=False, date=None):
                     )
                     sequence += 10
                 for move_line in self.env["account.move.line"].browse(move_line_ids):
```
simahawk, line 60: we overlooked here... this browse requires a sudo as well :face_holding_back_tears:

Issue #1775, opened 2024-08-13
The Constraint will be triggered on duplicating the invoice, if the pricelist is not the partner_id -> pricelist_id.


agent-z28:
When we assign a custom pricelist, with a diffrent currency, the invoice cannot be duplicated.

"Pricelist and Invoice need to use the same currency."

https://github.com/OCA/account-invoicing/blob/4be442bf1284c76472ddb1750f7a4b9b3d0699ae/account_invoice_pricelist/models/account_move.py#L35

I think the pricelist_id should be copy=True, but the pricelist is computed..

agent-z28:
@bizzappdev

PR #1777, opened 2024-08-14
[FIX] account_invoice_pricelist: Added copy attribute in the pricelist_id field

bizzappdev:
Fixing: #1775


agent-z28 (approved):
Works

legalsylvain (approved):
Makes sense.

/ocabot merge patch

PR #1781, opened 2024-08-23
[16.0][MIG] account_invoice_customer_no_autofollow

sbidoul:


sbidoul:
We may want to rename this module to `account_invoice_partner_no_autofollow` since it equally applies to vendor bills.

account_invoice_customer_no_autofollow/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+
+{
+    "name": "Invoice Customer No Autofollow",
+    "summary": "Do not add customer as follower in Invoices",
+    "author": "Cetmix,Odoo Community Association (OCA)",
+    "version": "16.0.1.0.0",
+    "website": "https://github.com/OCA/account-invoicing",
+    "category": "Account",
+    "depends": ["account"],
+    "maintainers": ["dessanhemrayev", "CetmixGitDrone"],
```
ivs-cetmix: Pls remove these maintainers or replace with the ones you would like to add.

ivs-cetmix (approved):
Code and functional LGTM. Pls check the comment regarding maintainers.

ivs-cetmix:
> We may want to rename this module to `account_invoice_partner_no_autofollow` since it equally applies to vendor bills.

@sbidoul this is a good idea. However I wouldn't do this to avoid potential migration issues.

account_invoice_customer_no_autofollow/__manifest__.py
sbidoul: Done

sbidoul:
I wonder why such module is not more popular. I kind of feel I'm missing something obvious...

@StefanRijnhart since you reviewed the same for sale orders, you may be interested in this one.

StefanRijnhart (approved):
Thanks! I see no issues with the rename, because there is no data associated with this module. It would be nice to see the config parameter renamed on migration (after using openupgrade's rename_modules) or (re)installation.

ivs-cetmix:
> 

Yup, this means that there we should add a migration script 

> Thanks! I see no issues with the rename, because there is no data associated with this module. It would be nice to see the config parameter renamed on migration (after using openupgrade's rename_modules) or (re)installation.

@StefanRijnhart Yup, actually we can keep the existing config parameter name or add a migration script it we really need to rename it.

dessanhemrayev (approved):
LGTM

rafaelbn:
Hello @sbidoul 

> I wonder why such module is not more popular. I kind of feel I'm missing something obvious...

We don't have and request in clients. They have customers as followers and there are not any problem reported in many years.

🤔 

Best regards!

legalsylvain:
@StefanRijnhart / @rafaelbn : could you merge this one ?

PR #1804, opened 2024-09-20
[17.0][FW] account_invoice_section_sale_order: multiple ports from 16.0

sebalix:
Same port than:
- #1803

Port of the following PRs from 14.0 to 17.0:
- #1050
- #1117
- #1135
- #1384 (one commit updating `account_invoice_section_sale_order` regarding tests)

### DIFF
--- a/account_invoice_section_sale_order/__manifest__.py
+++ b/account_invoice_section_sale_order/__manifest__.py
@@ -10,4 +10,9 @@
     "license": "AGPL-3",
     "category": "Accounting & Finance",
     "depends": ["account", "sale"],
+    "data": [
+        "security/res_groups.xml",
+        "views/res_config_settings.xml",
+        "views/res_partner.xml",
+    ],
 }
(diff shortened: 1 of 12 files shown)


henrybackman (approved):

grindtildeath (approved):
LG but as your fixup is touching multiple commits, IMO it's better to keep it as a separated commit for the migration of previous ones. Then you only need to reword it

sebalix:
@grindtildeath that would be better yes, but the downside of doing that is that we are still generating conflicts we need to resolve in a separate commit on each version, while updating the current patch one version after the other makes the port simpler.

In an ideal world, when someone propose a patch on a given version, he could take the opportunity to port it straight to the next versions so that will be same commit+author and he could amend it without issue (here with what I have done, I'm not the author of the adaptations).

EDIT: I didn't understood well your comment, it's OK now

sebalix:
@grindtildeath fixup fixed & squashed :)

PR #1809, opened 2024-10-04
[17.0][ADD] acount_move_tags: add module

joepsanders:


chrisb-c01 (approved):

chrisb-c01:
LGTM

@OCA/accounting-maintainers can you review?

pedrobaeza:
Thanks for the contribution. Please avoid plurals on module names as stated in https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#modules. You can call it `account_move_tag` for example (although we know that it can be more than one tag). It's a technical name, and thus, it shouldn't reflect the "language correction". Why plurals should be avoided is for avoiding frequent typos typing the module name. Imagine an extension module `account_move_tags_images`, and then `account_move_tags_images_relative`, and so on: you'll end up putting s where it's not and not putting them when it should.

Please also squash both commits.

joepsanders:
> Thanks for the contribution. Please avoid plurals on module names as stated in https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#modules. You can call it `account_move_tag` for example (although we know that it can be more than one tag). It's a technical name, and thus, it shouldn't reflect the "language correction". Why plurals should be avoided is for avoiding frequent typos typing the module name. Imagine an extension module `account_move_tags_images`, and then `account_move_tags_images_relative`, and so on: you'll end up putting s where it's not and not putting them when it should.
> 
> Please also squash both commits.

Thanks @pedrobaeza for the feedback and explanation! Review of the updated module is much appreciated :)

account_move_tag/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+    "summary": """
+        Adds tags to the account move so vendor bills and customer invoices can be
+        easily distinguished in the accounting journal.
+    """,
+    "author": "Codeforward B.V., Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/account-invoicing",
+    "category": "Hidden",
+    "version": "17.0.1.0.0",
+    "depends": ["account"],
+    "license": "LGPL-3",
```
pedrobaeza: Why not putting it `AGPL-3` for avoiding private extensions on this?

pedrobaeza (approved):

account_move_tag/__manifest__.py
joepsanders: Hi Pedro, we'll dive into the licensing question ;). Thanks

chrisb-c01 (approved):

account_move_tag/__manifest__.py
joepsanders: @pedrobaeza the licence has been updated. Happy to see this one merged! Thanks

pedrobaeza (approved):

marcelsavegnago (approved):
LGTM

joepsanders:
@pedrobaeza can this be merged?

PR #1813, opened 2024-10-09
[16.0][ADD] stock_account_move_reset_to_draft: New module

victoralmau:
New module

Please @pedrobaeza and @carlosdauden can you review it?

@Tecnativa TT51201


stock_account_move_multi_reset_to_draft/models/account_move.py
```diff
@@ -0,0 +1,32 @@
...
+    _inherit = "account.move"
+
+    def button_draft(self):
+        """If it is a purchase invoice, we will delete the SVLS linked to it."""
+        res = super().button_draft()
+        for item in self.filtered(
+            lambda x: x.is_inbound
+            and any(line.stock_valuation_layer_ids for line in x.sudo().line_ids)
+        ):
+            item.sudo().line_ids.stock_valuation_layer_ids.unlink()
```
carlosdauden, line 17: Perhaps a better option would be to create a copy of stock_valuation_layer_ids but with opposite sign.
This will ensure that if someone print a stock valuation report on a certain date, that report will not be affected by later invoice changes.
Otherwise users may say "How come I have this report in pdf with one value and if I pull it now it comes out with a different value?

carlosdauden (changes requested):

stock_account_move_multi_reset_to_draft/models/account_move.py
carlosdauden: Perhaps a better option would be to create a new SVL with sum of stock_valuation_layer_ids values but with opposite sign.
This will ensure that if someone print a stock valuation report on a certain date, that report will not be affected by later invoice changes.
Otherwise users may say "How come I have this report in pdf with one value and if I pull it now it comes out with a different value?

carlosdauden (changes requested):

stock_account_move_multi_reset_to_draft/models/account_move.py
victoralmau: Is that really necessary? The SVL is removed because when the invoice is changed to draft and confirmed again the extra SVL will (maybe) be generated for the difference (without the need to create an extra one with opposite signs).

Can you describe the use case you indicate that may be confusing?

yostashiro:
Looks like we should converge the ideas. https://github.com/OCA/account-financial-tools/pull/1946 :)
CC: @rousseldenis

stock_account_move_multi_reset_to_draft/models/account_move.py
pedrobaeza: I agree with Carlos. The key is not doing changes in operations in the past, so you have the regularization SVL, then reset again to draft, so you neutralize the regularization SVL with another one in the date you are resetting, and then create a new one when finally validating again.

pedrobaeza:
This one seems more complete, handling all the SVLs adjustments.

yostashiro (commented):
Why "multi" in the module name?

pedrobaeza:
Because you can reset to draft "multiple" times. The standard only allows one.

yostashiro:
> The standard only allows one.

I don't see it this way. You can reset a bill to draft multiple times with the standard as long as there is no adjustment SVL, no?

pedrobaeza:
Yes, correct, but we are referring if you change the price multiple times. That's difficult to be reflected in the technical name, but if you have a nicer name, we are all ears.

yostashiro:
My preference is stock_account_move_force_reset_to_draft but it's not a strong one. I was just curious.

stock_account_move_multi_reset_to_draft/models/account_move.py
victoralmau: Ok, if I understood correctly, the change is already done.

stock_account_move_multi_reset_to_draft/models/account_move.py
```diff
@@ -0,0 +1,41 @@
...
+            lambda x: x.is_inbound
+            and any(line.stock_valuation_layer_ids for line in x.sudo().line_ids)
+        ):
+            for line in item.sudo().line_ids.filtered("stock_valuation_layer_ids"):
+                svls = line.stock_valuation_layer_ids
+                value = sum(svls.mapped("value"))
+                if not float_is_zero(
+                    value, precision_rounding=line.currency_id.rounding
+                ):
+                    svls[0].copy({"value": value * -1})
```
carlosdauden: You can add `-` as a prefix to the variable to change sign, but leave it as it is if you think it is better understood the way it is coded.
```suggestion
                    svls[0].copy({"value": -value})
```

stock_account_move_multi_reset_to_draft/models/account_move.py
```diff
@@ -0,0 +1,41 @@
...
+    def button_draft(self):
+        """If it is a purchase invoice, we will create a new SVL for each line with
+        the sum of the value in opposite sign.
+        """
+        res = super().button_draft()
+        for item in self.filtered(
+            lambda x: x.is_inbound
+            and any(line.stock_valuation_layer_ids for line in x.sudo().line_ids)
+        ):
+            for line in item.sudo().line_ids.filtered("stock_valuation_layer_ids"):
```
carlosdauden: Try this to call sudo only once for all lines.
```suggestion
        for item in self.sudo().filtered(
            lambda x: x.is_inbound
            and any(line.stock_valuation_layer_ids for line in x.line_ids)
        ):
            for line in item.line_ids.filtered("stock_valuation_layer_ids"):
```

carlosdauden (approved):
Code LGTM

rousseldenis:
> Looks like we should converge the ideas. [OCA/account-financial-tools#1946](https://github.com/OCA/account-financial-tools/pull/1946) :) CC: @rousseldenis

Yes, I think this one is better in terms of reliability and avoid to enforce additional protections on prices changes or ...

stock_account_move_multi_reset_to_draft/models/account_move.py
```diff
@@ -0,0 +1,41 @@
...
+    def button_draft(self):
+        """If it is a purchase invoice, we will create a new SVL for each line with
+        the sum of the value in opposite sign.
+        """
+        res = super().button_draft()
+        for item in self.sudo().filtered(
+            lambda x: x.is_inbound
+            and any(line.stock_valuation_layer_ids for line in x.line_ids)
+        ):
+            for line in item.sudo().line_ids.filtered("stock_valuation_layer_ids"):
```
carlosdauden:
```suggestion
            for line in item.line_ids.filtered("stock_valuation_layer_ids"):
```

rousseldenis:
> Yes, correct, but we are referring if you change the price multiple times. That's difficult to be reflected in the technical name, but if you have a nicer name, we are all ears.

I'm also in favor of not making 'multi' in the name. As, if you do it once, you can dot it more. That's the same principle for all existing modules that reset objects to draft state. But you can add it to USAGE which is a good approach IMHO.

pedrobaeza:
Then which name?

rousseldenis:
> stock_account_move_force_reset_to_draft

Preference for `stock_account_move_reset_to_draft` which means what it does.

pedrobaeza:
OK, let's rename it, but first please do the review for performing all the possible changes at the same time.

stock_account_move_multi_reset_to_draft/models/account_move.py
```diff
@@ -0,0 +1,41 @@
...
+            and any(line.stock_valuation_layer_ids for line in x.line_ids)
+        ):
+            for line in item.line_ids.filtered("stock_valuation_layer_ids"):
+                svls = line.stock_valuation_layer_ids
+                value = sum(svls.mapped("value"))
+                if not float_is_zero(
+                    value, precision_rounding=line.currency_id.rounding
+                ):
+                    svls[0].copy({"value": -value})
+        return res
```
yostashiro: There are certain cases where we shouldn't be allowing the reset-to-draft operation, and a few considerations should be added. 

```suggestion
        for move in self.sudo():
            if not move.is_inbound:
                continue
            for line in move.line_ids.filtered("stock_valuation_layer_ids"):
                origin_svls = line.stock_valuation_layer_ids.stock_valuation_layer_id
                if len(origin_svls.stock_valuation_layer_ids.account_move_line_id.filtered(lambda x: x.parent_state == "posted")) > 1:
                    raise UserError(_("Invenotory valuation records are intertwined for %(line_name)s.", line_name=line.display_name))
                for origin_svl in origin_svls:
                    if origin_svl.quantity != origin_svl.remaining_qty:
                        raise UserError(_("The inventory has already been (partially) consumed for %(line_name)s.", line_name=line.display_name))
                    svls = origin_svl.stock_valuation_layer_ids
                    value = sum(svls.mapped("value"))
                    if not float_is_zero(
                        value, precision_rounding=line.currency_id.rounding
                    ):
                        origin_svl.remaining_value -= value
                        revert_svl = svls[0].copy({"value": -value})
                        revert_svl._validate_accounting_entries()
                product = line.product_id.with_company(move.company_id.id)
                if product.cost_method == "average":
                    product.sudo().with_context(disable_auto_svl=True).write(
                        {"standard_price": product.value_svl / product.quantity_svl}
                    )
        return super().button_draft()
```
The suggested code has survived my functional test, although I may have missed some corner cases.

I did the following steps with FIFO and AVCO products and checked the valuation at each step.

- PO for a product: 2 pcs at EUR10
- Receive 1 pc and create a backorder
- Receive 1 pc
- Create a bill for 2 pcs at EUR12 and post
- Reset the bill to draft
- Change the bill content to 1 pc at EUR15 and post
- Create another bill for 1 pc at EUR8 and post
- Reset the first bill to draft  -> User error to prevent valuation inconsistencies
- Delivery 1 pc

yostashiro (changes requested):
Can you please check my suggestion?

stock_account_move_multi_reset_to_draft/models/account_move.py
victoralmau: I am trying to add this change and creating a test with the use case you describe but there are several considerations:
- I failed to reproduce the error in the step “_Reset the first bill to draft -> User error to prevent valuation inconsistencies_”.
- I think that the `_validate_accounting_entries()` method should not be called.
- I think changing the `standard_price` is something this module should not do.

In general I think it is adding too much logic to something that should not have it, you should be able to change to draft as many times as you want (at least that's my opinion).

yostashiro: @victoralmau Thanks for checking.

> * I failed to reproduce the error in the step “_Reset the first bill to draft -> User error to prevent valuation inconsistencies_”.

I think the error should be reproducible if you follow the steps exactly as described.

> * I think that the `_validate_accounting_entries()` method should not be called.

The inventory / interim receipt account balance will be incorrect if you don't do this.

> * I think changing the `standard_price` is something this module should not do.

Why not? The value of the outgoing shipment will be incorrect without this, if the shipment is processed before the next update of `standard_price` (repost of the bill or a receipt of the product).

> In general I think it is adding too much logic to something that should not have it, you should be able to change to draft as many times as you want (at least that's my opinion).

It can't be that simple, unfortunately, and that's why there has been no proper "fix" by Odoo until today after this comment https://github.com/odoo/odoo/issues/118687#issuecomment-1510927107. Inventory valuation and account balances will go wrong with the current code.

rousseldenis:
@victoralmau @pedrobaeza Could you check @yostashiro comments ?

victoralmau:
Renamed to `stock_account_move_reset_to_draft` and applied the changes suggested in https://github.com/OCA/account-invoicing/pull/1813#discussion_r1795483401 (personally I think it is adding logic that this module should not have).

yostashiro:
> personally I think it is adding logic that this module should not have

@victoralmau Thanks for the update but I'm not sure where your comment is coming from. If an implementation risks compromising data integrity/consistency, we must do our best to prevent that. No question about it, IMO.

stock_account_move_multi_reset_to_draft/models/account_move.py
victoralmau: I have added all the changes you indicate, but in the test this step _Reset the first bill to draft -> User error to prevent valuation inconsistencies_ does not work.

gdgellatly:
@pedrobaeza @victoralmau I added some comments here https://github.com/OCA/account-financial-tools/pull/1947#issuecomment-2428475423 about an alternative approach which is strictly speaking less accurate, but more tolerant of intervening transactions (e.g. valuation already fixed, sold though etc) and avoids editing svl's. I make no claim that it is better or worse, just food for thought.

ivantodorovich (approved):

kanda999 (approved):
Functional test.

pedrobaeza (approved):
We have a good start with some constraints and amount rectifications code, so let's merge it, and we can improve it later when used in real cases.

/ocabot merge nobump

yostashiro:
Maybe it's too late now, but there is an issue with the code. We will propose a fix in another PR if it's too late.

pedrobaeza:
I have cancelled the CI. Tell us the problem.

stock_account_move_reset_to_draft/models/account_move.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    def button_draft(self):
+        """If it is a purchase invoice, we will create a new SVL for each line with
+        the sum of the value in opposite sign.
+        """
+        res = super().button_draft()
```
yostashiro:
```suggestion
```

stock_account_move_reset_to_draft/models/account_move.py
```diff
@@ -0,0 +1,73 @@
...
+                    ):
+                        origin_svl.remaining_value -= value
+                        revert_svl = svls[0].copy({"value": -value})
+                        revert_svl._validate_accounting_entries()
+                product = line.product_id.with_company(item.company_id.id)
+                if product.cost_method == "average":
+                    product.sudo().with_context(disable_auto_svl=True).write(
+                        {"standard_price": product.value_svl / product.quantity_svl}
+                    )
+        return res
```
yostashiro:
```suggestion
        return super().button_draft()
```

yostashiro (changes requested):
@pedrobaeza Thanks!

This is why the error didn't reproduce: https://github.com/OCA/account-invoicing/pull/1813#discussion_r1800531724

victoralmau:
Changes done.

stock_account_move_reset_to_draft/models/account_move.py
```diff
@@ -0,0 +1,72 @@
+# Copyright 2024 Tecnativa - Víctor Martínez
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
```
yostashiro:
```suggestion
# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
```

yostashiro (approved):
LGTM just with a correction of the license. Thanks!

rrebollo:
> we can improve it later when used in real cases.

We have a customer being hit by this. We deployed OCA/account-financial-tools#1962 for them but if @yostashiro looks our code and PR description there and grant us this addon cover the same use case we are open to deploy this and we'll give U feedback about it.

PR #1827, opened 2024-10-29
[ADD] account_move_picking_match_or_create: new addon

DiegoParadeda:


account_move_stock_create/models/account_move.py
```diff
@@ -0,0 +1,100 @@
...
+    def action_generate_pickings_from_invoices(self):
+        """Generate pickings from invoices."""
+        for record in self:
+            if record.picking_ids and len(self) == 1:
+                raise UserError(
+                    _("There's already a picking created for this account move.")
+                )
+            elif record.picking_ids:
+                continue
+            record.generate_picking_from_invoice()
```
rvalyi: It is common to have more than one picking grouped in the same invoice...

But also more generally, I think it would be cool to have a wizard or maybe an o2m field were one could select existing pickings. If so, the module could try to match the existing stock moves with some kind of key (product_id, possibly quantity...) and simply link matched stock moves, possibly with some match flag that would visually show what was matched or not. If the line could not be matched, then it might be generated as you did...

It is indeed common that PO will generate incoming pickings that one will mostly want to match when electronic invoices are imported. With your current proposals you would need to delete manually these incoming pickings generated from the PO's and your PO will miss their delivery and invoicing status...

Bonus, it should be flexible so that it will keep working with the new PO/Bill matching system from v18.

rvalyi (changes requested):

account_move_stock_create/__manifest__.py
```diff
@@ -0,0 +1,21 @@
+# Copyright (C) 2024-Today - KMEE (<http://www.kmee.com.br>).
+# @author Diego Paradeda <diego.paradeda@kmee.com.br>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Account Move Stock Create",
+    "summary": """This addon creates stock transfers from an account move""",
+    "version": "14.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "KMEE INFORMATICA LTDA,Odoo Community Association (OCA)",
```
mileo: KMEE only plz.

account_move_stock_create/models/account_move.py
DiegoParadeda: @rvalyi  I've moved the picking check to the invoice line, allowing the creation of multiple stock pickings for the same invoice as long as there's a new invoice line to transfer (one that hasn't already created a stock transfer).

Regarding the wizard idea, that is a goal we have in mind. We plan to add functionalities similar to the stock_invoice_onshipping wizard in the future. I'm changing the module to alpha and including that in the roadmap. For now, the goal is to determine which modules will cover the most basic cases when handling the creation of any documents that have an invoice as the "source document."

rvalyi (changes requested):
So given the new matching logic, I think the account_move_stock_create name can be improved... What about something like account_move_picking_match  or something like that?

PR #1836, opened 2024-11-08
[13.0][FIX] account_invoice_show_currency_rate: use only debit from currency lines to get rate

JordiMForgeFlow:
With anglo-saxon accounting, the booking of the COGS and stock interim account is done always in company currency. With the current logic, we are considering the extra debit of the COGS in the invoices which has no amount currency and the resulting rate is incorrect.


GuillemCForgeFlow (approved):

JoanSForgeFlow (approved):
LGTM!

JordiMForgeFlow:
Hi @victoralmau @pedrobaeza 

Could you check this one? :)

MiquelRForgeFlow (approved):

victoralmau (approved):
I am not familiar with anglo-saxon accounting but I believe that the change is coherent.

PR #1837, opened 2024-11-08
[16.0][FW][FIX] account_invoice_show_currency_rate: use only debit from currency lines to get rate

JordiMForgeFlow:
FW of #1836


MiquelRForgeFlow (approved):

victoralmau (approved):

pedrobaeza (approved):
/ocabot merge patch

pedrobaeza:
Wait, I have now the doubt about customer refunds. What happen with them?

JordiMForgeFlow:
@pedrobaeza it still works as we are only comparing the lines having amount_currency > 0, so having a debit amount. Of course both in the invoice and refund some lines are ignored because they are credit lines instead of debits.

PR #1838, opened 2024-11-11
[13.0][IMP] account_invoice_show_currency_rate: add journal item rate and consider journal item balance for credit amounts

JordiMForgeFlow:
- Add currency rate field on journal items
- Compute rate based on the balance instead of looking just at the debit amount

### DIFF
--- a/account_invoice_show_currency_rate/models/account_move.py
+++ b/account_invoice_show_currency_rate/models/account_move.py
@@ -47,3 +47,46 @@ class AccountMove(models.Model):
             item.show_currency_rate_amount = (
                 item.currency_id and item.currency_id != item.company_id.currency_id
             )
+
+
+class AccountMoveLine(models.Model):
+    _inherit = "account.move.line"
+
+    # TODO: This field should be removed in V16 as it is already
+    # added by Odoo in the account module
+    currency_rate_amount = fields.Float(
+        string="Rate amount", compute="_compute_currency_rate_amount", digits=0,
+    )
+
+    @api.depends(
+        "move_id.state",
+        "move_id.date",
+        "amount_currency",
+        "debit",
+        "move_id.company_id",
+        "currency_id",
+    )
+    def _compute_currency_rate_amount(self):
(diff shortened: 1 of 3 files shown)


JordiMForgeFlow:
@victoralmau could you check these changes? :)

AaronHForgeFlow:
Is the new field because the currency rate in the lines can be different from the currency rate of the journal entries? is that because of manual editions of the amount currency?

JordiMForgeFlow:
@AaronHForgeFlow you can create misc entries with lines having different currency. Odoo already provides this field from >=V16

AaronHForgeFlow (approved):
Code review LGTM

JoanSForgeFlow (approved):
LGTM!

victoralmau (changes requested):
I don't know exactly the specific use case that is solved, can you add a test to confirm it?

What is the real need to add the `currency_rate_amount` field all `account.move.line` records (You must set a migration script to create the column and set the values that corresponds).

JordiMForgeFlow:
@victoralmau the rate in the journal items is not stored.

The currency rate is actually added to journal items in V16 (https://github.com/odoo/odoo/blob/e614f082a95a249e4abc29ee7a7e3a3bcaedd9ea/addons/account/models/account_move_line.py#L115). It is useful for entries where different currencies may be involved.

victoralmau:
> @victoralmau the rate in the journal items is not stored.
> 
> The currency rate is actually added to journal items in V16 (https://github.com/odoo/odoo/blob/e614f082a95a249e4abc29ee7a7e3a3bcaedd9ea/addons/account/models/account_move_line.py#L115). It is useful for entries where different currencies may be involved.

OK sorry, I hadn't seen it. I think it would be interesting to add a TODO comment to indicate that this field and method should be removed in v16 because account already adds it with the `currency_rate` field.

JordiMForgeFlow:
@victoralmau now it should be ready :)

victoralmau (approved):

PR #1839, opened 2024-11-11
[15.0][FIX] account_invoice_mass_sending: Users have extra notifications when the invoice has sent by the job

sergio-teruel:
cc @Tecnativa TT51673

ping @pilarvargas-tecnativa @chienandalu

### DIFF
--- a/account_invoice_mass_sending/models/account_move.py
+++ b/account_invoice_mass_sending/models/account_move.py
@@ -45,6 +45,7 @@ class AccountInvoice(models.Model):
                 "active_model": self._name,
                 "active_ids": self.ids,
                 "discard_logo_check": True,
+                "account_invoice_mass_sending": True,
             }
         )
         wiz = self.env["account.invoice.send"].with_context(**wiz_ctx).create({})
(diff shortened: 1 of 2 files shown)


pilarvargas-tecnativa (approved):

chienandalu (approved):

jguenat:
thanks @sergio-teruel 
This is to avoid other follower of the invoice to get notification when it is send by the job right?
Kind of hard to test it on runboat because queue_job is not enabled there.
Does it has any side effect when module mass_mailing is installed?

sergio-teruel:
Hi @jguenat I do not try with this module but I don't think it should be a problem.
The compute method is altered only if 'account_invoice_mass_sending' key comes in context

jguenat (approved):
Indeed no issue with mail_marketing.
Tested it and it works well thanks
If making codecov happy is mendatory then we just need a test sending an invoice without enqueue_invoices

pedrobaeza:
Fw-port it to upper versions:

/ocabot merge patch

PR #1846, opened 2024-11-14
[16.0][FP] account_invoice_mass_sending: Users have extra notifications when the invoice has sent by the job

jguenat:


pedrobaeza (approved):
Thanks for doing the ports.

sergio-teruel (approved):
Thanks

PR #1847, opened 2024-11-18
[18.0][MIG] account_invoice_fixed_discount: Migration to 18.0

achulii:


JessBrandl (approved):
Code LGMT, tested locally and worked as intended. 
Small improvement maybe, when putting in a discount without setting a product or price first, it throws an undescriptive error, so perhaps we can add a message to tell the user that this needs to be done first. 
image

pedrobaeza:
Closing in favor of #1949

PR #1852, opened 2024-11-25
[16.0][MIG] account_invoice_view_payment: Migration to 16.0

anusriNPS:
Migrating account_invoice_view_payment to 16.0


account_invoice_view_payment/models/account_move.py
```diff
@@ -0,0 +1,94 @@
...
+                            ),
+                            "amount_foreign_currency": foreign_currency
+                            and formatLang(
+                                self.env,
+                                abs(counterpart_line.amount_currency),
+                                currency_obj=foreign_currency,
+                            ),
+                        }
+                    )
+        return reconciled_vals
```
HekkiMelody: **issue**: I don't think we should reintroduce this function. It changed name and structure slightly, but it's `_compute_payments_widget_reconciled_info` in v16 and I think you should be able to use it in the same way.

HekkiMelody (changes requested):

account_invoice_view_payment/models/account_move.py
anusriNPS: Introduced new function as reconciled_vals is not obtainable from existing _compute_payments_widget_reconciled_info() or any other available methods in v16.

However, account_invoice_view_payment only uses payment related information hence using get_reconciled_payments method to achieve the same in order to view payments from invoices.

account_invoice_view_payment/models/account_move.py
```diff
@@ -0,0 +1,37 @@
...
+                "account.action_account_payments_payable"
+            )
+        else:
+            action = self.env["ir.actions.act_window"]._for_xml_id(
+                "account.action_account_payments"
+            )
+        payment = []
+        for move in self:
+            if move._get_reconciled_payments().id:
+                payment.append(move._get_reconciled_payments().id)
```
HekkiMelody: **issue**: `_get_reconciled_payments()` could return multiple payments, so `.id` could then give a SingletonError

I've tested that `_get_reconciled_payments` is safe to use on a recordset with multiple records, so:

```suggestion
        payments = move._get_reconciled_payments().ids
```

HekkiMelody (commented):
everything else LGTM

account_invoice_view_payment/models/account_move.py
HekkiMelody: ah, nice catch!

account_invoice_view_payment/models/account_move.py
anusriNPS: Updated using ids

HekkiMelody (approved):
Code review, LGTM

account_invoice_view_payment/models/account_move.py
```diff
@@ -22,16 +22,16 @@ def action_view_payments(self):
...
         # choose the view_mode accordingly
-        if len(reconciles) != 1:
-            action["domain"] = "[('id', 'in', " + str(payment) + ")]"
+        if len(payments) != 1:
+            action["domain"] = "[('id', 'in', " + str(payments) + ")]"
         else:
             res = self.env.ref("account.view_account_payment_form", False)
             action["views"] = [(res and res.id or False, "form")]
-            action["res_id"] = payment and payment[0] or False
+            action["res_id"] = payments and payments[0] or False
```
Honeyxilia: Regarding these conditionnal branches : I believe that rather than testing if the length of `payments` is not equal to one, it should check if it's strictly above 1 
Indeed, if the list is empty, then it's length is 0, which would pass the test, but not work
Also, in the else branch, there's seemingly a line for whether payments is an empty list (L36)

Honeyxilia (commented):
Overall, this migration looks good, but I do one have question about one section

account_invoice_view_payment/models/account_move.py
anusriNPS: Updated as suggested

sergiocorato (approved):

HekkiMelody:
Thank you @sergiocorato !

PR #1854, opened 2024-11-26
[17.0][mig]stock_account_move_reset_to_draft

leemannd:
Standard migration of the module to 17.0


stock_account_move_reset_to_draft/__manifest__.py
```diff
@@ -4,7 +4,7 @@
     "name": "Stock account move reset to draft",
     "author": "Tecnativa, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/account-invoicing",
-    "version": "16.0.1.0.1",
+    "version": "17.0.1.0.1",
```
leemannd: Some maintainers want to maintain the 3 last version numbers inbetween odoo versions.
I can update it to `17.0.1.0.0` if required

cyrilmanuel (approved):

stock_account_move_reset_to_draft/__manifest__.py
yostashiro: I thought we were supposed to make it `17.0.1.0.0`. What is the rationale of keeping the last three digits?

leemannd: Some maintainers keep the 3 digits up to date inbetween versions. The rationale behind this is to be able to know which fixes/features are included.
The usual procedure is to set as you say `17.0.1.0.0`.

Beside that @yostashiro , if you have the time for a review, you're welcome.

pedrobaeza: That's not the rule, and it shouldn't be, as it's very easy to desynchronize such versions and a maintenance burden to synchronize them with no added benefit. Anyway, the only I know doing it is `mis_builder`, which is more a Python engine than Odoo code, and then it's more possible (although I also don't like it).

leemannd: @pedrobaeza Agreed. I have updated the version number. If you have time to drop a review, you're welcome.

pedrobaeza (approved):
It seems OK looking at the diff review, but @yostashiro should confirm all the valuation stuff.

ivantodorovich (approved):

PR #1855, opened 2024-12-02
[FIX] account_invoice_mass_sending: retry on server disconnect

yajo:
Gmail usually disonnects when there are simultaneous logins.

This allows automatic retries on such situation.

- [ ] Draft until testing if this really fixes the problem in production.

@moduon MT-8120


yajo:
After some testing, the problem seems to have disappeared when changing this config:

```diff
 [queue_job]
-channels = root:4
+channels = root:4,root.account_invoice_mass_sending_channel:1
```

PR #1856, opened 2024-12-02
[FIX] account_invoice_mass_sending: wrong channel name

yajo:
This is how it looks like without the fix:

image

@moduon MT-8120


rafaelbn (approved):
👍🏼 

Ping @Shide

Shide (approved):

rafaelbn (approved):
/ocabot merge patch

PR #1857, opened 2024-12-02
[16.0][MIG][sale_stock_picking_invoicing]

rvalyi:
cc @mbcosta @renatonlima 

EDIT: I put it in draft as I'm not too sure yet what should should be done for analytical fields around https://github.com/OCA/account-invoicing/pull/1857/commits/65fa88301f030e8c386481bbf2bf80c1f6d4347c#diff-d8a5810924c9e2509dedcd08b9b6aea083a8543de434b845310fbe7f0c501b58R163
Suggestions are welcome.


mbcosta:
@rvalyi @renatonlima about the changes:

* Update README in History about migration 
   Should we change the README files from **.rst** to **.md** or this change depends of a repo update?

* For get Analytic Lines I use the value from **analytic_distribution** in the Dict of **_prepare_invoice_line** here https://github.com/OCA/OCB/blob/16.0/addons/sale/models/sale_order_line.py#L1122 

```bash
        analytic_account_id = self.order_id.analytic_account_id.id
        if self.analytic_distribution and not self.display_type:
            res['analytic_distribution'] = self.analytic_distribution
        if analytic_account_id and not self.display_type:
            analytic_account_id = str(analytic_account_id)
            if 'analytic_distribution' in res:
                res['analytic_distribution'][analytic_account_id] = res['analytic_distribution'].get(analytic_account_id, 0) + 100
            else:
                res['analytic_distribution'] = {analytic_account_id: 100}
```

This value and the **sale_line_ids** are the only two fields informed when the ''Invoice Type" are **out_refund** ( Should be informed other fields? Is there another "Invoice Type" that should not copy Sale fields? I let a TODO there https://github.com/akretion/account-invoicing/blob/16.0-mig-sale_stock_picking_invoicing/sale_stock_picking_invoicing/wizards/stock_invoice_onshipping.py#L169 )

Test the copy of field:

image

image

image

Analytic Lines seems to be change as Mixin in the module https://github.com/OCA/OCB/tree/16.0/addons/analytic 

* Use Common Tests from **stock_picking_invoicing** https://github.com/OCA/account-invoicing/blob/16.0/stock_picking_invoicing/tests/common.py , reduce duplicate code 

I did the changes in new commits to make easy to see the diff, but can be just a **amend** in one MIG commit, let me know if will be better only one commit.

With this changes I think the PR can be move from status **Draft** to **Ready to Review**

rvalyi:
@mbcosta thank you for the fixes. As it was a pretty simple commit I squashed your Analytic lines fix https://github.com/OCA/account-invoicing/pull/1857/commits/6d9450cde7bfa49ef0136d8cac51a5e5e99c8102 into the previous migration to 16.0 commit. I believe it is now ready indeed...

renatonlima (approved):

mbcosta:
Thank you @rvalyi @renatonlima for review

PR #1860, opened 2024-12-06
[FIX] account_invoice_fixed_discount: Fixed devision by zero error to…

fd-oerp:
… v17.0

Case: When an invoice line includes a product with a unit price of 0 and a fixed discount is applied (even by mistake), it results in a traceback error.

Sometimes users may not be aware of the proper configurations for applying discounts. If this occurs, the system prevents further entry and shows a traceback.

This scenario has already been addressed in sale_fixed_discount previously.


fd-oerp:
Hi @JordiBForgeFlow 

Could you please review this PR and let me know any updates are needed from my end ?

I see the failed checks but seems it's related other module, Please advice!

image

Thank you

PR #1873, opened 2025-01-03
[18.0][MIG] account_invoice_show_currency_rate

JordiMForgeFlow:
Migration to V18:

- Odoo adds a new field to display the currency rate on invoices when the currency is different than the company's currency (https://github.com/odoo/odoo/blob/e3d54695862f73d9be6c35121def57ed641a1a45/addons/account/models/account_move.py#L462). However, the computed rate is simply based on the rate configured in the system at the time of computing. It could be the case that a different rate was existing when the invoice was posted, so this module ensures that the rate computed for posted invoices is based on the actual currency amount applied.

FW of https://github.com/OCA/account-invoicing/pull/1838:

- Odoo computes the currency rate on journal items (https://github.com/odoo/odoo/blob/e3d54695862f73d9be6c35121def57ed641a1a45/addons/account/models/account_move_line.py#L133). However, the computed rate is based on the rate configured in the system. It could be the case for manually created entries that the user has manually modified the amount in currency, so this module ensures that the computed rate is based on the actual currency amount applied.

FW of https://github.com/OCA/account-invoicing/pull/1899 and https://github.com/OCA/account-invoicing/pull/1901


GuillemCForgeFlow (approved):
makes sense to me 👍🏿

AaronHForgeFlow (approved):
Functional + code

JasminSForgeFlow (approved):
Code review

AlexPForgeFlow (approved):
LGTM!

PR #1875, opened 2025-01-06
[14.0][PERF] account_invoice_refund_link: translate post_init_hook into SQL

Ricardoalso:
The key changes include replacing the Python logic with a SQL query to handle the matching process more efficiently.

With a large database ( > 65 million account.move.line records) I faced a memory error when executing the post_init_hook.
By translating the python code to a SQL query it took me less than 10 minutes to execute it (postgresql V16.0)

### DIFF
--- a/account_invoice_refund_link/hooks.py
+++ b/account_invoice_refund_link/hooks.py
@@ -5,34 +5,51 @@
 from odoo import SUPERUSER_ID, api
 
 
-def match_origin_lines(refund):
-    """Try to match lines by product or by description."""
-    invoice = refund.reversed_entry_id
-    invoice_lines = invoice.invoice_line_ids
-    for refund_line in refund.invoice_line_ids:
-        for invoice_line in invoice_lines:
-            match = (
-                refund_line.product_id
-                and refund_line.product_id == invoice_line.product_id
-                or refund_line.name == invoice_line.name
-            )
-            if match:
-                invoice_lines -= invoice_line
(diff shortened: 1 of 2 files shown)


account_invoice_refund_link/hooks.py
```diff
@@ -5,34 +5,51 @@
...
+    WHERE
+        (rl.product_id IS NOT NULL AND rl.product_id = il.product_id)
+        OR rl.name = il.name
+)
+UPDATE account_move_line aml
+SET origin_line_id = ml.invoice_line_id
+FROM matched_lines ml
+WHERE
+    aml.id = ml.refund_line_id
+    AND ml.match_priority = 1;
```
StefanRijnhart, line 53: If I understand correctly, you're only matching lines by product. Is it intended to ignore matches by aml name?

Ricardoalso: The match by product and aml name is made on line 45 & 46

StefanRijnhart: Yes, but aren't these assigned a value for `match_priority` of 2 in line 36? Where as you are filtering on match_priority 1 here.

Ricardoalso: Got it; I am probably missing an OR in here 
 `AND (ml.match_priority = 1 OR ml.match_priority = 2)` 

I will take some additional time when possible to review the query

PR #1876, opened 2025-01-08
[16.0][IMP] account_invoice_mass_sending: Add EDI documents to mass sending email

sergiobstoj:
Add edi attachments to mass sending email, just like normal sending

### DIFF
--- a/account_invoice_mass_sending/models/account_move.py
+++ b/account_invoice_mass_sending/models/account_move.py
@@ -1,7 +1,7 @@
 # Copyright 2019 ACSONE SA/NV
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import _, fields, models
+from odoo import Command, _, fields, models
 
 
 class AccountInvoice(models.Model):
@@ -58,6 +58,7 @@ class AccountInvoice(models.Model):
                 "is_email": True,
                 "template_id": template.id,
                 "composition_mode": "comment",
+                "attachment_ids": [Command.set(self._get_edi_documents())],
             }
         )
         wiz.onchange_template_id()
@@ -67,3 +68,13 @@ class AccountInvoice(models.Model):
             }
         )
         return wiz.send_and_print_action()
+
+    def _get_edi_documents(self):
+        self.ensure_one()
+        attachment_ids = []
+        if getattr(self, "edi_document_ids", False):
+            for edi in self.edi_document_ids:
+                attachment = self._get_edi_attachment(edi.edi_format_id)
+                if attachment:
+                    attachment_ids.append(attachment.id)
+        return attachment_ids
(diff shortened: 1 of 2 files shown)


account_invoice_mass_sending/models/account_move.py
```diff
@@ -67,3 +68,14 @@ def _send_invoice_individually(self, template=None):
             }
         )
         return wiz.send_and_print_action()
+
+    def _get_edi_documents(self):
+        self.ensure_one()
+        attachment_ids = []
+        if getattr(self, "edi_document_ids", False):
+            for edi in self.edi_document_ids:
+                if self._get_edi_attachment(edi.edi_format_id):
```
aritzolea: Could you refactor a bit the code?

```
attachment = self._get_edi_attachment(edi.edi_format_id)
if attachment:
    attachment_ids.append(attachment.id)
```

sergiobstoj: Done, thanks.

aritzolea (approved):

sergiobstoj:
@jguenat is this ok for you to merge?

jguenat:
Hello @sergiobstoj ,
It seems the EDI, at least factur-x, is already embedded in the pdf file when mass sending.
if I understand correctly you are also adding it in attachment of the email.
correct ?

sergiobstoj:
@jguenat Yes, that's correct. With normal sending it already attached it. So in order to be consistent, it will also be attached on mass sending.

jguenat:
@sergiobstoj 
I just tested it

- With Normal sending you have a pdf with embedded factur-x. Only the pdf in the mail attachment.
- With Mass sending you have a pdf with embedded factur-x. Only the pdf in the mail attachment.

I don't see any difference between the two.
What you are doing here is adding the factur-x document as attachment of the email. Which is interesting but maybe here is not the place to do it.

sergiobstoj:
Maybe is miss to give more context.

factur-x document is generated as attachment on invoice (and stored in `edi_document_ids`) while having:

image

> UBL Bis 3 is only available for companies which country is present in the [EAS list](https://docs.peppol.eu/poacc/billing/3.0/codelist/eas/).

From there, that file is attached to normal sending. And with this PR, in mass sending.

jguenat:
Okay I see you are right.
This module's intend is to mimic the user action when he sends an invoice manually  -> Open the wizard and click on send.
This should make it as much compatible with localization as possible.
So yes if the country permits it and the settings are right, the ubl_bis3 xml file should be there.

The issue we have here is an unforeseen side effect introduced here #1839 
before this modification the ubl_bis3 was there. 

So we should correct it.

jguenat:
hello @sergio-teruel 
Seems the modification made to avoid notifications causes issue with edi attachments.
What do you think ?

sergio-teruel:
@jguenat Are you sure??
@sergiobstoj Can you check if @jguenat are right, please??

sergiobstoj:
Hi @sergio-teruel 
@jguenat is right, just checked. Sorry for the delay.
I'll open a PR that fixes this problem and also this issue: https://github.com/OCA/account-invoicing/issues/2028

sergiobstoj:
Closed, as it's solved in https://github.com/OCA/account-invoicing/pull/2052

PR #1763, opened 2024-07-14
[17.0][MIG] account_billing: Migration to 17.0

antonioburic:
based on https://github.com/OCA/account-invoicing/pull/1644

PR #1879, opened 2025-01-14
[18.0][MIG] account_billing: Migration to 18.0

kevinkhao:
based on https://github.com/OCA/account-invoicing/pull/1763


account_billing/views/account_billing_views.xml
```diff
@@ -0,0 +1,302 @@
...
+                                        nolabel="1"
+                                        readonly="state != 'draft'"
+                                    />
+                                </group>
+                                <group />
+                            </group>
+                        </page>
+                    </notebook>
+                </sheet>
+                <div class="oe_chatter" />
```
AungKoKoLin1997, line 253: You should use like this.
```suggestion
                <chatter/>
```

account_billing/views/account_billing_views.xml
```diff
@@ -0,0 +1,302 @@
...
+                            invisible="state == 'draft'"
+                        />
+                    </h1>
+                    <group>
+                        <group>
+                            <field name="partner_id" readonly="state != 'draft'" />
+                            <field
+                                name="bill_type"
+                                invisible="1"
+                                readonly="state != 'draft'"
```
AungKoKoLin1997, line 177: Why invisible field needs to add `readonly` attribute.

PR #1871, opened 2024-12-31
[17.0][FIX] account_invoice_section_sale_order: FWD port and fix test-case

LucasTran380381:
### FWD ports
https://github.com/OCA/account-invoicing/pull/1774
 https://github.com/OCA/account-invoicing/pull/1862
 https://github.com/OCA/account-invoicing/pull/1770
 
 ### Fix test-case
 `test_create_invoice_with_currency` change compare 
 ```
self.assertEqual(invoices.amount_total, 80) => self.assertEqual(invoices.amount_untaxed, 80)
 ```

PR #1880, opened 2025-01-14
[FIX] account_move_tier_validation: do not overwrite line values

len-foss:
Edit a new journal entry with lines in currency X different from company currency Y. Say you have a line of value (amount_currency) N, which is converted to Y(N) (debit/credit). Then `amount_total` is also N. Because the `readonly="not(review_ids)"` was added, `amount_total` is dirty and sent at save. Therefore, Odoo uses it and sets the debit/credit values of the lines to N (instead of Y(N), the converted value). It's as if it's ignoring the currency conversion (note that the values are correct through onchange before the save).

To avoid this, exclude the field from being modified in `get_view`.

### DIFF
--- a/account_move_tier_validation/models/account_move.py
+++ b/account_move_tier_validation/models/account_move.py
@@ -22,6 +22,14 @@ class AccountMove(models.Model):
     def _get_under_validation_exceptions(self):
         return super()._get_under_validation_exceptions() + ["needed_terms_dirty"]
 
+    def _get_validation_exceptions(self, extra_domain=None, add_base_exceptions=True):
+        res = super()._get_validation_exceptions(extra_domain, add_base_exceptions)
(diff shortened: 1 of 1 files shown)


len-foss:
It looks like the branch is borked? It's module `account_invoice_section_sale_order` that fails before the tests are run for the modified module.

StefanRijnhart (approved):
Thanks, makes sense!
Tests of this module are running fine, but indeed those of account_invoice_section_sale_order are broken because of https://github.com/odoo/odoo/commit/1443e5503c7396512a. A fix is already proposed in https://github.com/OCA/account-invoicing/pull/1871.

Shide (approved):

StefanRijnhart:
I'm now proposing the test fix separately in https://github.com/OCA/account-invoicing/pull/1888

PR #1881, opened 2025-01-16
[16.0][ADD] partner_invoicing_mode_at_shipping_by_partner_by_carrier

rousseldenis:
Glue module between `partner_invoicing_mode_at_shipping` and `stock_picking_group_by_partner_by_carrier` modules.

This to avoid no invoice creation after:

- https://github.com/OCA/stock-logistics-workflow/pull/1789

is merged.


jbaudoux (changes requested):
Related fix is not the right approach.
So this module should not be necessary

nihelgabsi-acsone (approved):
Code review

PR #1887, opened 2025-01-22
[17.0][IMP]: account_invoice_section_sale_order add option to always …

victorvermot:
We want to be able to create sections even if only one so is selected when creating the invoice

### DIFF
--- a/account_invoice_section_sale_order/models/res_company.py
+++ b/account_invoice_section_sale_order/models/res_company.py
@@ -21,3 +21,8 @@ class ResCompany(models.Model):
         default="sale_order",
         required=True,
     )
+
+    always_create_invoice_section = fields.Boolean(
+        help="Defines when to create sections",
+        default=False,
+    )
--- a/account_invoice_section_sale_order/views/res_config_settings.xml
+++ b/account_invoice_section_sale_order/views/res_config_settings.xml
@@ -28,6 +28,14 @@
                         />
                         <field name="invoice_section_name_scheme" />
                     </div>
+                    <div class="row">
+                        <label
+                            for="always_create_invoice_section"
+                            class="col-lg-4 o_light_label"
+                            string="Create section even if an invoice is created from a single sale order"
+                        />
+                        <field name="always_create_invoice_section" />
+                    </div>
                 </setting>
             </xpath>
         </field>
--- a/account_invoice_section_sale_order/models/res_config_settings.py
+++ b/account_invoice_section_sale_order/models/res_config_settings.py
@@ -16,3 +16,8 @@ class ResConfigSettings(models.TransientModel):
         readonly=False,
         required=True,
     )
+
+    always_create_invoice_section = fields.Boolean(
+        related="company_id.always_create_invoice_section",
+        readonly=False,
+    )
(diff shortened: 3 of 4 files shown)


StefanRijnhart:
Tests are fixed in https://github.com/OCA/account-invoicing/pull/1888

victorvermot:
@StefanRijnhart Thanks a lot

StefanRijnhart:
@victorvermot A review of that PR would be appreciated.

account_invoice_section_sale_order/models/res_company.py
```diff
@@ -21,3 +21,13 @@ class ResCompany(models.Model):
         default="sale_order",
         required=True,
     )
+
+    invoice_section_treshold = fields.Selection(
+        [
+            ("1", "Create section if multiple sale orders are selected"),
```
StefanRijnhart: It seems to me this could be a boolean. Or are there additional options imaginable?

victorvermot: Did not want to move away from the current behavior -> https://github.com/OCA/account-invoicing/blob/a8c70203b456180db792980f8628ab2aed122b53/account_invoice_section_sale_order/models/sale_order.py#L23
cc @simahawk

StefanRijnhart: TBH I would find that section more readable if it said something like
```
            if (
                not invoice.company_id.account_invoice_section_sale_order_single_section
                and len(invoice.line_ids.mapped(invoice.line_ids._get_section_grouping()))
                == 1
            ):
 ```

**Comment by MiquelRForgeFlow** [comment]
Fix was merged, you can rebase.

File: account_invoice_section_sale_order/models/res_company.py (hunk 1 above)
Comment (victorvermot) [reply account_invoice_section_sale_order/models/res_company.py]: I tend to agree with you, I went into this direction after discussing with Simone, would be interesting to get his input

Comment (simahawk) [reply account_invoice_section_sale_order/models/res_company.py]: IMO you can go ahead w/ the boolean. Another option is to have an integer, instead of a selection, and simply do
```
            if (
                len(invoice.line_ids.mapped(invoice.line_ids._get_section_grouping()))
                == invoice.company_id.sale_order_invoice_section_treshold
            ):
```
Although I doubt you want to do it w/ 2 or 3 orders instead of 1.

File: account_invoice_section_sale_order/models/res_company.py [hunk 2]
```diff
@@ -21,3 +21,9 @@ class ResCompany(models.Model):
         default="sale_order",
         required=True,
     )
+
+    always_create_invoice_section = fields.Boolean(
+        help="Defines when to create sections",
+        default=False,
+        required=True,
```
StefanRijnhart: Requiredness is not meaningful in the case of boolean fields.

account_invoice_section_sale_order/views/res_config_settings.xml
```diff
@@ -28,6 +28,14 @@
                         />
                         <field name="invoice_section_name_scheme" />
                     </div>
+                    <div class="row">
+                        <label
+                            for="always_create_invoice_section"
+                            class="col-lg-4 o_light_label"
+                            string="When to create sections"
```
StefanRijnhart: Can you clarify in the UI that the option concerns invoices for a single sale order?

StefanRijnhart (approved):
Thanks!

simahawk (approved):

PR #1717, opened 2024-04-29
[17.0][MIG] account_invoice_force_number

David-Luis-Mora:

PR #1890, opened 2025-01-23
[MIG] account_invoice_force_number: Migration to 17.0

luc-demeyer:


luc-demeyer:
This PR is based upon https://github.com/OCA/account-invoicing/pull/1717
I have rebased that PR and fixed the pre-commit error.

houssine78 (approved):

AaronHForgeFlow (commented):
Is it intended to change the sequence (ir.sequence) when Forcing a number? See those cases:

In runboat next number was INV/2025/00005 but I forced it to be KKK/123. Then duplicated the original invoice again, and in the posting the number was set to KKK/124. Is that intended?

Also, I duplicated the invoice with number KKK/124 and force the number to be INV/2025/00006 and it was good. However, when duplicating that one, and not forcing any number, the number set was KKK/125. It is like the sequence was changed once only.

andreampiovesana (approved):
LGTM

PR #1894, opened 2025-01-27
[16.0][FW][IMP] account_invoice_show_currency_rate: use absolute balance to get rate on credit amounts

JordiMForgeFlow:
FW of https://github.com/OCA/account-invoicing/pull/1838

- Odoo computes the currency rate on journal items (https://github.com/odoo/odoo/blob/e3d54695862f73d9be6c35121def57ed641a1a45/addons/account/models/account_move_line.py#L133). However, the computed rate is based on the rate configured in the system. It could be the case for manually created entries that the user has manually modified the amount in currency, so this module ensures that the computed rate is based on the actual currency amount applied.

### DIFF
--- a/account_invoice_show_currency_rate/models/account_move.py
+++ b/account_invoice_show_currency_rate/models/account_move.py
@@ -20,6 +20,7 @@ class AccountMove(models.Model):
         "state",
         "date",
         "line_ids.amount_currency",
+        "line_ids.balance",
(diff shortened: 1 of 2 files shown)


MiquelRForgeFlow (approved):

AaronHForgeFlow (approved):
Thank you for the FW port!

PR #1901, opened 2025-02-04
[16.0][FIX] account_invoice_show_currency_rate: do not round currency rate

JordiMForgeFlow:
Currency rates are not rounded in Odoo. Applying the rounding causes the currency rate of the line to be different when the move is posted.


JordiMForgeFlow:
CC @AaronHForgeFlow

AaronHForgeFlow (approved):
Tested

victoralmau (approved):

PR #1903, opened 2025-02-07
[18][MIG] sale_line_refund_to_invoice_qty

florian-dacosta:
This migration from v16 include https://github.com/OCA/account-invoicing/pull/1748 that was not ported from 14 to 16.

I also made 2  changes : 
- Hide the `sale_qty_to_reinvoice` field if the refund is not linked to an invoice.
It was quite strange to see this "Sale Qty To Reinvoice" checked for refund created manually or from an other source (RMA...). And it does nothing if no sale is linked.
- Take the company configuration in the default of the field. It was also confusin IMO to have the field checked by default when it was deactivated on the company configuration

@PicchiSeba Could you have a look at the change on default to see if you agree since it kind of completes your addition ?

@JordiMForgeFlow @LoisRForgeFlow @rousseldenis It could interests you I guess


rousseldenis:
> * It was quite strange to see this "Sale Qty To Reinvoice" checked for refund created manually or from an other source (RMA...). And it does nothing if no sale is linked.

@florian-dacosta That's because this is the standard Odoo behavior. It is intended to not change that.

If people don't want to reinvoice, uncheck the box

florian-dacosta:
> > * It was quite strange to see this "Sale Qty To Reinvoice" checked for refund created manually or from an other source (RMA...). And it does nothing if no sale is linked.
> 
> @florian-dacosta That's because this is the standard Odoo behavior. It is intended to not change that.
> 
> If people don't want to reinvoice, uncheck the box

I know it is standard behavior, but this behavior exists only for refund linked to sale orders.
So if your refund is not linked to a sale order, this box does not do anything. Then, it is confusing for user to have some option checked, that say it will be re-invoiced when it does not has any effect anywhere right ?
Does it seem right/do you agree to hide it anyway in this case ?

sale_line_refund_to_invoice_qty/models/account.py
```diff
@@ -0,0 +1,26 @@
+# Copyright 2021 ForgeFlow (http://www.forgeflow.com)
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
+from odoo import api, fields, models
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    @api.returns("self", lambda value: value.id)
```
glitchov: This decorator here causes a js bug when duplicating an invoice

glitchov (changes requested):

sale_line_refund_to_invoice_qty/models/account.py
JordiMForgeFlow: yes, same as was done in https://github.com/OCA/sale-workflow/pull/3723

LoisRForgeFlow: @florian-dacosta could you attend this?

florian-dacosta: I missed the comment, thanks for the reminder

JordiMForgeFlow (approved):

LoisRForgeFlow (approved):
Thanks!

AaronHForgeFlow (approved):
Code LGTM

AaronHForgeFlow:
> This migration from v16 include https://github.com/OCA/account-invoicing/pull/1748 that was not ported from 14 to 16.

If this is still not ported to v16 could you please create the PR? Thank you.

BhaveshHeliconia (approved):
Functional review LGTM!

glitchov (approved):
LGTM

PR #1905, opened 2025-02-07
[IMP] stock_picking_invoicing: set reversed_entry in return pick invoicing

DiegoParadeda:
This PR automatically references a picking's invoice in it's return invoice using the field `reversed_entry_id`

```
Stock Move  --->  Invoice 1
     |
     v
Return Stock Move  --->  Invoice 2 (references Invoice 1)


e.g. invoice_2.reversed_entry_id = invoice_1
```

### DIFF
--- a/stock_picking_invoicing/models/stock_picking.py
+++ b/stock_picking_invoicing/models/stock_picking.py
@@ -45,3 +45,11 @@ class StockPicking(models.Model):
         if any(m.invoice_state == "2binvoiced" for m in self.mapped("move_lines")):
             self.write({"invoice_state": "2binvoiced"})
         return super().action_assign()
+
+    def get_return_origin_invoice_ids(self):
+        """
+        Override this method if you need to set a specific invoice to return
+        :return: invoices
+        """
+        source_picking_id = self.mapped("move_lines.origin_returned_move_id.picking_id")
+        return source_picking_id.mapped("invoice_ids")
--- a/stock_picking_invoicing/wizards/stock_invoice_onshipping.py
+++ b/stock_picking_invoicing/wizards/stock_invoice_onshipping.py
@@ -528,12 +528,26 @@ class StockInvoiceOnshipping(models.TransientModel):
         invoices = self.env["account.move"].browse()
         for pickings in pick_list:
             moves = pickings.mapped("move_lines")
+            # Origin moves are set when invoicing a return picking
+            origin_moves = moves.mapped("origin_returned_move_id")
+            origin_account_move_id = False
+            if origin_moves:
+                # Core does not suport multi returns:
+                # pick first and set as invoice.reversed_entry_id
+                # TODO: check if this is approach changes in next versions
+                if pickings.get_return_origin_invoice_ids():
+                    origin_account_move_id = pickings.get_return_origin_invoice_ids()[
+                        0
+                    ].id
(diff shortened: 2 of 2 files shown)


mbcosta:
Thank you @DiegoParadeda for your PR, some points about:

* Why not infom the field **reversed_entry_id** inside the method [def _build_invoice_values_from_pickings](https://github.com/OCA/account-invoicing/blob/14.0/stock_picking_invoicing/wizards/stock_invoice_onshipping.py#L349) ? It's seems the right place even to be inherit

* Should we consider include the **move_lines.origin_returned_move_id.picking_id** as a **KEY** in [def _get_picking_key](https://github.com/OCA/account-invoicing/blob/14.0/stock_picking_invoicing/wizards/stock_invoice_onshipping.py#L302) because in this case there are no Grouped Invoice as **reversed_entry_id** is a **Many2One** ?

*  Are you sure about the necessity to change the **Origin Invoice with def get_return_origin_invoice_ids**? It's ssems unnecessary or can you give a example about it

* You can also include a check in the tests about this case to avoid regressions after the line https://github.com/OCA/account-invoicing/blob/14.0/stock_picking_invoicing/tests/test_picking_invoicing.py#L858

       self.assertEqual(
            invoice_devolution.reversed_entry_id, invoice
       )

mbcosta:
@DiegoParadeda just a example about the suggested changes, this seems work as expect but let me know if there something about the problem that I didn't see or understand

```bash
--- a/stock_picking_invoicing/wizards/stock_invoice_onshipping.py
+++ b/stock_picking_invoicing/wizards/stock_invoice_onshipping.py
@@ -316,6 +316,13 @@ class StockInvoiceOnshipping(models.TransientModel):
                 picking.picking_type_id,
                 picking.partner_id,
             )
+            # Refund Case should not be Group
+            refund_picking = picking.mapped(
+                "move_lines.origin_returned_move_id.picking_id"
+            )
+            if refund_picking:
+                key = key + (refund_picking,)
+        print('=========== KEY =====', key)
         return key


     def _group_pickings(self, pickings):
@@ -383,6 +390,17 @@ class StockInvoiceOnshipping(models.TransientModel):
                 "picking_ids": [(4, p.id, False) for p in pickings],
             }
         )
+        refund_picking = picking.mapped(
+                "move_lines.origin_returned_move_id.picking_id"
+        )
+        if refund_picking:
+            values["reversed_entry_id"] = refund_picking.mapped(
+                "invoice_ids"
+            )[0].id
+        print('======== Build Values Refund Picking ===',
+              values.get('reversed_entry_id')
+        )
+
 
         invoice, values = self._simulate_invoice_onchange(values)


--- a/stock_picking_invoicing/tests/test_picking_invoicing.py
+++ b/stock_picking_invoicing/tests/test_picking_invoicing.py
@@ -856,6 +856,13 @@ class TestPickingInvoicing(SavepointCase):
         wizard.action_generate()
         domain = [("picking_ids", "=", picking_devolution.id)]
         invoice_devolution = self.invoice_model.search(domain)
+        print('========= Invoice Devolution ===',
+              invoice,
+              invoice_devolution.reversed_entry_id
+        )
+        self.assertEqual(
+            invoice_devolution.reversed_entry_id, invoice
+        )
         # Confirm Return Invoice

```

PR #1907, opened 2025-02-11
[FIX] account_invoice_tax_note: Fix test

etobella:
For example, if we have install `l10n_es` the test raised an error of multicompany issues.

With this change, we avoid it.

### DIFF
--- a/account_invoice_tax_note/tests/test_account_invoice.py
+++ b/account_invoice_tax_note/tests/test_account_invoice.py
@@ -49,7 +49,8 @@ class TestAccountInvoice(TransactionCase):
         self.product1.write({"taxes_id": [(6, False, self.tax1.ids)]})
         self.product2.write({"taxes_id": [(6, False, self.tax2.ids)]})
         account = self.env["account.account"].search(
-            [("account_type", "=", "income")], limit=1
+            [("account_type", "=", "income"), ("company_id", "=", self.env.company.id)],
+            limit=1,
         )
         account.write({"tax_ids": [(4, self.tax1.id, False), (4, self.tax2.id, False)]})
         journal = self.env["account.journal"].create(


luisDIXMIT (approved):
LGTM!

account_invoice_tax_note/tests/test_account_invoice.py
```diff
@@ -48,8 +48,12 @@ def setUp(self):
...
         self.product1.write({"taxes_id": [(6, False, self.tax1.ids)]})
         self.product2.write({"taxes_id": [(6, False, self.tax2.ids)]})
-        account = self.env["account.account"].search(
-            [("account_type", "=", "income")], limit=1
+        account = self.env["account.account"].create(
+            {
+                "name": "account",
+                "code": "ACCOUNT",
+                "account_type": "income",
+            }
```
pedrobaeza: I think the fix should be:

```suggestion
        account = self.env["account.account"].search(
            [("account_type", "=", "income"), ("company_id", "=", self.env.company.id)], limit=1
        )
```

etobella: Sure, creating it ensured that it worked in any case (because we might have no localization installed), but your suggestion is less agressive :wink:

pedrobaeza: And avoids other problems: the account creation itself may fail, due to constraint, missing data...

AaronHForgeFlow (approved):

account_invoice_tax_note/tests/test_account_invoice.py
luisDIXMIT: @pedrobaeza We have followed your suggestions, tested them, and they work properly with l10n_es. Thanks!

pedrobaeza (approved):
/ocabot merge patch

PR #1919, opened 2025-02-20
[16.0][REF] sale_stock_picking_invoicing: Allow other modules inform Fields that should not be used from Sale 'prepare' methods, FWD 1906

mbcosta:
Foward Port https://github.com/OCA/account-invoicing/pull/1906

Included a method to allow other modules inform the Fields that should not be used from Sale 'prepare' methods, there are cases where Localization Projects or maybe custom modules need to remove some fields, for example to give priority to the fields came from stock.picking instead of from sale.order.

This change not affect the actual behavior just move parts of the code for this new method to be able inherit it.


rvalyi (approved):

rvalyi:
cc @renatonlima @marcelsavegnago @antoniospneto @DiegoParadeda @mileo

mileo (approved):

PR #1830, opened 2024-11-02
[18.0][MIG] account_move_tier_validation: Migration to 18.0

kevinkhao:
depends on https://github.com/OCA/server-ux/pull/966
EDIT: above PR superseded by https://github.com/OCA/server-ux/pull/992

PR #1921, opened 2025-02-21
[18.0][MIG]  account_move_tier_validation:Migration to 18

aarosales-spearhead:
Replace https://github.com/OCA/account-invoicing/pull/1830


account_move_tier_validation/tests/test_tier_validation.py
```diff
@@ -1,14 +1,21 @@
...
+from odoo.addons.base.tests.common import DISABLED_MAIL_CONTEXT
 
 
 @tagged("post_install", "-at_install")
-class TestAccountTierValidation(common.TransactionCase):
+class TestAccountTierValidation(TransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.env = cls.env(context=dict(cls.env.context, **DISABLED_MAIL_CONTEXT))
```
celm1990: This class can inherit from [BaseCommon](https://github.com/odoo/odoo/blob/6caa783991dad7272391ae8ea0e76f32eadc9401/odoo/addons/base/tests/common.py#L20), allowing `setUpClass` to be removed.

celm1990 (changes requested):
LGTM, just  a minor change

@xaviedoanhduy @pedrobaeza @bosd @StefanRijnhart could you please review this?

bosd (approved):
Functional test on runboat

account_move_tier_validation/tests/test_tier_validation.py
aarosales-spearhead: Done sr

account_move_tier_validation/tests/test_tier_validation.py
```diff
@@ -88,6 +88,4 @@ def test_03_move_post(self):
         with self.assertRaisesRegex(
             ValidationError, self.env._("You are not allowed to write those fields")
         ):
-            invoice._post()
-        # Calls _post method by passing context skip_validation_check set to True
-        invoice.action_post()
+            invoice.action_post()
```
StefanRijnhart: This change looks like it might be related to the test failure. Was it intended?

TheerayutEncoder:
Hi @aarosales-spearhead @StefanRijnhart 
I just fix test script from this PR [https://github.com/aarosales-spearhead/Arosales-18.0-mig-account_move_tier_validation/pull/1](https://github.com/aarosales-spearhead/Arosales-18.0-mig-account_move_tier_validation/pull/1)

Could you merge and rebase your commit?
It's not `[18.0][MIG] sale_tier_validation: Migration to version 18.0`.

StefanRijnhart:
Good to see tests green again! Can you squash all migration commits (and the fixing commit) so that we can check what has actually changed now?

aarosales-spearhead:
@StefanRijnhart

account_move_tier_validation/tests/test_tier_validation.py
```diff
@@ -66,25 +69,25 @@ def test_03_move_post(self):
...
         )
-        invoice.with_user(self.test_user_2.id).request_validation()
+        reviews = invoice.with_user(self.test_user_2.id).request_validation()
+        self.assertTrue(reviews)
+        with self.assertRaises(ValidationError):
+            invoice.action_post()
         invoice = invoice.with_user(self.test_user_1.id)
-        invoice.invalidate_model()
         invoice.validate_tier()
-        with self.assertRaisesRegex(
```
StefanRijnhart: You are removing a meaningful test here. Is it no longer necessary to pass skip_validation_check in `action_post` (in which case that override can be removed), or have the internals of `action_post` changed so that the test needs to be updated?

account_move_tier_validation/tests/test_tier_validation.py
```diff
@@ -66,25 +69,25 @@ def test_03_move_post(self):
...
+                "invoice_date_due": fields.Date.to_date("2024-01-01"),
                 "invoice_line_ids": [
                     (0, 0, {"product_id": product.id, "quantity": 1, "price_unit": 30})
                 ],
             }
         )
-        invoice.with_user(self.test_user_2.id).request_validation()
+        reviews = invoice.with_user(self.test_user_2.id).request_validation()
+        self.assertTrue(reviews)
+        with self.assertRaises(ValidationError):
```
StefanRijnhart: Please use assertRaisesRegex (plus savepoint) to give an indication which error occurs here.

TheerayutEncoder: Hi @StefanRijnhart 

Could we write test like this?
```
with self.assertRaisesRegex(ValidationError, "You cannot post an invoice that requires tier validation"):
            invoice.action_post()
```

PS. IMHO we do not need to skip_validation_check because it will skip the error so this ensure if user try to post invoice while it on validation process, it should raise some error.

@aarosales-spearhead 
Can you try with this?

StefanRijnhart: @TheerayutEncoder Yes, that's how assertRaisesRegex can be used. You can also replace some of the text with `.*`.

Your humble opinion aside :smile:, I just tested without passing the context and then `action_post()` fails on `You are not allowed to write those fields under validation.`. The original test code remains valid and should be restored.

bosd:
@aarosales-spearhead Can you please fix the tests?

celm1990 (approved):
@StefanRijnhart please update your review

StefanRijnhart (approved):
@celm1990 thanks for tidying this one up!

bosd:
Oops, This one appears to be broken rn. Just tested on runboat.
Create a tier validation for mr Demo on all invoices.
On creation of an invoice and clicking on the button  "request validation." 
There is an error:
image


Even mr Demo is not allowed to change state of that customer invoice.

PR #1934, opened 2025-03-04
[16.0][ADD] account_currency: Module added

geomer198:
Set the default currency for the company's invoices and vendor bills.


Bearnard21 (approved):
LGTM

account_currency/models/account_move.py
```diff
@@ -0,0 +1,20 @@
...
+    def _compute_currency_id(self):
+        result = super()._compute_currency_id()
+        company = self.env.company
+        for invoice in self.filtered(
+            lambda rec: rec.move_type in ["out_invoice", "in_invoice"]
+        ):
+            invoice.currency_id = (
+                company.default_account_currency_id or invoice.currency_id
+            )
+        return result
```
GabbasovDinar, line 20:
```suggestion
    @api.depends("journal_id", "statement_line_id")
    def _compute_currency_id(self):
        invoices = self.filtered(
            lambda rec: rec.move_type in ["out_invoice", "in_invoice"] and rec.company_id.default_account_currency_id
        )
        for invoice in invoices:
            invoices.currency_id = invoice.company_id.default_account_currency_id
        
        return super(AccountMove, (self-invoices))._compute_currency_id()
```

GabbasovDinar (changes requested):

account_currency/models/account_move.py
GabbasovDinar: I think your way is better, I mean you need call super before, then apply changes:
```python
res = super()
....
return res
```

GabbasovDinar (approved):

dreispt (approved):

Bearnard21:
@pedrobaeza Could you please merge the PR. Thank you

pedrobaeza:
About the module feature itself, I think this is already handled if you put a currency at journal level, isn't it?

PR #1840, opened 2024-11-12
[16.0] [REF] account_invoice_triple_discount Consolidate discount in std field (Follow-up of #1638)

legalsylvain:
Finish https://github.com/OCA/account-invoicing/pull/1638

Basically 
- **Get all the commits from the previous @grindtildeath PR.**

- **remove** this commit https://github.com/OCA/account-invoicing/pull/1638/commits/01df667c1c3fdae39671a5effc0cb54ef6700fd5 that was introducing incompatibility between ``account_invoice_triple_discount`` and ``account_invoice_global_discount`` that can works together in fact. 

- **Add** 
    - e03dba34: a simple commit to make ``account_global_discount`` working if ``account_invoice_triple_discount`` is installed. With that commit, the [remark of @pedrobaeza](https://github.com/OCA/account-invoicing/pull/1638#issuecomment-2342817865) should be answered.
    - 37c8e087: another commit to exclude mutually ``account_invoice_triple_discount`` and ``account_invoice_fixed_discount``. This commit is a fix in fact, as the exclusion [is written in the documentation](https://github.com/OCA/account-invoicing/tree/14.0/account_invoice_fixed_discount#installation), but the CI and the copier answers don't mention that exclusion.

Let's see if all CI is green.

PR #1938, opened 2025-03-05
[16.0][FIX] account invoice triple discount report after refactoring

legalsylvain:
- [FIX] account_invoice_triple_discount: do not hide column that are hidden, because display_discount is False
- [FIX] account_invoice_triple_discount: discounts fields was hidden, because was t-if display_discount (that is False).
- [IMP] account_invoice_triple_discount: display only discount column with not null discount

This is regression introduced by #1638 / #1840.

CC : @grindtildeath

### DIFF
--- a/account_invoice_triple_discount/report/invoice.xml
+++ b/account_invoice_triple_discount/report/invoice.xml
@@ -10,6 +10,18 @@
         </xpath>
         <xpath expr="//t[@t-set='display_discount']" position="after">
             <t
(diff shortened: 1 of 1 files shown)


grindtildeath (approved):
Looks legit :+1: 

(thanks for taking this over :hugs: )

luisDIXMIT (approved):
Tested on runboat, LGTM!

PR #1941, opened 2025-03-07
[REM] account_invoice_fiscal_position_update: Remove code from account_invoice_fiscal_position_update as is not longer required

rov-adhoc:


pedrobaeza:
Thanks for the PR. Instead of setting it as installable=False, which will lead to the same problem on instances with it installed, it's to "neutralize" the module removing any code, and rewriting the README to say that it's obsolete.

rov-adhoc:
> Thanks for the PR. Instead of setting it as installable=False, which will lead to the same problem on instances with it installed, it's to "neutralize" the module removing any code, and rewriting the README to say that it's obsolete.

Perfect! I just made the modification you requested.

StefanRijnhart (approved):
Ah yes, Pedro is right. I vaguely remembered that's what Odoo used it to deprecate obsolete payment modules at some point but that seems to have been purged in all versions now.

Thanks for the PR!

.pre-commit-config.yaml
```diff
@@ -1,6 +1,7 @@
 exclude: |
   (?x)
   # NOT INSTALLABLE ADDONS
+  ^account_invoice_fiscal_position_update/|
```
pedrobaeza: This entry should go out

rov-adhoc: done! It was added automatically by pre-commit

StefanRijnhart: @pedrobaeza OK for you now?

pedrobaeza (approved):
/ocabot merge major

rafaelbn:
Hello @rov-adhoc @pedrobaeza @StefanRijnhart @legalsylvain @alexis-via : 

In this moment Odoo dont's make the same as this module. This module chages automaticalle the taxes.

Actually Odoo warn with a button that should be clicked manually.

We would like to maintain the automatically function and not having to press "uptadte taxes each time"

Please, what do you think about this? Should be in our private? or is something interesting in OCA?

Thank you!
Regards,
Rafael

StefanRijnhart:
@rafaelbn No particular interest in that functionality from my side, I think.

alexis-via:
I agree that the native feature is not exactly the same. And it's strange that odoo implemented it via a button on which you need to click, whereas their made it automatic on sale.order and purchase.order. But, in order to avoid confusing users with a double implementation, I'm ok with the removal of the module and stick to the native implementation.

PR #1892, opened 2025-01-23
[16.0][FIX] stock_picking_invoicing: Invoice Type Map Dropshipping Case, FWD 1823

mbcosta:
Foward Port https://github.com/OCA/account-invoicing/pull/1823

The Invoice Type Map was missing for Dropshipping case, should solve Issue https://github.com/OCA/account-invoicing/issues/1790 , the module don't has direct dependecie of [stock_dropshipping module](https://github.com/OCA/OCB/tree/15.0/addons/stock_dropshipping)

cc @rvalyi @renatonlima @anpartner

PR #1942, opened 2025-03-07
[18.0][MIG] stock_picking_invoicing: Migration to 18.0

sergioM-S73:
Migration to 18.0

Add fix: 
https://github.com/OCA/account-invoicing/pull/2048
https://github.com/OCA/account-invoicing/pull/1819
https://github.com/OCA/account-invoicing/pull/1892


ppyczko:
Hi @sergioM-S73, could you please rebase so we can relaunch runboat for testing? Thanks!

ppyczko:
Thanks for the work @sergioM-S73, tested and everything works well overall.
The only suggestion is to consider including the changes proposed here:
https://github.com/OCA/account-invoicing/pull/2048

Currently, when invoicing a picking, its state remains "To Be Invoiced" instead of being updated to "Invoiced".

ppyczko (approved):
LGTM!

ppyczko:
Regarding the failing test in the pipeline, I’ve tried to reproduce the issue locally but couldn’t — all tests pass on my machine.

To investigate further, I used this draft PR: https://github.com/OCA/account-invoicing/pull/2051. Through debugging, I noticed that after calling [button_draft()](https://github.com/OCA/account-invoicing/blob/18.0/account_invoice_show_currency_rate/tests/test_account_move.py#L93), the `invoice_currency_rate` field was not being recomputed as expected.

Eventually, I got it to work. Here's what I believe is happening:
- Both `account_invoice_show_currency_rate` and `account_manual_currency` override the `_compute_invoice_currency_rate` method:
  - [account_invoice_show_currency_rate](https://github.com/OCA/account-invoicing/blob/18.0/account_invoice_show_currency_rate/models/account_move.py#L19)
  - [account_manual_currency](https://github.com/OCA/account-invoicing/blob/18.0/account_manual_currency/models/account_move.py#L119)

- However, these two modules are independent and do not declare a dependency on each other.
- Because of that, the order in which they're installed affects which override takes precedence. If `account_invoice_show_currency_rate` is loaded first, its logic can be silently overridden by `account_manual_currency`, without merging both behaviors.

This would explain why, in the pipeline, when executing `button_draft()` the compute is silently skipped because `state` is no longer a dependency in `account_manual_currency`'s version of the compute.

To confirm this, I tried triggering the compute manually ([check here](https://github.com/OCA/account-invoicing/pull/2051/commits/7aca673648bba9a4120f48d866cca3b3bce62449)) by modifying a common dependency field (invoice_date), and the pipeline tests passed.

Does this explanation make sense? I'd appreciate your input on how to proceed. Thank you!

cc: @pedrobaeza @victoralmau @JordiMForgeFlow

pedrobaeza:
As said other times, for me this module deviates too much the standard flows, so I don't use it in any customer.

ppyczko:
Thanks for your feedback, @pedrobaeza! Just to clarify, the `stock_picking_invoicing` module does not depend on either `account_manual_currency` or `account_invoice_show_currency_rate`.

The test errors I mentioned actually stem from those two modules, not from `stock_picking_invoicing`. My main concern is whether these issues related to those modules need to be fixed within this PR before it can eventually be merged, once it has the required reviews.

rvalyi:
@sergioM-S73 thank you for porting this module. Could just please cherry pick this important bugfix from 16.0:

- https://github.com/OCA/account-invoicing/pull/1819
- https://github.com/OCA/account-invoicing/pull/1892
- https://github.com/OCA/account-invoicing/pull/2048 (as commented by @ppyczko https://github.com/OCA/account-invoicing/pull/1942#issuecomment-3056715904 )

@pedrobaeza yes it differs from the "standard flow". But I believe this is very useful in industries or wholesalers who deal with lot's of stock moves where entries and deliveries are often fractioned along the time. 

Moreover, in countries that are federations such as Brazil it is also often mandatory. Because moving goods (even with no sales or purchases) in federations like Brazil is a bit like moving goods between different countries in Europe where you should create intrastat documents (so the federal tax accounting can be done). In countries like Brazil these intrastat documents are pretty much kind of regular invoices (simples remessa) that even plug into the usual electronic accounting. But I mean chances are this modules is or will be useful in other federations.

pedrobaeza:
Yeah, Raphäel, we have talked about that other times, and that's why I don't block it, but not recommend it either.

eugenios73:
Hello @ppyczko @rvalyi and @pedrobaeza  I've applied the mentioned fixes and added a small test for _get_custom_move_fields() in stock_rule. I've also rebased the branch. Could you please review it?

Add fixes:

- https://github.com/OCA/account-invoicing/pull/1819
- https://github.com/OCA/account-invoicing/pull/1892

ppyczko (approved):
LGTM! 👍

BernatObrador (approved):
LGTM! Functional Review

mpascuall (approved):
LGTM!

eugenios73:
@pedrobaeza Can this PR be merged or does it need further review? What do you think?

rvalyi (approved):
LGTM, sorry for the delay.

PR #1946, opened 2025-03-17
[16.0][ADD] sale_invoice_date_from_picking

AungKoKoLin1997:
This module sets the invoice date based on the date or datetime field configured in the 'Picking Date Field for Invoice Date' setting. If the selected field has no value, the system will fallback to using the `date_done` of the latest completed picking as the invoice date.

@qrtl QT5227


sale_invoice_date_from_picking_actual_date/models/res_company.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2025 Quartile (https://www.quartile.co)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    use_picking_actual_date_for_invoice = fields.Boolean()
```
yostashiro: It would be better to let the admin select a date field from `stock.picking` instead. We can remove the dependency on `stock_move_actual_date` that way.

AungKoKoLin1997: It is a good idea!

sale_invoice_date_from_picking/models/sale_order.py
```diff
@@ -0,0 +1,32 @@
...
+        if not picking:
+            return invoice_vals
+        date_value = getattr(picking, date_field.name, False)
+        if date_field.ttype == "date" and date_value:
+            invoice_vals["invoice_date"] = date_value
+        else:
+            invoice_vals["invoice_date"] = fields.Datetime.context_timestamp(
+                self.with_context(tz=self.company_id.partner_id.tz),
+                date_value or picking.date_done,
+            )
```
yostashiro: To be slightly safer.
```suggestion
        date_value = picking[date_field.name] or picking.date_done
        if date_field.ttype == "date":
            invoice_vals["invoice_date"] = value
        else:
            tz = self.company_id.partner_id.tz or self.env.user.tz
            invoice_vals["invoice_date"] = fields.Datetime.context_timestamp(self.with_context(tz=tz), date_value).date()
```

sale_invoice_date_from_picking/models/sale_order.py
```diff
@@ -0,0 +1,32 @@
...
+        return self.picking_ids.filtered(lambda p: p.state == "done").sorted(
+            key=lambda p: p.date_done, reverse=True
+        )[:1]
+
+    def _prepare_invoice(self):
+        invoice_vals = super()._prepare_invoice()
+        date_field = self.company_id.picking_date_field_for_invoice_date
+        if not date_field:
+            return invoice_vals
+        picking = self._get_picking_for_actual_date()
```
yostashiro:
```suggestion
        picking = self._get_last_completed_picking()
```

sale_invoice_date_from_picking/views/res_config_settings_views.xml
```diff
@@ -0,0 +1,27 @@
...
+        <field name="model">res.config.settings</field>
+        <field name="inherit_id" ref="sale.res_config_settings_view_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//div[@id='proforma_configuration']" position="after">
+                <div class="col-12 col-lg-6 o_setting_box">
+                    <div class="o_setting_right_pane">
+                        <label for="picking_date_field_for_invoice_date" />
+                        <div class="text-muted">
+                            Select the date or datetime field from the picking.
+                            The value of this field from the last completed picking associated with the sales order will be used as the invoice date.
```
yostashiro:
```suggestion
                            If set, the value of this field or the effective date of the last completed picking associated with the sales order will be used as the invoice date.
```

yostashiro (changes requested):

sale_invoice_date_from_picking/__manifest__.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2025 Quartile (https://www.quartile.co)
+# Licence AGPL-3.0 or later (https://www.gnu.org/licenses/agpl-3.0).
```
kanda999:
```suggestion
# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
```

sale_invoice_date_from_picking/models/sale_order.py
```diff
@@ -0,0 +1,32 @@
+# Copyright 2025 Quartile (https://www.quartile.co)
+# Licence AGPL-3.0 or later (https://www.gnu.org/licenses/agpl-3.0).
```
kanda999:
```suggestion
# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
```

sale_invoice_date_from_picking/models/sale_order.py
AungKoKoLin1997: As we talked, this will not respect the time of date_value if we choose date type field in the setting(eg.actual_date) and there is no value for this field in picking.

yostashiro: @AungKoKoLin1997 We still need to let it fall back on user's timezone when company timezone is not available. Please take care of that.

yostashiro (commented):
@AungKoKoLin1997 Please add USAGE.md and CONTEXT.md, or it can be a bit hard to understand how/when the module should be used.

sale_invoice_date_from_picking/tests/test_sale_invoice_date_from_picking.py
```diff
@@ -0,0 +1,73 @@
+# Copyright 2025 Quartile (https://www.quartile.co)
+# Licence AGPL-3.0 or later (https://www.gnu.org/licenses/agpl-3.0).
```
kanda999:
```suggestion
# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
```

sale_invoice_date_from_picking/readme/CONTEXT.md
yostashiro, line 1: Can we add the following information please.

---
One example of such an implementation case is as follows:
- Install OCA modules sale_automatic_wofklow and stock_move_actual_date, along with this module.
- Select the actual date field as the Picking Date Field for Invoice Date.
- Configure automatic workflow domain so that the delivery is validated and the invoice is created when the date set in the actual date has arrived.

AungKoKoLin1997: Updated.

yostashiro (approved):
LGTM.

kanda999 (approved):
Functional and code review.

Ryoko04 (approved):
Functional review on runboat. LGTM.

AungKoKoLin1997:
@oca/account-invoicing-maintainers
Could we get this PR merged?

rvalyi (approved):
LGTM

rvalyi:
There is a failure in the last Runboat but it seems completely unrelated. Let's see if we can merge it.

/ocabot merge nobump

PR #1949, opened 2025-03-20
[18.0][MIG] account_invoice_fixed_discount: Migration to 18.0

MohamedOsman7:
This is a new PR since no further work was done on the other one (https://github.com/OCA/account-invoicing/pull/1847)


JessBrandl (approved):
Code LGMT, tested in runboat and works as intended.

CRogos (approved):
code + tested LGTM

There is an edge case when the unit price is 0 and you give $5.00 discount. You could argument that the resulting unit price should be %-5.00.
But because the price calculation is always using the discount % value, this cannot be calculated in this case, because of a divide by 0.

CRogos:
@pedrobaeza could you merge?

PR #1952, opened 2025-03-20
[15.0][OU-FIX] account_invoice_supplier_self_invoice: Fix Migration script

etobella:
The scripts fail if we migrate from 14.0


AaronHForgeFlow (approved):
:+1:

etobella:
The failing of the tests is completly unrelated as 15 is failing :cry:

AaronHForgeFlow:
I am checking, the error is in the test of the module account_invoice_anglo_saxon_no_cogs_deferral, it must be in some conflict with other module because by itself tests won't fail.

AaronHForgeFlow:
I think this should fix it: https://github.com/OCA/account-invoicing/pull/1953 let's see :)

luisDIXMIT (approved):
LGTM!

PR #1960, opened 2025-03-28
[18.0][MIG] account_invoice_check_total: Migration to 18.0

jdidderen:


jbaudoux (approved):
Tested

account_invoice_check_total/models/account_move.py
```diff
@@ -0,0 +1,62 @@
...
+                and inv.move_type in ("in_invoice", "in_refund")
+                and float_compare(
+                    inv.check_total,
+                    inv.amount_total,
+                    precision_rounding=inv.currency_id.rounding,
+                )
+                != 0
+            ):
+                raise ValidationError(
+                    _(
```
AaronHForgeFlow:
```suggestion
                    self.env._(
```

AaronHForgeFlow (approved):
LGTM

account_invoice_check_total/models/account_move.py
jdidderen: Done 👍

ThomasBinsfeld (approved):

matteotognini (approved):
LGTM! Merge?

jbaudoux:
Is this module still of any use? It's now a standard odoo feature you can enable in the settings (quick edit mode for supplier invoices)

jdidderen:
@jbaudoux I know, but I’m a bit puzzled about this, because the quick edit mode isn’t limited to that feature.
In our case, this module is mainly used by customers whose administrative staff have limited accounting knowledge and are only responsible for encoding invoices. The module acts as a safeguard, but, for example, in quick edit mode the bill date is automatically filled in which isn’t always desired. Some customers prefer it to remain empty to avoid having a default date applied. So, enabling quick edit mode will likely require additional custom code to handle these cases properly.

alexis-via (approved):
Using format_amount() in the error message would be a great addition. I have a PR for this on 16.0 https://github.com/OCA/account-invoicing/pull/2165

jdidderen:
@alexis-via I have cherry picked your commit

Can we have this merged ?

PR #1973, opened 2025-04-09
[18.0][MIG] sale_credit_note_reversal: Migration to 18.0

BhaveshHeliconia:


JordiMForgeFlow:
Technical migration looks good but I think small adjustments are needed to make the user experience better.

As it is now, when the wizard is opened to revert a credit note the following happens:

- The new label of the reason field is confusing, because in our case it is applying also to a reason for the invoice, not the credit note.
- Two reverse buttons appear: one to only reverse and another one to fully reverse and create automatically a new draft. In our case the second button would be to reverse and create a new credit note so the wording is confusing

image

BhaveshHeliconia:
@JordiMForgeFlow : I've addressed your suggestion—thank you! Please review when you have a moment.

JordiMForgeFlow:
@BhaveshHeliconia we should also address the second point I mentioned:

Two reverse buttons appear: one to only reverse and another one to fully reverse and create automatically a new draft. In our case the second button would be to reverse and create a new credit note so the wording is confusing

CI is failing also if you could check

BhaveshHeliconia:
@JordiMForgeFlow : I've updated the button label to better reflect its purpose. If you have a better approach or wording in mind, I’m happy to adjust—feel free to suggest!

Yes, I checked the CI logs. The failure is coming from the account_invoice_show_currency_rate module. I also tested this on a fresh DB with only that module installed, and its test case still fails. What would you suggest as the next step?

JordiMForgeFlow:
@BhaveshHeliconia I am sorry it does not look right:

image

You have modified the first button, which was already fine because it only said 'Reverse'. The problem is the second button, which depending on what you are reversing it will create a new invoice or a new credit note. Maybe you can simply rename the second button to 'Reverse and create new invoice/credit note'

The CI will be fixed once https://github.com/OCA/account-invoicing/pull/2056 is merged

BhaveshHeliconia:
@JordiMForgeFlow : Please review. Thanks

sale_credit_note_reversal/wizard/account_move_reversal_views.xml
```diff
@@ -0,0 +1,18 @@
...
+        <field name="model">account.move.reversal</field>
+        <field name="inherit_id" ref="account.view_account_move_reversal" />
+        <field name="arch" type="xml">
+            <field name="reason" position="attributes">
+                <attribute name="string">Reversal reason</attribute>
+            </field>
+            <button name="modify_moves" position="attributes">
+                <attribute
+                    name="string"
+                >Reverse and create new invoice/credit note</attribute>
```
JordiMForgeFlow: Could you align with how Odoo uses the uppercases? Reverse and Create Invoice/Credit Note

BhaveshHeliconia:
@JordiMForgeFlow : Sure, I’ve updated the labels to align with Odoo’s capitalization style. Please review—thanks!

JordiMForgeFlow (approved):

JoanSForgeFlow (approved):

AaronHForgeFlow (approved):
Functional ok!

PR #1959, opened 2025-03-28
[18.0][MIG] account_invoice_pricelist_sale: Migration to 18.0

jdidderen:

PR #1982, opened 2025-04-12
[18.0][FIX] account_invoice_pricelist - pricelist_id set as not readonly and compute update not triggered when already filled

jdidderen:
In the child module account_invoice_pricelist_sale, adding the pricelist to the invoice from the sale order is broken.

Related in 17.0 : https://github.com/OCA/account-invoicing/pull/1981

Needed for https://github.com/OCA/account-invoicing/pull/1959

### DIFF
--- a/account_invoice_pricelist/models/account_move.py
+++ b/account_invoice_pricelist/models/account_move.py
@@ -15,6 +15,7 @@ class AccountMove(models.Model):
         tracking=True,
         store=True,
         precompute=True,
+        readonly=False,
     )
 
     @api.constrains("pricelist_id", "currency_id")
@@ -39,6 +40,7 @@ class AccountMove(models.Model):
                 invoice.partner_id
                 and invoice.is_sale_document()
                 and invoice.partner_id.property_product_pricelist
+                and not invoice.pricelist_id
             ):
                 invoice.pricelist_id = invoice.partner_id.property_product_pricelist
 


sebalix (approved):

account_invoice_pricelist/models/account_move.py
```diff
@@ -15,6 +15,7 @@ class AccountMove(models.Model):
         tracking=True,
         store=True,
         precompute=True,
+        readonly=False,
```
sebalix, line 18: Do we need to set it as readonly on the view?

jdidderen: In the view, there’s a more specific domain based on the invoice state, so I believe it does.

account_invoice_pricelist/models/account_move.py
```diff
@@ -39,6 +40,7 @@ def _compute_pricelist_id(self):
                 invoice.partner_id
                 and invoice.is_sale_document()
                 and invoice.partner_id.property_product_pricelist
+                and not invoice.pricelist_id
```
SilvioC2C, line 43: This changes the old behavior, where changing the customer/company of an invoice triggered an update of the pricelist, and it doesn't make sense: why would we want to keep the old pricelist if the customer/company changed?

I opened https://github.com/OCA/account-invoicing/pull/2085 to revert this change and only introduce the `readonly=False` attribute for the field.

jdidderen: I understand, but the readonly is not sufficient for account_invoice_pricelist_sale (https://github.com/OCA/account-invoicing/pull/1959), because if the sale order does not have the same pricelist as the partner, the pricelist is recomputed when creating the invoice from the sale order. However, I am fine with another workaround if we can find one.

SilvioC2C: @jdidderen I tested it locally and I couldn't replicate the bug:
- I created a SO for a given customer
- I changed the pricelist to one that's not the customer's default pricelist
- I created an invoice from the SO
- the invoice had the correct pricelist (the one from the SO)

Care to explain how to reproduce the bug?

grindtildeath: @jdidderen Would be great if you can add a test (more likely in https://github.com/OCA/account-invoicing/pull/1959) that would highlight the issue you mention

jdidderen:
Closed to be replaced with https://github.com/OCA/account-invoicing/pull/2085 

Sorry I didn't had the time to give feedback on this. I will recheck all that when migrating the module to 19.0 later this year :(

PR #1983, opened 2025-04-14
[FIX] account_invoice_fiscal_position_update: fix tax incl. to tax excl.

### LINKED ISSUE #1896: [16.0] account_invoice_fiscal_position_update wrong computation if B2C / B2B mapping.
## Module
account_invoice_fiscal_position_update

## Describe the bug

when fiscal position is mapping VAT Inc -> VAT Exc (or no vat), the computations of the prices are wrong.

As a result, select product then set fiscal position generates a differente invoice as select fiscal position then, select product.

## To Reproduce
**Affected versions**: 
- 16.0 
- 17.0 (on runboat, with the current migration PR, here https://github.com/OCA/account-invoicing/pull/1630)

**Steps to reproduce the behavior:**

1. Configure a fiscal position with VAT excl -> VAT excl mapping and Vat INCL -> Vat Excl mapping.

(this is the case, by default with INTRA-EU B2B fiscal position installed by odoo l10n_fr localization).

Image

2. Create 2 product, one with vat Incl and the other with vat Excl

Image

3. Create an Invoice, add the two products. (The untaxed amount is 834.33)

Image

4. Set the fiscal position. **The Untaxed amount of the invoice changed**. (Now 1001.00)

Image

**Expected behavior**

If we set first the fiscal position and then select product the result is the following, and is the good one. (834.33)

Image

legalsylvain:
backport of https://github.com/adhoc-dev/oca-account-invoicing/pull/1/commits/bd836c9c9d37304af97304483b3ce0b9820bafd0 (V18 PR)

work done by @StefanRijnhart 

# Description 

Fixes #1896

This change causes the subtotal amount to remain the same when a fiscal position change triggers a tax update from tax included in price to tax excluded from price.

Tests are added for conversion between tax excluded to tax included as well, but in this case the behaviour is not 'fixed'. The standard behaviour of Odoo in case of tax excl. to tax incl. is to adjust the subtotal amount. The same behaviour is present in the function that this module provides. It might not be what users expect, but it seems that Odoo does not support to preserve the subtotal amount in this case.

### DIFF
--- a/account_invoice_fiscal_position_update/models/account_move.py
+++ b/account_invoice_fiscal_position_update/models/account_move.py
@@ -23,7 +23,14 @@ class AccountMove(models.Model):
             if not line.product_id:
                 lines_without_product |= line
             else:
+                # Preserve the subtotal when converting from tax incl to excl
+                price_unit = line.product_id._get_tax_included_unit_price_from_price(
+                    line.price_unit,
+                    line.tax_ids,
+                    fiscal_position=line.move_id.fiscal_position_id,
+                )
                 line._compute_tax_ids()
+                line.price_unit = price_unit
                 line._compute_account_id()
         if lines_without_product:
             res["warning"] = {"title": _("Warning")}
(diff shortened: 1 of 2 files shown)


StefanRijnhart (approved):
Thanks for the port!

rafaelbn (approved):
👍🏼 Thank you!

PR #1999, opened 2025-05-08
[14.0][FIX] account_invoice_purchase_picking_selection: Use commercial_partner_id instead of partner_id

carlos-lopez-tecnativa:
When a contact has children and one is marked as the invoice address, the partner_id is, by default, set to this invoice contact. As a result, the pickings only display records related to that invoice recipient. With this commit, we now use commercial_partner_id instead, ensuring all related records are correctly shown.

TT56280
@Tecnativa @pedrobaeza could you review this please

### DIFF
--- a/account_invoice_purchase_picking_selection/views/account_move_views.xml
+++ b/account_invoice_purchase_picking_selection/views/account_move_views.xml
@@ -12,7 +12,7 @@
                     attrs="{'invisible': ['|', ('state','!=','draft'), ('move_type', '!=', 'in_invoice')]}"
                     class="oe_edit_only"
                     placeholder="Select a purchase reception"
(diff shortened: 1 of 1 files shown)


pedrobaeza (commented):
Searches are done with `child_of` operator?

carlos-lopez-tecnativa:
> Searches are done with `child_of` operator?

No, searches are done using `"partner_id.commercial_partner_id", "=", partner_id"`, in the same way Odoo does in the purchase module.
https://github.com/OCA/account-invoicing/blob/3c56a7eb3b8dfc7fc0f719b3ada97dc40c23aeb1/account_invoice_purchase_picking_selection/models/stock_picking.py#L137

https://github.com/odoo/odoo/blob/cc0060e889603eb2e47fa44a8a22a70d7d784185/addons/purchase/views/account_move_views.xml#L17

pedrobaeza (approved):
/ocabot merge patch

PR #2002, opened 2025-05-08
[18.0][FIX]account_move_tier_validation: Allow to send invoice after validation

### LINKED ISSUE #1998: [18.0][MIG] account_move_tier_validation: Unable to mail invoice
## Describe the bug
After validation of an invoice it is not possible to send it by email.

## To Reproduce
**Affected versions**: 18.0

1. Create a simple tier validation on account move. e.g. validate all by specific user mr demo.
2. Create an invoice
3. Log in as mr Demo validate the invoice
4. Log in again as admin
5. Go to invoice click "Print & Send"

Image

**Expected behavior**
Being able to send the invoice to the customer

**Additional context**
Add any other context about the problem here. (e.g. OS, Python version, ...)
base_tier_validation version `18.0.1.1.0`

As discovered in https://github.com/OCA/account-invoicing/pull/1921

bosd:
Fixes: https://github.com/OCA/account-invoicing/issues/1998

Depends on: https://github.com/OCA/server-ux/pull/1076

Before this PR: Unable to send an invoice by email after validation has finished.
After this PR: Invoice can be sent.

Included unit tests, so if/when module breaks in the future it can be detected.

### DIFF
--- a/account_move_tier_validation/models/account_move.py
+++ b/account_move_tier_validation/models/account_move.py
@@ -28,7 +28,16 @@ class AccountMove(models.Model):
         # otherwise editing manually the values on lines dirties the field at onchange
         # since it's not in readonly because readonly="not(review_ids)", it's then
         # sent at save, and will override the values set by the user
-        return res + ["amount_total"]
+        # The other exclusions are needed to be able to generate the pdf
+        # and send the invoice by email
+        am_exceptions = [
+            "amount_total",
+            "needed_terms_dirty",
+            "is_manually_modified",
+            "is_move_sent",
+            "sending_data",
+        ]
+        return res + am_exceptions
 
     def _get_to_validate_message_name(self):
         name = super()._get_to_validate_message_name()
--- a/account_move_tier_validation/models/account_move.py
+++ b/account_move_tier_validation/models/account_move.py
@@ -36,6 +36,8 @@ class AccountMove(models.Model):
             "is_manually_modified",
             "is_move_sent",
             "sending_data",
+            "matched_payment_ids",
+            "payment_state",
         ]
         return res + am_exceptions
 
--- a/account_move_tier_validation/tests/test_tier_validation.py
+++ b/account_move_tier_validation/tests/test_tier_validation.py
@@ -30,6 +30,32 @@ class TestAccountTierValidation(BaseCommon):
         )
         cls.account_move_model = cls.env["ir.model"]._get("account.move")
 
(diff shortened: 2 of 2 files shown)


account_move_tier_validation/models/account_move.py
```diff
@@ -28,7 +28,15 @@ def _get_validation_exceptions(self, extra_domain=None, add_base_exceptions=True
...
         # sent at save, and will override the values set by the user
-        return res + ["amount_total"]
+        # The other exclusions are needed to be able to generate the pdf
+        # and send the invoice by email
+        am_exceptions = [
+            "needed_terms_dirty",
+            "is_manually_modified",
+            "is_move_sent",
+            "sending_data",
+        ]
```
Saran440, line 41: Should we add `["amount_total"]` in am_exceptions? and return with res + am_exceptions
```suggestion
            "amount_total",
        ]
        return res + am_exceptions
```

bosd: Yes, that would be more clean fix

bosd:
Would be nice if we could fasttrack this one

bosd:
@davidwul Can you please review?

Saran440 (approved):
Code reviewed :+1:

davidwul (approved):

account_move_tier_validation/models/account_move.py
```diff
@@ -28,7 +28,16 @@ def _get_validation_exceptions(self, extra_domain=None, add_base_exceptions=True
...
-        return res + ["amount_total"]
+        # The other exclusions are needed to be able to generate the pdf
+        # and send the invoice by email
+        am_exceptions = [
+            "amount_total",
+            "needed_terms_dirty",
+            "is_manually_modified",
+            "is_move_sent",
+            "sending_data",
+        ]
```
len-foss, line 41:
```suggestion
            "matched_payment_ids",
            "payment_state",
        ]
```

I think these should also directly be integrated. Since validation is before post (and even if validation was used after post to block payments), payment will necessarily be registered after, so these fields should never be blocking.

len-foss (approved):
:+1:

account_move_tier_validation/models/account_move.py
bosd: Could be :) I'll take your word for it and commit this suggestion :+1:

StefanRijnhart (approved):
Thanks!

/ocabot merge patch

PR #2006, opened 2025-05-16
[REF] account_invoice_triple_discount : integrate triple.discount.mixin, previously in purchase_triple_discount module. (OCA/purchase-workflow)

legalsylvain:
Rational : 

- @pedrobaeza in https://github.com/OCA/purchase-workflow/pull/2599/ and @ramiadavid in https://github.com/OCA/purchase-workflow/pull/2317 introduced a new triple.discount.mixin.
- This PR move this mixin in base account_invoice_triple_discount, and makes the account.move.line model inherit from this mixin.

Once merge, we could merge https://github.com/OCA/purchase-workflow/pull/2686

### DIFF
--- a/account_invoice_triple_discount/models/__init__.py
+++ b/account_invoice_triple_discount/models/__init__.py
@@ -1,1 +1,2 @@
+from . import triple_discount_mixin
 from . import account_move_line
--- a/account_invoice_triple_discount/models/account_move_line.py
+++ b/account_invoice_triple_discount/models/account_move_line.py
@@ -2,86 +2,9 @@
 # Copyright 2023 Simone Rubino - Aion Tech
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-import functools
-
-from odoo import api, fields, models
+from odoo import models
 
 
 class AccountMoveLine(models.Model):
-
-    _inherit = "account.move.line"
-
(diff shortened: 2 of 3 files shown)


legalsylvain:
@pedrobaeza, @ramiadavid, @Borruso : could you take a look here ?

thanks !

arnaudlayec (approved):
seems ok for me
shouldn't the tests & translations also be imported too?

legalsylvain:
- Translation will be updated by weblate. 

- There is no test for the mixin, son nothing to move IMO.
Dont you think ?
Sée https://github.com/grap/purchase-workflow/blob/16.0-REF-MOVE-triple_discount_mixin/purchase_triple_discount/tests/test_purchase_discount.py

arnaudlayec:
> * Translation will be updated by weblate.
> * There is no test for the mixin, son nothing to move IMO.
>   Dont you think ?
>   Sée https://github.com/grap/purchase-workflow/blob/16.0-REF-MOVE-triple_discount_mixin/purchase_triple_discount/tests/test_purchase_discount.py

Oh yes you're right for tests, sorry. For sure there are some in both modules and they stay whatever place the mixin is
I didn't know about weblate
all good IMO then

pedrobaeza (approved):
/ocabot merge major

PR #2024, opened 2025-06-04
[IMP] account_invoice_section_sale_order: speed up

gurneyalex:
Forward port of https://github.com/OCA/account-invoicing/pull/2022

On large invoices, at least on 17.0, rewriting the sequence of the invoice lines is very slow, as it triggers a flush to disk for each line (this is even worse when the database is lot running on the same server).

This commit implements a different method for inserting the sections in the invoice lines: we precreate the lines with spaced out sequence numbers, and then we can add the sections between the existing lines, which is much faster.

Test on 17.0: 47 sale orders for 2 customers, generation of 2 invoices with 1200 and 900 lines respectively

Without patch: 9678 SQL queries, 6.868s in SQL 461.969s in Python, total time: 469s

With patch: 5660 SQL queries 1.934 in SQL, 24.167s in Python, total time 26s

### DIFF
--- a/account_invoice_section_sale_order/models/sale_order.py
+++ b/account_invoice_section_sale_order/models/sale_order.py
@@ -2,7 +2,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
 from collections import OrderedDict
 
-from odoo import models
+from odoo import Command, models
 from odoo.tools.safe_eval import safe_eval, time
 
 
@@ -16,6 +16,13 @@ class SaleOrder(models.Model):
         the group name.
         Only do this for invoices targetting multiple groups
         """
+        # the context key 'current_line_sequence' is used by
+        # sale.order.line:_prepare_invoice_line to generate spaced apart sequence number
+        # for invoice lines. We use a mutable in the context because the context is an
+        # immutable dictionary: this way we can update the value of the first element on
+        # the list it and see the updated values in the different calls to
+        # _prepare_invoice_line
+        self = self.with_context(current_line_sequence=[10])
         invoices = super()._create_invoices(grouped=grouped, final=final, date=date)
         for invoice in invoices.sudo():
             if invoice.line_ids and (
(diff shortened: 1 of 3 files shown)


account_invoice_section_sale_order/models/sale_order_line.py
```diff
@@ -0,0 +1,21 @@
...
+    _inherit = "sale.order.line"
+
+    def _prepare_invoice_line(self, **optional_values):
+        values = super()._prepare_invoice_line(**optional_values)
+        current_sequence = self.env.context.get("current_line_sequence")
+        if current_sequence is None:
+            return values
+        # update the sequence of the invoice lines, add some space between lines so that
+        # we can insert the sections afterwards.
+        # we use a mutable in the context to be able to update it and have the updated
```
simahawk: but why a list? It's because of frozendict?

simahawk: I'd move this comment to where you inject it ;)

gurneyalex: fixed

account_invoice_section_sale_order/models/sale_order.py
```diff
@@ -16,34 +16,36 @@ def _create_invoices(self, grouped=False, final=False, date=None):
...
             ):
                 continue
-            sequence = 10
-            # Because invoices are already created, this would require
-            # an extra read access in order to read order fields.
-            move_lines = invoice._get_ordered_invoice_lines()
             # Group move lines according to their sale order
             section_grouping_matrix = OrderedDict()
-            for move_line in move_lines:
+            for move_line in invoice.invoice_line_ids:
```
santostelmo, line 35:
```suggestion
            for move_line in invoice.invoice_line_ids.sorted(key=lambda l: l.sale_line_ids.order_id.date_order):
```

account_invoice_section_sale_order/models/sale_order.py
```diff
@@ -16,34 +16,36 @@ def _create_invoices(self, grouped=False, final=False, date=None):
         the group name.
         Only do this for invoices targetting multiple groups
         """
+        # the context key 'current_line_sequence' is used by
+        # sale.order.line:_prepare_invoice_line to generate spaced apart sequence number
+        # for invoice lines. We use a mutable in the context because the context is an
+        # immutable dictionary: this way we can update the value of the first element on
+        # the list it and see the updated values in the different calls to
+        # _prepare_invoice_line
+        self = self.with_context(current_line_sequence=[10])
```
santostelmo, line 25:
```suggestion
        self = self.with_context(current_line_sequence=[10]).sorted(key=lambda order: order.date_order)
```
Oldest orders are shown in invoice first

PR #1986, opened 2025-04-18
[16.0][IMP] account_billing: Add billing_ids to invoice list and multi-company record rule

AungKoKoLin1997:
This PR does the following:

- **Add billing_ids to invoice list**

Screenshot 2025-05-07 151114

- **Add multi-company record rule**

This is the billing created for YourCompany.

Screenshot 2025-05-07 174310

This is not able to see from another company.

Screenshot 2025-05-07 151657

@qrtl QT5339 QT5371

PR #2041, opened 2025-07-01
[18.0][MIG] account_billing: Migration to 18.0

AungKoKoLin1997:
Supersedes https://github.com/OCA/account-invoicing/pull/1879

Based on: https://github.com/OCA/account-invoicing/pull/1763
Includes: https://github.com/OCA/account-invoicing/pull/1986

@qrtl QT5088


yostashiro (commented):
The report output doesn't look good in runboat. Can you please check?

image

AungKoKoLin1997:
> The report output doesn't look good in runboat. Can you please check?

@yostashiro I updated except for table border part. I don't think we need to do the same with old versions because this is the same with odoo18 table design(the table border of this module for odoo16 uses the same design with odoo16 core table border).

yostashiro (approved):
LGTM.

The report output now looks like this.
image

The table presentation of 16.0 report below looks better to me (it may be a matter of taste), but it's not because of the borders.
image

kanda999 (approved):
Functional test: It works as expected in runboat

Ryoko04:
If I understand correctly, the "Threshold Date Type" does not seem to be working properly.
Even if I select “Due Date” in Threshold Date Type, when I press the "Compute items" button, it seems that the invoice with “accounting date” before the Threshold Date is taken from invoices , and “Due Date” is not referenced.
image

AungKoKoLin1997:
> If I understand correctly, the "Threshold Date Type" does not seem to be working properly.

@Ryoko04 Thanks for the review. I fixed it.

account_billing/models/account_billing.py
```diff
@@ -0,0 +1,257 @@
...
+    def name_get(self):
+        result = [(billing.id, (billing.name or "Draft")) for billing in self]
+        return result
+
+    def validate_billing(self):
+        for rec in self:
+            if not rec.billing_line_ids:
+                raise UserError(_("You need to add a line before validate."))
+            date_type = dict(self._fields["threshold_date_type"].selection).get(
+                self.threshold_date_type
```
yostashiro: https://github.com/OCA/account-invoicing/pull/2061#issuecomment-3100660380

Please quickly review the validity of the entire code.

AungKoKoLin1997: Yes. It should be `rec`.
Fixed it.

account_billing/models/account_billing.py
```diff
@@ -0,0 +1,257 @@
...
+    _description = "Billing Line"
+
+    billing_id = fields.Many2one(comodel_name="account.billing")
+    move_id = fields.Many2one(
+        comodel_name="account.move",
+        index=True,
+    )
+    name = fields.Char(related="move_id.name")
+    invoice_date = fields.Date(related="move_id.invoice_date")
+    threshold_date = fields.Date(related="move_id.invoice_date_due")
```
yostashiro: I find the field name `threshold_date` quite misleading. I'd consolidate these two into a compute field (with `invoice_date` as the surviving one), and simplify the relevant part of `validate_billing()`.

AungKoKoLin1997: Done!

account_billing/models/account_billing.py
```diff
@@ -0,0 +1,260 @@
...
+        string="Amount Due",
+    )
+    state = fields.Selection(related="move_id.state")
+    payment_state = fields.Selection(related="move_id.payment_state")
+
+    @api.depends(
+        "move_id.invoice_date",
+        "move_id.invoice_date_due",
+        "billing_id.threshold_date_type",
+    )
```
yostashiro: I think we can remove these triggers.
```suggestion
```

AungKoKoLin1997: You are right. Updated!

yostashiro (approved):
Code review. LGTM.

However, it would be good to improve the commit messages for fixes and improvements added.

kanda999 (approved):
Functional review: It works as expected.

AungKoKoLin1997:
@OCA/account-invoicing-maintainers 
Could we get this PR merged?

TheerayutEncoder (approved):
LGTM 👍

yostashiro:
@OCA/accounting-maintainers Could this be merged? It's been blocking the progress of the localization modules in l10n-japan. 🙏

PR #2042, opened 2025-07-02
[16.0][FIX] account_invoice_date_due: sync and update es.po translation

carlossainznetkia:
This PR updates the es.po translation file for the account_invoice_date_due module.

Although the standard procedure is to use Weblate for translation contributions, in this case it was not possible, because the current es.po file in the repository is not in sync with the .pot template. As a result, some field strings do not appear in Weblate and cannot be translated from there.

The issue likely stems from the migration of the module, where the .pot file may have been updated, but the es.po file was not regenerated accordingly.

To fix this:

    The .pot file was used to regenerate a fresh and complete es.po.

    The translation file was then updated with the correct Spanish translations for all available terms.

This approach ensures that:

    All strings present in the module are now available in the Spanish translation.

    Weblate will be able to work correctly after this sync.

Let me know if there's any objection to this manual update — I’m happy to adjust the approach if needed.

### DIFF
--- a/account_invoice_date_due/i18n/es.po
+++ b/account_invoice_date_due/i18n/es.po
@@ -1,21 +1,20 @@
 # Translation of Odoo Server.
 # This file contains the translation of the following modules:
-#    * account_invoice_date_due
+# 	* account_invoice_date_due
 #
 msgid ""
 msgstr ""
-"Project-Id-Version: Odoo Server 12.0+e\n"
+"Project-Id-Version: Odoo Server 16.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2019-11-14 19:17+0000\n"
-"PO-Revision-Date: 2021-03-11 13:45+0000\n"
-"Last-Translator: Ana Suárez <ana.suarez@qubiq.es>\n"
-"Language-Team: \n"
+"PO-Revision-Date: 2025-07-02 12:03+0000\n"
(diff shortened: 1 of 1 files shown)


pedrobaeza:
You have to do these changes in Weblate: https://odoo-community.org/resources/translate

carlossainznetkia:
Hi @pedrobaeza, 
As I mentioned in the PR description, I tried to use Weblate, but it wasn’t possible because the es.po file is not properly synced with the .pot file. Several field strings are missing and do not appear in Weblate for translation.

Could you please advise on how to proceed in this case?

pedrobaeza:
OK, I see, reopening then. @sbidoul do you know why the weblate job updating translation files wasn't executed?

pedrobaeza (approved):
/ocabot merge nobump

sbidoul:
> OK, I see, reopening then. @sbidoul do you know why the weblate job updating translation files wasn't executed?

For questions about weblate, it's best to send an email to transbot@odoo-community.org. We track all issues there and handle them eventually.

pedrobaeza:
@carlossainznetkia please expose the problem to that mail. It also happens with other languages for this same module as I have seen.

carlossainznetkia:
> @carlossainznetkia please expose the problem to that mail. It also happens with other languages for this same module as I have seen.

I'll do so @pedrobaeza. Thank you!

Issue #2028, opened 2025-06-10
[>=16.0] account_invoice_mass_sending: Missing info when sending invoices


sergiobstoj:
When sending invoices through `account.invoice.send` wizard but with `Email Mass Sending (Queue Job)` button, record on chatter is missing the envelope icon, where we can check info.

## Module
account_invoice_mass_sending

## Describe the bug
If we send an Invoice with this option:

![Image](https://github.com/user-attachments/assets/3513c1ef-92f8-4b78-af83-5db6debf6b61)

The record on chatter is viewed with envelope icon, where we can check related info:

![Image](https://github.com/user-attachments/assets/e6f8dc41-32c5-446c-a78e-ec6f00553540)

But using enqueue option:

![Image](https://github.com/user-attachments/assets/1cdd51a2-5367-405f-955d-cadc5184ff3e)

We are losing that info:

![Image](https://github.com/user-attachments/assets/4313ed1b-6ec0-4161-a9ee-2925407e7d53)

## To Reproduce
**Affected versions**:
Checked on 16.0

1. Create invoice
2. Click on Send and Print
3. Send Invoice with `Email Mass Sending (Queue Job)` button

**Expected behavior**
We can visualize envelope icon on chatter record and check the usual info, just like when sending with normal process.

rousseldenis:
@sergiobstoj Have you checked the queue job has ended ? Because I suspect job to send the email has not started when you look int the chatter.

sergiobstoj:
@rousseldenis yes

It was checked on local environment only installing that module.

Here you can see the finished job:

![Image](https://github.com/user-attachments/assets/52b5aced-7407-42d4-ae33-66315dad1748)

And the chatter remains that way:

![Image](https://github.com/user-attachments/assets/384ab345-8995-4706-84f4-3c84a1459fdf)

PR #2052, opened 2025-07-10
[16.0][FIX] account_invoice_mass_sending: Send as comment to preserve mail tracking and keep avoiding extra notifications

sergiobstoj:
Solves https://github.com/OCA/account-invoicing/issues/2028
Also, fixes problem detected in here: https://github.com/OCA/account-invoicing/pull/1876#issuecomment-2671848093

Message is again sent as comment, in order to preserve mail track widget. And it avoids extra notifications for users, to keep the fix made here: https://github.com/OCA/account-invoicing/pull/1839

### DIFF
--- a/account_invoice_mass_sending/models/account_move.py
+++ b/account_invoice_mass_sending/models/account_move.py
@@ -13,7 +13,7 @@ class AccountInvoice(models.Model):
         "and it will prevent the sending of a duplicated mail.",
     )
 
-    def mass_sending(self, template=None):
+    def mass_sending(self, template=None, include_followers=False):
         """
         This method triggers the asynchronous sending for the selected
         invoices for which there is no asynchronous sending in progress
@@ -33,10 +33,13 @@ class AccountInvoice(models.Model):
                 invoice.with_delay(
                     description=description,
                     channel="root.account_invoice_mass_sending_channel",
-                )._send_invoice_individually(template=template)
+                )._send_invoice_individually(
+                    template=template,
+                    include_followers=include_followers,
+                )
         return invoices_to_send
 
-    def _send_invoice_individually(self, template=None):
+    def _send_invoice_individually(self, template=None, include_followers=False):
         self.ensure_one()
         res = self.action_invoice_sent()
         wiz_ctx = res["context"] or {}
@@ -48,7 +51,7 @@ class AccountInvoice(models.Model):
                 "active_ids": self.ids,
                 "active_id": self.id,
                 "discard_logo_check": True,
-                "account_invoice_mass_sending": True,
+                "include_followers": include_followers,
             }
         )
         wiz = self.env["account.invoice.send"].with_context(**wiz_ctx).create({})
@@ -67,3 +70,21 @@ class AccountInvoice(models.Model):
             }
         )
         return wiz.send_and_print_action()
+
+    def _notify_get_recipients(self, message, msg_vals, **kwargs):
+        recipients_data = super()._notify_get_recipients(message, msg_vals, **kwargs)
+        if self.env.context.get("include_followers", False):
+            return recipients_data
+        else:
+            msg_sudo = message.sudo()
+            pids = (
+                msg_vals.get("partner_ids", [])
+                if msg_vals
+                else msg_sudo.partner_ids.ids
+            )
+            filtered_recipients_data = [
+                recipient
+                for recipient in recipients_data
+                if recipient.get("id") in pids
+            ]
+            return filtered_recipients_data
--- a/account_invoice_mass_sending/tests/test_account_invoice_mass_sending.py
+++ b/account_invoice_mass_sending/tests/test_account_invoice_mass_sending.py
@@ -169,7 +169,7 @@ class TestAccountInvoiceMassSending(TransactionCase):
             trap.assert_jobs_count(1)
             trap.assert_enqueued_job(
                 self.first_eligible_invoice._send_invoice_individually,
-                kwargs={"template": self.mail_template_obj},
+                kwargs={"template": self.mail_template_obj, "include_followers": False},
             )
             trap.perform_enqueued_jobs()
             self.assertFalse(self.first_eligible_invoice.sending_in_progress)
--- a/account_invoice_mass_sending/wizards/account_invoice_send.py
+++ b/account_invoice_mass_sending/wizards/account_invoice_send.py
@@ -1,15 +1,24 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import _, api, models
+from odoo import _, fields, models
 
 
 class AccountInvoiceSend(models.TransientModel):
     _inherit = "account.invoice.send"
 
+    include_followers = fields.Boolean(
+        default=False,
+        help="If checked, this email will be also send to document followers. "
+        "Else, the email will be sent just to the invoice partner."
+        "\nNote: This functionality only applies for mass sending.",
+    )
+
     def enqueue_invoices(self):
         active_ids = self._context.get("active_ids")
         invoices = self.env["account.move"].browse(active_ids)
-        invoices_to_send = invoices.mass_sending(self.template_id)
+        invoices_to_send = invoices.mass_sending(
+            template=self.template_id, include_followers=self.include_followers
+        )
         ineligible_invoices = invoices - invoices_to_send
         title = _("Invoices: Mass sending")
         msg = _(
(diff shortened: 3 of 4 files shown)


jguenat:
Thanks @sergiobstoj for fixing this. 

maybe this feature could be improved?

I think removing recipient when using mass sending should be an option that the user can enable or not.
In my opinion doing stuff like that under the hood is confusing for the user. The goal of this module is to send invoices by batch with the same output as if you send all of them manually.

One solution would be to add a field in the wizard with a help explaining what the feature does?

What do you think? @sergio-teruel

aritzolea (approved):
Code review is OK, but I think adding the field as @jguenat suggests could be a good idea

sergiobstoj:
Hi @jguenat 
The field has been added as you suggested. Can you check if it's ok please?

account_invoice_mass_sending/wizards/account_invoice_send.py
```diff
@@ -1,15 +1,23 @@
...
 
 
 class AccountInvoiceSend(models.TransientModel):
     _inherit = "account.invoice.send"
 
+    include_followers = fields.Boolean(
+        default=False,
+        help="If checked, this email will be also send to document followers. "
+        "Else, the email will be sent just to the invoice partner.",
+    )
```
jguenat: To avoid confusion with classic send in the wizard view, maybe add to the help that this parameter will only be applied to mass sending?

jguenat (commented):
Thanks for the work, small comment but otherwise looks good

account_invoice_mass_sending/wizards/account_invoice_send.py
sergiobstoj: Make sense, applied!

jguenat: thanks but you put it on the account.move field not the wizard account.invoice.send ^

sergiobstoj: my bad, i'm so sorry. fixed it!

jguenat (approved):

pedrobaeza:
Is this using the same features as standard send & print dialog?

jguenat:
> Is this using the same features as standard send & print dialog?

No, in standard send & print dialog you cannot remove followers from the recipients.



A feature was added to not send the email to the followers in case of mass sending by changing composition mode to "mass_mail". Later it was found that this "mass_mail" composition mode prevents the attachment of EDI documents.

This attempts to fix it.

sergiobstoj:
@pedrobaeza do you think we can merge this?

sergiobstoj:
Can this be merged?

account_invoice_mass_sending/tests/test_account_invoice_mass_sending.py
```diff
@@ -169,7 +169,7 @@ def test_invoice_mass_sending_3(self):
             trap.assert_jobs_count(1)
             trap.assert_enqueued_job(
                 self.first_eligible_invoice._send_invoice_individually,
-                kwargs={"template": self.mail_template_obj},
+                kwargs={"template": self.mail_template_obj, "include_followers": False},
```
jguenat, line 172:
```suggestion
                kwargs={"template": self.mail_template_obj, "exclude_followers": True},
```

account_invoice_mass_sending/models/account_move.py
```diff
@@ -13,7 +13,7 @@ class AccountInvoice(models.Model):
         "and it will prevent the sending of a duplicated mail.",
     )
 
-    def mass_sending(self, template=None):
+    def mass_sending(self, template=None, include_followers=False):
```
jguenat, line 16:
```suggestion
    def mass_sending(self, template=None, exclude_followers=False):
```

account_invoice_mass_sending/models/account_move.py
```diff
@@ -33,10 +33,13 @@ def mass_sending(self, template=None):
                 invoice.with_delay(
                     description=description,
                     channel="root.account_invoice_mass_sending_channel",
-                )._send_invoice_individually(template=template)
+                )._send_invoice_individually(
+                    template=template,
+                    include_followers=include_followers,
```
jguenat, line 38:
```suggestion
                    exclude_followers=exclude_followers,
```

account_invoice_mass_sending/models/account_move.py
```diff
@@ -33,10 +33,13 @@ def mass_sending(self, template=None):
...
+                )
         return invoices_to_send
 
-    def _send_invoice_individually(self, template=None):
+    def _send_invoice_individually(self, template=None, include_followers=False):
```
jguenat, line 42:
```suggestion
    def _send_invoice_individually(self, template=None, exclude_followers=False):
```

account_invoice_mass_sending/models/account_move.py
```diff
@@ -48,7 +51,7 @@ def _send_invoice_individually(self, template=None):
                 "active_ids": self.ids,
                 "active_id": self.id,
                 "discard_logo_check": True,
-                "account_invoice_mass_sending": True,
+                "include_followers": include_followers,
```
jguenat, line 54:
```suggestion
                "exclude_followers": exclude_followers,
```

account_invoice_mass_sending/models/account_move.py
```diff
@@ -67,3 +70,21 @@ def _send_invoice_individually(self, template=None):
             }
         )
         return wiz.send_and_print_action()
+
+    def _notify_get_recipients(self, message, msg_vals, **kwargs):
+        recipients_data = super()._notify_get_recipients(message, msg_vals, **kwargs)
+        if self.env.context.get("include_followers", False):
```
jguenat, line 76:
```suggestion
        if not self.env.context.get("exclude_followers", False):
```

account_invoice_mass_sending/wizards/account_invoice_send.py
```diff
@@ -1,15 +1,24 @@
...
+        "Else, the email will be sent just to the invoice partner."
+        "\nNote: This functionality only applies for mass sending.",
+    )
+
     def enqueue_invoices(self):
         active_ids = self._context.get("active_ids")
         invoices = self.env["account.move"].browse(active_ids)
-        invoices_to_send = invoices.mass_sending(self.template_id)
+        invoices_to_send = invoices.mass_sending(
+            template=self.template_id, include_followers=self.include_followers
```
jguenat, line 20:
```suggestion
            template=self.template_id, exclude_followers=self.exclude_followers
```

account_invoice_mass_sending/wizards/account_invoice_send.py
```diff
@@ -1,15 +1,24 @@
...
 
 class AccountInvoiceSend(models.TransientModel):
     _inherit = "account.invoice.send"
 
+    include_followers = fields.Boolean(
+        default=False,
+        help="If checked, this email will be also send to document followers. "
+        "Else, the email will be sent just to the invoice partner."
+        "\nNote: This functionality only applies for mass sending.",
+    )
```
jguenat, line 14:
```suggestion
    exclude_followers = fields.Boolean(
        default=True,
        help="If checked, the email will be send only to the invoice "
        "partner, excluding followers."
        "\nNote: This functionality only applies for mass sending.",
    )
```

account_invoice_mass_sending/wizards/account_invoice_send.xml
```diff
@@ -5,6 +5,9 @@
         <field name="model">account.invoice.send</field>
         <field name="inherit_id" ref="account.account_invoice_send_wizard_form" />
         <field name="arch" type="xml">
+            <field name="template_id" position="after">
+                <field name="include_followers" />
```
jguenat, line 9:
```suggestion
                <field name="exclude_followers" />
```

account_invoice_mass_sending/models/account_move.py
jguenat: You need to change how you manage context check in _notify_get_recipients
otherwise you disable notifying followers when there is no context (e.g. in the chatter)

jguenat (changes requested):
Some adjustment to do, sorry I didn't detect this first review

account_invoice_mass_sending/models/account_move.py
jguenat: (to avoid the " if not condition"  which is less readable than "if condition" you could do the return filtered_recipients_data in the if and the return recipients_data in the else.)

PR #2053, opened 2025-07-10
[ADD] account_custom_account_by_invoice_address

sbejaoui:
This module allows to define custom receivable and payable accounts per **invoice address** for a given customer or vendor.

When the option is enabled on the commercial partner, the account defined on the selected invoicing address will be used instead of the default from the parent company.

cc/ @lmignon , @rousseldenis 

image


rousseldenis (approved):
LGTM

account_custom_account_by_invoice_address/models/account_move_line.py
```diff
@@ -0,0 +1,25 @@
...
+class AccountMoveLine(models.Model):
+
+    _inherit = "account.move.line"
+
+    def _compute_account_id(self):
+        res = super()._compute_account_id()
+        for rec in self.filtered(lambda r: r.display_type == "payment_term"):
+            move = rec.move_id
+            if not move.partner_id.parent_use_invoice_address_accounts:
+                continue
```
lmignon: When you use filtered as input of your loop, you create a double loop. A first time into filtered and a second time on the recordset return by filtered. It's more efficient to test the display_type directly into a loop on self....
```suggestion
        for rec in self:
            if rec.display_type != "payment_term":
                continue
            move = rec.move_id
            if not move.partner_id.parent_use_invoice_address_accounts:
                continue
```

lmignon (changes requested):

lmignon (approved):

rousseldenis:
@OCA/accounting-maintainers @jbaudoux

PR #2061, opened 2025-07-21
[16.0][Fix] account_billing test_create_billing_from_selected_invoices

santostelmo:
### Context 
CI unit test is currently falling : 
```
TestAccountBilling.test_4_create_billing_from_selected_invoices
  File "/__w/account-invoicing/account-invoicing/account_billing/tests/test_account_billing.py", line 128, in test_4_create_billing_from_selected_invoices
    with self.assertRaises(ValidationError):
  File "/usr/lib/python3.10/contextlib.py", line 142, in __exit__
    next(self.gen)
  File "/opt/odoo/odoo/tests/common.py", line 387, in _assertRaises
    with ExitStack() as inner:
  File "/usr/lib/python3.10/contextlib.py", line 576, in __exit__
    raise exc_details[1]
  File "/usr/lib/python3.10/contextlib.py", line 561, in __exit__
    if cb(*exc_details):
AssertionError: ValidationError not raised
```
Because payment terms are setup to 15days, depending on the day you run this test you get a different result.

[threshold_date](https://github.com/OCA/account-invoicing/blob/7207e6e26a87963ceb058e74180001ba9c9b703c/account_billing/models/account_billing.py#L148) is calculated based on [invoice_date_due](https://github.com/OCA/account-invoicing/blob/7207e6e26a87963ceb058e74180001ba9c9b703c/account_invoice_date_due/models/account_move.py#L14). That field is computed based on account.move `date_maturity` that depends of today's date and the payment terms

https://github.com/odoo/odoo/blob/b0be89a67d0ce458db99f4331e21ed56d3c4774a/addons/account/models/account_move.py#L819

### DIFF
--- a/account_billing/tests/test_account_billing.py
+++ b/account_billing/tests/test_account_billing.py
@@ -37,6 +37,7 @@ class TestAccountBilling(AccountTestInvoicingCommon):
             invoice_amount=100,
             currency_id=cls.currency_eur_id,
             partner_id=cls.partner_a.id,
+            date_invoice=fields.Date.context_today(cls.env.user),
             payment_term_id=cls.payment_term.id,
             auto_validate=True,
         )


rvalyi (approved):

yostashiro:
I didn't quite get how switching the payment terms from 15 days to 30 days should solve the issue. Could you please clarify with some concrete examples?

I'm not sure if it's related but I thought we should fix the following lines to use `rec` instead of `self`:
https://github.com/OCA/account-invoicing/blob/7207e6e/account_billing/models/account_billing.py#L145
https://github.com/OCA/account-invoicing/blob/7207e6e/account_billing/models/account_billing.py#L151

santostelmo:
@yostashiro if you execute the test today `22-07-2025` with payment terms **15 days**
threshold_date = 22-07-2025
customer_billing1.threshold_date_type = invoice_date_due
customer_billing1.billing_line_ids[0].move_id.invoice_date_due = 16-07-2025

invoice_date_due calculated [here](https://github.com/odoo/odoo/blob/b0be89a67d0ce458db99f4331e21ed56d3c4774a/addons/account/models/account_move.py#L1017)

invoice_date = 2025-07-01
invoice_date_due = 2025-07-01 + 15 days = 16-07-2025
invoice_date_due < 22-07-2025 therefore no validation error is raised.

With payment terms **30 days** 
invoice_date_due = 01-07-2025 + 30 days = 31-07-2025 => validation error is raised

Note: I am not currently user of account-billing module.Try to have CI fixed but if you know well this module fell free to investigate deeper

yostashiro (commented):
@santostelmo Thanks for your explanation. What puzzled me was how the invoice date was set to 2025-07-01, but now I understand it comes from https://github.com/odoo/odoo/blob/86436a0/addons/account/tests/common.py#L439.

IIUC, switching the payment term from 15 days to 30 days will only solve the problem temporarily, until 30 July, and the issue will revive starting 31 July.

The proper fix should be to assign the current date to the test invoices as the invoice date in the following manner:

```.py
        cls.inv_1 = cls._create_invoice(
            cls,
            move_type="out_invoice",
            invoice_amount=100,
            currency_id=cls.currency_eur_id,
            partner_id=cls.partner_a.id,
            date_invoice=fields.Date.context_today(cls.env.user),  # Add this
            payment_term_id=cls.payment_term.id,
            auto_validate=True,
        )
```

Could you please update the relevant parts accordingly?

santostelmo:
Thank you @yostashiro I updated the code

yostashiro (approved):
Thanks @santostelmo ! Would be great if you could do the same to the rest of test invoices (i.e., cls.inv_2 - cls.inv_6) just to avoid future surprises in case we fiddle with the test content.

rvalyi:
@santostelmo would be cool to consider https://github.com/OCA/account-invoicing/pull/2061#pullrequestreview-3051797181

But as it blocks the tests on the whole branch, I propose to merge this for now to unblock things.

/ocabot merge patch

PR #2063, opened 2025-07-25
[18.0][MIG] sale_timesheet_invoice_description: Migration to 18.0

chrisb-c01:


matteotognini (approved):
LGTM

carlos-lopez-tecnativa (approved):
LGTM
TT55917
@pedrobaeza could you review this please?

sale_timesheet_invoice_description/models/account_move.py
```diff
@@ -33,25 +33,3 @@ def _link_timesheets_to_invoice_line(self, start_date=None, end_date=None):
 
                 timesheets = self.env["account.analytic.line"].sudo().search(domain)
                 timesheets.write({"timesheet_invoice_line_id": aml.id})
-
```
pedrobaeza: Please keep both models in the same file, as they are linked through one2many field, so they are very tied. Check https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#id12, where it's stated.

chrisb-c01: Hi @pedrobaeza thanks for the review. I didn't notice it before but the docs indeed state:
> An exception can be made when the model is a model intended to be used only as a one2many model nested on the main model.

Not trying to ruffle your feathers but I would argue for keeping it in a separate file because:
* The docs state that "an exception _can_ be made", so apparently it's not mandatory.
* In general models are split per file. Conventions are more easy to follow if there are less exceptions. Is the advantage of putting it in 1 file really greater than adhering to the default?
* In the Odoo codebase I can find sparse examples but not necessarily applying the logic around one2many fields.

Also, the docs do not state _why_ the exception can/must be made. It's not clear what the advantage would be. Can you explain it to me?

pedrobaeza: Apart from the rule, this one should be the valid one: if they are together, keep them together. This is only a migration, so do only the minimum changes for the migration. Right now, you are dirtying the migration commit with that changes that make the review harder.

Later, a refactoring can be discussed for separating them, but not on the migration.

pedrobaeza: > Also, the docs do not state why the exception can/must be made. It's not clear what the advantage would be. Can you explain it to me?

This is for having together the 2 tied models, as they are usually sharing things (like `many2one` / `inverse` information, calling from one to the other, so you don't have to look into 2 files to find things.

pedrobaeza: @chrisb-c01 can you undo the change for proceeding with the merge?

pedrobaeza (approved):
Thanks for the changes.

/ocabot migration sale_timesheet_invoice_description
/ocabot merge nobump

sale_timesheet_invoice_description/models/account_move.py
chrisb-c01: @pedrobaeza sure, just did so. 

Thanks for providing the feedback. I understand that code changes while migrating should be kept to a minimum.

I do think the convention for putting the models in 1 file is too subjective, and personally I would argue that in terms of consistency it would be better to put each model in their own file. But let's have that discussion another time :-)

pedrobaeza: Yes, there are not white/black thing here. Anyway, the main lead should be to preserve the will of the original contributor, unless you upscale to maintain the code, and then, you can go your own direction (in a `[REF]` commit) for being comfortable with that codebase.

pedrobaeza:
@chrisb-c01 can you check the red CI?

chrisb-c01:
> @chrisb-c01 can you check the red CI?

It's tests failing from other modules (`sale_stock_picking_invoicing`, `account_move_cancel_confirm`). The `account_move_cancel_confirm` is failing in other PR's as well.

I can take a look, unfortunately not right now.

pedrobaeza:
Let's try with the merged patch...

/ocabot merge nobump

jusle:
Could rerun of failed test jobs allow this PR to be merged?

grojof:
Update for rerun context: related stabilizing PRs are now green and ready in 18.0:

- #2271
- #2272
- #2273
- #2274

This should reduce matrix noise observed previously. Could CI for #2063 be re-run against current 18.0 context?

sergiocorato:
@chrisb-c01 could you do a rebase?

jans23:
@chrisb-c01  please rebase

PR #2074, opened 2025-08-08
[16.0][FIX] account_invoice_pricelist: prevent price_unit changes when qty inverted for credit notes

nobuQuartile:
When action_switch_invoice_into_refund_credit_note inverts line quantities for invoices with a negative
amount_total, it triggers _compute_price_unit and unintentionally updates price_unit. This change prevents
such updates while preserving intended qty changes.

action_switch_invoice_into_refund_credit_note 
https://github.com/odoo/odoo/blob/16.0/addons/account/models/account_move.py#L3727

@qrtl
QT5644, QT5480

### DIFF
--- a/account_invoice_pricelist/models/account_move.py
+++ b/account_invoice_pricelist/models/account_move.py
@@ -57,6 +57,10 @@ class AccountMove(models.Model):
                 invoice.currency_id = self.pricelist_id.currency_id
         return res
 
+    def action_switch_invoice_into_refund_credit_note(self):
+        self = self.with_context(skip_compute_price_unit=True)
+        return super().action_switch_invoice_into_refund_credit_note()
+
     def button_update_prices_from_pricelist(self):
         self.filtered(
             lambda r: r.state == "draft"
@@ -68,6 +72,8 @@ class AccountMoveLine(models.Model):
 
     @api.depends("quantity")
     def _compute_price_unit(self):
+        if self.env.context.get("skip_compute_price_unit"):
+            return
         res = super()._compute_price_unit()
         for line in self:
             if not line.move_id.pricelist_id:
--- a/account_invoice_pricelist/tests/test_account_move_pricelist.py
+++ b/account_invoice_pricelist/tests/test_account_move_pricelist.py
@@ -316,3 +316,10 @@ class TestAccountMovePricelist(BaseCommon):
             self.invoice.with_context(force_check_currecy=True).write(
                 {"currency_id": self.usd_currency.id}
             )
+
+    def test_account_invoice_creditnote_line_qty_flip_when_total_negative(self):
+        self.invoice.pricelist_id = self.sale_pricelist
+        self.invoice.invoice_line_ids.write({"quantity": -1, "price_unit": 100.00})
+        self.invoice.action_switch_invoice_into_refund_credit_note()
+        for invoice_line in self.invoice.invoice_line_ids:
+            self.assertAlmostEqual(invoice_line.price_unit, 100.00)


account_invoice_pricelist/models/account_move.py
```diff
@@ -87,45 +91,44 @@ def _calculate_discount(self, base_price, final_price):
 
     def _get_price_with_pricelist(self):
```
yostashiro: Please revert the changes on this method. We don't want to change the conditions here to be safe.

account_invoice_pricelist/models/account_move.py
```diff
@@ -70,7 +70,11 @@ class AccountMoveLine(models.Model):
     def _compute_price_unit(self):
         res = super()._compute_price_unit()
         for line in self:
-            if not line.move_id.pricelist_id:
+            if (
+                not line.product_id
+                or not line.move_id.move_type == "out_invoice"
+                or not line.move_id.pricelist_id
+            ):
```
yostashiro: Please revert this change as these conditions don't look correct.

The issue of the sales order price not persisting in credit notes is triggered by this write: https://github.com/odoo/odoo/blob/f6db02f/addons/account/models/account_move.py#L3744

Therefore, my suggestion is to skip the price recomputation when the quantity change comes from this method.

1. Add this to account.move:
```.py
    def action_switch_invoice_into_refund_credit_note(self):
        self = self.with_context(skip_compute_price_unit=True)
        return super().action_switch_invoice_into_refund_credit_note()
```

2. Turn this:
```.py
        res = super()._compute_price_unit()
```
into this:
```.py
        if self.env.context.get("skip_compute_price_unit"):
            return
        res = super()._compute_price_unit()
```

yostashiro (changes requested):
Please also update tests to use action_switch_invoice_into_refund_credit_note().

account_invoice_pricelist/models/account_move.py
nobuQuartile: Done!

account_invoice_pricelist/models/account_move.py
nobuQuartile: Done
I have updated the test code.

nobuQuartile: Sorry. I misunderstood the issue.
I have updated the test, the commit message and the PR name.

account_invoice_pricelist/tests/test_account_move_pricelist.py
```diff
@@ -316,3 +334,9 @@ def test_check_currency(self):
             self.invoice.with_context(force_check_currecy=True).write(
                 {"currency_id": self.usd_currency.id}
             )
+
+    def test_account_invoice_creditnote_line_qty_flip_when_total_negative(self):
+        self.negative_invoice.action_switch_invoice_into_refund_credit_note()
+        invoice_line = self.negative_invoice.invoice_line_ids[:1]
+        self.assertAlmostEqual(invoice_line.price_unit, 100.00)
+        self.assertEqual(invoice_line.discount, 0.00)
```
yostashiro: I suppose we don't necessarily have to have the negative invoice in the setup.

```suggestion
        self.invoice.pricelist_id = self.sale_pricelist
        self.invoice.invoice_line_ids.write({"quantity": -1, "price_unit": 100.00})
        self.invoice.action_switch_invoice_into_refund_credit_note()
        for invoice_line in self.invoice.invoice_line_ids
            self.assertAlmostEqual(invoice_line.price_unit, 100.00)
```

If we keep the negative invoice, we should still assign a pricelist to it.

yostashiro (approved):
LGTM. 👍

account_invoice_pricelist/models/account_move.py
```diff
@@ -68,6 +72,8 @@ class AccountMoveLine(models.Model):
 
     @api.depends("quantity")
     def _compute_price_unit(self):
+        if self.env.context.get("skip_compute_price_unit"):
```
AungKoKoLin1997, line 75: I don't understand this module feature fully. But I don't prefer using context inside the compute method.
Is there any way to solve the issue without using context?
In my opinion, compute method should be worked with attributes of records.
This is also opinion from one of OCA maintainers.
https://github.com/OCA/stock-logistics-warehouse/pull/1711#discussion_r1530059486

AungKoKoLin1997 (approved):
After checking, passing context from action_switch_invoice_into_refund_credit_note into the compute is safe. Please disregard my previous comment.

PR #2076, opened 2025-08-12
[16.0][FIX] account_invoice_date_due: Handle multi records on write

sergiobstoj:
Fixes singleton error when writing over `invoice_date_due` on multiple records.

Error is reproduced in shell, couldn't reproduce it on runboat without `server_action_mass_edit` from `server-ux`

Without this fix:

```bash
>>> invoice1 = self.env["account.move"].search([], limit=2, order='id desc')[0]
>>> invoice2 = self.env["account.move"].search([], limit=2, order='id desc')[1]
>>> invoice1.invoice_date_due
datetime.date(2025, 8, 11)
>>> invoice2.invoice_date_due
datetime.date(2025, 8, 11)
>>> invoices = invoice1 + invoice2
>>> invoices.write({"invoice_date_due": "2025-09-11"})

  File "/mnt/code/development/account-invoicing/account_invoice_date_due/models/account_move.py", line 39, in write
    if "invoice_date_due" in vals and self.state == "posted":
  File "/mnt/environment/odoo-server/odoo/fields.py", line 1154, in __get__
    record.ensure_one()
  File "/mnt/environment/odoo-server/odoo/models.py", line 5251, in ensure_one
    raise ValueError("Expected singleton: %s" % self)
ValueError: Expected singleton: account.move(50331, 50330)
```
With this fix:

```bash
>>> invoice1 = self.env["account.move"].search([], limit=2, order='id desc')[0]
>>> invoice2 = self.env["account.move"].search([], limit=2, order='id desc')[1]
>>> invoice1.invoice_date_due
datetime.date(2025, 8, 11)
>>> invoice2.invoice_date_due
datetime.date(2025, 8, 11)
>>> invoices = invoice1 + invoice2
>>> invoices.write({"invoice_date_due": "2025-09-11"})
True
```

PR #2080, opened 2025-08-14
[17.0][FIX] account_invoice_date_due: Handle multi records on write

sergiobstoj:
The write method failed with a singleton error when updating a batch of records.
This was caused by accessing `self.state` on the entire recordset.

The logic is now fixed to first filter for `posted` moves, and then apply the due date propagation only to that safe subset. This ensures batch operations no longer crash.

fw-port from https://github.com/OCA/account-invoicing/pull/2076

### DIFF
--- a/account_invoice_date_due/models/account_move.py
+++ b/account_invoice_date_due/models/account_move.py
@@ -35,8 +35,9 @@ class AccountMove(models.Model):
         res = super().write(vals)
         # Propagate due date to move lines
         # that correspond to the receivable/payable account
-        if "invoice_date_due" in vals and self.state == "posted":
-            payment_term_lines = self.line_ids.filtered(
+        posted_moves = self.filtered(lambda x: x.state == "posted")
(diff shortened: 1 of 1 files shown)


CarlosRoca13 (approved):

account_invoice_date_due/models/account_move.py
```diff
@@ -34,9 +34,10 @@ def _onchange_invoice_date_due_payment_term(self):
     def write(self, vals):
         res = super().write(vals)
         # Propagate due date to move lines
-        # that correspond to the receivable/payable account
-        if "invoice_date_due" in vals and self.state == "posted":
-            payment_term_lines = self.line_ids.filtered(
+        # that correspont to the receivable/payable account
```
pedrobaeza: Why changing the comment line introducing a typo (t instead of d).

pedrobaeza (approved):
/ocabot merge patch

PR #2083, opened 2025-08-18
[18.0][MIG] account_invoice_supplierinfo_update_discount: Migration to 18.0

victoralmau:
Migration to 18.0 (Merged into `account_invoice_supplierinfo_update`)

@Tecnativa TT57250


pedrobaeza:
As `discount` is now a native field, this shouldn't be in a separate module.

account_invoice_supplierinfo_update/models/account_move_line.py
```diff
@@ -23,18 +23,21 @@ def _is_matching_supplierinfo(self, supplierinfo):
...
         self.ensure_one()
-        return (
+        res = (
             not self.product_uom_id or self.product_uom_id == supplierinfo.product_uom
         ) and not float_compare(
             self.price_unit,
             supplierinfo.price,
             precision_rounding=self.move_id.currency_id.rounding,
         )
+        discount = self.discount1 if "discount1" in self._fields else self.discount
```
pedrobaeza: `discount1` is not something native, so don't do this ugly patches. Change also the docstring.

account_invoice_supplierinfo_update/models/account_move_line.py
```diff
@@ -18,18 +18,16 @@ def _get_supplierinfo(self):
...
         self.ensure_one()
-        return (
+        res = (
             not self.product_uom_id or self.product_uom_id == supplierinfo.product_uom
         ) and not float_compare(
             self.price_unit,
             supplierinfo.price,
             precision_rounding=self.move_id.currency_id.rounding,
         )
+        return res and (self.discount == supplierinfo.discount)
```
legalsylvain: Maybe we should use float compare to compare discount field. Dont you think ?

(I'm not sûre)

legalsylvain (approved):
A remark inline. Otherwise, lgtm.

account_invoice_supplierinfo_update/models/account_move_line.py
victoralmau: Although this was already the case in v16 (https://github.com/OCA/account-invoicing/blob/a53031cb87274f60a408d630e069c0d0c6a4709a/account_invoice_supplierinfo_update_discount/models/account_move_line.py# L14) and the fields have the same defined digits (https://github.com/odoo/odoo/blob/f42fe80576c8b419b5215ff17859e0dfebcd996a/addons/product/models/product_supplierinfo.py# L66 + https://github.com/odoo/odoo/blob/f42fe80576c8b419b5215ff17859e0dfebcd996a/addons/account/models/account_move_line.py#L373), I think it makes sense to use float_compare, so it has been changed.

pedrobaeza (approved):
/ocabot merge minor

PR #2064, opened 2025-07-30
[17.0][FIX] stock_account_move_reset_to_draft: only consider generated SVLs from account move line

AungKoKoLin1997:
Forward-port of https://github.com/OCA/account-invoicing/pull/2032.

@qrtl

PR #2090, opened 2025-08-28
[18.0] [MIG] stock_account_move_reset_to_draft: Migration to 18.0

ivantodorovich:
It includes:
- https://github.com/OCA/account-invoicing/pull/2064


henrybackman (approved):

sebalix (approved):

stock_account_move_reset_to_draft/models/account_move.py
```diff
@@ -0,0 +1,75 @@
...
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    def button_draft(self):
+        """If it is a purchase invoice, we will create a new SVL for each line with
+        the sum of the value in opposite sign.
+        """
+        for item in self.sudo().filtered(
+            lambda x: x.is_inbound
```
grindtildeath:
```suggestion
            lambda x: x.is_inbound()
```
`is_inbound` being a function it would always evaluate to True if not called.

stock_account_move_reset_to_draft/models/account_move.py
```diff
@@ -0,0 +1,75 @@
...
+            and any(line.stock_valuation_layer_ids for line in x.line_ids)
+        ):
+            for line in item.line_ids.filtered("stock_valuation_layer_ids"):
```
grindtildeath: I feel like the condition in first if is unneeded with the filtering from the for loop

stock_account_move_reset_to_draft/models/account_move.py
ivantodorovich: Indeed, good catch!

Moreover `is_inbound` was not the right method as that's based on the money flow.
The correct one is `is_purchase_document`, so I changed it

AungKoKoLin1997 (approved):
Code Review: LGTM

grindtildeath (approved):

PR #2095, opened 2025-09-03
[16.0][IMP] account_invoice_show_currency_rate: add option to choose currency rate display format

AungKoKoLin1997:
This PR adds an option to display the exchange rate either as company currency per unit or unit per company currency.

Example (company currency: JPY; 1 USD = 150 JPY):

- (Default) Unit per Company Currency → 0.0066 (USD/JPY)
- Company Currency per Unit → 150 (JPY/USD)

@qrtl QT5684

### DIFF
--- a/account_invoice_show_currency_rate/models/account_move.py
+++ b/account_invoice_show_currency_rate/models/account_move.py
@@ -32,6 +32,7 @@ class AccountMove(models.Model):
         - Case C: Get expected rate (according to date) to show some value in creation.
         """
         self.currency_rate_amount = 1
+        inverse = self.env.company.invoice_rate_display_type == "inverse_rate"
         for item in self.filtered("show_currency_rate_amount"):
             lines = item.line_ids.filtered(lambda x: abs(x.amount_currency) > 0)
             if item.state == "posted" and lines:
@@ -40,11 +41,19 @@ class AccountMove(models.Model):
                 )
                 total_balance_positive = sum([abs(b) for b in lines.mapped("balance")])
                 item.currency_rate_amount = (
-                    amount_currency_positive / total_balance_positive
+                    (amount_currency_positive / total_balance_positive)
+                    if not inverse
+                    else (total_balance_positive / amount_currency_positive)
                 )
             else:
                 rates = item.currency_id._get_rates(item.company_id, item.date)
-                item.currency_rate_amount = rates.get(item.currency_id.id)
+                item.currency_rate_amount = (
+                    rates.get(item.currency_id.id)
+                    if not inverse
+                    else item.currency_id._convert(
+                        1.0, item.company_id.currency_id, item.company_id, item.date
+                    )
+                )
 
     @api.depends("currency_id", "currency_id.rate_ids", "company_id")
     def _compute_show_currency_rate_amount(self):
--- a/account_invoice_show_currency_rate/README.rst
+++ b/account_invoice_show_currency_rate/README.rst
@@ -53,6 +53,11 @@ Some rates must be defined (and be distinct to 1.0) for currencies different fro
 #. Go to Rates smart-button
 #. Update 01/01/2010 record and change rate to 1.5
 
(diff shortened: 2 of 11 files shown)


account_invoice_show_currency_rate/models/res_currency.py
```diff
@@ -0,0 +1,24 @@
...
+class ResCurrency(models.Model):
+    _inherit = "res.currency"
+
+    invoice_currency_rate_display = fields.Selection(
+        [
+            ("unit_per_company", "Unit per Company Currency"),
+            ("company_per_unit", "Company Currency per Unit"),
+        ],
+        default="unit_per_company",
+        string="Invoice Rate Display",
```
yostashiro: Not a strong suggestion.

```suggestion
    invoice_rate_display_type = fields.Selection(
        [
            ("normal", "Unit per Company Currency"),
            ("inverse_rate", "Company Currency per Unit"),
        ],
        default="normal",
        string="Invoice Rate Display Type",
```

account_invoice_show_currency_rate/models/account_move.py
```diff
@@ -12,6 +12,14 @@ class AccountMove(models.Model):
...
         digits=0,
     )
+    company_currency_rate_amount = fields.Float(
+        string="Rate Amount",
+        compute="_compute_currency_rate_amount",
+        digits=0,
+    )
+    currency_rate_display = fields.Selection(
+        related="currency_id.invoice_currency_rate_display"
+    )
```
yostashiro: IMO, we don't need to add these fields. We can just use the existing `currency_rate_amount` field, as the field is only for the display purpose.

```suggestion
```

account_invoice_show_currency_rate/models/account_move.py
```diff
@@ -57,6 +72,40 @@ def _compute_show_currency_rate_amount(self):
 class AccountMoveLine(models.Model):
     _inherit = "account.move.line"
 
+    company_currency_rate_amount = fields.Float(
+        string="Rate Amount",
+        compute="_compute_company_currency_rate_amount",
+        digits=0,
+    )
```
yostashiro: Likewise, we don't need to add a new field.
```suggestion
```

account_invoice_show_currency_rate/views/res_currency_views.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="view_currency_form" model="ir.ui.view">
+        <field name="name">res.currency</field>
+        <field name="model">res.currency</field>
+        <field name="inherit_id" ref="base.view_currency_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='symbol']" position="before">
```
yostashiro:
```suggestion
            <xpath expr="//field[@name='position']" position="after">
```

account_invoice_show_currency_rate/models/account_move.py
```diff
@@ -34,17 +34,26 @@ def _compute_currency_rate_amount(self):
         self.currency_rate_amount = 1
         for item in self.filtered("show_currency_rate_amount"):
             lines = item.line_ids.filtered(lambda x: abs(x.amount_currency) > 0)
+            inverse = item.company_id.invoice_rate_display_type != "normal"
```
yostashiro: I prefer this style. It could be more "correct" to use the company of the invoice, but using the env company should be sufficient, IMO.

```suggestion
        inverse = self.env.company.invoice_rate_display_type == "inverse_rate"
        for item in self.filtered("show_currency_rate_amount"):
            lines = item.line_ids.filtered(lambda x: abs(x.amount_currency) > 0)
```

yostashiro (approved):
LGTM. 👍

kanda999 (approved):
It works as expected in my local env.

victoralmau (approved):

PR #2099, opened 2025-09-08
[15.0][FIX] account_manual_currency: recompute tax include

Saran440:
Step to test
1. create move with **tax include** and use manual currency
Selection_017

2. Total (USD) is 50 but in Journal Items line tax still compute with old currency. So, **total amount is 63.42 is not equal 50**
Selection_018

This PR fixed recompute tax when manual currency
Selection_019


AungKoKoLin1997 (commented):
Could you please a test case for the fix?

PR #2100, opened 2025-09-08
[IMP] account_invoice_supplier_self_invoice: Optional reference auto generation

lmignon:
Makes the generation of a client reference on self billed invoices optional.

In a lot of context, the reference given by the supplier for the self billed invoice comes from its own system. By default we should not generate this information into Odoo. A flag allows you to adapt this behavior at partner level.


lmignon:
A migration script is required to ensure that the current behavior remains the default one on existing installation.

lmarion-source (approved):

Shide:
I don't understand this improvement. You could do this behavior right now:

1. Change default prefix on company if you want
2. Activate debug mode
3. Go to a partner
4. Enable self billing
5. Click on the button to create the sequences
6. Enter on the sequences and change prefix

image

https://www.loom.com/share/20e008dbe23f45c195fcea9117dd037b

What is changing with this improvement?

nihelgabsi-acsone (approved):
Functional test + code review

Shide (changes requested):
Can you review my comment? https://github.com/OCA/account-invoicing/pull/2100#issuecomment-3654550199

jbaudoux (approved):

jbaudoux:
> Can you review my comment? [#2100 (comment)](https://github.com/OCA/account-invoicing/pull/2100#issuecomment-3654550199)

@Shide According to the standard for electronic invoicing, we should not generate a sequence for the other party. The company generating the self invoice generates a number in its own journal and that's the only number of the document. When the self-invoice is processed by the other party in its accounting, he will freely generate a sequence in his journal.
Laurent is removing here the default self-billing sequence and makes it optional
https://docs.peppol.eu/poacc/self-billing/3.0/bis-sb/

jbaudoux (commented):
In my opinion, the invoice report should be adapted to always display the invoice number (name).

Shide:
> In my opinion, the invoice report should be adapted to always display the invoice number (name).

But company and partner changes position in self billing report: https://github.com/OCA/account-invoicing/pull/2248#issuecomment-3823278977

PR #2102, opened 2025-09-09
[IMP] sale_line_refund_to_invoice_qty: exclude refunds without reinvoicing qty from invoiced amount

Chionne27:


sale_line_refund_to_invoice_qty/models/sale.py
```diff
@@ -54,3 +54,30 @@ def _compute_qty_refunded_not_invoiceable(self):
...
+        """
+        res = super()._compute_untaxed_amount_invoiced()
+        for line in self:
+            amount_invoiced = line.untaxed_amount_invoiced
+            for invoice_line in line.invoice_lines:
+                if (
+                    invoice_line.move_id.state == "posted"
+                    and invoice_line.move_id.move_type == "out_refund"
+                    and not invoice_line.sale_qty_to_reinvoice
+                ):
```
HekkiMelody, line 72: **suggestion** (non-blocking): This check is almost the same as the checks in the other two methods in this file.

1. Maybe `state` could be the same as the others for consistency
2. Maybe the check could be extracted into its own method and reused, to make the module more DRY?

HekkiMelody (approved):
A suggestion, but otherwise code LGTM

HekkiMelody (approved):
Code review, LGTM

PR #2103, opened 2025-09-11
[IMP] account_invoice_merge: Added so that the invoice_incoterm_id field is considered and must match when merging invoices.

dferest:
Added so that the invoice_incoterm_id field is considered and must match when merging invoices.


dferest:
Now when you try to merge two invoices with different incoterm, Odoo display an error
image

Additionaly when you merge the invoices the incoterm of the invoices move to the new invoice.
image
image

PR #2109, opened 2025-09-23
[16.0][FIX] account_invoice_fixed_discount,account_invoice_triple_discount: Remove excludes

hildickethan:
Follow-up to the changes merged in https://github.com/OCA/account-invoicing/pull/1840
Since having these modules in excludes ends up being a bit problematic and messy in github and existing installations that might have both installed, I think the better approach is nullifying the functionality of one of them when both are installed.
IMO if both are set the fixed discount should take priority, since the general sentiment when adding a fixed discount you're thinking "I just want to discount 20€ from this line".

Optimally, you would have a way of both co-existing and computing taking each other into account (ideally also the general and global discount modules would be included in this), but that would require a glue module. Since these modules have co-existed until now without that feature taken care of, I don't think it's a practical use case and not worth the time to implement it until someone needs it.

With these changes the modules can be co-installed and you can use either, just not simultaneously on the same line as the fixed will take priority.

If this seems OK, I will supersede https://github.com/OCA/sale-workflow/pull/2885 and add them so we can get that merged also and finish this refactor

<details><summary>Screenshots</summary>
<p>

Captura desde 2025-09-23 16-22-15
Captura desde 2025-09-23 16-22-23
Captura desde 2025-09-23 16-22-31

</p>
</details>


legalsylvain (changes requested):
Hi @hildickethan Thanks a lot to try to fix this problem. 

Unfortunately, the current design you propose generate trouble in many places I think. For exemple, the data in the reports will be wrong, when discount_fixed is set, also the PDF will display bad things to customers. 

> Since these modules have co-existed until now without that feature taken care of,

In fact no. AFAIK, both modules never worked together. But the problem was hidden. If you don't think so, please mention the version where both module installed was working correctly. Thanks !

hildickethan:
> > Since these modules have co-existed until now without that feature taken care of,
> 
> In fact no. AFAIK, both modules never worked together. But the problem was hidden. If you don't think so, please mention the version where both module installed was working correctly. Thanks !

What I meant is that they could be co-installed even if they didn't work together, sorry I could've worded it better

> Unfortunately, the current design you propose generate trouble in many places I think. For exemple, the data in the reports will be wrong, when discount_fixed is set, also the PDF will display bad things to customers.

While this is true, it's easily fixable by the user setting the triple fields to 0 after deciding they want a fixed discount instead. Realistically since they never worked together, I don't think there is anyone that is going to be upset about that.

But this comment did make me check the report and I saw that the modules messed with each other there, never showing the fixed discount. I just pushed some changes to fix that

If you have a fixed discount
image
With both, yes it looks bad but like I said IMO if someone ever got themselves into this position they can also fix it
image

legalsylvain:
> What I meant is that they could be co-installed even if they didn't work together, sorry I could've worded it better

Sorry, but I'm not conviced by the objective of this PR. Could we talk about it in the general issue, to avoid you loose time implementing something ? https://github.com/OCA/sale-workflow/issues/3638

account_invoice_triple_discount/models/triple_discount_mixin.py
```diff
@@ -51,10 +51,17 @@ class TripleDiscountMixin(models.AbstractModel):
 
     @api.depends(lambda self: self._get_multiple_discount_field_names())
     def _compute_discount(self):
+        res = True
+        if hasattr(super(), "_compute_discount"):
+            res = super()._compute_discount()
         for line in self:
+            if "discount_fixed" in line._fields:
+                if line.discount_fixed != 0:
+                    continue
```
legalsylvain:
```suggestion
        lines = self
        if "discount_fixed" in line._fields:
            lines = self.filtered(lambda x: x.discount_fixed != 0)            
        for line in lines:
`

Comment (legalsylvain) [reply account_invoice_triple_discount/models/triple_discount_mixin.py]: Another simplier implementation is : 

    def _get_lines_compute_discount(self):
        return self

and, then, in the glue module, implement.

    def _get_lines_compute_discount(self):
        return self.filtered(lambda x: x.discount_fixed != 0)

File: account_invoice_triple_discount/readme/DESCRIPTION.rst [hunk 2]
```diff
@@ -1 +1,3 @@
 This module allows to have three successive discounts on each invoice line.
+
+If the module `account_invoice_fixed_discount` is installed and a fixed discount is set, it will take priority over anything set in the triple discount fields.
```
legalsylvain, line 3: I think that for end users, having the field hidden by a glue module is a better UX.

account_invoice_triple_discount/models/triple_discount_mixin.py
hildickethan: Good change, currently in draft because I'm testing it more thoroughly and seeing other things that need to be adapted

legalsylvain (commented):
if I summary, do you think it's possible to add a little glue module that 
- adds a warning for end users, as you proposed here https://github.com/OCA/sale-workflow/issues/3638#issuecomment-3327042441 (or better, to hide triple discount fields), if fixed discount is != 0.


Thanks !

hildickethan:
@legalsylvain With the new changes and the glue module this is what I have ended up with, what do you think? Screenshots inside the details

<details><summary>Example invoice PDF with a line with fixed discount and triple (wrong), normal triple discount and a normal fixed discount</summary>
<p>

image
image

</p>
</details> 
I chose to hide the fixed discount cell when it has to show triple discount, else it would do something like `0,00€ (or 77.50%)`, but it does look a bit ugly so I'm not sure about it. Not keen on calculating back the € value of the total discount to show in the field. Maybe a value like N/A would be better UX.
But it is the most extreme case of using both discounts on the same invoice albeit different lines (in this case the user would correct the first one after seeing the warning and avoid the 99% on the pdf)

<details><summary>Invoice with only fixed discounts</summary>
<p>

image
image

</p>
</details> 

<details><summary>Invoice with only % discounts</summary>
<p>

image
image

</p>
</details>

hildickethan:
~~Using the copier template it deleted the readme addons content, last time I just copied it back but since now there's a new module being added I'm not sure. Does the bot generate the readme when merging and I can just leave it as is, or should I run the script myself?~~
Generated it myself

legalsylvain (approved):
Thanks a lot !

Code review. No test. Lgtm !

Thanks you a lot to take time for this compatibility

hildickethan:
One last push for i18n files so the warning can be translated

legalsylvain:
@OCA/accounting-maintainers : could you take a look over there ? 

thanks !

lmignon (approved):
LGTM Thank you for the fix

lmignon (approved):
LGTM (Functional tests) + Code review

legalsylvain (commented):
/ocabot merge nobump

legalsylvain:
hi @OCA/accounting-maintainers Could you merge this one ? 
thanks !

account_invoice_fixed_triple_discount/tests/test_account_move.py
```diff
@@ -0,0 +1,72 @@
...
+                line.discount_fixed = 10.0
+        lines = invoice.invoice_line_ids._get_lines_to_compute_discount()
+        self.assertEqual(lines, invoice.invoice_line_ids - invoice.invoice_line_ids[0])
+
+    def test_03_should_copy_discount_to_discount1(self):
+        invoice = self._create_invoice(discount=10.00)
+        self.assertAlmostEqual(invoice.invoice_line_ids[0].discount1, 10)
+        self.assertAlmostEqual(invoice.invoice_line_ids[0].discount, 10)
+        invoice_2 = self._create_invoice(discount=10.00, discount_fixed=10.00)
+        self.assertAlmostEqual(invoice_2.invoice_line_ids[0].discount1, 0)
```
sergiocorato, line 73: Could you add a test for `discount` here?

account_invoice_fixed_triple_discount/tests/test_account_move.py
```diff
@@ -0,0 +1,72 @@
...
+        with Form(invoice) as invoice_form:
+            with invoice_form.invoice_line_ids.edit(0) as line:
+                line.discount_fixed = 10.0
+        lines = invoice.invoice_line_ids._get_lines_to_compute_discount()
+        self.assertEqual(lines, invoice.invoice_line_ids - invoice.invoice_line_ids[0])
+
+    def test_03_should_copy_discount_to_discount1(self):
+        invoice = self._create_invoice(discount=10.00)
+        self.assertAlmostEqual(invoice.invoice_line_ids[0].discount1, 10)
+        self.assertAlmostEqual(invoice.invoice_line_ids[0].discount, 10)
```
sergiocorato, line 68: Could you add a test for `discount_fixed` here?

sergiocorato (approved):
Only a little remark, only code check, thanks!

sergiocorato:
depends on  #2268

sergiocorato:
@hildickethan a rebase now should be very useful!

hildickethan:
@sergiocorato tests added + rebase

sergiocorato:
Good work!
/ocabot merge major

PR #2112, opened 2025-09-29
[18.0][MIG] account_partner_journal: Migration to 18.0

P-H-Phuc:
This module was initially part of the [Foodcoop set of addons](https://github.com/AwesomeFoodCoops/odoo-production/tree/12.0). We're now migrating it to 18.0 and taking this occasion to contribute it to OCA.

**Features**: This module allows to define a default Purchase Journal on a partner. This journal will be used by default when an invoice is created against this partner.
**Changes**: `account.invoice` has been replaced by `account.move` (see [odoo/odoo@bc131c0](https://github.com/odoo/odoo/commit/bc131c0cfb51c953de8ec41fb820c8c7831eefb5)).


account_partner_journal/models/account_move.py
```diff
@@ -0,0 +1,32 @@
...
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+from odoo import api, models
+
+
+class AccountInvoice(models.Model):
```
lmignon:
```suggestion
class AccountMove(models.Model):
```

account_partner_journal/models/account_move.py
```diff
@@ -0,0 +1,32 @@
...
+    _inherit = "account.move"
+
+    @api.onchange("partner_id", "company_id")
+    def _onchange_partner_id(self):
+        res = super()._onchange_partner_id()
+        if self.partner_id.default_purchase_journal_id:
+            self.journal_id = self.partner_id.default_purchase_journal_id
+        return res
```
lmignon: In Odoo 18, journal_id is a computed field with an inverse_method. https://github.com/odoo/odoo/blob/19.0/addons/account/models/account_move.py#L160

You should not declare an `onchange` method but override the `_search_default_journal` method declared into the base addon https://github.com/odoo/odoo/blob/19.0/addons/account/models/account_move.py#L886

```suggestion
def _search_default_journal(self):
    if self.partner_id.default_purchase_journal_id:
        return self.partner_id.default_purchase_journal_id
    return super()._search_default_journal()
```

account_partner_journal/models/res_partner.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    default_purchase_journal_id = fields.Many2one(
+        "account.journal",
+        "Default Purchase Journal",
+        domain="[('type', '=', 'purchase')]",
+    )
```
lmignon, line 19: This field should be company dependent

account_partner_journal/models/account_move.py
```diff
@@ -0,0 +1,32 @@
...
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
```
lmignon: Can you shorten the headers as it's done into others OCA addons
```
# Copyright 2016-Today: La Louve (<http://www.lalouve.net/>)
```

lmignon (changes requested):
Thank you for the contribution. Some comments....

account_partner_journal/models/account_move.py
lmignon: In Odoo 18, journal_id is a computed field with an inverse_method. https://github.com/odoo/odoo/blob/19.0/addons/account/models/account_move.py#L160

You should not declare an `onchange` method but override the `_search_default_journal` method declared into the base addon https://github.com/odoo/odoo/blob/19.0/addons/account/models/account_move.py#L886

```suggestion
def _search_default_journal(self):
    if self.partner_id.default_purchase_journal_id:
        return self.partner_id.default_purchase_journal_id
    return super()._search_default_journal()
``

**Comment by P-H-Phuc** [comment]
Thank you @lmignon , resolved!

File: account_partner_journal/models/res_partner.py [hunk 5]
```diff
@@ -0,0 +1,39 @@
...
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
```
lmignon: This header should also be shortened

rousseldenis:
@P-H-Phuc

account_partner_journal/models/res_partner.py
P-H-Phuc: @rousseldenis , @lmignon Sorry for the mistake. I’ve updated it and .po file

account_partner_journal/views/res_partner_view.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- ********************************************************************** -->
```
rousseldenis: In OCA, it's not the way. You should add licence here too. Like : https://github.com/OCA/account-invoicing/blob/16.0/account_global_discount/views/account_invoice_views.xml#L2

account_partner_journal/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+# Copyright (C) 2019-Today: La Louve (<https://cooplalouve.fr>)
+# Copyright (C) 2019-Today: Druidoo (<https://www.druidoo.io>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
```
rousseldenis:
```suggestion
# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
```

account_partner_journal/views/res_partner_view.xml
```diff
@@ -0,0 +1,23 @@
...
+    <record id="view_partner_property_form" model="ir.ui.view">
+        <field name="name">res.partner.form.journal</field>
+        <field name="model">res.partner</field>
+        <field name="priority">2</field>
+        <field name="inherit_id" ref="account.view_partner_property_form" />
+        <field name="arch" type="xml">
+            <field name="property_account_payable_id" position="after">
+                <field
+                    name="default_purchase_journal_id"
+                    required="supplier_rank &gt; 0"
```
rousseldenis: @P-H-Phuc This should not be mandatory as the code will fallback to the standard behavior if not filled in.

account_partner_journal/models/res_partner.py
```diff
@@ -0,0 +1,22 @@
...
+
+    default_purchase_journal_id = fields.Many2one(
+        "account.journal",
+        "Default Purchase Journal",
+        company_dependent=True,
+        domain="[('type', '=', 'purchase')]",
+    )
+
+    @api.onchange("parent_id")
+    def _onchange_parent_id(self):
```
rousseldenis: You should not use onchanges anymore.

Prefer changing the field to computed/stored/readonly=False and this into compute.

P-H-Phuc:
@rousseldenis Thank you, I just updated it

PR #2115, opened 2025-10-02
[18.0][IMP] account_invoice_warn_message: show warning for vendor invoices too

AaronHForgeFlow:
The warning already shows in the standard Odoo onchange_partner method but the banner is not in vendor bills. I think it should be there.

It shows the warning anyway:
image

no harm to show the banner:
image

### DIFF
--- a/account_invoice_warn_message/models/account_move.py
+++ b/account_invoice_warn_message/models/account_move.py
@@ -19,7 +19,8 @@ class AccountMove(models.Model):
         for rec in self:
             if (
                 rec.partner_id
-                and rec.move_type in ("out_invoice", "out_refund")
+                and rec.move_type
+                in ("out_invoice", "out_refund", "in_invoice", "in_refund")
                 and rec.state == "draft"
             ):
                 if (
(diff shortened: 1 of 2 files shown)


luisDIXMIT (commented):
Hi @AaronHForgeFlow, the tests are failing because they assert that the invoices do not raise the warning. If you want the warning to be raised, you should change the test to use assertTrue.
Anyway, if the test was originally written like this, I believe the warning should not be raised. Despite that, it works properly from a functional point of view. :+1:

AaronHForgeFlow:
@luisDIXMIT Thank you! I updated the tests. The warning is showing anyway, it is just the banner that is not there, so i think it is consistent.

luisDIXMIT (approved):
Tested on runboat and LGTM!!

yostashiro:
This is still for 17.0, but I thought you might find it interesting: https://github.com/OCA/web/pull/3305

We are working on adding functionality to configure context-aware banner messages on any form view. We plan to migrate this to 18.0 in the coming days.

PR #2135, opened 2025-10-14
[18.0][FIX] account_receipt_journal: Create journals in multicompany

EmilioPascual:
Since version 17 after this commit https://github.com/odoo/odoo/commit/512574861691f425ec6a17f20fe4b586bb88a299, journals are created in method with decorator `@template`, by this way the journals are created with company xml_id and also when you load or reload chart template.

@eLBati @yajo @Shide @rafaelbn please review, thank you.

MT-12135 @moduon


yajo (approved):

chienandalu (approved):

EmilioPascual:
Tests are red because of what seems to be an unrelated error in module `account_move_cancel_confirm`.

```
2025-10-14 11:35:49,193 286 ERROR odoo odoo.addons.account_move_cancel_confirm.tests.test_account_move_cancel_confirm: FAIL: TestAccountMoveCancelConfirm.test_02_cancel_payment
  File "/__w/account-invoicing/account-invoicing/account_move_cancel_confirm/tests/test_account_move_cancel_confirm.py", line 87, in test_02_cancel_payment
    self.assertEqual(payment.state, "paid")
AssertionError: 'in_process' != 'paid'
- in_process
+ paid
```

@kittiu as maintainer of the module, can you fix the test, please? I saw more CI down.

eLBati (approved):
LGTM

account_receipt_journal/data/account_journal_data.xml
```diff
@@ -1,18 +0,0 @@
-<?xml version="1.0" encoding="utf-8" ?>
-<odoo noupdate="1">
-    <record id="sale_receipts_journal" model="account.journal">
```
celm1990, line 3: What happens if you have this module installed and there are account moves using these journals? Please preserve these XML IDs, at least for the main company.

celm1990 (changes requested):

account_receipt_journal/data/account_journal_data.xml
EmilioPascual: So, I prefer to create a migration script to assign the new xml_id to these journals. What do you think @celm1990?

celm1990: > So, I prefer to create a migration script to assign the new xml_id to these journals. What do you think @celm1990?


Yes, it seems reasonable for these records in the main company. For new companies, it’s OK to generate new journals without XML-IDs. Thanks!

Shide (approved):

account_receipt_journal/__manifest__.py
celm1990, line 1: You need to increase the module version so that the migration script is executed correctly; otherwise, the migration script won’t run

celm1990 (changes requested):

account_receipt_journal/__manifest__.py
EmilioPascual: You're right, thanks for letting me know 😀

account_move_cancel_confirm/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 
 {
     "name": "Account Move Cancel Confirm",
-    "version": "18.0.1.0.0",
+    "version": "18.0.1.0.1",
```
celm1990, line 6: The manifest version must match the migration folder. In this case, the version should be `18.0.1.0.3`, or you should rename the migration folder to `18.0.1.0.1`.

celm1990 (changes requested):

account_move_cancel_confirm/__manifest__.py
EmilioPascual: Updated

celm1990 (approved):
LGTM

yajo (approved):

rafaelbn (approved):
/ocabot merge minor

PR #2139, opened 2025-10-16
[19.0][MIG] account_invoice_refund_reason: Migrate to 19.0

Tatider:
- Ensured that only newly created refunds would get the reason updated:
   - This is fixing the problem when refunds created before module installation would get updated, because they don't have "reason_id" set.
   - Ensured with extra asserts in tests, that the field I am using instead ("new_move_ids") contains all created invoices.
 - An update of tests was required because in Odoo 19.0, the database by default doesn't have Demo Data
 - Also I have removed duplicated action in tests:
   - old function `create_refund_wizard` was already doing the refund, and then tests were calling it again (https://github.com/odoo/odoo/blob/19.0/addons/account/wizard/account_move_reversal.py#L184)

Screenshot 2025-10-16 at 15 20 32

Screenshot 2025-10-16 at 14 52 40
Screenshot 2025-10-16 at 15 15 54


cav-adhoc (approved):
LGTM! Functional test video: https://drive.google.com/file/d/11-piUUNHcnVYHgmR-F4OHOfe5wCUhboK/view

bosd (commented):
Looks good,
Can you please squash the administrative commits?

Tatider:
> Looks good, Can you please squash the administrative commits?

I have squashed the commit, but with the latest Odoo code, another module from this repo is failing the tests - `account_invoice_pricelist` with rounding issue...
I have made the correction https://github.com/OCA/account-invoicing/pull/2139/changes/f28ad61a399a3cd7450d48bf1af083526d5239df, but someone need to check, that it is the correct change... Could @pablo-cort-s73 take a look as the person who migrated the "account_invoice_pricelist" module?

PR #2140, opened 2025-10-16
[18.0][ADD] account_tax_legal_notes_translate: added translate to tax legal notes

sabrinaRMartin:
@moduon @EmilioPascual MT-12165

 This module allows you to have translations for the Legal Notes field for taxes.

https://www.loom.com/share/f05273ed508b459aacaab140c1ad11af?sid=54f3d2ae-8ec3-4bf7-a729-0131e6e88e79

PR #2133, opened 2025-10-14
[18.0][FIX] account_move_tier_validation: fix revert to draft raising error

SilvioC2C:
When reverting a move back to draft, ``ValidationError`` was raised due to field ``state`` being written after validating the move. This commit fixes the issue.

PR #2141, opened 2025-10-17
[18.0][FIX] account_move_cancel_confirm: remove assert related to payment state in test

carlos-lopez-tecnativa:
This assert is not related to the module, because this module doesn’t change anything related to the payment state, and this assert becomes problematic, so we removed it.

Technically: when a vendor bill is paid, the payment state remains in_process instead of paid. This occurs because in _compute_state, when considering whether all invoices are paid, it uses the field reconciled_invoice_ids, which only takes sales documents.
See https://github.com/odoo/odoo/blob/1cac54db8634267a780b4011291f1e8a80ac5f5b/addons/account/models/account_payment.py#L428

https://github.com/odoo/odoo/blob/1cac54db8634267a780b4011291f1e8a80ac5f5b/addons/account/models/account_payment.py#L699

image

This PR fixes all CI failures in V18, such as: https://github.com/OCA/account-invoicing/pull/2140 https://github.com/OCA/account-invoicing/pull/2135 https://github.com/OCA/account-invoicing/pull/2133 https://github.com/OCA/account-invoicing/pull/2063

@Tecnativa @pedrobaeza @sabrinaRMartin could you review this please?

### DIFF
--- a/account_move_cancel_confirm/tests/test_account_move_cancel_confirm.py
+++ b/account_move_cancel_confirm/tests/test_account_move_cancel_confirm.py
@@ -84,7 +84,6 @@ class TestAccountMoveCancelConfirm(TransactionCase):
         payment = payment_register_form.save()
         payment.action_create_payments()
         payment = self.payment_model.search([], order="id desc", limit=1)
-        self.assertEqual(payment.state, "paid")
         # Click cance, cancel confirm wizard will open. Type in cancel_reason
         res = payment.action_cancel()
         ctx = res.get("context")


EmilioPascual (approved):

EmilioPascual (approved):
I don't think it's right to change a test to pass the CI, but I approve because it's blocking everyone's CI.

pedrobaeza:
For me that assert should go out. This module is not modifying that state, only asking for a confirmation on cancel, so please remove completely that assert.

pedrobaeza (approved):
/ocabot merge patch

PR #2149, opened 2025-10-30
19.0 [MIG] account move tier validation

Wodran14:


lmignon (changes requested):
@Wodran14 Thank you for the migration. It should be nice to temporarily add a test dependency on https://github.com/OCA/server-ux/pull/1176 to allow functional tests with runboat and makes checks for test :heavy_check_mark:  into the CI. Moreover should this addon be moved to the new repo https://github.com/OCA/tier-validation? see https://github.com/OCA/purchase-workflow/pull/2857

bosd:
superseeded by https://github.com/OCA/tier-validation/pull/3

BT-jmichaud:
What is the state here? Is someone working on it?

PR #2158, opened 2025-11-07
[18.0][ADD] partner_account_move_line

P-H-Phuc:
This module was initially part of the [Foodcoop set of addons](https://github.com/AwesomeFoodCoops/odoo-production/tree/12.0). We're now migrating it to 18.0 and taking this occasion to contribute it to OCA.

image


luisDIXMIT (approved):
Tested on runboat and LGTM. I'm not sure if the commit history is correct.

PR #2163, opened 2025-11-12
[16.0][FIX] account_invoicing_pricelist: compute price recalcule

adrip-s73:
This change is to prevent the unit price from being recalculated since when making a refund invoice the price is recalculated and the price of the product is taken.

### DIFF
--- a/account_invoice_pricelist/models/account_move.py
+++ b/account_invoice_pricelist/models/account_move.py
@@ -62,12 +62,19 @@ class AccountMove(models.Model):
             lambda r: r.state == "draft"
         ).invoice_line_ids._compute_price_unit()
 
+    def action_switch_invoice_into_refund_credit_note(self):
+        return super(
+            AccountMove, self.with_context(avoid_price_unit_compute=True)
+        ).action_switch_invoice_into_refund_credit_note()
+
 
 class AccountMoveLine(models.Model):
     _inherit = "account.move.line"
 
     @api.depends("quantity")
     def _compute_price_unit(self):
+        if self.env.context.get("avoid_price_unit_compute", False):
+            return True
         res = super()._compute_price_unit()
         for line in self:
             if not line.move_id.pricelist_id:
--- a/account_invoice_pricelist/tests/test_account_move_pricelist.py
+++ b/account_invoice_pricelist/tests/test_account_move_pricelist.py
@@ -316,3 +316,11 @@ class TestAccountMovePricelist(BaseCommon):
             self.invoice.with_context(force_check_currecy=True).write(
                 {"currency_id": self.usd_currency.id}
             )
+
+    def test_switch_invoice_into_refund_credit_note(self):
+        self.invoice.pricelist_id = self.sale_pricelist_with_discount.id
+        self.invoice.invoice_line_ids.write({"price_unit": -100.00})
+        self.invoice.action_switch_invoice_into_refund_credit_note()
+        self.assertEqual(self.invoice.move_type, "out_refund")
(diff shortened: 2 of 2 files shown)


luisDIXMIT (commented):
Tittle name must be: [16.0][FIX] account_invoice_pricelist: compute price recalcule

luisDIXMIT (commented):
Hi @adrip-s73, I don’t understand your change. Before your modification, when I tried to switch a customer invoice to a refund, the unit prices on the lines weren’t updated unless I modified the line or clicked the “Update Price” button. Could you explain how to test this? Maybe I’m missing something.

santostelmo:
@luisDIXMIT I have the same issue on my project. If you run the test bellow without this change it will fail
@adrip-s73 your code LGTM. Could you please add a unit test ? Something like:
```
def test_switch_invoice_into_refund_credit_note(self):
        """ Test that switching an invoice into a refund credit note does not
        update the price_unit according to the invoice's pricelist.
        """
        self.invoice.pricelist_id = self.sale_pricelist_with_discount.id
        self.invoice.invoice_line_ids.write({"price_unit": -100.00})
        self.assertEqual(self.invoice.amount_total, -230.00)
        self.assertEqual(self.invoice.move_type, "out_invoice")

        # Switch to refund credit note
        self.invoice.action_switch_invoice_into_refund_credit_note()

        self.assertEqual(self.invoice.move_type, "out_refund")
        self.assertEqual(self.invoice.amount_total, 230.00)
```

cyrilmanuel (approved):

cyrilmanuel (approved):

santostelmo (approved):

account_invoice_pricelist/models/account_move.py
```diff
@@ -62,12 +62,19 @@ def button_update_prices_from_pricelist(self):
...
+            AccountMove, self.with_context(avoid_price_unit_compute=True)
+        ).action_switch_invoice_into_refund_credit_note()
+
 
 class AccountMoveLine(models.Model):
     _inherit = "account.move.line"
 
     @api.depends("quantity")
     def _compute_price_unit(self):
+        if self.env.context.get("avoid_price_unit_compute", False):
```
simahawk, line 76: I'm out of context here but... is the value already available for the cache?
I would expect that, either you call super before returning or you do `self.price_unit = False`
Am I missing anything? 🤔

simahawk (commented):
nitpicking: can you pls drop the odoo version from the commit and include the description of the PR in the commit itself?
Note: if next time you put it directly in the commit, you'll have it automatically in the description 😉

account_invoice_pricelist/models/account_move.py
adrip-s73: The goal here is to prevent the pricelist logic from overwriting the prices during the refund process. In this specific flow, the price_unit is already correctly set on the lines before this compute is triggered we ensure the ORM keeps the existing value in the cache/database instead of recalculating it

adrip-s73:
@simahawk commit changed !!

cyrilmanuel:
hi, @adrip-s73 this branch was fix with commit https://github.com/OCA/account-invoicing/commit/b2bdef25050e729ce98582a55afc5cc9eb3b9da6 , could you rebase the PR and re-call Simahawk to merge this one please ???

adrip-s73:
@simahawk can u merge this pr ?

cyrilmanuel:
@OCA/accounting-maintainers could you merge this PR please ?

PR #2167, opened 2025-11-20
18.0 mig account invoice mass sending from 17

marionumza:


marionumza:
I was finally able to install it and the work queue is actually being created

PR #2172, opened 2025-11-24
[18.0][FIX] sale_stock_picking_invoicing: Review models on test in order to ensure the equality when other modules are installed

etobella:
For example, this test fails if sale_commission_oca and sale_stock_picking_invoicing are installed


luisDIXMIT (approved):
LGTM!

LoisRForgeFlow (approved):
LGTM

etobella:
The mistake comes from an Odoo change...

https://github.com/odoo/odoo/commit/7f810d306aaad9e6a89a412cedfde48232b4c4dd

So nice that they don't do breaking changes in production....

PR #2177, opened 2025-11-26
[17.0][ADD] account_invoice_currency_by_partner: New module

eduezerouali-tecnativa:
cc @Tecnativa TT59474

New module that lets you set a default currency on a partner, which is then automatically applied to vendor bills.

ping @sergio-teruel @pedrobaeza


account_vendor_bill_currency_by_partner/models/res_partner.py
```diff
@@ -0,0 +1,14 @@
...
+from odoo import fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    vendor_currency_id = fields.Many2one(
+        comodel_name="res.currency",
+        string="Currency",
+        help="This field set currency_id in vendor bill automatically",
```
pedrobaeza:
```suggestion
        help="If field, this value is set as currency in vendor bills for this partner",
```

pedrobaeza (approved):

account_vendor_bill_currency_by_partner/models/account_move.py
```diff
@@ -0,0 +1,19 @@
...
+    @api.depends("partner_id")
+    def _compute_currency_id(self):
+        for invoice in self:
+            if invoice.partner_id.vendor_currency_id and invoice.move_type in [
+                "in_invoice",
+                "in_refund",
+            ]:
+                invoice.currency_id = invoice.partner_id.vendor_currency_id
+            else:
+                return super()._compute_currency_id()
```
victoralmau:
```suggestion
    @api.depends("partner_id")
    def _compute_currency_id(self):
        move_types = ["in_invoice", "in_refund"]
        _self = self.filtered(
            lambda x: x.partner_id.vendor_currency_id 
            and invoice.move_type in move_types
        )
        for invoice in _self:
            invoice.currency_id = invoice.partner_id.vendor_currency_id
        return super(AccountMove, (self - _self))._compute_currency_id()
```

victoralmau (changes requested):

victoralmau (approved):
Code review OK

eduezerouali-tecnativa:
I made improvements, and now its working with customer invoice too. Also add some test.
ping @victoralmau @pedrobaeza @sergio-teruel

sergio-teruel (approved):

pedrobaeza (changes requested):
Please check independently if each field is filled, or you may assign an empty value.

account_invoice_currency_by_partner/models/res_partner.py
```diff
@@ -0,0 +1,21 @@
...
+
+    vendor_currency_id = fields.Many2one(
+        comodel_name="res.currency",
+        string="Currency",
+        help="If field, this value set currency in vendor bills for this partner",
+        company_dependent=True,
+    )
+    customer_currency_id = fields.Many2one(
+        comodel_name="res.currency",
+        string="Currency",
```
sergio-teruel: don't use the same label for both fields...

eduezerouali-tecnativa: done, anyways open to suggestions for the label.

eduezerouali-tecnativa:
@pedrobaeza did add another condition before assigning the value.

account_invoice_currency_by_partner/models/account_move.py
```diff
@@ -0,0 +1,27 @@
+# Copyright 2025 Tecnativa - Eduardo Ezerouali
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+
+from odoo import api, models
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    @api.depends("partner_id.vendor_currency_id", "partner_id.customer_currency_id")
```
pedrobaeza: Watch out! You shouldn't put a dependency that affects previous records. Imagine all the past invoices being changed because you now change the currency in the partner...

```suggestion
    @api.depends("partner_id")
```

eduezerouali-tecnativa: change done.

account_invoice_currency_by_partner/tests/test_currency_by_partner.py
```diff
@@ -0,0 +1,87 @@
+# Copyright 2025 Tecnativa - Eduardo Ezerouali
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
+
+from odoo.addons.base.tests.common import BaseCommon
+
+
+class TestAccountMoveCurrency(BaseCommon):
```
pedrobaeza: Why not depending on `AccountInvoicingCommon` for avoiding most of the stuff and be sure the CoA is loaded?

eduezerouali-tecnativa: Good point, did not know much about `AccountTestInvoicingCommon`, but after a quick research this will make sure `account.move` works correctly.

luisDIXMIT (commented):
I have this setting:

image

But when I create a vendor bill for that partner, the GBP currency is used.

image

On the other hand, I created a purchase order and the currency was USD, which was expected. But then I created an invoice from that purchase order and the currency was still USD, even though I’ve set GBP for vendor bills. Maybe this is the expected behaviour to maintain consistency between the purchase order and the invoice. Let me know.

eduezerouali-tecnativa:
@luisDIXMIT I have checked, and when you create a vendor bill for that partner, EUR is used. Am I missing something?
We did not intend to change anything in the purchase workflow, but in any case, there is a `supplier currency` field that actually manages this.

luisDIXMIT:
> @luisDIXMIT I have checked, and when you create a vendor bill for that partner, EUR is used. Am I missing something? We did not intend to change anything in the purchase workflow, but in any case, there is a `supplier currency` field that actually manages this.

Yes, you're right. I’ve now run the same test and it works. However, when I tested with a customer invoice, the currency wasn’t updated, and I even got this error when saving the invoice. The pricelist currency is USD, and the invoice shows USD as well (maybe the correct GBP currency is being used in the background? I’m not sure). Anyway, I don’t intend to block this PR.

image

sergio-teruel:
@eduezerouali-tecnativa Can you review the tests please?

eduezerouali-tecnativa:
> > @luisDIXMIT I have checked, and when you create a vendor bill for that partner, EUR is used. Am I missing something? We did not intend to change anything in the purchase workflow, but in any case, there is a `supplier currency` field that actually manages this.
> 
> Yes, you're right. I’ve now run the same test and it works. However, when I tested with a customer invoice, the currency wasn’t updated, and I even got this error when saving the invoice. The pricelist currency is USD, and the invoice shows USD as well (maybe the correct GBP currency is being used in the background? I’m not sure). Anyway, I don’t intend to block this PR.
> image

This is because it is not compatible with `account_invoice_pricelist`. That module force the `currency_id` to be the same as pricelist one https://github.com/OCA/account-invoicing/blob/5b94ba3ba1d48ffbe2344fb12c75f26819facdae/account_invoice_pricelist/models/account_move.py#L46, moreover it also need to be define at partner level. https://github.com/OCA/account-invoicing/blob/5b94ba3ba1d48ffbe2344fb12c75f26819facdae/account_invoice_pricelist/models/account_move.py#L36.

eduezerouali-tecnativa:
ping @sergio-teruel

eduezerouali-tecnativa:
@sergio-teruel @pedrobaeza I have put the model to run on separated test container, but I'm not sure if i can do it that way. In case that is the way ,should put it on an other commit/PR? other option it is to enable test flag on `account_invoice_pricelist`
https://github.com/OCA/account-invoicing/blob/5b94ba3ba1d48ffbe2344fb12c75f26819facdae/account_invoice_pricelist/models/account_move.py#L46

pedrobaeza:
I prefer to not do that, as this is only hidden the incompatibility problem, and it will raise again if both conflicting modules are used in the same instance. Instead, you can test `config["test_enable"]` variable to check if you are in test mode, and pass context keys for not triggering the behavior in that cases. Check for other examples across OCA.

eduezerouali-tecnativa:
ping @pedrobaeza @juancarlosonate-tecnativa could you review please?

juancarlosonate-tecnativa (changes requested):
I'm not sure this module is the right approach.

First, on the vendor side, we are duplicating functionality that already exists in core: the property_purchase_currency_id field, which already fixes the currency on vendor bills:

https://github.com/odoo/odoo/blob/17.0/addons/purchase/models/res_partner.py#L35

With purchase installed, vendor_currency_id and property_purchase_currency_id end up coexisting in the same Purchase section of the contact form while effectively doing the same thing.

Second, on the customer side, if the pricelist is properly configured on the customer, we already have the correct currency. The pricelist provides the currency for both the sales order and the invoice. Therefore, customer_currency_id overlaps with existing functionality and, more importantly, decouples the pricelist from the currency, which is exactly where the problem starts.

A pricelist has a single currency and prices are computed in that currency. If the pricelist is in EUR but customer_currency_id is forced to USD, you end up with amounts calculated in EUR being displayed as USD, which is clearly incorrect.

Shouldn't these use cases be solved through a properly configured pricelist, rather than introducing a currency field on the partner that duplicates core functionality on purchases and conflicts with the pricelist logic on sales?

eduezerouali-tecnativa:
Well I understand your point, but what happens if you only want to use `account`. In that case even if module `product` is install you will not have access to pricelist unless you install `sale`. 

Anyways I have add  this to roadmap.

juancarlosonate-tecnativa (approved):

eduezerouali-tecnativa:
ping @pedrobaeza

pedrobaeza (approved):
/ocabot merge nobump

pedrobaeza:
@sbidoul :pray:

sbidoul:
name reserved

PR #2180, opened 2025-11-28
[16.0][ADD] account_invoice_partner_company_only: Restrict partner…

maisim:
… to companies


luisDIXMIT (commented):
I tested it on runboat and it doesn’t seem to be working.

image

maisim:
> I tested it on runboat and it it doesn’t seem to be working.
image

The domain seems oververvrited by another module in the repo. That means this is probably not the right way to do it. Maybe with a Python constraint? Let me take a look.

maisim:
After some tests, using a constraint that throws an exception is too intrusive, so I just left the contact filter for "company" types in the form.

I'm moving this pull request to draft status because this module will be an optional dependency (using preferences) of https://github.com/OCA/sale-workflow/pull/4039, and I'll be making some changes. This time, I'll wait until everything is ready before requesting a review. :)

rrebollo:
@maisim the name of the addon should be `account_invoice_partner_company_only` or similar, I think the `_id` part should go off.

> The domain seems oververvrited by another module in the repo. That means this is probably not the right way to do it. Maybe with a Python constraint? Let me take a look.

Look at my review comments on the sale-workflow addon.

maisim:
Hi @rrebollo,
I made the same changes as in https://github.com/OCA/sale-workflow/pull/4042 and https://github.com/OCA/project/pull/1611
- Rename the module
- Add negative tests
- Fix copyright
- use base_view_inheritance_extension

Let me know if that's okay with you

account_invoice_partner_company_only/tests/test_account_invoice_partner_company_only.py
```diff
@@ -0,0 +1,83 @@
...
+        domain_found = False
+        field = partner_fields[0]
+        domain = field.get("domain")
+        if "('is_company', '=', True)" in domain:
+            domain_found = True
+
+        self.assertTrue(
+            domain_found,
+            "partner_id field should have a domain restricting to companies",
+        )
```
rrebollo, line 50: Here, you're testing whether Odoo is functioning as expected, but what you should be testing is the feature of this addon—not Odoo's built-in functionality.

account_invoice_partner_company_only/tests/test_account_invoice_partner_company_only.py
```diff
@@ -0,0 +1,83 @@
...
+            {
+                "partner_id": self.partner_company.id,
+                "move_type": "out_invoice",
+            }
+        )
+        self.assertEqual(
+            invoice.partner_id,
+            self.partner_company,
+            "Invoice should be created with company partner",
+        )
```
rrebollo, line 64: Idem

account_invoice_partner_company_only/tests/test_account_invoice_partner_company_only.py
```diff
@@ -0,0 +1,83 @@
...
+        self.assertIn(
+            self.partner_company,
+            results,
+            "Company partner should be included in restricted domain search",
+        )
+        self.assertNotIn(
+            self.individual_contact,
+            results,
+            "Individual contact should be excluded by the is_company domain",
+        )
```
rrebollo, line 83: Similar to before. You're testing whether domains are correctly handled by Odoo—that's not a feature of your addon.

account_invoice_partner_company_only/views/account_move_views.xml
```diff
@@ -0,0 +1,16 @@
...
+        <field name="name">account.move.form.partner.company.only</field>
+        <field name="model">account.move</field>
+        <field name="inherit_id" ref="account.view_move_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='partner_id']" position="attributes">
+                <attribute
+                    name="domain"
+                    operation="domain_add"
+                >[('is_company', '=', True)]</attribute>
+            </xpath>
```
rrebollo, line 13: Are you sure this view is used only for invoices? I believe it's generally used for journal entries. Is that also your goal?

During the tests, you seemed interested only in the `out_invoice` type, so is this the correct implementation? Also, you're only limiting at this level. What about `account.move` records created via the API?

You need to think more carefully about your design.

rrebollo (changes requested):
The design and implementation still feel quite naive at this point. The base requirement may be valid, but you need to work on it a bit longer.

Take a look at already merged addons to grasp the philosophy behind OCA contributions.

Thanks for your effort.

account_invoice_partner_company_only/tests/test_account_invoice_partner_company_only.py
maisim: Hi ! 

As mentioned here https://github.com/OCA/sale-workflow/pull/4039#discussion_r3062814725 , I'm going to close this PR because I agree it's strange to have this module on its own, without a strong link to the "main" module.

However, and with the performance goal we all share here, I don't quite understand your comment: this test verifies that the filter is indeed present, which is precisely the role of this module.

maisim:
Hi @rrebollo, thanx for your review.

>  The design and implementation still feel quite naive at this point. The base requirement may be valid, but you need to work on it a bit longer.

As mentioned in the first comment of this PR and in the documentation for this module:

> This module is used by ``sale_partner_sale_contact`` (available in the
> sale-workflow repository) to enforce proper commercial relationships while
> allowing contact person tracking.

But I admit that:

- The PR message should have described how it works.
- This is probably not the right approach; it might be better to integrate it into the main module as an option that can be enabled in the settings.

PR #2183, opened 2025-12-02
[18.0][FIX] account_invoice_show_currency_rate: prevent ZeroDivisionError

juhenriquez:
When computing invoice_currency_rate on moves with zero balance, a ZeroDivisionError occurred. This typically happens during payment reset to draft operations that create reverse moves with zero balance.

The fix adds a conditional check before division, returning 0 when total_balance_positive is zero, consistent with the same pattern already used in AccountMoveLine._compute_currency_rate().

Added test case to verify the fix prevents the error.

### DIFF
--- a/account_invoice_show_currency_rate/models/account_move.py
+++ b/account_invoice_show_currency_rate/models/account_move.py
@@ -29,6 +29,8 @@ class AccountMove(models.Model):
             total_balance_positive = sum([abs(b) for b in lines.mapped("balance")])
             move.invoice_currency_rate = (
                 amount_currency_positive / total_balance_positive
+                if total_balance_positive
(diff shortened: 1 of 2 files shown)


victoralmau (commented):
Thanks, can you combine all the changes into a single commit (the first one)?

luisDIXMIT (approved):
LGTM! Squash commits please.

JordiMForgeFlow (approved):
would be nice to get this one merged soon

RicardCForgeFlow (approved):

JordiMForgeFlow (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #2186, opened 2025-12-05
[18.0][ADD] account_billing_portal

AungKoKoLin1997:
Depends on #2407

This module adds a customer portal view for account billings. It also lets you configure, per company, the report used to display the billing in the portal — which is additionally used as the PDF attached to billing emails when the billing is sent.

Note: the "Send by Email" action itself was moved out of this module into `account_billing` (see #2407), since it is not portal-specific. This module only adds the portal view and the portal report configuration on top of it.

@qrtl QT6177


account_billing_portal/models/res_company.py
```diff
@@ -0,0 +1,14 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    billing_portal_report = fields.Many2one(
+        "ir.actions.report",
```
yostashiro, line 17:
```suggestion
    billing_portal_report_id = fields.Many2one(
        "ir.actions.report",
        domain="[('model', '=', 'account.billing')]",
```

account_billing_portal/controllers/portal.py
```diff
@@ -0,0 +1,164 @@
...
+        date_end,
+        sortby,
+        filterby,
+        domain,
+        searchbar_filters,
+        default_filter,
+    ):
+        values = self._prepare_portal_layout_values()
+        Billing = request.env["account.billing"]
+        domain = self._billing_domain()
```
yostashiro: Domain passed from the caller is overridden here. Should be combined?

account_billing_portal/views/account_billing_portal_templates.xml
```diff
@@ -0,0 +1,190 @@
...
+        <xpath expr="//div[hasclass('o_portal_docs')]" position="inside">
+            <t t-call="portal.portal_docs_entry">
+                <t t-set="icon" t-value="'/account/static/src/img/Bill.svg'" />
+                <t t-set="title">Your Billings</t>
+                <t t-set="url" t-value="'/my/billings'" />
+                <t t-set="text">Follow or download your billings</t>
+                <t t-set="placeholder_count" t-value="'billing_count'" />
+            </t>
+        </xpath>
+    </template>
```
yostashiro, line 53: This doesn’t seem to show my `/my` page. I suppose we need to follow the same structure as https://github.com/odoo/odoo/blob/5922822/addons/account/views/account_portal_templates.xml#L15-L38.

account_billing_portal/models/account_billing.py
```diff
@@ -0,0 +1,74 @@
...
+                "mail.email_compose_message_wizard_form"
+            )[1]
+        except ValueError:
+            compose_form_id = False
+        ctx = dict(self.env.context or {})
+        report_template = self.company_id.billing_portal_report or self.env.ref(
+            "account_billing.report_account_billing"
+        )
+        template = self.env["mail.template"].browse(template_id)
+        template.report_template_ids = report_template.ids
```
yostashiro: We shouldn't be writing every time the action is run. Can we instead write to the template when the company setting is updated?

account_billing_portal/controllers/portal.py
```diff
@@ -0,0 +1,164 @@
...
+
+    def _billing_domain(self):
+        partner = request.env.user.partner_id
+        commercial = partner.commercial_partner_id
+        return [("partner_id", "child_of", commercial.id)]
+
+    def _prepare_home_portal_values(self, counters):
+        values = super()._prepare_home_portal_values(counters)
+        if "billing_count" in counters:
+            Billing = request.env["account.billing"].sudo()
```
yostashiro: Do we need `sudo()` here?

yostashiro (changes requested):
I think a record rule needs to be added for portal users.

account_billing_portal/views/account_billing_portal_templates.xml
AungKoKoLin1997: I believe this occurs because the current user has no billing records. Regardless, I updated the implementation to match the invoice structure, but it will still be visible only when the current user has billing records.
If we want it to appear even when there are no billing records, we would need to extend the JavaScript and add our own counters. To keep the module maintainable, I’m not adding JS to this module.

https://github.com/odoo/odoo/blob/eed2bf97536cc9dc2fb471936e2422e2313e3009/addons/portal/static/src/js/portal.js#L72-L79

account_billing_portal/models/account_billing.py
AungKoKoLin1997: Since the template isn’t company-dependent, updating it based on company settings could select the wrong report in a multi-company environment.

account_billing_portal/security/account_billing_portal_security.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="account_billing_rule_portal" model="ir.rule">
+        <field name="name">Portal Billings</field>
+        <field name="model_id" ref="account_billing.model_account_billing" />
+        <field
+            name="domain_force"
+        >[('state', '=', 'billed'), ('message_partner_ids','child_of',[user.commercial_partner_id.id])]</field>
```
AungKoKoLin1997, line 8: As per this record rule, we need to improve `account_billing` to add the partner as a follower when we validate the billing like the invoice validation.

account_billing_portal/controllers/portal.py
```diff
@@ -0,0 +1,175 @@
...
+        default_filter,
+    ):
+        values = self._prepare_portal_layout_values()
+        Billing = request.env["account.billing"]
+        domain = self._get_billing_domain()
+        if date_begin and date_end:
+            domain += [
+                ("create_date", ">", date_begin),
+                ("create_date", "<=", date_end),
+            ]
```
yostashiro: There seems to be no corresponding feature in the UI to use date_begin and date_end at the moment (even in the corresponding code in account). Can we remove these to simplify the code?

account_billing_portal/controllers/portal.py
```diff
@@ -0,0 +1,175 @@
...
+        searchbar_sortings = self._get_billing_searchbar_sortings()
+        if not sortby:
+            sortby = "date"
+        order = searchbar_sortings[sortby]["order"]
+        if searchbar_filters:
+            if not filterby:
```
yostashiro: To prevent server errors when the param value is manually changed.
```suggestion
            if not filterby or filterby not in searchbar_filters:
```

account_billing_portal/views/account_billing_portal_templates.xml
```diff
@@ -0,0 +1,203 @@
...
+                            </td>
+                            <td>
+                                <span t-field="billing.date" />
+                            </td>
+                            <td
+                                class='d-none d-md-table-cell'
+                                t-att-class="'text-danger' if billing.date and billing.date &lt; datetime.date.today() else ''"
+                            >
+                                <span t-field="billing.date" />
+                            </td>
```
yostashiro: Can we remove this? I wasn't sure what the intention was.

account_billing_portal/models/account_billing.py
```diff
@@ -0,0 +1,84 @@
...
+        self.ensure_one()
+        ir_model_data = self.env["ir.model.data"]
+        try:
+            template_id = ir_model_data._xmlid_lookup(
+                "account_billing_portal.email_template_billing"
+            )[1]
+        except ValueError:
+            template_id = False
+        if not template_id:
+            return
```
yostashiro: The design doesn't look ideal. Could we make the template assignment configurable? We also shouldn't do a silent return when no template is found.

account_billing_portal/models/account_billing.py
yostashiro: We still shouldn't write to the template. Just use the value from company.billing_portal_report to generate an attachment and assign it via ctx.

account_billing_portal/models/res_company.py
```diff
@@ -0,0 +1,33 @@
...
+        for company in self:
+            if company.billing_portal_report:
+                template.write(
+                    {
+                        "report_template_ids": [
+                            (6, 0, [company.billing_portal_report.id])
+                        ]
+                    }
+                )
+        return res
```
yostashiro: We shouldn't do this as email templates are not company specific.
```suggestion
```

account_billing_portal/models/account_billing.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def validate_billing(self):
+        res = super().validate_billing()
+        for rec in self.filtered(lambda x: x.state == "billed"):
+            subscribers = (
+                [rec.partner_id.id]
+                if rec.partner_id not in rec.message_partner_ids
+                else None
+            )
+            rec.message_subscribe(subscribers)
```
yostashiro:
```suggestion
            if rec.partner_id not in rec.message_partner_ids:
                rec.message_subscribe([rec.partner_id.id])
```

yostashiro (changes requested):

account_billing_portal/models/account_billing.py
```diff
@@ -0,0 +1,96 @@
...
+        ctx = dict(self.env.context or {})
+        report = self.company_id.billing_portal_report or self.env.ref(
+            "account_billing.report_account_billing"
+        )
+        if not report:
+            raise UserError(
+                _("Please configure the Billing Portal Report in the settings.")
+            )
+        pdf_content, _type = report._render_qweb_pdf(report.id, self.ids)
+        name = self.display_name if self.display_name else "BILL"
```
yostashiro:
```suggestion
        name = self.display_name if self.display_name else "BILLING"
```

account_billing_portal/models/account_billing.py
```diff
@@ -0,0 +1,96 @@
...
+        )
+        email_xml_id = "mail.mail_notification_layout_with_responsible_signature"
+        ctx.update(
+            {
+                "default_model": "account.billing",
+                "default_res_ids": self.ids,
+                "default_template_id": template.id,
+                "default_composition_mode": "comment",
+                "default_email_layout_xmlid": email_xml_id,
+                "default_attachment_ids": [(6, 0, [attach.id])],
```
yostashiro:
```suggestion
                "default_attachment_ids": [Command.set([attach.id])],
```

account_billing_portal/tests/test_account_billing_portal.py
yostashiro, line 1: Can we add a test for action_billing_send()?

yostashiro (commented):
Almost there. I think we need to add payment status to billings but we need to follow up on this separately.

yostashiro (approved):
Code review. LGTM. 👍

nobuQuartile (approved):
Functional review: LGTM

BhaveshHeliconia (approved):
Functional review LGTM!

AungKoKoLin1997:
Heads-up on a rework (force-pushed): the "Send by Email" feature that was originally part of this module has been **moved to `account_billing`** in #2407, since sending a billing to the partner by email is a core part of the billing workflow and is not portal-specific (it only depends on `mail`, already available through `account`).

This PR now only adds the customer portal view and the portal report configuration. The portal report setting still applies to the emailed PDF, via a small `_get_billing_report` override on top of the hook introduced in #2407.

As a result, this PR now **depends on #2407** and should be merged after it.

PR #2190, opened 2025-12-12
[16.0][FIX] purchase_invoicing_no_zero_line: use default UoM precision when filtering zero-qty lines

sbejaoui:
When avoid_zero_lines is enabled, zero-quantity invoice lines are filtered
using the product UoM rounding. If the invoice line has no UoM, the rounding
value is 0, which causes float_is_zero to fail

Fallback to the default "Product Unit of Measure" decimal precision to ensure
a strictly positive rounding value


phschmidt (approved):
Functional review done

luisDIXMIT (approved):
Code review and tested on runboat, LGTM!

dreispt (approved):

PR #2192, opened 2025-12-15
[16.0][ADD] account_edi_no_product_name_match

sbejaoui:
This module disable the automatic matching of products based on their name when importing EDI invoices, ensuring products are only matched using explicit identifiers (such as vendor references, barcodes, or internal codes)


phschmidt (approved):
Functional review done

sbejaoui:
moved to https://github.com/OCA/edi/pull/1269

PR #2195, opened 2025-12-17
[16.0][FIX] account_billing: fix issue when create billings from refunds

AungKoKoLin1997:
This PR fixes an issue that occurs when creating billings from the list view of credit notes, or when creating credit notes and invoices together by calling `_create_billing` from dependency modules or test cases.

@qrtl QT6260

### DIFF
--- a/account_billing/models/account_move.py
+++ b/account_billing/models/account_move.py
@@ -21,11 +21,16 @@ class AccountMove(models.Model):
             billing_lines = bl_obj.search([("move_id", "=", rec.id)])
             rec.billing_ids = billing_lines.mapped("billing_id")
 
+    def _get_billing_type(self):
+        outbound_types = {"out_invoice", "out_refund", "out_receipt"}
+        move_types = set(self.mapped("move_type"))
+        return "out_invoice" if move_types.issubset(outbound_types) else "in_invoice"
+
     def _create_billing(self, partner):
         billing = self.env["account.billing"].create(
             {
                 "partner_id": partner.id,
-                "bill_type": list(set(self.mapped("move_type")))[0],
+                "bill_type": self._get_billing_type(),
                 "billing_line_ids": [
                     Command.create(
                         {


account_billing/models/account_move.py
```diff
@@ -21,11 +21,16 @@ def _compute_billing_ids(self):
             billing_lines = bl_obj.search([("move_id", "=", rec.id)])
             rec.billing_ids = billing_lines.mapped("billing_id")
 
+    def _get_billing_type(self):
+        outbound_types = {"out_invoice", "out_refund"}
```
yostashiro: How about `out_receipt`?

AungKoKoLin1997: I missed that. Updated!

yostashiro (approved):
Code review. LGTM. 👍

Saran440 (approved):
LGTM 👍 

In my opinion, should we add a new method so other modules can hook into it?

```
{
    "partner_id": partner.id,
    "bill_type": self._get_billing_type(),
    "billing_line_ids": [
        Command.create(
            {
                "move_id": m.id,
                "amount_total": m.amount_total * (-1 if m.move_type in ["out_refund", "in_refund"] else 1),
                "amount_residual": m.amount_residual * (-1 if m.move_type in ["out_refund", "in_refund"] else 1),
            }
        )
        for m in self
     ],
}
```

AungKoKoLin1997:
> In my opinion, should we add a new method so other modules can hook into it?

@Saran440 I think it is better to have a method `_prepare_billing_vals()` instead of `def _create_billing():` and do the creation inside `def action_create_billing():` by calling  `_prepare_billing_vals()`.
How do you think?

Saran440:
I agree with you 👍

AungKoKoLin1997:
@yostashiro @Saran440 I added the new commit to follow up on my above comment. 
https://github.com/OCA/account-invoicing/pull/2195#issuecomment-3667893734

nobuQuartile (approved):
LGTM
Functional review

yostashiro (approved):
LGTM.

Saran440 (approved):

account_billing/models/account_move.py
```diff
@@ -59,7 +61,8 @@ def action_create_billing(self):
                 )
             )
 
-        billing = self._create_billing(partner)
```
etobella, line 62: At least maintain the old function. Maybe someone used it as a hook 🙏

etobella (changes requested):

account_billing/models/account_move.py
AungKoKoLin1997: Yes. I will remove that changes and it is better to handle in the future migration.

etobella (approved):
LGTM

yostashiro:
@AungKoKoLin1997 Can you please forward-port this.

PR #2198, opened 2025-12-22
[18.0][ADD] account_move_partner_bank

AungKoKoLin1997:
This module adds a bank account field to partners and sets it as the Recipient Bank on account moves based on the configuration of the invoice’s company.

@qrtl QT6258 QT7070


account_move_partner_bank_source/models/res_company.py
```diff
@@ -0,0 +1,27 @@
...
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    account_move_bank_source_ids = fields.One2many(
+        "account.move.bank.source",
+        "company_id",
+        string="Bank Account Suggestion Sources for Invoices",
+    )
+
+    def get_bank_for_move(self, move):
```
yostashiro: I wasn't sure why this method should be in res.company instead of account.move.

account_move_partner_bank/models/res_partner.py
```diff
@@ -0,0 +1,28 @@
...
+
+from odoo import api, fields, models
+
+
+class ResPartner(models.Model):
+    _name = "res.partner"
+    _inherit = ["res.partner", "bank.account.mixin"]
+
+    bank_account_id = fields.Many2one(
+        company_dependent=True,
```
yostashiro, line 12: I guess we can set `company_depenedent=True` in the mixin instead, and remove this field override. Not sure if I'm missing something.

account_move_partner_bank_source/models/account_move_bank_source.py
```diff
@@ -0,0 +1,51 @@
...
+        related="move_rel_field_id.relation", store=True
+    )
+    target_bank_field_id = fields.Many2one(
+        comodel_name="ir.model.fields",
+        string="Bank Field on Related Record",
+        required=True,
+        ondelete="cascade",
+        help="Many2one field on the related record that points to res.partner.bank "
+        "(e.g. bank_account_id).",
+    )
```
yostashiro: Do we need this field when the bank account field is expected to be added to models by mixin?

yostashiro (commented):
Partial review.

account_move_partner_bank/models/res_partner.py
AungKoKoLin1997: I don’t feel comfortable putting this in the mixin because it would introduce unnecessary overhead for models that don’t need per-company behavior.
Even though Odoo 18 stores company-dependent values in JSONB instead of `ir.property`, it may bring unexpected issue.

account_move_partner_bank_source/models/account_move_bank_source.py
AungKoKoLin1997: I want to avoid potential issues in the future. For example, if we later rely on` bank_account_id` from journals and that field changes in a future Odoo version, this design may no longer fit.
It may never happen, but it’s safer to avoid.
Since this configuration is not part of daily operations, I prefer to keep it.

account_move_partner_bank_source/models/res_company.py
AungKoKoLin1997: I moved it to account.move.

account_move_partner_bank_source/models/account_move_bank_source.py
yostashiro: How about having a char field such as `bank_account_path` instead? All these config fields are just to get the path to the bank account field. I think switching to a char field will simplify the design a lot.

AungKoKoLin1997: That makes sense to me. I will change to this approach.

yostashiro (approved):
@AungKoKoLin1997 Please review my last update and squash commits. Otherwise, the design should be good now.

hoangtrann (approved):
Code LGTM

AungKoKoLin1997:
> @AungKoKoLin1997 Please review my last update and squash commits. Otherwise, the design should be good now.

@yostashiro I reviewed and LG. I squash commits and add you as a co-author.

account_move_partner_bank/views/res_company_views.xml
```diff
@@ -0,0 +1,26 @@
...
+            <xpath expr="//notebook" position="inside">
+                <page string="Bank Account Sources">
+                    <group>
+                        <field name="bank_account_source_ids" nolabel="1">
+                            <list editable="bottom">
+                                <field name="sequence" widget="handle" />
+                                <field name="source_model_id" />
+                                <field
+                                    name="bank_field_path"
+                                    placeholder="e.g. partner_id.bank_account_id"
```
nobuQuartile:
```suggestion
                                    placeholder="e.g. partner_id.commercial_partner_id.bank_account_id"
```
This is better I think.

nobuQuartile (commented):
Minor point

account_move_partner_bank/views/res_company_views.xml
AungKoKoLin1997: It is just a placeholder. So, I won't fix it since both can be used.

nobuQuartile: When a contact has a parent company, the bank_account_id field is hidden on the contact form view.
Therefore, using partner_id.bank_account_id as an example is not appropriate.

To avoid causing confusion, we should avoid using a placeholder that could be misleading.

yostashiro:
@AungKoKoLin1997 Can you follow up on the CI error, please.

AungKoKoLin1997:
> Can you follow up on the CI error, please.

@yostashiro Done!

nobuQuartile (approved):
LGTM: Functional review

account_move_partner_bank/views/res_partner_views.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="view_partner_property_form" model="ir.ui.view">
+        <field name="name">res.partner.property.form.inherit</field>
+        <field name="model">res.partner</field>
+        <field name="inherit_id" ref="account.view_partner_property_form" />
+        <field name="arch" type="xml">
+            <field name="invoice_sending_method" position="before">
+                <field name="bank_account_id" domain="bank_account_id_domain" />
+            </field>
```
nobuQuartile, line 11: This bank_account_id is invisible="not is_company and parent_id". As a result, the field can only be set on the parent company, not per contact.

```suggestion
<xpath expr="//page[@name='accounting_disabled']" position="inside">
    <group>
        <field name="bank_account_id" domain="bank_account_id_domain" />
    </group>
</xpath>
```

AungKoKoLin1997: @nobuQuartile This is intentional because accounting related information should take from parent. 
Then, I added `bank_account_id` to `_commercial_fields` for getting propagated to its child contacts.

nobuQuartile: @AungKoKoLin1997 
Sometimes, you register branches as individual contacts under a parent company. Each branch issues its own invoices, and the receiving bank account can differ from branch to branch.
To support this, contacts that share the same commercial_partner_id still need to be able to set their own separate Recipient Bank.

AungKoKoLin1997: @nobuQuartile Can you please review my latest design?

nobuQuartile: It looks good. Thank you.

nobuQuartile (approved):
Functional review: LGTM

AungKoKoLin1997:
Improved some points.

- Bank resolution — sources are now sorted explicitly by sequence (_order doesn't apply to records cached in the same transaction), and a resolved bank belonging to a different company is skipped instead of returned, since the target fields are check_company and the assignment would have been rejected.
- Field path validation — the constraint now rejects an unloaded, abstract, or transient source model with a clear message instead of raising KeyError, and rejects an empty path. create/write normalize the path so stray whitespace fails at configuration time rather than at runtime in attrgetter.
- Security — added a global multi-company record rule on bank.account.source, and dropped the public/portal read entries from the ACL.

PR #2203, opened 2025-12-30
[FIX] sale_stock_picking_invoicing: Narration could be false

etobella:
Otherwise, an error is raised if one of the Sale Orders has no note


rvalyi (approved):

mbcosta (approved):
Thanks @etobella

PR #2196, opened 2025-12-19
[19.0][ADD] account_move_to_send_attachment

rousseldenis:

PR #2204, opened 2025-12-30
[18.0][ADD] account_invoice_send_attachment

jbaudoux:
Automatically include all invoice attachments in the invoice sending wizard.

Put your attachment on the invoice. When sending the invoice (email or peppol), by default the attachment will be included. In the sending wizard, you can review included attachments.

cc @rousseldenis


rousseldenis:
It can be dangerous to include all attachments (you don't see them if you mass send).

See this : https://github.com/OCA/account-invoicing/pull/2196

jbaudoux:
It now depends on account_edi_ubl_cii.

Should now maybe move to edi repo ?

PR #2206, opened 2026-01-02
[16.0][FIX] partner_last_invoice_date test for > 2025

quentinDupont:
Test failed for an other PR, linked to this module and the fact that we are in 2026 now :-D 

I never used the freeze-time function, maybe that's not the right way to do it, thanks for your help

### DIFF
--- a/partner_last_invoice_date/tests/test_partner_last_invoice_date.py
+++ b/partner_last_invoice_date/tests/test_partner_last_invoice_date.py
@@ -3,6 +3,8 @@
 
 import datetime
 
+from freezegun import freeze_time
+
 from odoo.tests import tagged
 
 from odoo.addons.account.tests.common import AccountTestInvoicingCommon
@@ -11,6 +13,7 @@ from odoo.addons.account.tests.common import AccountTestInvoicingCommon
 @tagged("post_install", "-at_install")
 class TestPartnerLastInvoiceDate(AccountTestInvoicingCommon):
     @classmethod
+    @freeze_time("2025-01-01")
     def setUpClass(cls):
         super().setUpClass()
         cls.partner_model = cls.env["res.partner"]


AungKoKoLin1997 (approved):
LGTM

partner_last_invoice_date/tests/test_partner_last_invoice_date.py
```diff
@@ -11,6 +13,7 @@
 @tagged("post_install", "-at_install")
 class TestPartnerLastInvoiceDate(AccountTestInvoicingCommon):
     @classmethod
+    @freeze_time("2025-01-01")
```
yostashiro, line 16: I'm not against the use of freeze_time but I guess this particular issue (i.e. inconsistency between sequence and date in `cls.bill_1` creation) comes from the fact that `cls.invoice` is created with no specific `invoice_date` first (and then the field is updated after post, which appears to me a strange scenario). I suppose, therefore, assigning `2025-01-01` to `invoice_date` in creating `cls.invoice` will fix the issue.

yostashiro (approved):

partner_last_invoice_date/tests/test_partner_last_invoice_date.py
AungKoKoLin1997: This is also a valid approach.

atgalvez08:
Thanks for the PR 👍
This fix correctly addresses the failing test and using freeze_time is a valid approach to make the test deterministic.

I would like to add some context, as I also proposed an alternative fix in PR #2212 that addresses the same issue using a different approach.

In Odoo accounting, invoice_date is part of the business data and directly impacts:
	•	sequence generation,
	•	legal numbering constraints,
	•	accounting consistency.

The issue observed in this test is not caused by the system clock itself, but by the invoice being posted before its intended accounting date is explicitly set.

In the alternative approach, invoice_date is assigned before calling action_post(), which:
	•	mirrors real business usage,
	•	avoids relying on mocked system time,
	•	keeps the test independent from execution date,
	•	and removes implicit dependencies from the test environment.

While freeze_time is perfectly valid for tests that explicitly validate time-dependent logic, in this case the issue can be solved by correcting the order of operations, which makes the test clearer and closer to real Odoo flows.

Happy to adapt depending on maintainers’ preference for consistency across the test suite.

legalsylvain:
closing, as fixed in https://github.com/OCA/account-invoicing/pull/2212

Issue #2073, opened 2025-08-07
[16.0] partner_invoicing_mode: Using _get_invoice_grouping_keys() in read_group() leads to incompatibility with other modules


AungKoKoLin1997:
## Module
partner_invoicing_mode

**Affected versions**: 16.0
Didn't check in other versions.

## Describe the bug
Using _get_invoice_grouping_keys() in read_group() of sale.order leads to incompatibility with other modules if they want to add new keys from the account.move model.

IIUC, this method is for defining grouping keys for the invoice values (i.e., keys from _prepare_invoice()), not necessarily fields from the sale.order model itself.

However, in partner_invoicing_mode, these keys are passed directly to read_group(), which requires them to be actual fields on sale.order.

https://github.com/OCA/account-invoicing/blob/97f2065ac7f2a854482857d23ac178390ff6feff/partner_invoicing_mode/models/sale_order.py#L75-L80

You can see the incompatibility in this PR:
https://github.com/OCA/account-invoicing/pull/2071

@qrtl

AungKoKoLin1997:
@rousseldenis 
Could you please take a look at this issue?
I noticed that `_get_invoice_grouping_keys()` was introduced in your PR. Shouldn't a dedicated method be used instead of the standard Odoo one? I'm not fully familiar with the background of the `partner_invoicing_mode` module, so I might be missing some context.

https://github.com/OCA/account-invoicing/pull/1515

PR #2209, opened 2026-01-09
[16.0][FIX] partner_invoicing_mode: invoice grouping for manual invoicing

atgalvez08:
@BinhexTeam

Problem:

The partner_invoicing_mode module overrides the method _get_invoice_grouping_keys() in order to group sale orders by partner_invoice_id instead of partner_id.

This behavior is required and correct for automatic invoicing (cron / queue jobs), where invoices must be generated per invoicing partner and according to specific grouping rules.

However, this override is global and also affects manual invoicing triggered from the Sale Orders list view (Sales → Create Invoice).

As a consequence, when a user manually selects multiple sale orders belonging to different invoicing partners and creates invoices, Odoo may:
	•	generate a single invoice,
	•	assign it to the first partner in the recordset,
	•	include invoice lines coming from sale orders of different partners.

This breaks standard Odoo behavior and may result in invalid or legally incorrect accounting documents.

The issue is not visible when invoices are generated through the cron, but becomes evident during manual invoicing, which is a very common user flow.

This behavior has already been reported and discussed by users of the module
(see related issue: OCA/account-invoicing#2073).

Root Cause:

The method _get_invoice_grouping_keys() is used by both:
	•	automatic invoicing (generate_invoices, cron, queue jobs),
	•	manual invoicing (sale.order._create_invoices()).

The module removes partner_id from the grouping keys unconditionally, which is correct for the automatic invoicing use case, but unsafe for manual invoice creation, where Odoo core guarantees that invoices are never mixed across partners.

Solution:

Scope the custom grouping behavior only to automatic invoicing, and preserve the standard Odoo grouping logic for manual invoice creation.

Implementation details
	1.	Introduce a dedicated context key (partner_invoicing_mode_auto) when invoices are generated through the automatic invoicing flow (cron / queue jobs).
	2.	Apply the custom grouping logic (grouping by partner_invoice_id, payment_term_id, and removal of partner_id) only when this context key is present.
	3.	Fallback to the standard Sale Order grouping behavior in all other cases (manual invoicing).

This ensures that:
	•	Automatic invoicing keeps the intended and existing grouping behavior.
	•	Manual invoicing remains safe and consistent with Odoo core expectations.
	•	No existing workflows are broken.

Tests:

A new regression test has been added to cover the manual invoicing scenario and ensure that:
	•	manual invoice creation never mixes sale orders from different invoicing partners.

Existing tests have been adjusted to:
	•	remove assumptions about grouping rules that are not enforced by the module,
	•	avoid dependencies on demo data,
	•	make the test suite fully deterministic and CI-safe.

All tests pass successfully.

Tested Scenarios:

Automatic invoicing (cron)
	•	Same partner_invoice_id
	•	Same payment terms
	•	one_invoice_per_order = False

Invoices are correctly grouped.

Manual invoicing (Sales → Create Invoice)
	•	Sale orders belonging to different partners

Invoices are correctly split per partner (standard Odoo behavior).

Scope & Compatibility
	•	No change in public API.
	•	No behavioral change for existing cron-based invoicing flows.
	•	Restores standard Odoo safety guarantees for manual invoicing.
	•	Fully backward compatible.

Conclusion

This PR fixes a critical side effect caused by a global override of _get_invoice_grouping_keys() by scoping the custom behavior to its intended use case only.

The result is a safer, more predictable invoicing flow that remains fully aligned with both Odoo core behavior and the design goals of partner_invoicing_mode.

### DIFF
--- a/partner_invoicing_mode/models/sale_order.py
+++ b/partner_invoicing_mode/models/sale_order.py
@@ -72,12 +72,17 @@ class SaleOrder(models.Model):
         domain = self._get_generate_invoices_domain(
             companies=companies, invoicing_mode=invoicing_mode
         )
+        # Apply context BEFORE computing grouping keys
+        sale_order_auto = self.with_context(partner_invoicing_mode_auto=True)
+        # Compute grouping with correct context
+        grouping_keys = sale_order_auto._get_invoice_grouping_keys()
         saleorder_groups = self.read_group(
             domain,
             ["partner_invoice_id", "sale_ids:array_agg(id)"],
-            groupby=self._get_invoice_grouping_keys(),
+            groupby=grouping_keys,
             lazy=False,
         )
+        # Enqueue invoice generation jobs
         for saleorder_group in saleorder_groups:
             self.with_delay()._generate_invoices_by_partner(saleorder_group["sale_ids"])
         companies.write({last_execution_field: Datetime.now()})
@@ -90,6 +95,10 @@ class SaleOrder(models.Model):
         add some missing keys. We remove also the partner_id key.
         """
         keys = super()._get_invoice_grouping_keys()
+
+        if not self.env.context.get("partner_invoicing_mode_auto", False):
+            return keys
+
         if "partner_invoice_id" not in keys:
             keys.append("partner_invoice_id")
         if "payment_term_id" not in keys:
--- a/partner_invoicing_mode/tests/__init__.py
+++ b/partner_invoicing_mode/tests/__init__.py
@@ -1,1 +1,2 @@
 from . import test_invoice_mode
+from . import test_manual_invoice_grouping
--- a/partner_invoicing_mode/tests/test_manual_invoice_grouping.py
+++ b/partner_invoicing_mode/tests/test_manual_invoice_grouping.py
@@ -1,0 +1,52 @@
+# Copyright 2022 Opener B.V.
+# Copyright 2023 ACSONE SA/NV
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html)
+from odoo.tests.common import TransactionCase
+
+from .common import CommonPartnerInvoicingMode
+
+
+class TestManualInvoiceGrouping(CommonPartnerInvoicingMode, TransactionCase):
+    def test_manual_invoice_does_not_mix_partners(self):
+        """
+        Manual invoice creation must never mix sale orders
+        belonging to different partners, even if
+        partner_invoicing_mode is installed.
+        """
+        # Confirm and deliver both sale orders
+        self._confirm_and_deliver(self.so1)
+
+        # Create a second sale order for another partner
+        so_other_partner = self.SaleOrder.create(
+            {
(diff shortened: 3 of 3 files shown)


sheragar (approved):
LGTM. Funtional Review 👍

partner_invoicing_mode/models/sale_order.py
```diff
@@ -72,14 +72,21 @@ def generate_invoices(
...
-        saleorder_groups = self.read_group(
+        # Apply context BEFORE computing grouping keys
+        sale_order_auto = self.with_context(partner_invoicing_mode_auto=True)
+        # Compute grouping with correct context
+        grouping_keys = sale_order_auto._get_invoice_grouping_keys()
+        saleorder_groups = sale_order_auto.read_group(
             domain,
             ["partner_invoice_id", "sale_ids:array_agg(id)"],
-            groupby=self._get_invoice_grouping_keys(),
+            groupby=grouping_keys,
```
adasatorres, line 82: @atgalvez08 
The with_context(partner_invoicing_mode_auto=True) call is only required when invoking self._get_invoice_grouping_keys().

partner_invoicing_mode/models/sale_order.py
```diff
@@ -90,6 +97,10 @@ def _get_invoice_grouping_keys(self) -> list:
         add some missing keys. We remove also the partner_id key.
         """
         keys = super()._get_invoice_grouping_keys()
+
+        if not self.env.context.get("partner_invoicing_mode_auto"):
```
adasatorres: @atgalvez08 
I would recommend explicitly providing a default value when accessing partner_invoicing_mode_auto from the context. While self.env.context.get("partner_invoicing_mode_auto") does not raise an error if the key is missing, it returns None, which can introduce ambiguity when the value is intended to behave as a boolean flag. Setting a default value (e.g., False) makes the intent clearer and ensures consistent, predictable behavior.

adasatorres (approved):
LGTM. I’ve left a few suggestions, but nothing blocking.

partner_invoicing_mode/models/sale_order.py
atgalvez08: Thank you for the clarification.
The context usage has been adjusted accordingly so that it is only applied when invoking _get_invoice_grouping_keys().
The read_group() call now runs without the extra context, keeping the scope of the context limited to where it is strictly required.

partner_invoicing_mode/models/sale_order.py
atgalvez08: Thank you for the suggestion.
The context access has been updated to explicitly provide a default value:
self.env.context.get("partner_invoicing_mode_auto", False)
This makes the boolean intent clearer and ensures consistent and predictable behavior.

rrebollo (approved):
Code Review. LGTM!

adasatorres:
@OCA/accounting-maintainers, I think the PR is ready to be merged.

adasatorres:
@jelenapoblet, Could you please merge this PR?

PR #2216, opened 2026-01-12
[18.0][ADD] account_invoice_send_template_from_journal

Andrii9090-tecnativa:
This module extends the Accounting functionality by allowing users to define a default email template for each accounting journal.

A new field is added to the Accounting Journal (account.journal) that lets users select an email template. When sending invoices from the invoice journal, the selected template will be automatically available, making it easier and faster to send invoices using a predefined format.

image

image

image

MT-13301


account_invoice_send_template_from_journal/models/account_move.py
```diff
@@ -0,0 +1,18 @@
...
+    _inherit = "account.move"
+
+    def _get_mail_template(self):
+        # If account.journal has mail_template_id -> use this mail template
+        if all(move.journal_id.mail_template_id for move in self):
+            return self.env.ref(
+                self[0].journal_id.mail_template_id.get_external_id()[
+                    self[0].journal_id.mail_template_id.id
+                ]
+            )
```
chienandalu: You just need to return a mail template record when all the invoices in the recordset share the same one
```suggestion
        # Use the journal mail template if it's common to everyone
        mail_template = fields.first(self).journal_id.template_id
        if all(mail_template == move.journal_id.mail_template_id for move in self):
            return mail_template
```

chienandalu (changes requested):

PR #2224, opened 2026-01-16
[19.0][MIG] account_move_cancel_confirm

JasminSForgeFlow:
Standard Migration

@ForgeFlow

Depends on:

- [x] https://github.com/OCA/server-ux/pull/1219


AaronHForgeFlow (commented):
If I create an invoice and I cancel it without confirming it I am not getting the cancel confirm wizard. I am only getting the cancel confirmation wizard when I set it to draft first. That is the same behavior this module has in v18 so this could be a feature and not a bug., but in such case we should at least put it in the description of the module

JasminSForgeFlow:
> If I create an invoice and I cancel it without confirming it I am not getting the cancel confirm wizard. I am only getting the cancel confirmation wizard when I set it to draft first. That is the same behavior this module has in v18 so this could be a feature and not a bug., but in such case we should at least put it in the description of the module

Have you set the system parameter account.move.cancel_confirm_disable?
Once you set key = account.move.cancel_confirm_disable and value = False.
It will work as expected

AaronHForgeFlow:
> > If I create an invoice and I cancel it without confirming it I am not getting the cancel confirm wizard. I am only getting the cancel confirmation wizard when I set it to draft first. That is the same behavior this module has in v18 so this could be a feature and not a bug., but in such case we should at least put it in the description of the module
> 
> Have you set the system parameter account.move.cancel_confirm_disable? Once you set key = account.move.cancel_confirm_disable and value = False. It will work as expected

Ahh yes, you are right. Thank you.

AaronHForgeFlow (approved):
Functional review + code

MiquelRForgeFlow:
Please rebase.

LoisRForgeFlow (approved):
LGTM :+1:

luisDIXMIT (approved):
Code review, LGTM!

BhaveshHeliconia (approved):
LGTM!

JordiMForgeFlow (approved):
code + functional 👍🏼 @AaronHForgeFlow could this one get merged?

PR #2225, opened 2026-01-16
[17.0][FIX] account_invoice_pricelist: add test_enable to not block other test

eduezerouali-tecnativa:
cc @Tecnativa 
ping @pedrobaeza @sergio-teruel @christian-ramos-tecnativa 

I found that overriding `_compute_currency_id` in another module will cause a collision. To mitigate this, I added `test_enable` so it doesn’t block other tests.

However, any module that tries to override this method will still be incompatible with this module.

Origin: github.com/OCA/account-invoicing/pull/2177


pedrobaeza (approved):
Thanks!

/ocabot merge patch

PR #2226, opened 2026-01-16
[18.0][MIG] account_invoice_supplier_self_invoice: Migration to 18.0

sabrinaRMartin:
Migrated from version 16.0 to 18.0

@moduon MT-13041


Shide:
@etobella can you take a look here? Thanks 😄

Shide (approved):
I've continued the migration

Gelojr (changes requested):
The following test has been performed and requires review:
Send vendor bill / self-invoice (before and after posting): Following the README sending flow triggers two errors depending on the invoice state. NOK
Reproduction steps:
- Create a vendor bill for a vendor with “Set Self Invoice” enabled -> attempt to send it while still in draft (before validation/posting) -> error “Invalid operation: You can't generate invoices that are not posted.” 

Image

- After posting the vendor bill -> attempt to send it again -> error “Invalid operation: You can only generate sales documents.”

Image

Expected behaviour: According to the README, the self-invoice should be sendable before and after validation, and it should also be possible to send/print it without restricting it to sales documents only.

Shide:
@Gelojr review again pls

BhaveshHeliconia (approved):
Functional review LGTM!

PR #2234, opened 2026-01-24
[19.0][ADD] partner_invoicing_mail_subject

NICO-SOLUTIONS:
This module extends the functionality of Invoice mailing to allow partners to have individual invoice email subjects. It ensures that the partner-specific subject is used when sending invoices, both for single and batch invoice emails.


partner_invoicing_mail_subject/models/res_partner.py
```diff
@@ -0,0 +1,46 @@
...
+                allowed_keys_formatted = ", ".join(f"{{{k}}}" for k in allowed_keys)
+                msg_template = partner.env._(
+                    "Invalid placeholders in invoice mail subject: %(invalid)s\n"
+                    "Allowed placeholders are: %(allowed)s"
+                )
+                msg = msg_template % {
+                    "invalid": invalid_keys_formatted,
+                    "allowed": allowed_keys_formatted,
+                }
+                raise ValidationError(msg)
```
CRogos: checking against {{ object.name }} or {{ object.partner_id.name }} where object would be of kind account.move, would be more powerfull. But I also need to lookup how this can be done.

partner_invoicing_mail_subject/models/account_move_send.py
```diff
@@ -0,0 +1,31 @@
...
+                values = {
+                    "invoice_number": move.name.replace("/", "_"),
+                    "partner_name": commercial.name,
+                    "invoice_date": move.invoice_date.strftime("%Y-%m-%d")
+                    if move.invoice_date
+                    else "",
+                }
+                return commercial.invoice_email_subject.format(**values)
+
+        return super()._get_default_mail_subject(move, mail_template, mail_lang)
```
CRogos: Instead of Overwriting the subject, you could also add a new custom invoice email template and set it as default in the Print & Send dialog.

Image

The alternate email template could have another subject formular.

Image

partner_invoicing_mail_subject/models/res_partner.py
NICO-SOLUTIONS: Changed implementation to allow usage of all stored fields available on account.move

partner_invoicing_mail_subject/models/account_move_send.py
NICO-SOLUTIONS: I prefer overriding the subject in the current implementation.
Using a dedicated email template would apply the change globally, while this approach allows per-partner customization, which is the intended scope.

NICO-SOLUTIONS:
I’m noticing duplicated CI jobs on pull requests (3× “test with OCA” and 3× “test with Odoo”).
This looks like either a duplicated workflow definition or a matrix duplication in GitHub Actions.
Could someone confirm whether this is expected or a misconfiguration?

@CRogos do you know anything about this on your side?

CRogos:
> I’m noticing duplicated CI jobs on pull requests (3× “test with OCA” and 3× “test with Odoo”). This looks like either a duplicated workflow definition or a matrix duplication in GitHub Actions. Could someone confirm whether this is expected or a misconfiguration?
> 
> @CRogos do you know anything about this on your side?

The issue is not this PR, because also other PRs have the same behavior.

https://github.com/OCA/account-invoicing/blob/0e38db6a3afc0f072b7456c26ef2c80f0434b4ff/.github/workflows/test.yml#L36-L58

If the tests of tripple and fixed_discount are compatible, these could be merged into one job in the Metrix.

partner_invoicing_mail_subject/models/res_partner.py
```diff
@@ -0,0 +1,67 @@
...
+                return False
+
+            is_last = i == len(parts) - 1
+            if not is_last:
+                if not field.relational or not field.comodel_name:
+                    return False
+
+                current = self.env[field.comodel_name]
+
+        return True
```
CRogos, line 67: Have you checked if there is a suitable method somewere? Maybe linked here?
https://github.com/odoo/odoo/blob/6955370fd2d62c83f0ea24247abf7a9e4b4ebed3/addons/mail/models/mail_render_mixin.py#L650

CRogos (commented):
Basically the same comment as in https://github.com/OCA/account-invoicing/pull/2236

partner_invoicing_mail_subject/models/res_partner.py
NICO-SOLUTIONS: I briefly reviewed mail.render.mixin, but it looks like it is focused only on template rendering rather than validating field paths at constraint level.

It appears that Odoo does not provide a dedicated API to validate dot-notated field paths on models. Therefore, this implementation relies on _fields introspection to ensure safe and deterministic validation before rendering.

Please let me know if I may have missed an existing integrated method that already covers this case.

PR #2236, opened 2026-01-25
[19.0][ADD] partner_invoicing_mail_filename

NICO-SOLUTIONS:
This module extends invoice PDF generation by allowing individual PDF filename templates on partners. 
It ensures that partner-specific Jinja templates are used when generating invoice PDFs.


NICO-SOLUTIONS:
@CRogos 
I have applied the changes to the implementation similar to partner_invoicing_mail_subject to allow the use of all stored fields available on account.move.

If you have a moment for reviewing, I would appreciate your feedback.

partner_invoicing_mail_filename/models/account_move.py
```diff
@@ -0,0 +1,44 @@
...
+    _inherit = "account.move"
+
+    def _get_invoice_report_filename(self, extension="pdf", report=None):
+        self.ensure_one()
+        commercial_p = self.partner_id.commercial_partner_id
+        if not commercial_p.invoice_pdf_filename:
+            return super()._get_invoice_report_filename(
+                extension=extension,
+                report=report,
+            )
```
CRogos: I think you should not early exit here, if there is a case that requires calling super at the end.

partner_invoicing_mail_filename/models/res_partner.py
```diff
@@ -0,0 +1,75 @@
...
+                return False
+
+            is_last = i == len(parts) - 1
+            if not is_last:
+                if not field.relational or not field.comodel_name:
+                    return False
+
+                current = self.env[field.comodel_name]
+
+        return True
```
CRogos, line 75: maybe add @api.model to this function?

Maybe implement it recursive? Something like this (untested):
```suggestion
    @api.model
    def _field_exists(self, model, field_path):
        parts = field_path.split(".", 1)
        field = current._fields.get(parts[0])
        
        if len(parts) == 1 and field:
            return True
        if not field.relational or not field.comodel_name:
            return False
        return self._field_exists(self.env[field.comodel_name], parts[1])
```

partner_invoicing_mail_filename/models/account_move.py
CRogos, line 1: I am not sure if this entire file is needed. Have you tried to use an odoo standard mechanism to achive this goal? 

I think you could change how the invoice report template ref("account.account_invoices_without_payment") creates its file name? In this case you only need to add the invoice_pdf_filename field to res-partner and adapt the print_report_name via xml.

Image

partner_invoicing_mail_filename/models/account_move.py
NICO-SOLUTIONS: true.. removed the early exit

partner_invoicing_mail_filename/models/res_partner.py
NICO-SOLUTIONS: I’ve added @api.model since the method operates purely on model metadata and does not depend on any specific recordset.
I kept the implementation iterative for clarity over recursion, and to ensure invalid chains fail early.

partner_invoicing_mail_filename/models/account_move.py
NICO-SOLUTIONS: While print_report_name could be used, it would move business logic into the report XML and tie it to a specific report.

This implementation keeps the logic in Python, making it easier to maintain, extend, and test. It also allows partner-specific, dynamic filenames at model level without modifying the report XML.

PR #2246, opened 2026-01-29
[18.0][MIG] account_invoice_discount_date

Shide:
Straightforward migration

https://www.loom.com/share/240966538cf24414a681bc61df2da65d

MT-13040 @moduon @rafaelbn @yajo @fcvalgar @Gelojr please review if you want 😄


luisDIXMIT (approved):
The discount date on account move lines does not seem to be editable, even in the draft state.
On the other hand, not all lines update their discount date when it is set on the invoice. Is this correct?

Image

luisDIXMIT:
I don't want to approve the PR yet, it was a mistake, sorry

Shide:
@luisDIXMIT only Payment Term lines (aka the ones who can have maturity date/needs to be paid) can have discount date.

Only when the discount date on the invoice is different than the discount date of the payment terms, would show up.

I've made a short video to clarify this. You can see how i can modify discount date in draft state.

https://www.loom.com/share/240966538cf24414a681bc61df2da65d
Clarify that Discount Date is always shown on Bills to know when to pay and benefit for early payment discounts.

account_invoice_discount_date/readme/USAGE.md
```diff
@@ -0,0 +1,8 @@
+To use this module, you need to:
+
+1. Go to Accounting > Suppliers > Bills (or Customers > Invoices) and set the Discount Date on a desired bill.
+2. All Move Lines with maturity date set, will update its Discount Date with the Discount Date of the invoice.
+3. You can group invoices to know which Invoice should be paid to ensure the discount benefit.
+4. You can change the Discount Date on any move line. The lower one will be set on the bill.
```
luisDIXMIT, line 6: Thanks for the video and the explanation. What caused me confusion were these lines because they made me think I could change them whenever I wanted.

Anyway, I left it as approved because I saw it was correct after watching your video.

account_invoice_discount_date/models/account_move.py
```diff
@@ -0,0 +1,40 @@
...
+                    ("display_type", "=", "payment_term"),
+                    ("discount_date", "!=", False),
+                ]
+            ).mapped("discount_date")
+            new_discount_date = d_dates and sorted(d_dates)[0] or None
+            if new_discount_date != record.discount_date or not new_discount_date:
+                record.discount_date = new_discount_date
+
+    def _inverse_discount_date(self):
+        """When set Discount date, update all move lines with Date maturity"""
```
luisDIXMIT, line 35: This refers to all move lines, not only the payment terms lines. In the lines below, I can see that it is filtered, but maybe the explanation could be adjusted.

luisDIXMIT (approved):

yajo (approved):

rafaelbn (approved):
Functional test 👍🏼 

/ocabot merge patch

PR #2247, opened 2026-01-29
[19.0][FIX] account_invoice_pricelist fix test pricelist assignment issue

Vicent-S73:
Due to some changes in this odoo pr: (https://github.com/odoo/odoo/commit/fbdcb07f50c1c6ba3449afcf41f0144a08413a11)

The way it assign the default pricelist has changed, the solution is move the partner creation under all the pricelist, because the sequence was:
pricelist (1,2,3,4)
partner.create
When you create the partner this function is triggered. 
image
And set false because default is 1. 
Default pricelist is resetted as 8, or 9 (depends on the final sequence bellow)
pricelist (8, 9, 1, 2, 3, 4, 5, 6, 7, 10, 11, 12, 13)

### DIFF
--- a/account_invoice_pricelist/tests/test_account_move_pricelist.py
+++ b/account_invoice_pricelist/tests/test_account_move_pricelist.py
@@ -99,6 +99,7 @@ class TestAccountMovePricelist(common.TransactionCase):
         cls.sale_pricelist = cls.ProductPricelist.create(
             {
                 "name": "Test Sale pricelist",
+                "sequence": 14,
                 "item_ids": [
                     Command.create(
                         {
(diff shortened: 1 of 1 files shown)


Vicent-S73:
@pedrobaeza this pr fix a test blocking issue

pedrobaeza (commented):
The patch seems weird. Shouldn't the pricelist contains the lowest sequence for avoiding this problem?

cristina-hidalgo-tecnativa:
I was checking the module's test and I found a different solution to the problem:

imagen

It only verifies that the invoice has an assigned pricelist and it should match the partner's one.
With that, the tests should pass. I'd like to know what you think about this proposal @Vicent-S73  @pedrobaeza

pedrobaeza:
@cristina-hidalgo-tecnativa yeah, it can be a solution for making tests to pass, but it's true that the partner/invoice won't have the expected pricelist, so that can have other side effects.

pedrobaeza (approved):
Let's fix the branch.

/ocabot merge patch

Vicent-S73:
@pedrobaeza the other option would be set a sequence for every pricelist and set the first one with the higher of all, that would make it harder to be the default

PR #2250, opened 2026-01-30
[16.0][IMP] account_invoice_supplier_self_invoice: compute payment_reference in case of self invoice

nicolas-delbovier-acsone:
Ensures that the `payment_reference` field is automatically computed for supplier invoices generated as self-invoices.

### DIFF
--- a/account_invoice_supplier_self_invoice/models/account_move.py
+++ b/account_invoice_supplier_self_invoice/models/account_move.py
@@ -55,6 +55,20 @@ class AccountMove(models.Model):
             move.set_self_invoice = partner_self_invoice
             move.can_self_invoice = partner_self_invoice
 
+    def _compute_payment_reference(self):
+        # Put this before call to "super" so as not to trigger
+        # call to `_inverse_payment_reference` (from main function) too early
+        for move in self.filtered(
+            lambda m: m.state == "posted"
+            and not m.payment_reference
+            and m.move_type == "in_invoice"
+            and m.set_self_invoice
+        ):
+            move.payment_reference = move._get_invoice_computed_reference()
+
+        res = super()._compute_payment_reference()
+        return res
+
     def _post(self, soft=True):
         # Set today for invoice date in self invoices
         self.filtered(
--- a/account_invoice_supplier_self_invoice/tests/test_self_invoice.py
+++ b/account_invoice_supplier_self_invoice/tests/test_self_invoice.py
@@ -117,6 +117,7 @@ class TestSelfInvoice(common.TransactionCase):
         self.invoice.with_user(self.user.id).action_post()
         self.assertTrue(self.invoice.invoice_date)
         self.assertTrue(self.invoice.self_invoice_number)
+        self.assertTrue(self.invoice.payment_reference)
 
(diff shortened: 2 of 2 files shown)


nicolas-delbovier-acsone:
@jbaudoux

jbaudoux (approved):

jbaudoux:
cc @Shide

marcos-mendez (changes requested):
### 1. **Root Cause**

The test failures occur because `account.move` records are being created without required fields like `date` or `journal_id`, which are accessed during `_compute_payment_reference()` when it calls `_get_invoice_computed_reference()`. This leads to `CacheMiss` errors when the record is not yet fully initialized in the database.

---

### 2. **Suggested Fix**

In `account_move.py`, the method `_compute_payment_reference()` should ensure that the move has all required fields before trying to compute `payment_reference`.

**File:** `account_invoice_supplier_self_invoice/models/account_move.py`  
**Method:** `_compute_payment_reference()`  
**Lines:** 58–68  

**Fix:**
Add a check to avoid computing `payment_reference` for records that are not yet valid (e.g., missing `date` or `journal_id`) by filtering out such records before setting `payment_reference`.

```python
def _compute_payment_reference(self):
    # Put this before call to "super" so as not to trigger
    # call to `_inverse_payment_reference` (from main function) too early
    for move in self.filtered(
        lambda m: m.state == "posted"
        and not m.payment_reference
        and m.move_type == "in_invoice"
        and m.set_self_invoice
        and m.date  # Ensure date is set
        and m.journal_id  # Ensure journal is set
    ):
        move.payment_reference = move._get_invoice_computed_reference()

    res = super()._compute_payment_reference()
    return res
```

---

- **Potential `CacheMiss` in `_get_invoice_computed_reference()`**: If `_get_invoice_computed_reference()` depends on fields that are not yet computed or available, this could also raise `CacheMiss`. However, based on current diff, this is not directly part of the change and likely handled in base Odoo logic.
  
- **Missing `journal_id` and `date` in test setup**: The test suite does not initialize these fields properly for new invoices, leading to failures in `_compute_payment_reference()`.

---

### 4. **Test Improvements**

The tests currently do not cover edge cases where required fields are missing during creation or post. Improve test coverage by:

#### Add test cases to ensure:
- Records are created with required fields (`journal_id`, `date`)
- `payment_reference` is not set prematurely before posting
- Edge case: `payment_reference` is correctly computed **after** posting and only for valid self-invoices

#### Example test improvement:
```python
def test_self_invoice_payment_reference_computed_after_posting(self):
    # Ensure journal and date are set
    self.invoice.journal_id = self.env['account.journal'].search([('type', '=', 'purchase')], limit=1)
    self.invoice.date = fields.Date.today()
    self.invoice.with_user(self.user.id).action_post()
    self.assertTrue(self.invoice.payment_reference)
```

#### OCA Testing Patterns:
Use `SavepointCase` or `TransactionCase` for tests involving creation and posting of moves, especially when relying on computed fields and state transitions.

---

✅ This fix ensures that `_compute_payment_reference()` only runs on valid, initialized records, preventing `CacheMiss` errors.
---

PR #2251, opened 2026-02-02
[18.0][ADD] account_utility_invoicing

Saran440:
This module manages monthly utility charges for recurring services such as rentals and shared facilities. Billing is handled manually for each period, providing flexibility while enhancing operational efficiency

How to used:
1. Create a `Utility Type`. Example: Electricity Meter (Normal)
selection_021
2. Create a `Utility`. Example: Meter 1, Meter 2, Meter 3 Each utility can be linked to a product and a customer (if any).
selection_022
3. Go to Utility > Account Utility > Utility Invoicing and create a new document for utility billing.
4. Click `Retrive Lines` to get utilities. Then enter the meter readings for each room, such as water and electricity usage.
selection_024
5. Click Confirm > Create Invoice. It will create invoice group by customer
selection_025
selection_026
6. If the utility invoice is confirmed, the next billing document will automatically fill in the previous meter reading (Prev. Unit).
selection_027

TODO:
- [ ] Test Script


legalsylvain (commented):
Hi. Thanks for your contribution. could you add demo data, to understand better the use case ? 
thanks !

CRogos:
Hi thank you for sharing the code. Not sure if we have the same use case, but we are using [OCA/contract](https://github.com/OCA/contract/tree/18.0/contract) for monthly recurring bills/invoices? Just want to make sure you are aware of this module.

Saran440:
@legalsylvain Thank you for your interest this module. I added demo for this module.

Saran440:
@CRogos Thank you for the suggestion.
We cannot use [OCA/contract](https://github.com/OCA/contract/tree/18.0/contract) because utility billing is based on **actual usage per month**, not on fixed contracts.

Each billing period is calculated from meter readings, and the unit price may change from month to month (for example, due to an FT adjustment).

For example:

- Month 1: Usage is 50 units at 6 USD per unit -> total charge is 300 USD.
- Month 2: No usage -> no charge.
- Month 3: Usage is 20 units at 6 USD per unit -> total charge is 120 USD.
- Month 4: The FT rate is adjusted, so the price becomes 7 USD per unit. Usage is 35 units -> total charge is 245 USD.

Because of this behavior, contract-based recurring billing is not suitable for this use case.

PR #2254, opened 2026-02-02
[16.0][FIX] account_invoice_supplier_self_invoice: fix self-invoice report layout

nicolas-delbovier-acsone:
**Before:**

image

**After:**

image

---

**NB:** 
You can test the generated pdf using the "report" endpoint like so:

    http://localhost:8069/report/pdf/account.report_invoice_with_payments/<account_move_id>

Similarly, you can view the genrated html like so:

    http://localhost:8069/report/html/account.report_invoice_with_payments/<account_move_id>

### DIFF
--- a/account_invoice_supplier_self_invoice/views/report_self_invoice.xml
+++ b/account_invoice_supplier_self_invoice/views/report_self_invoice.xml
@@ -23,7 +23,7 @@
                     class="float-start"
                     alt="Logo"
                 />
-                <div t-if="o.partner_id" t-attf-class="'col-5 ' + {{image_offset}}">
+                <div t-if="o.partner_id" t-attf-class="col-5 {{image_offset}}">
                     <span
                         t-field="o.partner_id"
                         t-options='{"widget": "contact", "fields": ["name", "address"], "no_marker": True}'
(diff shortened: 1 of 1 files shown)


jbaudoux (approved):

jbaudoux (commented):
Your company is the company emitting the self billing?
It should then be in the header. And in the letter window (on the right), it should be the supplier to who you can send the document.

jbaudoux (changes requested):
The info in the letter window doesn't seem to contain the info to who the document is addressed to

jbaudoux (commented):
Regarding the footer, in my opnion, the company footer should always be shown. There is no reason to hide it. It should behave like a customer invoice. It's company info from the company emitting the document.

The partner self billing footer should be an extra info, in addition to the company footer.

It would also be a lot easier if there was a company wide self billing extra info as there is no fiscal position on a self billing invoice to add the required legal text. That could be displayed like the fiscal position text on a customer invoice. That would prevent to set that info on all the partners (i.e. in the self billing footer).

nicolas-delbovier-acsone:
> Regarding the footer, in my opnion, the company footer should always be shown. There is no reason to hide it. It should behave like a customer invoice. It's company info from the company emitting the document.
> 
> The partner self billing footer should be an extra info, in addition to the company footer.
> 
> It would also be a lot easier if there was a company wide self billing extra info as there is no fiscal position on a self billing invoice to add the required legal text. That could be displayed like the fiscal position text on a customer invoice. That would prevent to set that info on all the partners (i.e. in the self billing footer).

I implemented your suggestions here: https://github.com/OCA/account-invoicing/pull/2256

nicolas-delbovier-acsone:
See https://github.com/OCA/account-invoicing/pull/2256

PR #2264, opened 2026-02-12
[16.0][IMP] account_invoice_check_total: allow exempting journals from verification

hbrunn:
For more fine-grained control, you can exempt some journals from the verification by
checking the box `Exempt from verification total check` in the advanced settings tab
of the journal form. This is useful ie. for expenses journals.


hbrunn:
tests are fixed in https://github.com/OCA/account-invoicing/pull/2265

marcos-mendez (dismissed):
### 1. **Root Cause**

The test failure occurs because `self.invoice` in `test_post_with_exemption` is not properly initialized or updated after setting `journal_id.account_invoice_check_total_exempt = True`. The `test_post()` method expects a valid invoice with `check_total` set, but the test fails during `Form` initialization due to an invalid field state or missing data in the invoice form.

### 2. **Suggested Fix**

In `test_account_invoice.py`, the `test_post_with_exemption` method should properly re-initialize the invoice form or reassign the invoice object after modifying the journal. Specifically, after setting `self.invoice.journal_id.account_invoice_check_total_exempt = True`, the test should ensure that the invoice is saved or reloaded to reflect the new field value.

**Fix location**:  
`account_invoice_check_total/tests/test_account_invoice.py`  
Method: `test_post_with_exemption`  
Lines: ~42–49

**Suggested change**:
```python
def test_post_with_exemption(self):
    """
    Check that verification doesn't happen when the journal is exempted
    """
    self.invoice.journal_id.account_invoice_check_total_exempt = True
    self.invoice.flush()  # Ensure changes are written to DB
    self.invoice.invoice_date = fields.Date.today()
    # Reassign the invoice to ensure the new field value is reflected
    self.invoice = self.invoice.with_context(force_recompute=True)
    self.test_post()  # This should now pass without raising ValidationError
```

Alternatively, a cleaner approach is to avoid reusing `test_post()` and instead write a new test method that directly tests the exemption logic, as `test_post()` may not be designed for reuse in this context.

- **Field `check_total_exempt` is not initialized in `account_move.py`**:  
  The field `check_total_exempt` is a related field to `journal_id.account_invoice_check_total_exempt`. However, if the invoice is created without a journal or if the journal is not set properly, the related field may be `False` or undefined, which could lead to inconsistent behavior in some edge cases.

- **Missing `flush()` in test case**:  
  In `test_post_with_exemption`, the change to `journal_id.account_invoice_check_total_exempt` is not flushed to the database, which may lead to inconsistent state in tests. Adding `self.invoice.journal_id.flush()` ensures the change is reflected.

### 4. **Test Improvements**

The current test `test_post_with_exemption` is not well-structured and reuses `test_post()` in a way that can cause test failures.

**Suggested improvements**:

- **Use `TransactionCase` for full DB isolation**:  
  The current test uses `Form` and `TransactionCase`, which is correct. But `test_post_with_exemption` should be rewritten to avoid reusing `test_post()`.

**Suggested new test method**:
```python
def test_post_with_exemption(self):
    """
    Check that verification doesn't happen when the journal is exempted
    """
    self.invoice.journal_id.account_invoice_check_total_exempt = True
    self.invoice.flush()
    self.invoice.invoice_date = fields.Date.today()
    self.invoice.check_total = self.invoice.amount_total  # Ensure check_total matches
    # Should not raise ValidationError
    self.invoice.action_post()
```

**Test pattern suggestion**:
- Use `SavepointCase` for more complex tests involving multiple transactions.
- Add tests for `check_total_exempt = True` and `False` to ensure both code paths are tested.
- Add test for `move_type != in_invoice` to ensure exemption does not affect other invoice types.

**OCA Testing Pattern Reference**:
- Use `TransactionCase` for tests requiring full Odoo environment and DB access.
- Prefer `SavepointCase` when you need to rollback changes after a test.
- Avoid reusing methods like `test_post()` in other test methods unless they are explicitly designed to be reusable (e.g., by returning a clean state).
---

hbrunn:
@marcos-mendez please rein in your bullshit bot

marcos-mendez:
Hi @hbrunn, thank you for taking the time to respond to our automated review! Your feedback is valuable and helps us improve.

We've logged this as an issue to review and improve our process: https://git.pop.coop/pop/oca-reviewer/issues/1

We'll work on addressing the points you raised.

---
_Auto-reply by [oca-feedback](https://git.pop.coop/pop/oca-feedback)_

hildickethan (approved):

BhaveshHeliconia (approved):
Functional review LGTM!

PR #2272, opened 2026-02-21
[18.0][FIX] tests: stabilize account_billing/custom_rounding/show_currency_rate

grojof:
## Summary
- Stabilize tests for stricter OCB environments across:
  - `account_billing`
  - `account_invoice_custom_rounding`
  - `account_invoice_show_currency_rate`
  - `partner_invoicing_mode`
- Keep test intent unchanged while removing fragile setup assumptions.

## Why
In a strict local OCB matrix, additional failures appeared from environment-sensitive test setup (multi-company creation and queue side effects). This update keeps coverage intent while making test behavior deterministic.

## Changes
- `account_billing/tests/test_account_billing.py`
  - Use an existing secondary company (or skip) in the multi-company record-rule test instead of creating a new company with schema-dependent required fields.
- `partner_invoicing_mode/tests/test_invoice_mode.py`
  - Replace a brittle `assert_jobs_count(0)` with a functional assertion that no extra invoice is created on a second run.
- Existing files already in this PR remain unchanged in scope:
  - `account_invoice_custom_rounding/tests/common.py`
  - `account_invoice_show_currency_rate/tests/test_account_move.py`

## Validation
- Local simulation (quick_nodemo) on integrated scenario (`approved baseline + migration stack`) is green:
  - `odoo_exclude=0`
  - `ocb_exclude=0`
  - Run artifacts: `tmp-lab/tmp/runs/20260221-231040-pr2063_on_approved_after_fixes_v3`

## Merge order
2/4 (after #2271)

Depends on:
- https://github.com/OCA/account-invoicing/pull/2271

Follow-up PRs:
- https://github.com/OCA/account-invoicing/pull/2273
- https://github.com/OCA/account-invoicing/pull/2274

PR #2271, opened 2026-02-21
[18.0][FIX] account_invoice_auto_send_by_email: set trusted outgoing bank in tests

grojof:
## Summary
- Set the test bank account as allowed for outgoing payments in `account_invoice_auto_send_by_email` tests.
- Prevent setup failures when validating invoice email sending behavior.

## Why
The test setup must use a bank account that is valid for outgoing payments in this scenario.

## Changes
- `account_invoice_auto_send_by_email/tests/test_account_invoice_auto_send_by_email.py`

## Merge order
1/4

Related PRs (recommended order):
- https://github.com/OCA/account-invoicing/pull/2272
- https://github.com/OCA/account-invoicing/pull/2273
- https://github.com/OCA/account-invoicing/pull/2274

PR #2274, opened 2026-02-21
[18.0][FIX] sale_stock_picking_invoicing: stabilize matrix tests

grojof:
## Summary
- Stabilize `sale_stock_picking_invoicing` tests for strict matrix variants.
- Keep existing test goals while isolating environment-dependent side effects.

## Why
In strict OCB runs, company creation in one test could trigger internal project creation and fail due additional required fields in mixed stacks. The test now isolates that side effect and remains focused on the default policy assertion.

## Changes
- `sale_stock_picking_invoicing/tests/test_sale_stock.py`
  - In `test_default_value_sale_invoicing_policy`, isolate internal project side effects during company creation and keep assertion scope limited to `sale_invoicing_policy` default behavior.

## Validation
- Local simulation (quick_nodemo) on integrated scenario (`approved baseline + migration stack`) is green:
  - `odoo_exclude=0`
  - `ocb_exclude=0`
  - Run artifacts: `tmp-lab/tmp/runs/20260221-231040-pr2063_on_approved_after_fixes_v3`

## Merge order
4/4 (after #2271 and #2272)

Depends on:
- https://github.com/OCA/account-invoicing/pull/2271
- https://github.com/OCA/account-invoicing/pull/2272

### DIFF
--- a/sale_stock_picking_invoicing/tests/test_sale_stock.py
+++ b/sale_stock_picking_invoicing/tests/test_sale_stock.py
@@ -505,9 +505,15 @@ class TestSaleStock(TestPickingInvoicingCommon):
         company_vals = {"name": "Test"}
         if "po_lead" in self.env["res.company"]._fields:
             company_vals["po_lead"] = 1.0
-        with patch.object(
-            type(self.env.company), "_create_internal_project_task", autospec=True
-        ):
+        company_model_cls = type(self.env.company)
+        if hasattr(company_model_cls, "_create_internal_project_task"):
+            with patch.object(
+                company_model_cls,
+                "_create_internal_project_task",
+                autospec=True,
+            ):
+                company = self.env["res.company"].create(company_vals)
+        else:
             company = self.env["res.company"].create(company_vals)
         self.assertEqual(company.sale_invoicing_policy, "sale_order")
 
--- a/sale_stock_picking_invoicing/tests/test_sale_stock.py
+++ b/sale_stock_picking_invoicing/tests/test_sale_stock.py
@@ -502,11 +502,13 @@ class TestSaleStock(TestPickingInvoicingCommon):
 
     def test_default_value_sale_invoicing_policy(self):
         """Test default value for sale_invoicing_policy"""
-        company = self.env["res.company"].create(
-            {
-                "name": "Test",
-            }
-        )
+        company_vals = {"name": "Test"}
+        if "po_lead" in self.env["res.company"]._fields:
+            company_vals["po_lead"] = 1.0
+        with patch.object(
+            type(self.env.company), "_create_internal_project_task", autospec=True
+        ):
+            company = self.env["res.company"].create(company_vals)
         self.assertEqual(company.sale_invoicing_policy, "sale_order")
 
     def test_picking_invocing_without_sale_order(self):
--- a/sale_stock_picking_invoicing/tests/test_sale_stock.py
+++ b/sale_stock_picking_invoicing/tests/test_sale_stock.py
@@ -52,9 +52,9 @@ class TestSaleStock(TestPickingInvoicingCommon):
             and line.product_id.service_tracking == "task_global_project"
             and not line.product_id.project_id
         )
-        service_lines_with_project_on_product.mapped("product_id.product_tmpl_id").write(
-            {"project_id": project.id}
-        )
+        service_lines_with_project_on_product.mapped(
+            "product_id.product_tmpl_id"
+        ).write({"project_id": project.id})
 
     def test_01_sale_stock_return(self):
         """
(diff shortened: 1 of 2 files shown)


grojof:
Cross-reference: migration follow-up has been posted on #2063 for rerun with the updated 18.0 context (after #2271-#2274 stabilization).

Ref: https://github.com/OCA/account-invoicing/pull/2063#issuecomment-3939289855

grojof:
Update pushed to this PR branch:
- [18.0][FIX] sale_stock_picking_invoicing: isolate company default policy test

This was identified during a strict integrated local simulation (`approved baseline + migration stack`).
Current local result for that scenario is green (`odoo_exclude=0`, `ocb_exclude=0`).

grojof:
Follow-up fix pushed for the failing CI run on this PR:

- [18.0][FIX] sale_stock_picking_invoicing: guard optional project patch in test

Root cause:
- `test_default_value_sale_invoicing_policy` patched `_create_internal_project_task` unconditionally.
- In some matrix variants that method is not present on `res.company`, so the test failed with:
  `AttributeError: <class 'odoo.api.res.company'> does not have the attribute '_create_internal_project_task'`.

Fix:
- Apply the patch only when the method exists (`hasattr`), otherwise create the company directly.
- Test intent remains unchanged (assert default `sale_invoicing_policy`).

grojof:
Follow-up coverage-oriented refinement pushed:

- [18.0][FIX] sale_stock_picking_invoicing: simplify optional method patch

Details:
- Replaced the conditional branch with a single `patch.object(..., create=True)` flow.
- This keeps compatibility when `_create_internal_project_task` is absent and reduces uncovered diff lines.
- Also fixed a local issue (`autospec` cannot be used with `create=True`) by using a valid mock configuration.

Goal: improve Codecov patch coverage while preserving test behavior.

mbcosta:
Hi @grojof thanks for your PR.

There is a general task in the most part of the Tests in OCA repositories, **run tests without Demo Data**, in **v18.0** still be as a recommendation:

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst

image

But in in **v19.0** this change becomes necessary:

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-19.0

image

I created a PR with this goal:

I have tried to make changes only in **stock_picking_invoicing** but **sale_stock_picking_invoicing** are using some methods from stock_picking_invoicing so the PR needed to included changes in the both modules, if you can please consider to help in the Review in order to check if solve the problem you saw .

About your PR I have some questions to understand the problem:

**Where, or in which repository, you saw the error with project_ modules? Are in a Public or Private repository? What are the specific module?**

Or

**How we can reproduce the problem you try to solve here?**

There are a concept question here, the modules **sale_stock_picking_invoicing and stock_picking_invoicing** don't has direct dependence of any **project_** modules:

https://github.com/OCA/account-invoicing/blob/18.0/sale_stock_picking_invoicing/__manifest__.py#L15

image


https://github.com/OCA/account-invoicing/blob/18.0/stock_picking_invoicing/__manifest__.py#L10

image

And, by the PR mentioned, is possible to see the problem don't happening in this repository, so we don't see the problem you try to solve.

We can consider included your change if this problem occurs in a **OCA Public Repository** but if in a **Private Repository** probably not, only with a strong argument, because it can be happening by a custom module unknown by us, maybe you can consider create a new module to integrate **project** with **sale_stock_picking_invoicing**, something like **project_sale_stock_picking_invoicing**, but this no make sense because the module since was included here in v16 (this module was originaly in Brazilian Localization Project) was adapt to get the Data used to create the Invoice from the **sale** methods, so any module where will be included new fields in the Invoice the fileds should be also included in the Invoice created by a Picking related to a Sale Order, this change was made to avoid the necessity of "glue modules" and, for now, seems work as expect.

**OBS.:** There are too much commits to solve only a specfic problem, the recommendation is avoid "Commit Noise" if the goal is only included a test for project modules only one commit in this case should be enough, this is overall recommendation. In the commits message is unnecessary inform the version like **[18.0][REF]** and as far I know **[TEST]** is not used just **[FIX], [REF], [REM], [UPD], [MIG] and [ADD]**, in this case is [REF] because until now there is not a clear error just a change to work with a indirect dependecie in a unknown repository.

But will be important to listen your argues about this change, as I pointed maybe we just don't see the problem yet.

sale_stock_picking_invoicing/tests/test_sale_stock.py
```diff
@@ -26,6 +28,66 @@ def setUpClass(cls):
         for company in cls.companies:
             company.sale_invoicing_policy = "stock_picking"
 
+    @classmethod
+    def _ensure_project_for_service_lines(cls, sale_order):
+        if (
```
mbcosta, line 33: This IF can be more simple just use **hasattr** is enough, for exemple:
```bash
    if not hasattr(sale_order.order_line, "project_id")
```
For reference https://github.com/OCA/account-invoicing/blob/18.0/stock_picking_return_refund_option/models/stock_picking.py#L59

PR #2276, opened 2026-02-24
[16.0][IMP] account_billing: refactor _get_moves() for the extensibility

AungKoKoLin1997:
This PR includes

- Extract search domain into _get_moves_domain() method to allow downstream modules to extend the domain via super().

@qrtl QT5088

### DIFF
--- a/account_billing/models/account_billing.py
+++ b/account_billing/models/account_billing.py
@@ -117,18 +117,19 @@ class AccountBilling(models.Model):
                 line.payment_state == "paid" for line in rec.billing_line_ids
             )
 
+    def _get_moves_domain(self, date=False, types=False):
+        return [
+            ("partner_id", "=", self.partner_id.id),
+            ("state", "=", "posted"),
+            ("payment_state", "!=", "paid"),
+            ("currency_id", "=", self.currency_id.id),
+            ("date", "<=", self.threshold_date),
+            ("move_type", "in", types),
+        ]
+
     def _get_moves(self, date=False, types=False):
-        moves = self.env["account.move"].search(
-            [
-                ("partner_id", "=", self.partner_id.id),
-                ("state", "=", "posted"),
-                ("payment_state", "!=", "paid"),
-                ("currency_id", "=", self.currency_id.id),
-                ("date", "<=", self.threshold_date),
-                ("move_type", "in", types),
-            ]
-        )
-        return moves
+        domain = self._get_moves_domain(date=date, types=types)
+        return self.env["account.move"].search(domain)
 
     def _compute_invoice_related_count(self):
         self.invoice_related_count = len(self.billing_line_ids)


account_billing/models/account_billing.py
```diff
@@ -117,18 +117,26 @@ def _compute_payment_paid_all(self):
...
-                ("date", "<=", self.threshold_date),
-                ("move_type", "in", types),
-            ]
+    def _get_moves_domain(self, date=False, types=False):
+        billed_move_ids = (
+            self.env["account.billing.line"]
+            .search([("billing_id.state", "!=", "cancel")])
+            .mapped("move_id")
+            .ids
         )
```
yostashiro: I think this may degrade performance. An idea could be to move `is_not_for_billing` that we have in `l10n_jp_summary_invoice` here and mark it when a move is billed.

However, I wasn't sure this (i.e. an account move should not appear in multiple billings) would meet all use cases. This is fine for our market but there might be a practice of printing all unpaid invoices in a billing report somewhere?

yostashiro (changes requested):

account_billing/models/account_billing.py
AungKoKoLin1997: > This is fine for our market but there might be a practice of printing all unpaid invoices in a billing report somewhere?

I am not sure myself. So, I just removed that part.

yostashiro (approved):
👍

Saran440 (approved):
Code Review 👍 
Could you add it on v18 please?

kanda999 (approved):
Code review

marcos-mendez (approved):
**Automated Review -- Tests Passed**
All tests for `account_billing` passed successfully on Odoo 16.0.
Environment: Minikube + K8s Job + oca-ci/py3.10-odoo16.0
---
### Test Coverage Suggestions
### Coverage Gaps

The new method `_get_moves_domain()` is introduced and used in `_get_moves()`. While the logic is straightforward, it's a **domain-building helper** that could be tested for:

- Correct domain construction when `date` or `types` are passed as `False`
- Correct filtering behavior when `types` is a list with one or multiple move types
- Domain composition with `partner_id`, `state`, `payment_state`, `currency_id`, `date`, and `move_type`

### Suggested Test Cases

```python
def test_get_moves_domain_with_date_and_types(self):
    """Test that _get_moves_domain builds correct domain with date and move types."""

def test_get_moves_domain_with_false_values(self):
    """Test that _get_moves_domain handles False date and types correctly."""

def test_get_moves_returns_correct_moves(self):
    """Test that _get_moves correctly filters account moves based on domain."""
```

### Codecov Risk

- `_get_moves_domain()` is a new method and **not currently tested**.
- If not covered, it could reduce overall coverage in the module.
- The method is used internally by `_get_moves()` which is likely tested, but the domain-building logic should be tested independently for clarity and robustness.

✅ **Conclusion**: Add tests for `_get_moves_domain()` to ensure all code paths are covered and to avoid regressions.
---

PR #2279, opened 2026-02-25
[16.0][IMP] account_invoice_triple_discount: Improve migration performance

lmignon:
Use the upgrade util library lib from Odoo if available to migrate data to improve performances.

It allows accounting line data to be updated in parallel (10 threads) in ranges of 10,000 rows. This significantly reduces the time required for migration. On a database with more than 12,000,000 account.move.line entries, before the change, the process was killed after 3 hours because it was taking too long. After the change, the update was completed in 12 minutes.

### DIFF
--- a/account_invoice_triple_discount/migrations/16.0.2.0.0/pre-migrate.py
+++ b/account_invoice_triple_discount/migrations/16.0.2.0.0/pre-migrate.py
@@ -1,64 +1,91 @@
 # Copyright 2024 Camptocamp SA
+# Copyright 2026 ACSONE SA/NV (https://www.acsone.eu)
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
-from openupgradelib import openupgrade
+import logging
+
+from odoo import SUPERUSER_ID
+from odoo.api import Environment
+from odoo.tools import column_exists
+
+USE_OPENUPGRADE = False
+USE_ODOOUPGRADE = False
+
+try:
+    from odoo.upgrade import util
+
+    USE_ODOOUPGRADE = True
+except ImportError:
+    try:
+        from openupgradelib import openupgrade
+
+        USE_OPENUPGRADE = True
+    except ImportError as err:
+        raise ImportError(
+            "This migration script requires openupgradelib or odoo.upgrade.util.\n"
+            "Please install one of these libraries to proceed with the migration.\n"
+            "For better performances, it is recommended to use odoo.upgrade.util.\n"
+            "See https://github.com/odoo/upgrade-util/"
+        ) from err
+
+
+_logger = logging.getLogger(__name__)
 
 
 def migrate_discount_to_discount1(env):
-    openupgrade.add_fields(
-        env,
-        [
-            (
-                "discount1",
-                "account.move.line",
-                "account_move_line",
-                "float",
-                "numeric",
-                "account_invoice_triple_discount",
-                0.0,
-            )
-        ],
-    )
-    openupgrade.logged_query(
-        env.cr,
-        """
-        UPDATE account_move_line
-        SET discount1 = discount;
-        """,
-    )
-    # if discounts are : 10% - 20% - 30% main discount is : 49.6 %
-    # if discounts are : 05% - 09% - 13% main discount is : 24.7885 %
-    if "discount_fixed" not in env.registry.models["account.move.line"]._fields:
-        openupgrade.logged_query(
-            env.cr,
+    if not column_exists(env.cr, "account_move_line", "discount1"):
+        env.cr.execute(
             """
-            UPDATE account_move_line
-            SET discount = 100 * (
+            ALTER TABLE account_move_line
+            ADD COLUMN discount1 numeric;
+            """
+        )
+        _logger.info("Added column 'discount1' to 'account_move_line'")
+
+    query = """
+        UPDATE account_move_line
+        SET discount1 = discount,
+        """
+
+    if "discount_fixed" not in env.registry.models["account.move.line"]._fields:
+        query += """
+            discount = 100 * (
                 1 - (
-                        (100 - COALESCE(discount1, 0.0)) / 100
+                        (100 - COALESCE(discount, 0.0)) / 100
                         * (100 - COALESCE(discount2, 0.0)) / 100
                         * (100 - COALESCE(discount3, 0.0)) / 100
                     )
-            );
-            """,
-        )
+            )
+            """
+
     else:
         # don't touch lines with fixed discount
-        openupgrade.logged_query(
-            env.cr,
-            """
-            UPDATE account_move_line
-            SET discount = 100 * (
-                1 - (
-                        (100 - COALESCE(discount1, 0.0)) / 100
-                        * (100 - COALESCE(discount2, 0.0)) / 100
-                        * (100 - COALESCE(discount3, 0.0)) / 100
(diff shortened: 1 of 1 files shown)


lmignon:
ping @legalsylvain @grindtildeath @hildickethan

lmignon:
I convert it to draft the time to verify the data in my migrated database...

account_invoice_triple_discount/migrations/16.0.2.0.0/pre-migrate.py
```diff
@@ -1,64 +1,91 @@
...
-        openupgrade.logged_query(
-            env.cr,
-            """
-            UPDATE account_move_line
-            SET discount = 100 * (
+        query += """
+            discount = 100 * (
                 1 - (
-                        (100 - COALESCE(discount1, 0.0)) / 100
+                        (100 - COALESCE(discount, 0.0)) / 100
```
hildickethan, line 53: wouldnt this add discount2 and discount3 to discount again?

hildickethan (commented):
first time seeing odoo.upgrade, seems like it was created in 2023, does this replace openupgrade for migration scripts? the parallel features at least look powerful and it seems like it has most openupgrade features at a glance

account_invoice_triple_discount/migrations/16.0.2.0.0/pre-migrate.py
lmignon: when the request is executed, discount1 is not yet initialized and discount on the right side of the  operator contains the original value into the row...

lmignon:
> first time seeing odoo.upgrade, seems like it was created in 2023, does this replace openupgrade for migration scripts? the parallel features at least look powerful and it seems like it has most openupgrade features at a glance

This is the library used internally at Odoo for their migration scripts. It contains a lot of methods equivalent to those offered by OpenUpgrade-Utils. I now tend to prefer using it to OpenUpgrade-Utils because this library offers more advanced and powerful features. However, we can never be grateful enough for what OpenUpgrade-Utils has offered the community for so many years. The advantage of OpenUpgrade-Utils is that the community still controls its life cycle, which is not the case with the Odoo library....

account_invoice_triple_discount/migrations/16.0.2.0.0/pre-migrate.py
hildickethan: ah I see now

hildickethan (approved):

pedrobaeza (commented):
Maybe we can replicate that parallel query in `openupgradelib` in our OCA way...

lmignon:
> Maybe we can replicate that parallel query in `openupgradelib` in our OCA way...

@pedrobaeza  IMHO it's not a good idea. The odoo library is well maintained, heavily used, tested and improved by the Odoo team with performance and usability in mind. For example take a look to the commit message of one of the last change and you will see the complexity of the problems taken into account https://github.com/odoo/upgrade-util/commit/cef1907c6ef4c6aaa374186e9b2838663343bb32. 

=> We could maybe switch to the odoo upgrade lib for our migration script :thinking: 

However, I didn't want to open this debate or take sides for one lib or another in the context of this PR, just highlight an alternative while remaining compatible with the user's final choice. But it's definitely a topic that deserves to be discussed in the openupgrade group because the performance gains are huge and the tools offered cover all the cases encountered in migrations.

pedrobaeza:
Yeah, I see it's complex... but if we want to improve the performance in OU in general, we should think about it. Maybe including calls to these utils in some places. Something to think in general. Opened https://github.com/OCA/OpenUpgrade/issues/5543.

account_invoice_triple_discount/migrations/16.0.2.0.0/pre-migrate.py
```diff
@@ -1,64 +1,91 @@
...
+    else:
+        util.parallel_execute(
+            env.cr, util.explode_query_range(env.cr, query, table="account_move_line")
         )
 
 
-@openupgrade.migrate()
-def migrate(env, version):
+def migrate(cr, version):
+    env = Environment(cr, SUPERUSER_ID, {})
```
legalsylvain, line 90: Why this change ? 

I think we could simplify the script. 
-> Assuming that openupgrade exists. it's a regular OCA tools, used in many migrations scripts.
-> just implement USE_ODOOUPGRADE.
-> if openupgrade and odoopgrade exists, the script should use odooupgrade.

lmignon:
The script has been successfully applied and executed in production.

marcos-mendez (changes requested):
### 1. **Root Cause**

The test failures occur because the migration script modifies the `discount` field on `account.move.line` in a way that conflicts with Odoo's internal logic during form creation. Specifically, it attempts to compute a cumulative discount based on `discount1`, `discount2`, and `discount3`, but this logic is not properly handled when creating new invoice lines via the `Form` wizard or directly via `account.move.create()`.

### 2. **Suggested Fix**

In `pre-migrate.py`, the `migrate_discount_to_discount1()` function incorrectly sets `discount = 100 * (1 - ...)` without ensuring that the new `discount1` column is correctly populated before updating `discount`. The order of operations in the migration is flawed.

**Fix**:
- Ensure that `discount1` is populated from `discount` **before** updating `discount`.
- The `discount` field update should only be applied to records where `discount_fixed == 0` (as intended), but the query needs to be more robust to avoid cache misses.

**Specifically**, change the `migrate_discount_to_discount1()` function to:
1. First populate `discount1` from `discount`.
2. Then compute `discount` from the triple discount logic **only** for records where `discount_fixed == 0`.

- **Migration Logic Flaw**: The migration uses a single `UPDATE` query that modifies both `discount1` and `discount` in one go. This can cause race conditions or invalid data if not properly sequenced.
- **Missing `discount2` and `discount3` Column Checks**: The migration does not check for existence of `discount2` and `discount3` columns before using them in the `COALESCE()` expressions, which can lead to SQL errors.
- **Use of `util.parallel_execute` without proper error handling**: If `util.explode_query_range` fails or if the migration is interrupted, it may leave the DB in an inconsistent state.

### 4. **Test Improvements**

The current tests rely on `Form` wizard usage, which is fragile in Odoo when dealing with computed fields and migrations. To improve test coverage and robustness:

- **Use `SavepointCase`**: For tests involving complex migrations or field updates, `SavepointCase` is preferred over `TransactionCase` to isolate test changes.
- **Add Test Cases for Edge Cases**:
  - Invoice lines with `discount_fixed = True`
  - Invoice lines with `discount1`, `discount2`, or `discount3` set to zero or null
  - Invoice lines with `discount` values that result in rounding errors
  - Migration of existing data with mixed discount types
- **Test Migration Script Directly**:
  - Create a test that manually calls `migrate_discount_to_discount1()` and asserts the expected field values after migration.
  - Use `self.env['account.move.line'].search([])` to inspect data before and after migration.

**Example test pattern**:
```python
def test_migrate_discounts(self):
    # Create a line with discount = 10
    line = self.env['account.move.line'].create({
        'move_id': self.invoice.id,
        'name': 'Test Line',
        'price_unit': 100,
        'quantity': 1,
        'discount': 10,
    })
    # Simulate migration
    migrate_discount_to_discount1(self.env)
    self.assertEqual(line.discount1, 10)
```

This ensures the migration works as expected and doesn't break during form or create operations.
---

rrebollo (approved):
Code Review. LGTM!

I've been using `upgrade-util` along with OpenUpgrade for a while now. It's a pretty solid combination.

PR #2292, opened 2026-03-09
[18.0][IMP] sale_line_refund_to_invoice_qty: compute qty_invoice_posted

LauraCForgeFlow:
In v18, the core logic for computing the remaining amounts to invoice on SOs relies heavily on the `qty_invoiced_posted` field. It was introduced in https://github.com/odoo/odoo/commit/4bcfd5c4d212d2ed8a4c66e7fa6682ce41bcb2da.

Currently, the `sale_line_refund_to_invoice_qty` module only overrides the `_compute_qty_invoiced` method to revert the effect of refunds (with `sale_qty_to_reinvoice`). 

This PR applies the same logic and criteria to `_compute_qty_invoiced_posted`. By adding the refunded quantities back into `qty_invoiced_posted` field (when the reinvoice bool is false), we ensure the SO invoicing status calculation respects the module's original intention.

### DIFF
--- a/sale_line_refund_to_invoice_qty/models/sale_order_line.py
+++ b/sale_line_refund_to_invoice_qty/models/sale_order_line.py
@@ -38,6 +38,36 @@ class SaleOrderLine(models.Model):
         return res
 
     @api.depends(
+        "invoice_lines.move_id.state",
+        "invoice_lines.quantity",
+        "invoice_lines.sale_qty_to_reinvoice",
+    )
+    def _compute_qty_invoiced_posted(self):
+        res = super()._compute_qty_invoiced_posted()
+        # Revert effect of refunds in qty_invoiced_posted when
+        # `sale_qty_to_reinvoice` is not set.
+        for line in self:
+            qty_invoiced_posted = line.qty_invoiced_posted
+            for invoice_line in line.invoice_lines:
+                if (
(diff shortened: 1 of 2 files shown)


sale_line_refund_to_invoice_qty/models/sale_order_line.py
```diff
@@ -37,6 +37,36 @@ def _compute_qty_invoiced(self):
...
+            for invoice_line in line.invoice_lines:
+                if (
+                    invoice_line.move_id.state == "posted"
+                    or invoice_line.move_id.payment_state == "invoicing_legacy"
+                ):
+                    if (
+                        invoice_line.move_id.state != "cancel"
+                        and invoice_line.move_id.move_type == "out_refund"
+                        and not invoice_line.sale_qty_to_reinvoice
+                    ):
```
JordiMForgeFlow: the 'if' conditions could be unified maybe

LauraCForgeFlow: Done!

JordiMForgeFlow (approved):

AaronHForgeFlow (approved):
Code review 👍

JoanSForgeFlow (approved):

PR #2299, opened 2026-03-22
[MIG] account_portal_invoice_search_by_lot: migration to 18.0

jorgevalmalanetkia:
- Updated stock.production.lot model to stock.lot
- Changed ._search() to .search() for external API call
- Improved domain return with explicit [("id", "=", False)]
- Added tests


jvmntkia:
Depends on account_portal_invoice_search

mariancuadranetkia (approved):
LGTM!

Yoana-RB (approved):
Functional review, LGTM!! I think the correct PR name should be "[18.0][MIG] account_portal_invoice_search_by_lot: Migration to 18.0".

As a possible improvement, I would keep the search term after performing the search. For example, if I search for INV/2026, it should remain in the search field, as it is currently cleared after searching.

Thanks for the contribution!

PR #2305, opened 2026-03-31
[19.0] [ADD] account_tax_fixed_amount_currency

ivantodorovich:
This module allows fixed-amount taxes to have their amount expressed in a specific currency instead of the document currency.

When a currency is set on the tax, the fixed amount is automatically converted to the document currency at invoice time using the appropriate exchange rate.

Three conversion scenarios are supported:

- **Tax currency = document currency**: no conversion is applied.
- **Tax currency = company currency**: the amount is converted using the document's exchange rate.
- **Tax currency is a third currency**: the amount is first converted to the company currency, then to the document currency.


account_tax_fixed_amount_currency/models/account_tax.py
```diff
@@ -0,0 +1,78 @@
...
+        currency_id = base_line.get("currency_id")
+        date = base_line.get("date")
+        base_line["tax_ids"] = base_line["tax_ids"].with_context(
+            document_currency=currency_id,
+            document_rate=rate,
+            document_date=date,
+        )
+        return super()._add_tax_details_in_base_line(
+            base_line, company, rounding_method=rounding_method
+        )
```
yankinmax, line 89: Do I correctly understand that we don't use just `["res.currency"]._convert` because we need a rate to be provided in this method?

account_tax_fixed_amount_currency/tests/test_fixed_amount_currency_js.py
```diff
@@ -0,0 +1,130 @@
...
+            quantity=5,
+            expected_values={
+                "total_excluded": 500.0,
+                "total_included": 550.0,
+                "taxes_data": [
+                    (500.0, 50.0),
+                ],
+            },
+        )
+        self._run_js_tests()
```
yankinmax, line 50: TIL: taxes `_run_js_tests` feature 😃  Thanks a lot

yankinmax (approved):

account_tax_fixed_amount_currency/models/account_tax.py
ivantodorovich: A document (e.g: invoice) may have a manually set currency rate, so we can't rely on the "current rate" provided by `["res.currency"]._convert`, we must use the rate provided in the `base_line['rate']` by the standard module

account_tax_fixed_amount_currency/models/account_tax.py
```diff
@@ -0,0 +1,78 @@
...
+        document_rate = self.env.context.get("document_rate")
+        # When called outside a document context (e.g. _get_tax_details
+        # directly), there is no conversion to apply.
+        if not document_currency:
+            return amount
+        # If the currency already matches, skip the conversion
+        if self.currency_id == document_currency:
+            return amount
+        # Convert: tax currency → company currency → document currency
+        return amount * self.currency_rate * document_rate
```
SilvioC2C, line 59: Any specific reason to store the conversion rate in a computed, non-stored field, to use it for conversion, instead of relying on the `res.currency` conversion methods?

SilvioC2C (approved):
Nice job, thanks!

account_tax_fixed_amount_currency/models/account_tax.py
ivantodorovich: The Python implementation must match the JavaScript implementation, and this is the easiest way I found to do it

SilvioC2C: Thanks for the explanation; can you add it to the code as an explanatory comment? I believe it could be useful to avoid future misunderstandings!

ivantodorovich: Done ✅  -> https://github.com/OCA/account-invoicing/compare/c76b461d50a1cebc7bb0dea3f423eb8375cef047..c62b5bef43fbb244b70d987959beb14cbd8554f5

cyrilmanuel:
hi @OCA/accounting-maintainers  could we merge this PR please  ?

account_tax_fixed_amount_currency/models/account_tax.py
```diff
@@ -0,0 +1,89 @@
...
+    )
+    currency_rate = fields.Float(
+        compute="_compute_currency_rate",
+        help="Conversion rate from the tax currency to the company currency.",
+    )
+
+    @api.depends_context("document_date")
+    @api.depends("currency_id", "company_id.currency_id")
+    def _compute_currency_rate(self):
+        date = self.env.context.get("document_date") or fields.Date.today()
```
grindtildeath, line 24: Just wondering, since `res.currency` expects ctx key "date" to define currency rate's date, wouldn't be easier to use the same name here for reusability?

grindtildeath (commented):
Interesting, but I'm wondering, is there any real world use case where a tax would have to be defined in another currency than the company's currency? 🤔

grindtildeath (commented):
Can you also please add "Alpha" as maturity level here? 🙏

account_tax_fixed_amount_currency/models/account_tax.py
ivantodorovich: IMO it makes it clearer with `document_date`. It aligns with the other variables `document_rate` / `document_currency`; and it avoids confusion with the current date.

ivantodorovich:
> Can you also please add "Alpha" as maturity level here? 🙏

Done!

> Interesting, but I'm wondering, is there any real world use case where a tax would have to be defined in another currency than the company's currency? 🤔

Our use case is a fee-like tax defined in a fixed CHF amount, that must be converted if the document is in another currency (EUR)

cyrilmanuel (approved):

ivantodorovich:
Hey @grindtildeath , is this ok to merge now ? 🙏🏻

grindtildeath (approved):

ivantodorovich:
Can you please retry again tomorrow? 🙏🏻
We reached the daily limit for new modules 🥲

PR #2306, opened 2026-03-31
[19.0] [ADD] account_tax_fixed_amount_multiplier

ivantodorovich:
This module adds a multiplier option to fixed-amount taxes, controlling how the line quantity is used in the tax computation.

By default, Odoo computes fixed taxes as `line quantity × amount`.
This module allows choosing between four multiplier modes:

- **No Multiplier**: the amount is applied once per line, regardless of quantity. Useful for per-document fees (e.g., stamp duties).
- **Quantity (default)**: standard behavior, equivalent to `line quantity × amount`.
- **Product Quantity**: the line quantity is first converted to the product's unit of measure before multiplying. This is useful when the tax is defined per unit of the product, but the line may use a different UoM (e.g., selling in dozens while the product is defined in units).
- **Product Weight**: multiplied by the total weight of the products on the line (product quantity × product weight). Useful for weight-based fees such as environmental levies.


yankinmax (approved):
Very cool feature, thanks

SilvioC2C (approved):
LGTM

cyrilmanuel:
hi @OCA/accounting-maintainers  could we merge this PR please  ?

account_tax_fixed_amount_multiplier/models/account_tax.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def _get_fixed_amount_quantity_ratio(self, evaluation_context):
+        """Return the ratio between the effective quantity and the line quantity.
+
+        This ratio is applied to the standard fixed tax result (which already
+        uses the line quantity) to adjust for the selected multiplier mode.
+        """
+        self.ensure_one()
+        quantity = evaluation_context["quantity"]
+        if not quantity or self.fixed_amount_multiplier == "quantity":
```
grindtildeath: Use `float_is_zero` instead of `not quantity` ⚠️

account_tax_fixed_amount_multiplier/models/account_tax.py
```diff
@@ -0,0 +1,77 @@
...
+            return 1.0
+        elif self.fixed_amount_multiplier == "none":
+            return 1.0 / quantity
```
grindtildeath: This looks dangerous, as if I understand correctly, you're dividing before multiplying to expect the same result.

IMO it would be safer to move this if condition to the `_eval_tax_amount_fixed_amount` function and only apply the multiplication in case `fixed_amount_multiplier != "none"`

account_tax_fixed_amount_multiplier/models/account_tax.py
```diff
@@ -0,0 +1,77 @@
...
+                self._get_fixed_amount_product_quantity(evaluation_context) / quantity
+            )
+            return ratio * evaluation_context["product"].get("weight", 0.0)
+        return 1.0
+
+    def _eval_tax_amount_fixed_amount(self, batch, raw_base, evaluation_context):
+        res = super()._eval_tax_amount_fixed_amount(batch, raw_base, evaluation_context)
+        if self.amount_type == "fixed":
+            ratio = self._get_fixed_amount_quantity_ratio(evaluation_context)
+            return res * ratio
```
grindtildeath: IMO it would be much safer to do `self.amount_type == "fixed" and self.fixed_amount_multiplier != "quantity"` and do the calculation without introducing any division to get a ratio to multiply.

Then you could do: 
```
sign = -1 if evaluation_context['price_unit'] < 0.0 else 1
if self.fixed_amount_multiplier == "none":
    return sign * self.amount
elif self.fixed_amount_multiplier == "product_quantity":
    return sign * self._get_fixed_amount_product_quantity(evaluation_context) * self.amount
elif self.fixed_amount_multiplier == "product_weight":
    ratio = self._get_fixed_amount_product_quantity(evaluation_context)
    return sign * ratio * evaluation_context["product"].get("weight", 0.0) * self.amount
```

grindtildeath (changes requested):
Can we please add an "alpha" maturity level to get it merged asap?

yankinmax:
Not really a module errors (do you have an idea for a fix):
```
2026-05-20 12:25:39,333 320 INFO odoo odoo.tests.result: 0 failed, 0 error(s) of 82 tests when loading database 'odoo' 
2026-05-20 12:25:39,333 320 INFO odoo odoo.service.server: Initiating shutdown 
2026-05-20 12:25:39,333 320 INFO odoo odoo.service.server: Hit CTRL-C again or send a second signal to force the shutdown. 
2026-05-20 12:25:39,460 320 INFO odoo odoo.sql_db: ConnectionPool(read/write;used=0/count=0/max=64): Closed 2 connections  
errors that caused failure (5):
2026-05-20 12:24:59,887 320 WARNING odoo odoo.addons.account_tax_fixed_amount_multiplier.tests.test_fixed_amount_multiplier_js.TestFixedAmountMultiplierJs.test_multiplier_none: Killing chrome descendants-or-self of 337: 6 remaining
- chrome (zombie)
- chrome (zombie)
- chrome (zombie)
- chrome (zombie)
- chrome (zombie)
- chrome (zombie)
```

ivantodorovich:
Weird. Seems like something's off with the oca-ci, as these warnings shouldn't exist.

As a workaround, maybe you can have them ignored by checklog-odoo: https://github.com/OCA/account-invoicing/blob/19.0/checklog-odoo.cfg

vvrossem (approved):
🐐

grindtildeath (approved):
LGTM 👌

PR #2309, opened 2026-04-02
Updated from template

SirPyTech:
To include https://github.com/OCA/oca-addons-repo-template/pull/333.

PR #2310, opened 2026-04-02
[FIX] account_invoice_supplierinfo_update: updating the right supplierinfo

SirPyTech:
Backporting the fix from https://github.com/OCA/account-invoicing/pull/1697.
I adapted the changes as little as possible to keep the versions aligned.

@sebastienbeau I kept you as author, let me know what you think.


SirPyTech:
`pre-commit` will be fixed after https://github.com/OCA/account-invoicing/pull/2309.

PR #2318, opened 2026-04-10
[18.0][IMP] account billing: improve billing line ordering

smorita7749:
1. Add a sequence field to account_billing_line to reorder lines.
2. Sort billing lines created via the Create Billing button in the same order as compute_lines

@qrtl QT6581

### DIFF
--- a/account_billing/tests/test_account_billing.py
+++ b/account_billing/tests/test_account_billing.py
@@ -90,16 +90,22 @@ class TestAccountBilling(TransactionCase):
         currency_id=None,
         partner=None,
         account_id=None,
+        invoice_date=None,
+        invoice_date_due=None,
     ):
         """Returns an open invoice"""
+        inv_date = invoice_date or fields.Date.context_today(self.env.user)
         invoice = self.invoice_model.create(
             {
                 "partner_id": partner or self.partner_agrolait.id,
                 "currency_id": currency_id or self.currency_eur_id,
                 "move_type": invoice_type,
-                "invoice_date": fields.Date.context_today(self.env.user),
-                "date": fields.Date.context_today(self.env.user),
-                "invoice_payment_term_id": self.payment_term.id,
+                "invoice_date": inv_date,
+                "date": inv_date,
+                "invoice_date_due": invoice_date_due,
+                "invoice_payment_term_id": (
+                    False if invoice_date_due else self.payment_term.id
+                ),
                 "invoice_line_ids": [
                     Command.create(
                         {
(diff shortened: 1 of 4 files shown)


yostashiro (commented):
@smorita7749 Sorry I changed the strategy (again) to sort billing lines after creating records (instead of sorting account moves). This is to cover the scenario of the user selecting invoices to create a billing and then switching threshold_date_type from due date to invoice date (i.e. billing lines already exist).

Can you please review my last commit, and follow up with tests?

account_billing/tests/test_account_billing.py
```diff
@@ -268,3 +270,49 @@ def test_8_check_move_button_draft(self):
...
+            {
+                "bill_type": "out_invoice",
+                "partner_id": self.partner_id.id,
+                "currency_id": self.currency_eur_id,
+                "threshold_date": datetime.now(),
+            }
+        )
+        empty_billing._sort_billing_lines()
+        self.assertFalse(empty_billing.billing_line_ids)
+
```
AungKoKoLin1997, line 285:
```suggestion
```

account_billing/tests/test_account_billing.py
```diff
@@ -268,3 +270,49 @@ def test_8_check_move_button_draft(self):
...
+            (inv_b + inv_c + inv_a).ids,
+        )
+
+        # Onchange triggers re-sort
+        billing.threshold_date_type = "invoice_date"
+        billing._onchange_threshold_date_type()
+        self.assertEqual(
+            billing.billing_line_ids.mapped("move_id").ids,
+            (inv_b + inv_c + inv_a).ids,
+        )
```
AungKoKoLin1997, line 318: Seems like the sequence order is the same with above.
Can you make it different by assigning `invoice_date_due` and assign values as different order with `invoice_date` when you create the invoice?

account_billing/tests/test_account_billing.py
```diff
@@ -268,3 +270,49 @@ def test_8_check_move_button_draft(self):
         customer_billing.action_cancel()
         self.inv_1.button_draft()
         self.assertEqual(self.inv_1.state, "draft")
+
+    def test_sort_billing_lines(self):
```
AungKoKoLin1997, line 274: Please add the scenario about having same `invoice_date`?

account_billing/tests/test_account_billing.py
smorita7749: Just to confirm - is your intent to drop the early-return check entirely? If yes, I'll also remove the empty_billing setup since it becomes unused.

account_billing/tests/test_account_billing.py
smorita7749: I added the scenario having same invoice_date (and invoice_date_due).

account_billing/tests/test_account_billing.py
smorita7749: I updated the test so that you can find the sort order changes when threshold_date_type is invoice_date_due or invoice_date.

account_billing/tests/test_account_billing.py
```diff
@@ -268,3 +274,59 @@ def test_8_check_move_button_draft(self):
...
+        empty_billing = self.billing_model.create(
+            {
+                "bill_type": "out_invoice",
+                "partner_id": self.partner_id.id,
+                "currency_id": self.currency_eur_id,
+                "threshold_date": datetime.now(),
+            }
+        )
+        empty_billing._sort_billing_lines()
+        self.assertFalse(empty_billing.billing_line_ids)
```
AungKoKoLin1997: That doesn't make sense to check. Please remove that.
```suggestion
```

smorita7749: Done!

AungKoKoLin1997 (approved):
Code Review: LG 👍

account_billing/tests/test_account_billing.py
```diff
@@ -268,3 +274,48 @@ def test_8_check_move_button_draft(self):
...
+            invoice_date=fields.Date.from_string("2024-04-02"),
+            invoice_date_due=fields.Date.from_string("2024-05-01"),
+        )
+        inv_d = self.create_invoice(
+            amount=400,
+            currency_id=self.currency_eur_id,
+            partner=self.partner_id.id,
+            invoice_date=fields.Date.from_string("2024-04-02"),
+            invoice_date_due=fields.Date.from_string("2024-05-01"),
+        )
```
yostashiro, line 315: We may drop this one as it doesn't add much value to confirm the ordering between inv_c and inv_d.

account_billing/tests/test_account_billing.py
```diff
@@ -268,3 +274,48 @@ def test_8_check_move_button_draft(self):
         customer_billing.action_cancel()
         self.inv_1.button_draft()
         self.assertEqual(self.inv_1.state, "draft")
+
+    def test_sort_billing_lines(self):
+        inv_a = self.create_invoice(
+            amount=100,
+            currency_id=self.currency_eur_id,
+            partner=self.partner_id.id,
```
yostashiro: We can remove these params to reduce noise and be more focused. The same applies to other invoices.
```suggestion
```

yostashiro (approved):

account_billing/tests/test_account_billing.py
smorita7749: The intent is to validate the name fallback.
When two invoices share the same invoice_date and invoice_date_due, the sort should fall back to name. That's why inv_d is included with the same dates as inv_c.

Or is validating the name fallback not worth it?

yostashiro: Technically valid, so it's okay to keep it. IMO, the name fallback is particularly important in this case, and I'd prefer to keep the code concise for maintainability.

yostashiro:
@smorita7749 Please rebase your branch to the latest 18.0 code.

smorita7749:
> @smorita7749 Please rebase your branch to the latest 18.0 code.

Done.

kanda999 (approved):
This update is already live in the production environment.

BhaveshHeliconia (approved):
LGTM!

PR #2319, opened 2026-04-10
[18.0][IMP] account_billing: Add One2many inverse relationship for billing lines

AungKoKoLin1997:
Add proper bidirectional relationship between account.move and account.billing.line:

- Add billing_line_ids One2many field as inverse of move_id Many2one
- Optimize _compute_billing_ids to use the relationship instead of search

This improvement will be useful when we want to add new field in `account.move` for the related billings and add in the search view because of `billing_ids` is non stored compute field and can't use in search view.

@qrtl QT6574

### DIFF
--- a/account_billing/models/account_move.py
+++ b/account_billing/models/account_move.py
@@ -8,6 +8,12 @@ from odoo.exceptions import UserError
 class AccountMove(models.Model):
     _inherit = "account.move"
 
+    billing_line_ids = fields.One2many(
+        comodel_name="account.billing.line",
+        inverse_name="move_id",
+        string="Billing Lines",
+        help="Billing lines that reference this invoice",
+    )
     billing_ids = fields.Many2many(
         comodel_name="account.billing",
         string="Billings",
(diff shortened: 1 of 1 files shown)


smorita7749 (approved):
Code review: LGTM!

Saran440 (approved):
Code Review 👍

account_billing/models/account_move.py
```diff
@@ -16,10 +22,8 @@ class AccountMove(models.Model):
     )
 
     def _compute_billing_ids(self):
-        bl_obj = self.env["account.billing.line"]
         for rec in self:
-            billing_lines = bl_obj.search([("move_id", "=", rec.id)])
-            rec.billing_ids = billing_lines.mapped("billing_id")
+            rec.billing_ids = rec.billing_line_ids.mapped("billing_id")
```
yostashiro, line 26:
```suggestion
            rec.billing_ids = rec.billing_line_ids.billing_id
```

yostashiro (approved):
LGTM.

Saran440 (changes requested):
I found some bug permission

Step to test:
1. Install module purchase and account_billing
2. Create new user with purchase only
Image

3. When `create bills` with purchase user, it will error permission
Image

AungKoKoLin1997:
> I found some bug permission

@Saran440 I just realized this issue is the existing one before my current improvement PR because of `_compute_billing_ids`.
That calls the billing line that causes the permission issue. So, I added `groups="account.group_account_invoice"` to `billing_ids` field definition. Could you please review my last commit?

Below image is the test result from runboat 18.0 branch.

Screenshot 2026-04-22 083927

nobuQuartile (approved):
LGTM: Functional review and code review. the issue is fixed.

Saran440 (approved):
Code Reivew 👍

PR #2320, opened 2026-04-10
[18.0][IMP] account_biliing: add total amount field in account billing list

Saralrat:
Add Total Amount and Total Amount Residual in the list view.

Screenshot 2569-04-20 at 10 31 58

and Add Untaxed Amount, Tax Amount, Total Amount, and Total Amount Residual in form view.

Screenshot 2569-04-20 at 10 30 47

### DIFF
--- a/account_billing/models/account_billing.py
+++ b/account_billing/models/account_billing.py
@@ -101,6 +101,25 @@ class AccountBilling(models.Model):
         compute="_compute_payment_paid_all",
         store=True,
     )
+    amount_untaxed = fields.Monetary(
+        string="Untaxed Amount",
+        compute="_compute_tax_totals",
+        store=True,
+    )
+    amount_tax = fields.Monetary(
+        string="Tax Amount",
+        compute="_compute_tax_totals",
+        store=True,
+    )
+    amount_total = fields.Monetary(
+        string="Total Amount",
+        compute="_compute_tax_totals",
+        store=True,
+    )
+    amount_due = fields.Monetary(
+        compute="_compute_amount_due",
+        store=True,
+    )
 
     @api.model
     def _get_default_threshold_date_type(self):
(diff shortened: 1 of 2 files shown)


account_billing/models/account_billing.py
```diff
@@ -99,6 +99,18 @@ class AccountBilling(models.Model):
         compute="_compute_payment_paid_all",
         store=True,
     )
+    total_amount = fields.Monetary(
+        compute="_compute_amount_total",
+        store=True,
+        string="Total Amount",
```
Saran440:
```suggestion
```

No need because it uses "Total Amount" already.
Add string when we use string difference variable

account_billing/models/account_billing.py
```diff
@@ -99,6 +99,18 @@ class AccountBilling(models.Model):
...
+        currency_field="currency_id",
```
Saran440:
```suggestion
```
No need because standard use currency_id.

account_billing/models/account_billing.py
```diff
@@ -99,6 +99,18 @@ class AccountBilling(models.Model):
...
+    )
+    total_amount_residual = fields.Monetary(
+        compute="_compute_amount_residual_total",
+        store=True,
+        string="Total Amount Due",
+        currency_field="currency_id",
```
Saran440:
```suggestion
```

Saran440 (commented):
Code Review:
1. Add `Total Amount`, `Total Amount Due` on the form view
2. If a billing is cancelled or still in draft, should the Total Amount be computed?

I noticed that `account_billing` has 2 separate views: 
`view_account_billing_tree` and `view_account_supplier_billing_tree`.

Should we consider combining them into a single view?
What do you think?
@AungKoKoLin1997 @yostashiro

AungKoKoLin1997:
> Should we consider combining them into a single view?

@Saran440 Yes. I think it would be better having a shared single view and having `default_bill_type` for each action.

yostashiro:
> > Should we consider combining them into a single view?
> 
> @Saran440 Yes. I think it would be better having a shared single view and having `default_bill_type` for each action.

Not an issue with us, but we may take care of this in future migrations as it might break existing implementations if we change the arrangement in this version.

By the way, we add some amount fields to account.billing in l10n_jp_summary_invoice, and we likely have to hide the fields proposed in this PR once merged (we can't use these fields as we have a special tax recomputation requirement). It's not a problem for us, but just giving some context...

https://github.com/OCA/l10n-japan/blob/d4f69ae/l10n_jp_summary_invoice/models/account_billing.py#L20-L40

Saran440:
@AungKoKoLin1997 @yostashiro Thank you for your feedback. I think this PR should be fixed only amount. and 2 separate views will open new PR.

We will fix this PR for support your case :)

Saran440 (approved):
Code Review 👍

Saran440:
@Saralrat Please check list view
1. Change name `Total Amount Residual ` to `Amount Due`
2. Add currency_id on list view (and add optional="hide") after `Amount Due`
3. Add `Tax Excluded` in list view before `Total Amount`
selection_463

Saralrat:
@Saran440, I have added `Tax Excluded` and `Currency` fields, and changed the name Total Amount Residual  to Amount Due.

Screenshot 2569-07-01 at 10 15 21

AungKoKoLin1997 (approved):
LGTM

AungKoKoLin1997:
@Saran440 I am checking the optimal way for not conflicting with the fields of l10n_jp_summary_invoice since the fields name same.

AungKoKoLin1997:
> I am checking the optimal way for not conflicting with the fields of l10n_jp_summary_invoice since the fields name same.

@Saran440 I found a optimal way to handle that. I will just override the compute method logic in our l10n_jp_summary_invoice module and current PR logic is fine.

PR #2298, opened 2026-03-20
[18.0][REF] stock_picking_invoicing, sale_stock_picking_invoicing: Run Tests without Demo Data

mbcosta:
Changed the Tests of **stock_picking_invoicing and sale_stock_picking_invoicing** to not use Demo Data, both modules need to be in the same PR to avoid error, the tests of sale_stock_picking_invoicing are using methods from stock_picking_invoicing.

This change is recommended by:

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst

image

And in **v19** this change becomes necessary:

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-19.0

image

I'm used the "Common Data" created in https://github.com/OCA/OCB/blob/18.0/addons/stock_account/tests/test_account_move.py#L10 and **Form** to create the other objects, by simulate the screen behaivor some defaults fields are be filled.

cc @rvalyi @renatonlima

PR #2176, opened 2025-11-26
[19.0][MIG] stock_picking_invoicing: Migration to 19.0

sergioM-S73:
stock_picking_invoicing: Migration to 19.0

PR #2325, opened 2026-04-16
[18.0][FIX] sale_stock_picking_invoicing: only service products should be invoiced form the Sale Order

RicardCForgeFlow:
Without this FIX, non-storable goods can be invoiced from the Sale Order. 

This is problematic because the Invoice Status of the Picking will be still "To Be Invoiced", so users can double invoice the same product.

That's why we should only invoice `product_id.type == "service"` from the Sale Order

### DIFF
--- a/sale_stock_picking_invoicing/models/sale_order.py
+++ b/sale_stock_picking_invoicing/models/sale_order.py
@@ -16,15 +16,14 @@ class SaleOrder(models.Model):
         if (
             self.company_id.sale_invoicing_policy == "stock_picking"
             and model != "stock.picking"
+            and lines
         ):
-            new_lines = lines.filtered(
-                lambda ln: not ln.product_id.is_storable and not ln.is_downpayment
+            so_invoiceable_lines = lines.filtered(
+                lambda ln: ln.product_id.type != "consu" and not ln.is_downpayment
             )
-            if new_lines:
-                # Case lines with Product Type 'service'
-                lines = new_lines
+            if so_invoiceable_lines:
+                lines = so_invoiceable_lines
             else:
-                # Case only Products Type 'product'
                 raise UserError(
                     self.env._(
                         "When 'Sale Invoicing Policy' is defined as"
--- a/sale_stock_picking_invoicing/tests/test_sale_stock.py
+++ b/sale_stock_picking_invoicing/tests/test_sale_stock.py
@@ -2,7 +2,7 @@
 # @author Magno Costa <magno.costa@akretion.com.br>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import exceptions, models
+from odoo import Command, exceptions, models
 from odoo.tests import Form
 
 from odoo.addons.stock_picking_invoicing.tests.common import TestPickingInvoicingCommon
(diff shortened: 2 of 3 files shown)


JordiMForgeFlow (approved):
makes sense 👍🏼

AaronHForgeFlow (approved):
👍

etobella (approved):
Makes sense

sale_stock_picking_invoicing/models/sale_order.py
```diff
@@ -16,13 +16,14 @@ def _get_invoiceable_lines(self, final=False):
         if (
             self.company_id.sale_invoicing_policy == "stock_picking"
             and model != "stock.picking"
+            and lines
         ):
-            new_lines = lines.filtered(
-                lambda ln: not ln.product_id.is_storable and not ln.is_downpayment
+            service_lines = lines.filtered(
+                lambda ln: ln.product_id.type == "service" and not ln.is_downpayment
```
etobella: After a second thought....

What about combo products?

Wouldn't be better to do type != 'consu'?

RicardCForgeFlow: I agree. The only "problem" now will be that the invoice from the picking do not have the line of the combo product:
image


But this is something that was already happening before this PR.

etobella: Yes, but if we are solving a problem, it makes sense to solve both, isn't it?

RicardCForgeFlow: Okay! I will use this PR to solve both

RicardCForgeFlow: Hi @etobella , the combo line problem should be fixed now with: https://github.com/OCA/account-invoicing/pull/2325/changes/998caf5bb206db4869fc4737ede4ae663df9e548

etobella (approved):

RicardCForgeFlow:
@etobella could we merge this? Thank you!

sale_stock_picking_invoicing/tests/test_sale_stock.py
```diff
@@ -441,6 +441,323 @@ def test_default_value_sale_invoicing_policy(self):
         )
         self.assertEqual(company.sale_invoicing_policy, "sale_order")
 
+    def test_consumable_product_invoicing_from_picking(self):
```
mbcosta, line 444: Hi @RicardCForgeFlow thank you for PR, recently I created a PR to refactoring the Tests

The goal is make the Tests independent of Demo Data, but also by inherited [TestAccountMoveStockCommon](https://github.com/OCA/OCB/blob/18.0/addons/stock_account/tests/test_account_move.py#L10) tried to reduce the creation and re-creation of the same or similar data several times in each test, maybe this is not will be merged in v18 because there was already a Migration to v19 PR

* https://github.com/OCA/account-invoicing/pull/2176

But in any case if not be merged in v18 I will propose the change in v19.

I need to adapt your changes here, but I decided to remove this specific test because seems very similar to the other tests, are you agree? If not can you please explain why should be keep it.

The other two tests about Combo Product was adapt [here the Data Creation](https://github.com/akretion/account-invoicing/blob/18.0-REF-stock_and_sale_pck_inv_tests/sale_stock_picking_invoicing/tests/common.py#L157) and [test_08_combo_product_invoicing](https://github.com/akretion/account-invoicing/blob/18.0-REF-stock_and_sale_pck_inv_tests/sale_stock_picking_invoicing/tests/test_sale_stock.py#L377) [test_09_combo_product_invoicing_picking_first](https://github.com/akretion/account-invoicing/blob/18.0-REF-stock_and_sale_pck_inv_tests/sale_stock_picking_invoicing/tests/test_sale_stock.py#L430).

RicardCForgeFlow: Hello @mbcosta , this test was specific for `"type": "consu", "is_storable": False,` products. Because without the fix of this PR, non-storable goods could be invoiced from the Sale Order. So I would say that we should keep it, as it tests something that was wrong in the past. In the end, we have to ensure that only service products are invoiced from the sale order.

mbcosta: Hi @RicardCForgeFlow thank you for the feedback, you are right, the PR was merged but I will create a new one to put back this test by create the Product Type Consu Storable and Not Storable, there some other small changes and an open debate about use Form to create the objects and wizards in Post Init Tests, if you see other things that, in your opinion, should be change please let us know.

I'm waiting the merge of the PR:

* https://github.com/OCA/account-invoicing/pull/2374

This PR seems simple and, for now, I'm thinking should be better make the changes after, so please, if you can, consider help in the Review of this PR and there are another one:

* https://github.com/OCA/account-invoicing/pull/2385

And just a important point about both modules **stock_picking_invoicing and sale_stock_picking_invoicing**, related to the way the module was implemented and can be useful to understand some parts of the code, and maybe help the PR:

* https://github.com/OCA/account-invoicing/pull/2326
 
Until **v8.0** the functionality to create Invoice from Picking was native, after that version the [Brazilian Localization Project](https://github.com/OCA/l10n-brazil) need to re-implement it, by Fiscal needs in Brazil any Transport of Goods need to has and carry with an Invoice/Fiscal Document even the cases where the Company are moving Goods between own companies address, the **sale_stock_picking_invoicing** was include here as an extraction of [l10n_br_sale_stock in v14.0](https://github.com/OCA/l10n-brazil/tree/14.0/l10n_br_sale_stock), the project has also another module to extend **stock_picking_invoicing** but for **Purchase** case [l10n_br_purchase_stock](https://github.com/OCA/l10n-brazil/tree/16.0/l10n_br_purchase_stock), so the code can reflect the point of view and the needs of Brazilian Cases of Use, something in the code that appears to be wrong can be just necessary in the Brazil case, for example, in Brazil, if a Sale Order has Services and Produts, we should create two Invoices/Fiscal Documents one for Services and another for Products (maybe its related to what you pointed as "double invoicing" in your PR).

But, and more important, I thinking the module can be adapt to attempted the most of Scenarios/Cases of Use by using Config, parameters, hook methods, contexts, etc we, as developers, just need to be careful and have in mind the different scenarios in order to not inviabilize one of the Scenario/Case of Use.

PR #2300, opened 2026-03-24
[18.0][REF] stock_picking_invoicing: Align 'To be Invoiced' in Kanban View

mbcosta:
Simple PR, just align the information about **'To be Invoiced'** in the Kanban View:

**Today**
image

**With this PR**

image

The change was based in the way the other informations are included at the view https://github.com/OCA/OCB/blob/18.0/addons/stock/views/stock_picking_type_views.xml#L279

Probably missing in the Migrations PRs:

* https://github.com/OCA/account-invoicing/pull/1794
* https://github.com/OCA/account-invoicing/pull/1942

cc @rvalyi @renatonlima

PR #2326, opened 2026-04-16
[18.0][IMP] sale_stock_picking_invoicing: allow invoicing from SO and prevent double invoicing

RicardCForgeFlow:
**Previously**, the module blocked invoice creation from the Sale Order when the policy was set to 'Stock Picking', forcing users to always invoice from the picking.                     
                                                                  
This PR adds the invoice policy: both. Invoices can now be created from both the **Sale Order** and the **Stock Picking**.


mbcosta:
Hi @RicardCForgeFlow thank you for your PR.

I think will be better preserve the case of 

"blocked invoice creation from the Sale Order when the policy was set to 'Stock Picking', forcing users to always invoice from the picking"

And include the case you are trying implemented as a third option

"Invoices can now be created from both the Sale Order and the Stock Picking."

What do you think? cc @renatonlima @rvalyi 

About the Additional Fixes, please if you can, consider take a look and help the Review in PR

This is important refactoring because change the tests to not use the Demo Data, maybe solve the problems you saw.

And if you can please consider help with another PR linked with **sale_stock_picking_invoiving**, just a simple thing related to View

* https://github.com/OCA/account-invoicing/pull/2300

RicardCForgeFlow:
> Hi @RicardCForgeFlow thank you for your PR.
> 
> I think will be better preserve the case of
> 
> "blocked invoice creation from the Sale Order when the policy was set to 'Stock Picking', forcing users to always invoice from the picking"
> 
> And include the case you are trying implemented as a third option
> 
> "Invoices can now be created from both the Sale Order and the Stock Picking."
> 
> What do you think? cc @renatonlima @rvalyi
> 
> About the Additional Fixes, please if you can, consider take a look and help the Review in PR
> 
> * [[18.0][REF] stock_picking_invoicing, sale_stock_picking_invoicing: Run Tests without Demo Data #2298](https://github.com/OCA/account-invoicing/pull/2298)
> 
> This is important refactoring because change the tests to not use the Demo Data, maybe solve the problems you saw.
> 
> And if you can please consider help with another PR linked with **sale_stock_picking_invoiving**, just a simple thing related to View
> 
> * [[18.0][REF] stock_picking_invoicing: Align 'To be Invoiced' in Kanban View #2300](https://github.com/OCA/account-invoicing/pull/2300)

Hello @mbcosta , thanks for your review. Now I have added the option "both" (Sale Order and Stock Picking).
I have tested different scenarios, Sale Orders with Goods + Service products, and everything looks good.

mbcosta (approved):
Thank you @RicardCForgeFlow for the work and for has taken my Review in consideration.

I'm thinking about include some "Background / Context" of this module and **stock_picking_invoicing** in order to alert other developers about the different scenarios, what can seems as a BUG / Error/ Problem maybe its something necessary for some scenarios / case of use, for example the "double invoincing" is necessary in Brazil but developers from other countries can see this as a something to fix.

rvalyi (approved):
LTGM

RicardCForgeFlow:
> Thank you @RicardCForgeFlow for the work and for has taken my Review in consideration.
> 
> I'm thinking about include some "Background / Context" of this module and **stock_picking_invoicing** in order to alert other developers about the different scenarios, what can seems as a BUG / Error/ Problem maybe its something necessary for some scenarios / case of use, for example the "double invoincing" is necessary in Brazil but developers from other countries can see this as a something to fix.

Yes! I agree :smiley:

PR #2331, opened 2026-04-22
[FIX] account_invoice_discount_date: Sync discount_dates between move and move lines

Shide:
Discount Date preserves on the first write the date given by the Early Payment Terms. Future modifications will be synced with the lines.

**Before:**
When you set a payment terms with early payments and select a new discount date before saving, discount date on the move and discount date on the lines are not synced.

**Now:**
If the Early Payment Term is set and you choose another discount date before saving the move, the Discount Date used will be the Early Payment Term Discount Date. No matter what you choose until move is saved.
Then, you can modify freely the discount date and sync like before.

https://www.loom.com/share/7edab76ed3e245debab2d8a0b9aa11d8

MT-14470 MT-14349 @moduon @rafaelbn @fcvalgar @Gelojr @EmilioPascual @chienandalu please review if you want 😄

### DIFF
--- a/account_invoice_discount_date/models/account_move.py
+++ b/account_invoice_discount_date/models/account_move.py
@@ -33,8 +33,10 @@ class AccountMove(models.Model):
 
     def _inverse_discount_date(self):
         """When set Discount date, update all move lines with Date maturity"""
+        discount_date_field = self._fields["discount_date"]
         for record in self:
             for line in record.line_ids.filtered_domain(
                 [("display_type", "=", "payment_term")]
             ):
                 line.discount_date = record.discount_date
+                self.env.add_to_compute(discount_date_field, record)
--- a/account_invoice_discount_date/readme/USAGE.md
+++ b/account_invoice_discount_date/readme/USAGE.md
@@ -1,8 +1,10 @@
 To use this module, you need to:
 
 1. Go to Accounting > Suppliers > Bills (or Customers > Invoices) and set the Discount Date on a desired bill.
-2. All Move Lines with maturity date set, will update its Discount Date with the Discount Date of the invoice.
-3. You can group invoices to know which Invoice should be paid to ensure the discount benefit.
-4. You can change the Discount Date on any move line. The lower one will be set on the bill.
+1. All Move Lines with maturity date set, will update its Discount Date with the Discount Date of the invoice.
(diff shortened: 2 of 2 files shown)


account_invoice_discount_date/models/account_move.py
```diff
@@ -33,8 +33,10 @@ def _compute_discount_date(self):
 
     def _inverse_discount_date(self):
         """When set Discount date, update all move lines with Date maturity"""
+        discount_date_field = self._fields["discount_date"]
         for record in self:
             for line in record.line_ids.filtered_domain(
                 [("display_type", "=", "payment_term")]
             ):
                 line.discount_date = record.discount_date
+                self.env.add_to_compute(discount_date_field, record)
```
chienandalu, line 42: Won't it enter in an infinite loop between the inverse/compute?

Shide: No, only this happens:
Call to _inverse
Call to _compute

In any case except when modifying only the payment terms, in this case only:
Call to _compute

account_invoice_discount_date/static/description/index.html
```diff
@@ -3,7 +3,7 @@
 <head>
 <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
 <meta name="generator" content="Docutils: https://docutils.sourceforge.io/" />
-<title>README.rst</title>
+<title>Account Invoice Discount Date</title>
```
chienandalu: I'd say to apply the same principle as in https://github.com/OCA/sale-workflow/pull/4129#discussion_r2736433809

chienandalu (approved):
thanks :)

fcvalgar (approved):
Functional review,

LGTM thank you @Shide

Gelojr (approved):
Great work on this contribution @Shide 

- Test 1: Visual check that Discount Date field appears and is populated correctly on supplier invoice creation – OK
- Test 2: Manual change of Discount Date on invoice correctly propagates to payment term lines after saving – OK
- Test 3: Changing Discount Date before saving does not update lines, but after saving it synchronizes correctly with payment term lines – OK
- Test 4: Invoices can be grouped by Discount Date to prioritize payments – OK
- Test 5: Updating Discount Date after invoice is saved correctly updates the related payment term line – OK

rafaelbn (approved):
tested functionally 👍🏼 

/ocabot merge patch

PR #2333, opened 2026-04-27
[18.0][IMP] stock_picking_invoicing: restrict picking invoice buttons to Invoicing group

RicardCForgeFlow:
image

The invoice buttons on the` stock.picking` form were visible to any internal user, while the `invoice_state`  field they act on is already restricted to `account.group_account_invoice`


MiquelRForgeFlow (approved):

AaronHForgeFlow (approved):
👍  It is rare this was not already like this.

stock_picking_invoicing/__manifest__.py
```diff
@@ -2,7 +2,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 {
     "name": "Stock Picking Invoicing",
-    "version": "18.0.1.0.2",
+    "version": "18.0.1.0.3",
```
AaronHForgeFlow: shouldn't we let the bot do it?

RicardCForgeFlow:
Hi @etobella, when you have time, could you review this? Thank you!

rvalyi (approved):

rvalyi:
makes sense. Thanks!

/ocabot merge minor

PR #2342, opened 2026-04-30
[19.0][MIG] partner_invoicing_mode: Migration to 19.0

yankinmax:


partner_invoicing_mode/tests/test_invoice_mode.py
```diff
@@ -0,0 +1,174 @@
...
+        self._confirm_and_deliver(self.so1)
+        self.assertFalse(self.so1.invoice_ids)
+        with trap_jobs() as trap:
+            self.SaleOrder.cron_generate_standard_invoices()
+            with trap_jobs() as trap_validate:
+                trap.perform_enqueued_jobs()
+            # TODO: understand is it the code issue or the test one?
+            # there should be only one job enqueued, because only `so1`
+            # was confirmed and delivered, but for some reason, there are two.
+            self.assertEqual(1, len(trap_validate.enqueued_jobs))
```
yankinmax: This is failing and I couldn't find a proper solution.
Help is appreciated.

partner_invoicing_mode/tests/test_invoice_mode.py
```diff
@@ -128,14 +119,15 @@ def test_invoicing_standard_cron_invoice_date(self):
...
         with trap_jobs() as trap:
-            self.cron.method_direct_trigger()
-            for job in trap.enqueued_jobs:
-                with trap_jobs() as trap_validate:
-                    job.perform()
-                self.assertEqual(1, len(trap_validate.enqueued_jobs))
-                for validate_job in trap_validate.enqueued_jobs:
-                    with freeze_time("2024-03-11"):
-                        validate_job.perform()
+            self.SaleOrder.cron_generate_standard_invoices()
```
yankinmax: To fix `psycopg2.InterfaceError: Cursor already closed` error

partner_invoicing_mode/tests/test_invoice_mode.py
yankinmax: Solved

Ricardoalso (approved):
LGTM

partner_invoicing_mode/models/sale_order.py
```diff
@@ -139,4 +133,4 @@ def _generate_invoices_by_partner(self, saleorder_ids):
 
     @api.model
     def _get_companies_standard_invoicing(self):
-        return self.env["res.company"].search([])
+        return self.env["res.company"].search(Domain.TRUE)
```
ivantodorovich, line 136: This change makes no sense IMO. `[]` to represent and empty domain is still the "odoo-way" AFAIK.

ivantodorovich (approved):
LG. Thanks!!

partner_invoicing_mode/models/sale_order.py
gdgellatly: I don't know, I've been using the new way a lot, it does help in inheritance chains and there is no way I have found of having an empty domain object. I would not change your code based on this comment. Yes it takes to get used to, but it is correct for new Domain objects based on everything I've seen and done.

Ricardoalso: This is a pattern I also discovered recently. You can actually pass a wide variety of parameter types  (bool, list, or tuple) to initialize an empty Domain object, and all of them will still evaluate to _TRUE_DOMAIN (if empty or true) . https://github.com/OCA/OCB/blob/4a754c7b31f14b810a6ffe12c06b2dd0adb6cd0b/odoo/orm/domains.py#L236

Example in an odoo shell:
```
>>> Domain(Domain.TRUE) & Domain("partner_id","=",1)
[('partner_id', '=', 1)]

>>> Domain(True) & Domain("partner_id","=",1)
[('partner_id', '=', 1)]

>>> Domain([]) & Domain("partner_id","=",1)
[('partner_id', '=', 1)]

>>> Domain(()) & Domain("partner_id","=",1)
[('partner_id', '=', 1)]
```

yankinmax: Here what `pre-commit` also says (I suppose it's already decided what to use :smile:):
```
pylint with optional checks..............................................Passed
- hook id: pylint_odoo
- duration: 3.21s

************* Module partner_invoicing_mode.models.sale_order
partner_invoicing_mode/models/sale_order.py:136: [W8163(no-search-all), SaleOrder._get_companies_standard_invoicing] Using an empty domain `search([])` without a `limit` will load all records, may impact performance.
```
So, I leave `Domain.TRUE`

Ricardoalso: That's a pre-commit loophole because search(Domain.TRUE) fetches the same records as search([]) :smile:

Issue #2154, opened 2025-11-03
Module account_invoice_blocking not migrated to 18.0 yet


lorenzoallegrucci:
There was a PR for the account_invoice_blocking migration to 18.0:
https://github.com/OCA/account-invoicing/pull/2049 
but it was closed, why? https://github.com/bjouini-acsone

## Module
account_invoice_blocking

## Describe the bug
account_invoice_blocking is not migrated to 18.0 yet

## To Reproduce
**Affected versions**: 18.0

**Expected behavior**
account_invoice_blocking migrated to 18.0

aviralgarg05:
I would like to solve this issue

solomonprabu:
@lorenzoallegrucci I'm willing to continue with `account_invoice_blocking` migration to 18.0

solomonprabu:
> [@lorenzoallegrucci](https://github.com/lorenzoallegrucci) I'm willing to continue with `account_invoice_blocking` migration to 18.0

Work discontinued, Check https://github.com/OCA/account-invoicing/pull/2345 comments.

PR #2345, opened 2026-05-08
[18.0][MIG] account_invoice_blocking: Migration to 18.0

solomonprabu:
On successful migration of this module, issue:https://github.com/OCA/account-invoicing/issues/2154 can be resolved.


solomonprabu:
The function invoice blocking has been removed from base Odoo account module in the following commit https://github.com/odoo/odoo/commit/67dc71588751c9f115e09959d774686c2207a1f9. As this module relies on it, this makes the module:`account_invoice_blocking` redundant and no longer needed.

PR #2350, opened 2026-05-15
[16.0][MIG] account_billing_portal: Migration to 16.0

AungKoKoLin1997:
Based on https://github.com/OCA/account-invoicing/pull/2186

@qrtl QT5383


account_billing_portal/security/account_billing_portal_security.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
```
yostashiro:
```suggestion
<odoo noupdate="1">
```

account_billing_portal/security/account_billing_portal_security.xml
```diff
@@ -0,0 +1,19 @@
...
+    <record id="account_billing_rule_portal" model="ir.rule">
+        <field name="name">Portal Billings</field>
+        <field name="model_id" ref="account_billing.model_account_billing" />
+        <field
+            name="domain_force"
+        >[('state', '=', 'billed'), ('message_partner_ids','child_of',[user.commercial_partner_id.id])]</field>
+        <field name="groups" eval="[(4, ref('base.group_portal'))]" />
```
yostashiro: Please use Command.

account_billing_portal/security/account_billing_portal_security.xml
AungKoKoLin1997: Done!

account_billing_portal/security/account_billing_portal_security.xml
AungKoKoLin1997: Done!

account_billing_portal/models/account_billing.py
```diff
@@ -0,0 +1,113 @@
...
+            "dateutil": tools.safe_eval.dateutil,
+            "timezone": tools.safe_eval.pytz.timezone,
+            "context_today": lambda: fields.Date.context_today(self),
+            "object": self,
+        }
+
+    def action_billing_send(self):
+        self.ensure_one()
+        template = self.company_id.billing_email_template_id or self.env.ref(
+            "account_billing_portal.email_template_billing"
```
yostashiro: Otherwise, it won't reach the UserError part due to ValueError.

```suggestion
        template = self.company_id.billing_email_template_id or self.env.ref(
            "account_billing_portal.email_template_billing", raise_if_not_found=False
```

account_billing_portal/models/account_billing.py
```diff
@@ -0,0 +1,113 @@
...
+            raise UserError(
+                _("Please configure the Billing Email Template in the settings.")
+            )
+        compose_form = self.env.ref(
+            "mail.email_compose_message_wizard_form", raise_if_not_found=False
+        )
+        compose_form_id = compose_form.id if compose_form else False
+        ctx = dict(self.env.context or {})
+        report = self.company_id.billing_portal_report or self.env.ref(
+            "account_billing.report_account_billing"
```
yostashiro:
```suggestion
            "account_billing.report_account_billing", raise_if_not_found=False
```

account_billing_portal/models/account_billing.py
```diff
@@ -0,0 +1,113 @@
...
+        if not report:
+            raise UserError(
+                _("Please configure the Billing Portal Report in the settings.")
+            )
+        pdf_content, _type = report.with_company(self.company_id)._render_qweb_pdf(
+            report.id, self.ids
+        )
+        if report.print_report_name:
+            eval_context = self._get_eval_context()
+            attachment_name = safe_eval(report.print_report_name, eval_context)
```
yostashiro: Can we do this and drop _get_eval_context() as most values are not used?

```suggestion
            try:
                attachment_name = safe_eval(report.print_report_name, {"object": self})
            except Exception:
                attachment_name = self.display_name
```

account_billing_portal/models/account_billing.py
```diff
@@ -0,0 +1,113 @@
...
+        attach = self.env["ir.attachment"].create(
+            {
+                "name": attachment_name,
+                "type": "binary",
+                "datas": base64.b64encode(pdf_content),
+                "mimetype": "application/pdf",
+                "res_model": "account.billing",
+                "res_id": self.id,
+            }
+        )
```
yostashiro: The current approach looks a bit unorthodox. Can we remove this and let email template generate the report by adding `report_template` and `report_name` fields to the template instead?
```suggestion
```

yostashiro (changes requested):

account_billing_portal/models/account_billing.py
AungKoKoLin1997: Done!

account_billing_portal/models/account_billing.py
AungKoKoLin1997: Done!

account_billing_portal/models/account_billing.py
AungKoKoLin1997: Yes. It is better.

account_billing_portal/models/account_billing.py
AungKoKoLin1997: Do you want to remove the company-level report configuration and allow users to update the report directly in the mail template if they want to use a different one?

If so, we will no longer be able to use different reports for each company, such as using the Summary Report for the JP company and the Billing Report for the US company.

yostashiro: Let's remove the company-level report config. Company-specific report can be achieved through company-level email template settings.

AungKoKoLin1997: I updated!

Issue #2317, opened 2026-04-10
[19.0] account_invoice_fixed_discount: discount_fixed getting set to 0


evanomnisoft:
When I set fixed discount the first time on a invoice line, everything goes well other than the fact that the fixed discount field gets set back to 0.

https://github.com/OCA/account-invoicing/blob/f6fd6b0a91560a8f634aad6ca241a7fa299562bb/account_invoice_fixed_discount/models/account_move_line.py#L77

It seems the initial intent is to "disable" the fixed discount field if discount% field is used by setting it to 0 but that leads inconsistency in data since the 2 discount fields will not tally. Would it be better to just reverse compute the field as such?

`self.discount_fixed = self.price_unit * (self.discount / 100.0)`

legalsylvain:
Could you make a PR with failing test and your solution ?

PR #2352, opened 2026-05-17
[FIX] account_invoice_fixed_discount: recalculate discount_fixed on percentage change (19.0)

fsmw:
[FIX] account_invoice_fixed_discount: recalculate discount_fixed on percentage change

When users set a fixed discount and then edit the percentage
discount field, the onchange unconditionally reset discount_fixed
to 0.0, silently destroying the user input.

Instead of zeroing out discount_fixed, recalculate it to its
equivalent fixed amount based on the percentage:
    discount_fixed = price_unit * (discount / 100.0)

This keeps both discount fields consistent at all times.

A new regression test is added to prevent future regression.

Fixes OCA/account-invoicing#2317


CRogos (approved):
Code + Test LGTM

@fsmw maybe rebase so runboat is available again?

PR #2354, opened 2026-05-18
[18.0][MIG] account_invoice_line_sequence: Standard migration

tendil:
Task: 5419


mmircoli-nexapp (approved):

BhaveshHeliconia (commented):
I reviewed the migration and noticed that Odoo 18 already provides a native sequence field on account.move.line.

Could you clarify what additional functional value this module still provides in 18.0?

If the module only adds ordering support for invoice lines, it may now be redundant because core Odoo already handles this behavior.

However, if the module introduces additional business logic (custom sequencing behavior, reporting logic, compliance constraints, migration compatibility, etc.), it would help to document that explicitly in the README or migration notes.

tendil:
> I reviewed the migration and noticed that Odoo 18 already provides a native sequence field on account.move.line.
> 
> Could you clarify what additional functional value this module still provides in 18.0?
> 
> If the module only adds ordering support for invoice lines, it may now be redundant because core Odoo already handles this behavior.
> 
> However, if the module introduces additional business logic (custom sequencing behavior, reporting logic, compliance constraints, migration compatibility, etc.), it would help to document that explicitly in the README or migration notes.

---

thanks for the review!

the module still provides a user-facing invoice line number, not only technical line ordering. It recalculates visible numbers for invoice lines, skips section/note lines, shows the number in the invoice line list, and prints it on the invoice report;

i clarified the module description to make this purpose more explicit

BhaveshHeliconia (commented):
@tendil Thanks for the clarification.

I understand the distinction now — the module provides functional/user-visible invoice line numbering rather than only relying on the technical sequence field already present in core Odoo 18.

The updated description makes the use case much clearer.

LGTM!

BhaveshHeliconia (approved):

mmircoli-nexapp:
merge?

tendil:
> merge?

---

there were two commits as part of the migration
i've already made one commit

Aldeigja (approved):
Functional LGTM

rvalyi (approved):
LGTM

PR #2360, opened 2026-05-20
[19.0][MIG] account_invoice_warn_message: Migration to 19.0

dnplkndll:
Port of `account_invoice_warn_message` from 18.0 to 19.0. Follows the [OCA migration guide](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-19.0).

Non-mechanical adaptations worth flagging:

- Odoo 19 dropped `res.partner.invoice_warn` (the Selection field that gated whether to warn). The module's compute now keys on `invoice_warn_msg` being non-empty instead of the gone Selection value.
- `res.partner.invoice_warn_msg` is re-added by this module (Odoo 19 removed it from core too) — new `models/res_partner.py` + a `views/res_partner_views.xml` group on the partner form to expose it.
- `_compute_invoice_warn_msg` rewritten to use `OrderedSet` for dedup + prefix each message with the partner name when both child and parent partners carry warnings, instead of f-string concatenation.


dnplkndll:
ready for review @BhaveshHeliconia @JasminSForgeFlow

JasminSForgeFlow:
Tests are failing, can you please fix that?

BhaveshHeliconia (approved):
LGTM!

PR #2364, opened 2026-05-26
[18.0][ADD] account_multi_discount

LorenzoC0:
This module adds a distribution-like field in the move line discount column, enabling infinite multiplicative discounts.

This module likely conflicts with some other modules that modify the behavior of the `discount` field, such as `account_invoice_triple_discount`. I could use some help figuring out which of these modules are incompatible, so I can add them as exclusions in the manifest. Thanks!


account_multi_discount/static/src/components/discount_distribution/discount_distribution.xml
```diff
@@ -0,0 +1,109 @@
...
+
+    <t t-name="account_multi_discount.DiscountDistributionPopup">
+        <div class="popover discount_distribution_popup dropdown-menu o-dropdown--menu show rounded py-0 overflow-x-hidden"
+             t-if="state.showDropdown"
+             t-ref="discountDropdown">
+
+            <!-- Header -->
+            <div class="popover-header sticky-top">
+                <div class="d-flex align-items-center">
+                    <div class="h5 mt-2 me-auto">Sconti</div>
```
eLBati: to be translated

account_multi_discount/static/src/components/discount_distribution/discount_distribution.xml
```diff
@@ -0,0 +1,109 @@
...
+                    <div>
+                        <span class="btn o_button"
+                              t-on-click.stop="() => this.forceCloseDropdown()"
+                              title="Chiudi">
```
eLBati: to be translated

account_multi_discount/static/src/components/discount_distribution/discount_distribution.xml
```diff
@@ -0,0 +1,109 @@
...
+                </div>
+            </div>
+
+            <!-- Body -->
+            <div class="popover-body p-2">
+                <table class="o_list_table table table-sm table-hover mb-2 table-striped">
+                    <thead>
+                        <tr class="border-bottom">
+                            <th class="text-center" style="width: 40px;">#</th>
+                            <th>Sconto (%)</th>
```
eLBati: to be translated

account_multi_discount/static/src/components/discount_distribution/discount_distribution.xml
```diff
@@ -0,0 +1,109 @@
...
+                                      t-on-click.stop="() => this.deleteLine(disc_index)"/>
+                            </td>
+                        </tr>
+
+                        <!-- Add line -->
+                        <tr>
+                            <td t-on-click.stop.prevent="addLine"
+                                class="o_field_x2many_list_row_add"
+                                colspan="3">
+                                <a href="#">Aggiungi sconto</a>
```
eLBati: to be translated

account_multi_discount/static/src/components/discount_distribution/discount_distribution.xml
```diff
@@ -0,0 +1,109 @@
...
+                            </td>
+                        </tr>
+                    </tbody>
+
+                    <!-- Total row -->
+                    <tfoot t-if="state.discounts.length > 1">
+                        <tr class="fw-bold border-top">
+                            <td/>
+                            <td class="text-end" colspan="2">
+                                Totale: <t t-esc="aggregatedDiscount"/>%
```
eLBati: to be translated

account_multi_discount/__manifest__.py
```diff
@@ -0,0 +1,27 @@
...
+    "category": "Accounting/Accounting",
+    "author": "Innovyou, Odoo Community Association (OCA)",
+    "maintainers": ["LorenzoC0"],
+    "website": "https://github.com/OCA/account-invoicing",
+    "license": "LGPL-3",
+    "depends": [
+        "account",
+    ],
+    "data": [
+        "views/account_move_views.xml",
```
eLBati, line 20: as in sale_multi_discount, wouldn't the printing modification also be necessary?

account_multi_discount/models/account_move_line.py
```diff
@@ -0,0 +1,64 @@
...
+        have configured through the widget, while still allowing simple
+        legacy writes (``line.discount = 10``) to round-trip correctly.
+        """
+        for line in self:
+            if not line.discount_distribution or len(
+                line.discount_distribution
+            ) <= 1:
+                line.discount_distribution = (
+                    [line.discount] if line.discount else []
+                )
```
eLBati: In my opinion, if the user writes discount, discount_distribution should always be overwritten, otherwise it would remain misaligned

account_multi_discount/__manifest__.py
```diff
@@ -0,0 +1,27 @@
...
+               "through a distribution edited via a dedicated widget.",
+    "version": "18.0.1.0.0",
+    "category": "Accounting/Accounting",
+    "author": "Innovyou, Odoo Community Association (OCA)",
+    "maintainers": ["LorenzoC0"],
+    "website": "https://github.com/OCA/account-invoicing",
+    "license": "LGPL-3",
+    "depends": [
+        "account",
+    ],
```
eLBati, line 15: I think you need excludes for account_invoice_triple_discount

account_multi_discount/static/src/components/discount_distribution/discount_distribution.xml
```diff
@@ -0,0 +1,109 @@
...
+                    <tbody>
+                        <tr t-foreach="state.discounts" t-as="disc" t-key="disc_index">
+                            <td class="text-center text-muted align-middle">
+                                <t t-esc="disc_index + 1"/>
+                            </td>
+                            <td>
+                                <input type="number"
+                                       class="o_input w-100"
+                                       step="0.01"
+                                       min="0"
```
eLBati: negative discounts are allowed if I'm not mistaken

account_multi_discount/models/account_move_line.py
```diff
@@ -0,0 +1,64 @@
...
+
+    @api.depends("discount_distribution")
+    def _compute_totals(self):
+        # Re-declare depends so the invoice totals refresh when the
+        # distribution is edited.
+        return super()._compute_totals()
```
eLBati: Since discount is stored and depends on discount_distribution, the standard depends chain already
  triggers totals recomputation. If you've verified this is actually needed , document why; otherwise drop it.

account_multi_discount/static/src/components/discount_distribution/discount_distribution.js
```diff
@@ -0,0 +1,217 @@
...
+            this.state.discounts = [...val];
+        }
+    }
+}
+
+export const discountDistribution = {
+    component: DiscountDistribution,
+    displayName: _t("Discount Distribution"),
+    supportedTypes: ["json", "jsonb", "char", "text"],
+    extractProps: ({ attrs, options }) => ({}),
```
eLBati: I think it can be eliminated

eLBati (changes requested):

account_multi_discount/__manifest__.py
eLBati: and account_global_discount

LorenzoC0: @eLBati We noticed that `account_global_discount` isn't needed since it only operates at the `account.move` level, not in the `account.move.line` model like we do in this new module. I removed `account_global_discount` from the `account_multi_discount` exclusions.

PR #2368, opened 2026-05-27
[18][ADD] account_move_restricted_group: Add new module to restrict access to account moves

DantePereyra:
This module overrides access control with `check_access()`. It ensures that:
- Creation, Edition, and Deletion on `account.move` are strictly blocked for users who are not part of the `account.group_account_invoice`.
- Users with `Technical/Show Accounting Features - Readonly` group or no Accounting role defined retain their original, restricted permissions, preventing other apps from escalating their privileges.


cgarcia-solvos (approved):
Code review. LGTM! Only minor comment:

account_move_restricted_group/tests/test_account_move_restricted_group.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def test_restricted_user_can_read_move(self):
+        """User can read an existing account.move"""
+        move_restricted = self.move.with_user(self.restricted_user)
+        try:
+            _ = move_restricted.read(["name", "state"])
+        except AccessError:
+            self.fail("""
+                The restricted user should be able to READ (read) the account moves.
+            """)
```
cgarcia-solvos: Codecov is failing because of the except block. I think you can remove the try-except block entirely and it should not affect the test logic and will fix the coverage.

```suggestion
        _ = move_restricted.read(["name", "state"])
```

DantePereyra: Change done!

PR #2373, opened 2026-06-02
[18.0][ADD] sale_invoiceability_on_payment_transaction: New Module

cristina-hidalgo-tecnativa:
Adds a new module that makes sale order lines invoiceable when a fully paid payment transaction exists, regardless of the product invoice policy. This allows creating invoices without a validated delivery when the customer has already paid.

@Tecnativa TT62394
@pedrobaeza @carlos-lopez-tecnativa


sale_invoiceability_on_payment_transaction/models/sale_order_line.py
```diff
@@ -0,0 +1,34 @@
+from odoo import api, models
+
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    def _get_paid_lines(self):
+        return self.filtered(
+            lambda line: line.product_id.type != "service"
```
pedrobaeza: The type of product doesn't matter. Only if it's paid or not.

sale_invoiceability_on_payment_transaction/models/sale_order_line.py
```diff
@@ -0,0 +1,34 @@
+from odoo import api, models
+
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    def _get_paid_lines(self):
```
pedrobaeza: To avoid also method names conflicts:

```suggestion
    def _filter_paid_lines_for_invocing(self):
```

sale_invoiceability_on_payment_transaction/models/sale_order_line.py
```diff
@@ -0,0 +1,34 @@
...
+            and line.state in ("sale", "done")
+            and line.order_id._is_paid()
+        )
+
+    @api.depends("order_id.transaction_ids.state")
+    def _compute_qty_to_invoice(self):
+        paid_lines = self._get_paid_lines()
+        for line in paid_lines:
+            line.qty_to_invoice = line.product_uom_qty - line.qty_invoiced
+        return super(SaleOrderLine, self - paid_lines)._compute_qty_to_invoice()
```
pedrobaeza: For clearer reading:

```suggestion
        self -= paid_lines
        return super()._compute_qty_to_invoice()
```

sale_invoiceability_on_payment_transaction/models/sale_order_line.py
```diff
@@ -0,0 +1,34 @@
...
+        return self.filtered(
+            lambda line: line.product_id.type != "service"
+            and line.state in ("sale", "done")
+            and line.order_id._is_paid()
+        )
+
+    @api.depends("order_id.transaction_ids.state")
+    def _compute_qty_to_invoice(self):
+        paid_lines = self._get_paid_lines()
+        for line in paid_lines:
```
pedrobaeza:
```suggestion
        for line in self._filter_paid_lines_for_invocing():
```

sale_invoiceability_on_payment_transaction/models/sale_order_line.py
```diff
@@ -0,0 +1,34 @@
...
+        paid_lines = self._get_paid_lines().filtered(
+            lambda line: line.product_id.invoice_policy == "delivery"
+        )
+        super(SaleOrderLine, self - paid_lines)._compute_untaxed_amount_to_invoice()
+        if paid_lines:
+            invoice_policy_field = self.env["product.product"]._fields["invoice_policy"]
+            with self.env.protecting([invoice_policy_field], paid_lines.product_id):
+                for product in paid_lines.product_id:
+                    product.invoice_policy = "order"
+                return super(
```
pedrobaeza: Do `return True` at the end for both branches.

sale_invoiceability_on_payment_transaction/models/sale_order_line.py
```diff
@@ -0,0 +1,34 @@
...
+    def _compute_untaxed_amount_to_invoice(self):
+        paid_lines = self._get_paid_lines().filtered(
+            lambda line: line.product_id.invoice_policy == "delivery"
+        )
+        super(SaleOrderLine, self - paid_lines)._compute_untaxed_amount_to_invoice()
+        if paid_lines:
+            invoice_policy_field = self.env["product.product"]._fields["invoice_policy"]
+            with self.env.protecting([invoice_policy_field], paid_lines.product_id):
+                for product in paid_lines.product_id:
+                    product.invoice_policy = "order"
```
pedrobaeza: Please don't change the field, that can cause other concatenated changes.

sale_invoiceability_on_payment_transaction/models/sale_order_line.py
cristina-hidalgo-tecnativa: Done.

sale_invoiceability_on_payment_transaction/models/sale_order_line.py
cristina-hidalgo-tecnativa: Done.

sale_invoiceability_on_payment_transaction/models/sale_order_line.py
cristina-hidalgo-tecnativa: Done.

sale_invoiceability_on_payment_transaction/models/sale_order_line.py
cristina-hidalgo-tecnativa: Modified.

sale_invoiceability_on_payment_transaction/models/sale_order_line.py
cristina-hidalgo-tecnativa: Modified.

sale_invoiceability_on_payment_transaction/models/sale_order_line.py
cristina-hidalgo-tecnativa: Removed the `env.protecting` + field mutation approach entirely. Now we call `super()` for all lines and then recalculate `untaxed_amount_to_invoice` directly for paid `delivery`-policy lines using `product_uom_qty` as the quantity base, without touching the field at all.

sale_invoiceability_on_payment_transaction/models/sale_order_line.py
```diff
@@ -0,0 +1,43 @@
...
+            line.qty_to_invoice = line.product_uom_qty - line.qty_invoiced
+        self -= paid_lines
+        return super()._compute_qty_to_invoice()
+
+    @api.depends("order_id.transaction_ids.state")
+    def _compute_untaxed_amount_to_invoice(self):
+        paid_lines = self._filter_paid_lines_for_invocing().filtered(
+            lambda line: line.product_id.invoice_policy == "delivery"
+        )
+        super()._compute_untaxed_amount_to_invoice()
```
pedrobaeza: Don't call super with all the records.

```suggestion
        self -= paid_lines
        super()._compute_untaxed_amount_to_invoice()
```

sale_invoiceability_on_payment_transaction/models/sale_order_line.py
```diff
@@ -0,0 +1,43 @@
...
+        for line in paid_lines:
+            if line.state != "sale":
```
pedrobaeza: Put a comment saying that this is a copy of the upstream code, adapted to directly use always `product_uom_qty`.

Always remember to document properly for your future you. You remember now perfectly the reason and the origin of the code, but in 2 years?

sale_invoiceability_on_payment_transaction/__manifest__.py
```diff
@@ -0,0 +1,12 @@
+{
+    "name": "Sale Invoiceability on Payment Transaction",
+    "summary": "Allow invoicing sale orders when a done payment transaction exists, "
```
pedrobaeza:
```suggestion
    "summary": "Allow invoicing sales orders when a done payment transaction exists."
```

sale_invoiceability_on_payment_transaction/__manifest__.py
```diff
@@ -0,0 +1,12 @@
+{
+    "name": "Sale Invoiceability on Payment Transaction",
```
pedrobaeza:
```suggestion
    "name": "Invoice orders with done payment transactions",
```

sale_invoiceability_on_payment_transaction/readme/CONTRIBUTORS.md
```diff
@@ -0,0 +1,2 @@
+- [Tecnativa](https://www.tecnativa.com):
+  - Cristina Hidalgo
```
pedrobaeza, line 3: When someone actively participates in design/review, it should be included here.

sale_invoiceability_on_payment_transaction/models/sale_order_line.py
cristina-hidalgo-tecnativa: Modified now

sale_invoiceability_on_payment_transaction/__manifest__.py
cristina-hidalgo-tecnativa: Fixed!

sale_invoiceability_on_payment_transaction/readme/CONTRIBUTORS.md
cristina-hidalgo-tecnativa: Modified!

sale_invoiceability_on_payment_transaction/__manifest__.py
cristina-hidalgo-tecnativa: Fixed!

sale_invoiceability_on_payment_transaction/models/sale_order_line.py
cristina-hidalgo-tecnativa: Thanks for the reminder. Added an inline comment above the block.

sale_invoiceability_on_payment_transaction/models/sale_order_line.py
```diff
@@ -0,0 +1,46 @@
+from odoo import api, models
+
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    def _filter_paid_lines_for_invocing(self):
+        return self.filtered(
+            lambda line: line.state in ("sale", "done") and line.order_id._is_paid()
+        )
```
carlos-lopez-tecnativa, line 12: The `done` option no longer exists. A confirmed sales order is in the `sale` state.

```suggestion
        return self.filtered(
            lambda line: line.state ==  "sale" and line.order_id._is_paid()
        )
```

sale_invoiceability_on_payment_transaction/models/sale_order_line.py
```diff
@@ -0,0 +1,46 @@
...
+    def _compute_untaxed_amount_to_invoice(self):
+        paid_lines = self._filter_paid_lines_for_invocing().filtered(
+            lambda line: line.product_id.invoice_policy == "delivery"
+        )
+        self -= paid_lines
+        super()._compute_untaxed_amount_to_invoice()
+        # Adapted from sale.order.line._compute_untaxed_amount_to_invoice (upstream),
+        # using product_uom_qty instead of qty_delivered for paid delivery-policy lines.
+        for line in paid_lines:
+            if line.state != "sale":
```
carlos-lopez-tecnativa: Based on my previous comment, `paid_lines` comes from `_filter_paid_lines_for_invoicing`, and that method only returns lines with `state == "sale"`.

Therefore, I don't see a case where `state != "sale"`. From my point of view, this condition is unnecessary and can be removed.

Or is there a case where the code can enter here with a different state?

carlos-lopez-tecnativa (dismissed):

sale_invoiceability_on_payment_transaction/models/sale_order_line.py
pedrobaeza: Please perform the needed changes, Carlos, as Cristina is off these days.

carlos-lopez-tecnativa: Done. I improved the test to cover all the fields involved. Additionally, I recompute the amount_to_invoice field, which was not taken into account before.

sale_invoiceability_on_payment_transaction/readme/CONTRIBUTORS.md
```diff
@@ -0,0 +1,3 @@
+- [Tecnativa](https://www.tecnativa.com):
+  - Pedro M. Baeza
+  - Cristina Hidalgo
```
pedrobaeza, line 3: Carlos, put yourself here.

sale_invoiceability_on_payment_transaction/models/sale_order_line.py
```diff
@@ -0,0 +1,58 @@
...
+        self -= paid_lines
+        return super()._compute_amount_to_invoice()
+
+    @api.depends("order_id.transaction_ids.state")
+    def _compute_untaxed_amount_to_invoice(self):
+        paid_lines = self._filter_paid_lines_for_invocing().filtered(
+            lambda line: line.product_id.invoice_policy == "delivery"
+        )
+        self -= paid_lines
+        super()._compute_untaxed_amount_to_invoice()
```
pedrobaeza: If the super can go at the end, put it.

pedrobaeza:
About returned goods, have you checked if the numbers are OK in that case?

carlos-lopez-tecnativa:
> About returned goods, have you checked if the numbers are OK in that case?

Sorry, but I don't completely understand what you mean. Are you referring to returning a stock.picking? If so, how should that impact this case? Could you please clarify?

carlos-lopez-tecnativa:
@pedrobaeza, please check again. I updated the code so that this module changes the logic only when `qty_delivered` is zero. In that case, the module allows the user to fully invoice the sales order. However, once any picking has been completed, the computations are handled by Odoo's standard flow, which takes into account delivered quantities, returned quantities, and other related scenarios.

Please let me know if this approach works for you.

pedrobaeza:
I'm afraid that doesn't serve, as you may deliver 1 qty, and return that 1 qty, so having quantity_delivered = 0.

carlos-lopez-tecnativa:
> I'm afraid that doesn't serve, as you may deliver 1 qty, and return that 1 qty, so having quantity_delivered = 0.

You are right. I changed the approach to apply these changes only when no stock moves have been processed. Once any move has been completed, Odoo's standard logic takes over and performs the computations. What do you think about this approach?

pedrobaeza (approved):
/ocabot merge nobump

sale_invoiceability_on_payment_transaction/models/sale_order_line.py
```diff
@@ -0,0 +1,64 @@
+from odoo import api, models
+
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    def _filter_paid_lines_for_invocing(self):
```
FrancoMaxime:
```suggestion
    def _filter_paid_lines_for_invoicing(self):
```

sale_invoiceability_on_payment_transaction/models/sale_order_line.py
```diff
@@ -0,0 +1,64 @@
...
+        )
+
+    def _is_delivery_started(self):
+        """Check if the delivery of the line has been started,
+        i.e. if there is at least one done move."""
+        return any(move.state == "done" for move in self.move_ids)
+
+    @api.depends("order_id.transaction_ids.state")
+    def _compute_qty_to_invoice(self):
+        paid_lines = self._filter_paid_lines_for_invocing()
```
FrancoMaxime:
```suggestion
        paid_lines = self._filter_paid_lines_for_invoicing()
```

sale_invoiceability_on_payment_transaction/models/sale_order_line.py
```diff
@@ -0,0 +1,64 @@
...
+        for line in paid_lines:
+            line.qty_to_invoice = line.product_uom_qty - line.qty_invoiced
+        self -= paid_lines
+        return super()._compute_qty_to_invoice()
+
+    @api.depends("order_id.transaction_ids.state")
+    def _compute_amount_to_invoice(self):
+        paid_lines = self._filter_paid_lines_for_invocing()
```
FrancoMaxime:
```suggestion
        paid_lines = self._filter_paid_lines_for_invoicing()
```

sale_invoiceability_on_payment_transaction/models/sale_order_line.py
```diff
@@ -0,0 +1,64 @@
...
+        for line in paid_lines:
+            # Compute the field based on product_uom_qty instead of qty_delivered.
+            if line.product_uom_qty:
+                uom_qty_to_consider = line.product_uom_qty
+                qty_to_invoice = uom_qty_to_consider - line.qty_invoiced_posted
+                unit_price_total = line.price_total / line.product_uom_qty
```
FrancoMaxime, line 35: Maybe its better to use  a float_round to avoid some FloatingPointError like (0.300000000000004) ?

What do you think @pedrobaeza @cristina-hidalgo-tecnativa ?

sale_invoiceability_on_payment_transaction/models/sale_order_line.py
```diff
@@ -0,0 +1,64 @@
...
+                line.amount_to_invoice = unit_price_total * qty_to_invoice
+            else:
+                line.amount_to_invoice = 0.0
+        self -= paid_lines
+        return super()._compute_amount_to_invoice()
+
+    @api.depends("order_id.transaction_ids.state")
+    def _compute_untaxed_amount_to_invoice(self):
+        paid_lines = self._filter_paid_lines_for_invocing().filtered(
```
FrancoMaxime:
```suggestion
        paid_lines = self._filter_paid_lines_for_invoicing().filtered(
```

FrancoMaxime (commented):
Thanks for your work 

LGTM: a typo in a method and a little question.

sbidoul:
I reserved odoo-addon-sale_invoiceability_on_payment_transaction on PyPI.

sale_invoiceability_on_payment_transaction/models/sale_order_line.py
carlos-lopez-tecnativa: This code is taken from the sales module. In the database, the value is stored without rounding, and Odoo applies the decimal precision when displaying it. Therefore, I don't think this is necessary.

https://github.com/odoo/odoo/blob/281658e86971687656f3235ac1ff8afcb52f2908/addons/sale/models/sale_order_line.py#L1148-L1152

FrancoMaxime: Ok perfect, thanks for the explanation

PR #1992, opened 2025-05-01
[14.0] [IMP] stock_picking_invoicing: restrict picking cancellation

rrebollo:
This PR proposes an improvement to the `stock_picking_invoicing` addon by introducing a restriction to prevent the cancellation of stock moves already linked to invoices or vendor bills. Only users who are members of the **"Allow to cancel stock moves linked to invoices/bills"** group will be authorized to perform this action.

## Why Restrict Cancelling Stock Moves Linked to Invoices/Bills

- **Data Integrity**: Prevents inconsistencies between inventory and accounting.
- **Controlled Access**: Cancellation is still allowed, but only for users in the designated security group.
- **Legal and Financial Compliance**: Avoids issues with invoiced or billed records.
- **Auditability**: Maintains a clear and traceable record of exceptional actions.
- **Best Practice**: Encourages using proper workflows like credit notes or returns.

## ROADMAP

- [x] Create security group
- [x] Restrict cancellation at the stock move model level
- [x] Update translations
- [x] Create a test

@BinhexTeam

PR #2374, opened 2026-06-03
[18.0] [IMP] stock_picking_invoicing: restrict picking cancellation

rrebollo:
This PR proposes an improvement to the `stock_picking_invoicing` addon by introducing a restriction to prevent the cancellation of stock moves already linked to invoices or vendor bills. Only users who are members of the **"Allow to cancel stock moves linked to invoices/bills"** group will be authorized to perform this action.

## Why Restrict Cancelling Stock Moves Linked to Invoices/Bills

- **Data Integrity**: Prevents inconsistencies between inventory and accounting.
- **Controlled Access**: Cancellation is still allowed, but only for users in the designated security group.
- **Legal and Financial Compliance**: Avoids issues with invoiced or billed records.
- **Auditability**: Maintains a clear and traceable record of exceptional actions.
- **Best Practice**: Encourages using proper workflows like credit notes or returns.

## ROADMAP

- Create security group
- Restrict cancellation at the stock move model level
- Update translations
- Create a test

## Note on Commit Structure

This PR contains a single commit:
- **[IMP]** : restricts picking cancellation + fixes test helper to work with OCA test infrastructure

@BinhexTeam T12289


rrebollo:
would you be so kind to review this @edescalona @szalatyzuzanna @rvalyi. It's a forward port of #1992

mbcosta:
Hi @rrebollo thank you for the work, I did the recently refactoring in the Tests,

There are already some update needs and a debate about the use of the Form in post-init tests, I want to make a new PR about this after the merge of yours PR, you are right about **invoice_policy** field https://github.com/OCA/OCB/blob/18.0/addons/sale/models/product_template.py#L40 thank you, I probably missing this inherits field problem by being testing with sale_stock_picking_invocing module.

I saw just some small changes needed to adapt the tests, remove the method **_create_picking_with_invoice_not_done** and just change the method **picking_move_state** to get a new parameter **state** used to move the Picking in to a specific State, it's seems unnecessary create the user **stock_manager** because the **self.env.user** has the both Groups and can be used with the same propose and keeping the Tests names methods with the sequential number **def test_16_X, def test_17_X** this seems a simple thing but make easy to debug.

EDIT. The problem with **invoice_policy** is already be solved, so you can removed your commit about this and make a rebase

* https://github.com/OCA/account-invoicing/pull/2376

stock_picking_invoicing/tests/test_picking_invoicing.py
```diff
@@ -129,7 +133,7 @@ def test_05_picking_out_invoicing_backorder(self):
                 inv_line.price_unit, 345.0, "Error in Price Unit informed by User."
             )
 
-    def test_06_picking_cancel(self):
+    def test_invoice_cancel(self):
```
mbcosta: Keeping the test with the sequential number **test_06_invoice_cancel**

stock_picking_invoicing/tests/test_picking_invoicing.py
```diff
@@ -390,5 +394,47 @@ def test_15_picking_extra_vals(self):
             line.quantity = 10
 
         picking.button_validate()
-        with self.assertRaises(exceptions.UserError):
+        with self.assertRaises(UserError):
             create_with_form_inv_onshipping(self.env, self.env["stock.picking"])
+
+    def _create_picking_with_invoice_not_done(self):
```
mbcosta: You can remove this method and change the [picking_move_state](https://github.com/OCA/account-invoicing/blob/18.0/stock_picking_invoicing/tests/common.py#L126) to:

```bash
    def picking_move_state(self, picking, state="done"):
        picking.action_confirm()
        if state != "confirmed":
            # Check product availability
            picking.action_assign()
            # Force product availability
            with Form(picking) as picking_form:
                i = 0
                while i != len(picking.move_ids_without_package):
                    with picking_form.move_ids_without_package.edit(i) as line:
                        line.quantity = line.product_uom_qty
                    i += 1

            picking_form.save()
            picking.button_validate()
            self.assertEqual(picking.state, "done")
```
It's seems will be less code and with the refactoring done in the tests the idea is in this file should be has only tests this type of method should be in the common.py file because it make possible inherit and re-used.

stock_picking_invoicing/tests/test_picking_invoicing.py
```diff
@@ -390,5 +394,47 @@ def test_15_picking_extra_vals(self):
...
+        picking.set_to_be_invoiced()
+        picking.action_confirm()
+        create_with_form_inv_onshipping(self.env, picking)
+        self.assertTrue(picking.move_ids.invoice_line_ids)
+        self.assertNotEqual(picking.state, "done")
+        return picking
+
+    @mute_logger("odoo.addons.auth_signup.models.res_users")
+    def test_picking_cancel_without_permission(self):
+        """User without the cancel group cannot cancel a picking with invoiced moves."""
```
mbcosta, line 401: Keeping the test with the sequential number **test_16_picking_cancel_without_permission**, with the remove proposed of the method **_create_picking_with_invoice not done** this test can be like:

```bash
    @mute_logger("odoo.addons.auth_signup.models.res_users")
    def test_16_picking_cancel_without_permission(self):
        """User without the cancel group cannot cancel a picking with invoiced moves."""
        picking = self.picking_out_1
        picking.set_to_be_invoiced()
        self.picking_move_state(picking, "confirmed")
        # self.assertNotEqual(picking.state, "done")
        create_with_form_inv_onshipping(self.env, picking)
        regular_user = new_test_user(
            self.env,
            login="regular_user_no_cancel_perm",
            groups="stock.group_stock_user,account.group_account_invoice",
        )
        with self.assertRaises(UserError):
            picking.with_user(regular_user).action_cancel()
        self.assertNotEqual(picking.state, "cancel")
```

rvalyi:
the invoice_policy regression was just fixed so I guess you can rebase and remove that fix https://github.com/OCA/account-invoicing/pull/2376

stock_picking_invoicing/tests/test_picking_invoicing.py
```diff
@@ -390,5 +394,47 @@ def test_15_picking_extra_vals(self):
...
+            self.env,
+            login="regular_user_no_cancel_perm",
+            groups="stock.group_stock_user,account.group_account_invoice",
+        )
+        with self.assertRaises(UserError):
+            picking.with_user(regular_user).action_cancel()
+        self.assertNotEqual(picking.state, "cancel")
+
+    @mute_logger("odoo.addons.auth_signup.models.res_users")
+    def test_picking_cancel_with_permission(self):
```
mbcosta: Keeping the test with the sequential number **test_17_picking_cancel_with_permission**, with the remove proposed of the method **_create_picking_with_invoice_not_done** and the use of the User **self.env.user** instead of create a new one, this test can be like:

```bash
    @mute_logger("odoo.addons.auth_signup.models.res_users")
    def test_17_picking_cancel_with_permission(self):
        """Stock Manager can cancel a picking with invoiced, non-done moves."""
        picking = self.picking_out_1
        picking.set_to_be_invoiced()
        create_with_form_inv_onshipping(self.env, picking)
        self.picking_move_state(picking, "confirmed")
        # self.assertNotEqual(picking.state, "done")
        # self.assertTrue(self.env.user.has_group("stock.group_stock_manager"))
        # self.assertTrue(self.env.user.has_group("account.group_account_invoice"))
        picking.with_user(self.env.user).action_cancel()
        self.assertEqual(picking.state, "cancel")
```

I let some checks commented, just to be sure if everything is alright, if you think be necessary you can include or remove it.

rrebollo:
@rvalyi @mbcosta, thanks for the suggestions. I'll ping you both once it's resolved.

rrebollo:
ping @rvalyi

mbcosta (approved):
@rrebollo thank you for your work and for take the Review in consideration.

PR #2376, opened 2026-06-07
[FIX] stock_picking_invoicing: invoice_policy is from sale, and it is not a dependency...

etobella:
It has no sense that we try to set the field if it does not exist....


rvalyi (approved):
Thanks, I had to make the exact same fix.

BhaveshHeliconia (approved):
LGTM!

mbcosta:
Thank you @etobella @rvalyi , I probably missing this inherits field problem by being testing with sale_stock_picking_invocing module.

PR #2378, opened 2026-06-08
[FIX] account_invoice_show_currency_rate: ignore COGS lines in rate

juhenriquez:
The invoice currency rate was computed as
sum(abs(amount_currency)) / sum(abs(balance)) over every move line with a non-zero amount_currency. Anglo-Saxon COGS lines are kept in company currency (amount_currency == balance, so their per-line rate is 1.0; see account.move.line._compute_currency_id), so including them dragged the company -> document currency rate towards 1.0.

On a foreign-currency customer invoice with automated inventory valuation this produced a wrong invoice_currency_rate. In Odoo 18 this field is used by the core tax totals to convert the untaxed base into company currency, so the printed subtotal in company currency came out wrong while the tax amount (taken from the posted line balance) stayed correct.

Exclude COGS lines from the computation, as the core itself does in AccountMove._get_lines_onchange_currency.

### DIFF
--- a/account_invoice_show_currency_rate/tests/test_account_move.py
+++ b/account_invoice_show_currency_rate/tests/test_account_move.py
@@ -101,3 +101,39 @@ class TestAccountMove(common.TransactionCase):
         invoice = self.env["account.move"].new({"move_type": "entry"})
         # Simply check the method doesn't crash
         invoice._compute_invoice_currency_rate()
+
+    def test_04_currency_rate_ignores_cogs_lines(self):
+        """Anglo-Saxon COGS lines are kept in company currency (rate 1.0) and
+        must be ignored when computing the invoice currency rate. Otherwise they
+        distort the sum(amount_currency)/sum(balance) ratio, dragging the rate
+        towards 1.0. The core ignores them too (see
+        AccountMove._get_lines_onchange_currency)."""
+        self.partner.property_product_pricelist = self.pricelist_currency_extra
+        invoice = self._create_invoice(self.currency_extra)
+        self.assertAlmostEqual(invoice.invoice_currency_rate, 2.0, 2)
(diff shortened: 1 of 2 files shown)


account_invoice_show_currency_rate/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 {
     "name": "Account Invoice Show Currency Rate",
     "summary": "Show currency rate in invoices.",
-    "version": "18.0.1.0.1",
+    "version": "18.0.1.0.2",
```
victoralmau, line 6: You don't need to change the version; when the merge takes place, the bot will do it automatically.

account_invoice_show_currency_rate/tests/test_account_move.py
```diff
@@ -93,3 +93,41 @@ def test_02_invoice_currency_extra(self):
...
+                    "display_type": "cogs",
+                    "account_id": self.other_account.id,
+                    "currency_id": company_currency.id,
+                    "balance": -1000.0,
+                    "amount_currency": -1000.0,
+                },
+            ]
+        )
+        invoice.invalidate_recordset()
+        invoice._compute_invoice_currency_rate()
```
victoralmau:
```suggestion
```
In my opinion, this shouldn't be necessary; the computation should be triggered and set automatically.

victoralmau (changes requested):
Although I'm not familiar with COGS lines, the change seems reasonable.

account_invoice_show_currency_rate/__manifest__.py
victoralmau: This comment is still pending.

victoralmau (approved):

PR #2381, opened 2026-06-10
[18.0][IMP]  account_invoice_tax_required: Skip requirement with context

david-banon-tecnativa:
Allow skipping if some operation needs it EX questionable but valid and legal external invoice importing
TT56336
@Tecnativa


pedrobaeza (approved):

victoralmau (approved):

pedrobaeza:
As this is just adding the option to skip it, we can fast-track it:

/ocabot merge patch

PR #2385, opened 2026-06-12
[18.0][FIX+REF] sale_stock_picking_invoicing: Grouping Pickings with and without Sale Orders related and change to use Command

mbcosta:
Simple PR with two changes:

* **[FIX]** - If the User try to create a Grouping Pickings by Partner/Product with and without Sale Orders related, the variable **sale_line** can be **[(<Command.SET: 6>, 0, [])]**, in this case will occur the error:

```bash
2026-06-12 21:06:03,548 14 ERROR db odoo.http: Exception during request handling. 
  File "/opt/venv/lib/python3.12/site-packages/odoo/http.py", line 2577, in __call__
    response = request._serve_db()
               ^^^^^^^^^^^^^^^^^^^

...

  File "/home/odoo/app/external-src/account-invoicing/sale_stock_picking_invoicing/wizards/stock_invoice_onshipping.py", line 318, in _create_invoice
    line[2]["sequence"] = invoice_item_seq_dict.get(sale_line[0][2][0])
                                                    ~~~~~~~~~~~~~~~^^^
IndexError: list index out of range
```

It's possible reproduce the error with Demo Data, configure to Create Invoice from Stock Picking, Confirm the Sale Orders and Pickings, select all Pickings go to Actions and Create Draft Invoices

image

Define Group as Partner/Product

image

image

Test was included to check this case.

* **[REF]** - The second change is just to use Command, instead of use (0, 0, {}) and (6, 0, []) to use Command.create and Command.set

CC @rvalyi @renatonlima

### DIFF
--- a/sale_stock_picking_invoicing/tests/test_sale_stock.py
+++ b/sale_stock_picking_invoicing/tests/test_sale_stock.py
@@ -655,3 +655,24 @@ class TestSaleStockPickingInvoicing(TestSaleStockPickingInvoicingCommon):
         self.assertEqual(picking.invoice_state, "invoiced")
         for move in picking.move_ids:
             self.assertEqual(move.invoice_state, "invoiced")
+
+    def test_16_grouping_pickings_with_and_without_sale_order(self):
+        """
+        Test the case of Grouping Pickings by Partner/Product
+        with and without Sale Orders.
+        """
+        picking_1 = self.picking_out_1
+        picking_1.set_to_be_invoiced()
+        self.picking_move_state(picking_1)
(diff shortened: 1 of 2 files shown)


RicardCForgeFlow:
Tested:
Before fix:
image
image
After fix:
image

RicardCForgeFlow (approved):

mbcosta:
Thank you @RicardCForgeFlow for review, after the merge of your PR

* https://github.com/OCA/account-invoicing/pull/2392

I need to make a rebase and only change the name of the test from **"test_10_"** to **"test_11_"**.

PR #2387, opened 2026-06-17
[19.0][FIX] stock_picking_return_refund_option: Recompute all PO lines to correctly handle returns with extra invoiceable lines

Andrii9090-tecnativa:
Without these changes, using this module together with other modules, such as `delivery_purchase`, raises this error:

```
ERROR prod odoo.addons.stock_picking_return_refund_option.tests.test_stock_picking_return_refund_options: FAIL: TestSaleOrderLineInput.test_return_po_wo_to_refund
  File "/opt/odoo/auto/addons/stock_picking_return_refund_option/tests/test_stock_picking_return_refund_options.py", line 137, in test_return_po_wo_to_refund
    self.assertEqual(po_order.invoice_status, "no")
AssertionError: 'to invoice' != 'no'
- to invoice
+ no
```

This change is needed for real purchase returns with extra invoiceable lines, such as delivery costs.
When `to_refund` is changed on a return picking, the received quantity of the related purchase line is recomputed. But other lines on the same purchase order, for example delivery_purchase lines marked as is_delivery, may also depend on whether the normal product lines are still invoiceable.
So after recomputing `qty_received`, we also recompute `qty_to_invoice` for all lines of the purchase order.
Without this, a fully returned purchase can still remain "to invoice" because an auxiliary line keeps an outdated `qty_to_invoice`.

@pedrobaeza @victoralmau can you review please?

@Tecnativa

### DIFF
--- a/stock_picking_return_refund_option/models/stock_picking.py
+++ b/stock_picking_return_refund_option/models/stock_picking.py
@@ -74,4 +74,11 @@ class StockPicking(models.Model):
             po_lines = self.mapped("move_ids.purchase_line_id").filtered(
                 lambda x: x.product_id.invoice_policy in ("order", "delivery")
             )
+            # Check if exists is_delivery field in purchase.order.line that has
+            # been added by delivery module, this module has not dependency of this.
+            # Otherwise, the PO will have an inconsistent `invoice_status`.
+            if hasattr(self.env["purchase.order.line"], "is_delivery"):
+                po_lines |= self.mapped("purchase_id.order_line").filtered(
+                    lambda x: x.is_delivery
+                )
             po_lines._compute_qty_received()


pedrobaeza (changes requested):
Don't getting the relation between the invoiced quantity and changing this option.

Anyway, this requires a regression tests that exercises the problem that the patch wants to fix.

Andrii9090-tecnativa:
I tested this case.

Without the fix:
We have a PO with a delivery line and a refund invoice with Refund Options set to `To Refund`
image

The Billing Status is set to `Waiting Bills`. I think it’s not correct.
image

After fix:
image
Now Billins status set to `Nothing to Bill`
image

pedrobaeza:
`qty_invoiced` depends directly from `qty_received`, so it should be auto-computed when there's a change:

https://github.com/odoo/odoo/blob/6a356cb0640ce20c71f20a8bd64ef99e76a82489/addons/purchase/models/purchase_order_line.py#L164

so something more should be going on.

Andrii9090-tecnativa:
The approach has been changed. @pedrobaeza could you review it?

pedrobaeza (commented):
Seeing your patch, isn't this a problem in `delivery_purchase`, not handling correctly `_compute_qty_received` when `is_delivery`?

Andrii9090-tecnativa:
delivery.purchase does not override the `_compute_qty_received` method. I’ll do a quick test.

pedrobaeza:
Then which is the problem? I'm still not fully understanding. Maybe the problem is that `delivery_purchase` is not overwriting that method for handling `is_delivery` lines.

PR #2393, opened 2026-06-19
[19.0][FIX] account_invoice_pricelist: make secondary currency tests deterministic

eduezerouali-tecnativa:
cc @Tecnativa TT63195

The base data ships an EUR rate (1.2834) that skewed the USD -> EUR conversion in the pricelist tests. test_06 and test_07 assumed an EUR rate of 1.0, so the converted price unit no longer matched the expected values (100 / 1.1913 = 83.94 EUR), making the tests fail.

ping @pedrobaeza @victoralmau could you review please?

### DIFF
--- a/account_invoice_pricelist/tests/test_account_move_pricelist.py
+++ b/account_invoice_pricelist/tests/test_account_move_pricelist.py
@@ -328,15 +328,13 @@ class TestAccountMovePricelist(common.TransactionCase):
                 ],
             }
         )
-        # Fix currency rate of EUR -> USD to 1.5289
-        usd_currency = cls.env["res.currency"].search([("name", "=", "USD")])
-        usd_rates = cls.env["res.currency.rate"].search(
-            [("currency_id", "=", usd_currency.id)]
-        )
-        usd_rates.unlink()
+        # Fix currency rate of EUR -> USD to 1.1913
+        cls.env["res.currency.rate"].search(
+            [("currency_id", "in", (cls.usd_currency.id, cls.euro_currency.id))]
+        ).unlink()
         cls.env["res.currency.rate"].create(
             {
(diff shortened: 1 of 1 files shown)


account_invoice_pricelist/tests/test_account_move_pricelist.py
```diff
@@ -328,15 +328,13 @@ def setUpClass(cls):
...
-        )
-        usd_rates.unlink()
+        # Fix currency rate of EUR -> USD to 1.1913
+        cls.env["res.currency.rate"].search(
+            [("currency_id", "in", (cls.usd_currency.id, cls.euro_currency.id))]
+        ).unlink()
         cls.env["res.currency.rate"].create(
             {
-                "currency_id": usd_currency.id,
+                "currency_id": cls.usd_currency.id,
```
pedrobaeza, line 337: Are you sure the currency of the company in your environment is EUR? The test may force it if not.

eduezerouali-tecnativa: https://github.com/OCA/account-invoicing/blob/ef461d587903fca81d3474dd44e584b52aa672a5/account_invoice_pricelist/tests/test_account_move_pricelist.py#L194-L197 

The test forces the currency to USD. Because of that, we should unlink the EUR rate; otherwise, it will fall back to 1.2834 (of the demo data), while the test is expecting 1.1913. 

https://github.com/odoo/odoo/blob/00118002bd6eab2f4c34a32e993a9219fded06ac/odoo/addons/base/data/res_currency_rate_demo.xml#L753-L757

pedrobaeza (approved):

victoralmau (approved):

PR #2401, opened 2026-06-29
[16.0][FIX] account_invoice_show_currency_rate: keep booked rate on reset to draft

almumu:
When a foreign-currency invoice that was already posted is reset to draft, _compute_currency_rate fell back to the table rate and re-derived the line balances, introducing a per-line rounding difference that wrongly raised an "unbalanced entry" error and blocked the reset.

Keep the real booked rate also for moves that were posted before (posted_before) in both _compute_currency_rate and _compute_currency_rate_amount, so the balance is preserved and the entry can be reset to draft and re-posted without error. Add tests covering a multi-line foreign-currency vendor bill reset to draft and re-posted.

### DIFF
--- a/account_invoice_show_currency_rate/models/account_move.py
+++ b/account_invoice_show_currency_rate/models/account_move.py
@@ -18,6 +18,7 @@ class AccountMove(models.Model):
 
     @api.depends(
         "state",
+        "posted_before",
         "date",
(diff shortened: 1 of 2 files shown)


aritzolea (approved):

victoralmau (approved):
Code review OK

PR #2402, opened 2026-07-01
[FIX] partner_last_invoice_date: avoid error when partner is created directly from

cubells:
… invoice

When a customer is created directly from invoice form, an error is raised because "TypeError: 'NewId' object is not iterable".

After this changes, user can create partners directly in invoice form.

ping @LauraCForgeFlow


flotho (approved):
code review,
Thanks for the unit test
LGTM

PR #2408, opened 2026-07-15
[18.0][ADD] account_invoice_refund_link_sale: new module to link refunds to…

carlos-lopez-tecnativa:
… their original invoices from sales orders

TT63711
@Tecnativa @pedrobaeza @sergio-teruel could you please review this?


account_invoice_refund_link_sale/models/sale_order.py
```diff
@@ -0,0 +1,37 @@
...
+            if len(candidate_invoices) == 1:
+                refund.reversed_entry_id = candidate_invoices
+                candidate_invoices.message_post(
+                    body=self.env._(
+                        "This entry has been %s",
+                        refund._get_html_link(title=self.env._("reversed")),
+                    ),
+                )
+                refund.message_post(
+                    body=candidate_invoices._get_copy_message_content(
```
pedrobaeza, line 34: Isn't the message should be the contrary?

account_invoice_refund_link_sale/models/sale_order.py
```diff
@@ -0,0 +1,37 @@
...
+    def _try_link_refund_with_invoice(self, new_moves):
+        """Try to link the refund to the original invoice
+        if there is exactly one candidate."""
+        for refund in new_moves.filtered(lambda move: move.move_type == "out_refund"):
+            candidate_invoices = self.env["account.move"]
+            for line in refund.invoice_line_ids:
+                candidate_invoices |= line.sale_line_ids.invoice_lines.move_id.filtered(
+                    lambda move: move.state != "cancel"
+                    and move.move_type == "out_invoice"
+                )
```
pedrobaeza: The loop is not needed

```suggestion
            candidate_invoices = refund.invoice_line_ids.sale_line_ids.invoice_lines.move_id.filtered(
                lambda move: move.state != "cancel"
                and move.move_type == "out_invoice"
            )
```

account_invoice_refund_link_sale/models/sale_order.py
carlos-lopez-tecnativa: No. In the `account` module, when a refund is created from an invoice or an invoice is duplicated, the record that calls the `_get_copy_message_content` method is the original record, in this case, the invoice, not the refund.

https://github.com/odoo/odoo/blob/5cbcf18762f08439c2dfeed17317a391fe89f074/addons/account/models/account_move.py#L5053-L5062

https://github.com/odoo/odoo/blob/5cbcf18762f08439c2dfeed17317a391fe89f074/addons/account/models/account_move.py#L3346

pedrobaeza: Yeah, but I mean the generated text. Is it going to say "The entry has been reversed", while this is the reversal. Or am I wrong?

Andrii9090-tecnativa (approved):
Code&functional review👍

eduezerouali-tecnativa (approved):
Code review + tested on runboat

account_invoice_refund_link_sale/models/sale_order.py
pedrobaeza: ping

carlos-lopez-tecnativa: As you can see, I posted the same two messages that Odoo generates when you create a credit note from an invoice. Please test it on Runboat if you have any doubts.

1- On the invoice, the message is: `This entry has been reversed.`
image
2- On the credit note, the message is: `This entry has been reversed from INV/2026/00009`
image

pedrobaeza: Oh, great, the methods are smart enough! :smile:

sbidoul:
Name reserved on PyPI

PR #2413, opened 2026-07-24
[17.0][ADD] product_average_price_invoice: new module

eduezerouali-tecnativa:
cc @Tecnativa TT63810
ping @pedrobaeza @carlos-lopez-tecnativa


product_average_price_invoice/models/product_product.py
```diff
@@ -0,0 +1,113 @@
...
+        "scheduled action or the button on the product.",
+    )
+    average_sale_price = fields.Monetary(
+        string="ASP",
+        currency_field="currency_id",
+        readonly=True,
+        help="Average sale unit price computed from posted customer "
+        "invoices, without taxes or added costs. Stored value, refreshed by the "
+        "scheduled action or the button on the product.",
+    )
```
carlos-lopez-tecnativa: Please add copy=False to prevent these fields (as well as all new fields) from being copied when duplicating a product. Otherwise, the user might think that the copied values are valid for the new product.
```suggestion
    average_purchase_price = fields.Monetary(
        string="APP",
        currency_field="cost_currency_id",
        readonly=True,
        copy=False,
        help="Average purchase unit price computed from posted vendor "
        "bills, without taxes or added costs. Stored value, refreshed by the "
        "scheduled action or the button on the product.",
    )
    average_sale_price = fields.Monetary(
        string="ASP",
        currency_field="currency_id",
        readonly=True,
        copy=False,
        help="Average sale unit price computed from posted customer "
        "invoices, without taxes or added costs. Stored value, refreshed by the "
        "scheduled action or the button on the product.",
    )
```

product_average_price_invoice/models/product_template.py
```diff
@@ -0,0 +1,33 @@
...
+        help="Average purchase unit price computed from posted vendor "
+        "bills of all the template variants, without taxes or added costs.",
+    )
+    average_sale_price = fields.Monetary(
+        string="ASP",
+        currency_field="currency_id",
+        readonly=True,
+        help="Average sale unit price computed from posted customer "
+        "invoices of all the template variants, without taxes or added costs.",
+    )
```
carlos-lopez-tecnativa: The same here, copy=False

product_average_price_invoice/models/product_product.py
```diff
@@ -0,0 +1,113 @@
...
+        date_to = fields.Date.context_today(self)
+        step_days, max_steps = self._get_average_price_window()
+        date_from = date_to - timedelta(days=step_days * max_steps)
+        groups = self.env["account.move.line"].read_group(
+            [
+                ("parent_state", "=", "posted"),
+                ("display_type", "=", "product"),
+                ("move_id.move_type", "in", ["in_invoice", "out_invoice"]),
+                ("move_id.invoice_date", ">=", date_from),
+                ("company_id", "in", self.env.companies.ids),
```
carlos-lopez-tecnativa: The cron runs as the Odoo Bot user. Is it correct to use the companies from the environment in this case?

What happens if a product has a company set?

carlos-lopez-tecnativa (changes requested):

product_average_price_invoice/data/ir_cron_data.xml
```diff
@@ -0,0 +1,14 @@
...
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html). -->
+<odoo noupdate="1">
+    <record id="ir_cron_update_average_prices" model="ir.cron">
+        <field name="name">Update product average prices</field>
+        <field name="model_id" ref="product.model_product_product" />
+        <field name="state">code</field>
+        <field name="code">model._cron_update_average_prices()</field>
+        <field name="interval_number">1</field>
+        <field name="interval_type">days</field>
+        <field name="active" eval="True" />
```
pedrobaeza, line 17: There's a way to set a dynamic date set here to a specific hour for the next scheduled date, so we do it by default on the night, not when the module is installed.

product_average_price_invoice/models/product_average_price_mixin.py
```diff
@@ -0,0 +1,166 @@
...
+
+from odoo import _, api, fields, models
+from odoo.tools import format_date
+
+
+class ProductAveragePriceMixin(models.AbstractModel):
+    _name = "product.average.price.mixin"
+    _description = "Product Average Invoiced Price Mixin"
+
+    average_purchase_price_date_from = fields.Date(readonly=True)
```
pedrobaeza: As the field is very long, you can use `avg` as an allowed abbreviate.

product_average_price_invoice/models/product_average_price_mixin.py
```diff
@@ -0,0 +1,166 @@
...
+            max(int(get_param("product_average_price_invoice.step_days")), 1),
+            max(int(get_param("product_average_price_invoice.max_steps")), 1),
+        )
+
+    def _average_price_product_ids(self):
+        """Variant ids whose invoices feed this record's averages."""
+        raise NotImplementedError
+
+    @api.model
+    def _invoice_average_domain(self, product_ids, move_type, date_from, date_to):
```
pedrobaeza: Please avoid indiscriminate AI usage. `product_ids` shouldn't be passed as parameter, but whenever possible, pass recordsets.

product_average_price_invoice/models/product_average_price_mixin.py
```diff
@@ -0,0 +1,166 @@
...
+                    if not quantity:
+                        continue
+                    balance = sum(amount[0] for amount in amounts)
+                    record.write(
+                        self._average_price_values(
+                            move_type, sign * balance / quantity, date_from, date_to
+                        )
+                    )
+                    del pending[record]
+            for record in pending:
```
pedrobaeza: No double loop. Less readable.

product_average_price_invoice/models/product_average_price_mixin.py
```diff
@@ -0,0 +1,166 @@
...
+        }
+
+    def _update_average_prices(self, move_types=None):
+        date_to = fields.Date.context_today(self)
+        step_days, max_steps = self._get_average_price_window()
+        products = {record: record._average_price_product_ids() for record in self}
+        for move_type in move_types or ["in_invoice", "out_invoice"]:
+            # Sale amounts sit on the credit side, so their balance is negative.
+            sign = -1 if move_type == "out_invoice" else 1
+            pending = dict(products)
```
pedrobaeza: This AI generated thing doesn't sound good. You only need to do one loop and do `break` when already found results or till the maximum steps, and then write after that.

product_average_price_invoice/models/product_product.py
pedrobaeza: The module is right now not multi-company compliant. You should state it, and do the cron only for the current company, or convert the fields to company dependent, and iterate through any existing company. The second one is better, but it's not our case. Please check the remaining timesheet to implement it or put the roadmap. In any case, it shouldn't be much more.

product_average_price_invoice/models/res_config_settings.py
```diff
@@ -0,0 +1,21 @@
...
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    average_price_step_days = fields.Integer(
+        string="Average Price Step (days)",
+        default=30,
+        config_parameter="product_average_price_invoice.step_days",
```
pedrobaeza: This should be a value per company according the previous comment.

product_average_price_invoice/readme/ROADMAP.md
```diff
@@ -0,0 +1 @@
+- Refunds/credit notes are not taken into account.
```
pedrobaeza: This is not a known issue. It's the proper thing to do, as the average price in refunds don't count.

product_average_price_invoice/tests/test_product_average_price_invoice.py
```diff
@@ -0,0 +1,72 @@
...
+        self.assertAlmostEqual(product.average_purchase_price, 25.0)
+        self.assertEqual(product.average_purchase_price_date_to, self.today)
+        self.assertIn("From", product.average_purchase_price_period)
+        self.assertAlmostEqual(product.average_sale_price, 0.0)
+        # The sale button only refreshes the sale average.
+        product.action_update_average_sale_price()
+        self.assertAlmostEqual(product.average_sale_price, 50.0)
+        self.assertAlmostEqual(product.average_purchase_price, 25.0)
+
+    def test_no_data(self):
```
pedrobaeza: Do this as the first step before creating an invoice, not as a separate test.

product_average_price_invoice/models/product_average_price_mixin.py
```diff
@@ -0,0 +1,166 @@
...
+                        "start": format_date(self.env, date_from),
+                        "end": format_date(self.env, date_to),
+                    }
+                    if date_from and date_to
+                    else no_data
+                )
+
+    @api.model
+    def _average_price_field_prefix(self, move_type):
+        if move_type == "in_invoice":
```
pedrobaeza: This should be done with a simple dictionary mapping, not a method.

product_average_price_invoice/models/product_average_price_mixin.py
```diff
@@ -0,0 +1,166 @@
...
+            pending = dict(products)
+            for step in range(1, max_steps + 1):
+                if not pending:
+                    break
+                date_from = date_to - timedelta(days=step_days * step)
+                totals = {
+                    group["product_id"][0]: (group["balance"], group["quantity"])
+                    for group in self.env["account.move.line"].read_group(
+                        self._invoice_average_domain(
+                            [pid for ids in pending.values() for pid in ids],
```
pedrobaeza: This is not differentiating between templates and products, and `pid` is not very verbose.

pedrobaeza (changes requested):

eduezerouali-tecnativa:
Changes made:
- Now it is multi company support.
- As fields now are `company_dependent` now longer compatible with `currency_field` so moved them to the mixin.
- Now we relay on quantity for the loop instead of doing the break.
- Change to average to avg
- Add `nextcall` to cron

product_average_price_invoice/models/product_average_price_mixin.py
```diff
@@ -0,0 +1,172 @@
...
+    "in_invoice": "avg_purchase_price",
+    "out_invoice": "avg_sale_price",
+}
+
+
+class ProductAveragePriceMixin(models.AbstractModel):
+    _name = "product.average.price.mixin"
+    _description = "Product Average Invoiced Price Mixin"
+
+    avg_purchase_price = fields.Float(
```
pedrobaeza, line 18: You should keep them `Monetary`, and add `currency_id` in this mixin as a computed one (non stored), and put `@api.depends_context("allowed_company_ids")` and get `self.env.company`.

product_average_price_invoice/tests/test_product_average_price_invoice.py
```diff
@@ -0,0 +1,87 @@
...
+            }
+        )
+        invoice.action_post()
+        return invoice
+
+    def test_stored_after_buttons_independent(self):
+        product = self.product_a
+        # Without invoices there's neither average nor period.
+        product._update_avg_prices()
+        self.assertAlmostEqual(product.avg_purchase_price, 0.0)
```
pedrobaeza: Can it be `self.assertEqual` in all of them?

product_average_price_invoice/models/product_average_price_mixin.py
eduezerouali-tecnativa: It is not possible to set a `company_dependent` as `Monetary` at least in 17.0
https://github.com/odoo/odoo/blob/7676723dbb47c09a71925831b13ced97eb855297/odoo/addons/base/models/ir_property.py#L52 
Maybe in upper versions will be possible.

product_average_price_invoice/tests/test_product_average_price_invoice.py
eduezerouali-tecnativa: Done

product_average_price_invoice/models/product_average_price_mixin.py
pedrobaeza: Then can you put the widget at screen level and do the same for currency_id?

product_average_price_invoice/views/product_template_views.xml
```diff
@@ -0,0 +1,83 @@
...
+                    class="o_row"
+                    name="avg_purchase_price_row"
+                    groups="account.group_account_invoice"
+                >
+                    <field
+                        name="avg_purchase_price"
+                        class="oe_inline"
+                        readonly="1"
+                        widget="monetary"
+                        options="{'currency_field': 'cost_currency_id'}"
```
eduezerouali-tecnativa, line 27: @pedrobaeza already done when i did the change to float.

pedrobaeza: OK, sorry for the misunderstanding.

pedrobaeza (approved):

carlos-lopez-tecnativa (approved):

pedrobaeza:
@sbidoul would you?

sbidoul:
name reserved

PR #2419, opened 2026-07-31
[IMP] stock_account_move_reset_to_draft: Allow advanced users to reset the vendor bill

rousseldenis:
Even if there are stock valuations with modifications, some advanced accounting users may want to modify other information that do not have influence on stock valuations.

The user will be warned with a wizard about moves that are already consumed or intertwined and the vendor bill will be reset to draft without reversing stock valuations. Then, at revalidation, no stock valuations will be generated as well. This is maintained with a flag on stock move.

A warning is displayed on stock moves that have stock valuations and should be revalidated.

### DIFF
--- a/stock_account_move_reset_to_draft/models/account_move.py
+++ b/stock_account_move_reset_to_draft/models/account_move.py
@@ -1,63 +1,120 @@
 # Copyright 2024 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
-
-from odoo import _, models
-from odoo.exceptions import UserError
-from odoo.tools.float_utils import float_is_zero
+from odoo import _, api, fields, models
+from odoo.exceptions import AccessError, ValidationError
+from odoo.fields import Command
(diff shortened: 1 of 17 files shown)


rousseldenis:
@jbaudoux

stock_account_move_reset_to_draft/models/account_move.py
```diff
@@ -1,64 +1,121 @@
 # Copyright 2024 Tecnativa - Víctor Martínez
```
lmignon, line 1: @rousseldenis Your contribution deserve copyright to acsone ;-)

stock_account_move_reset_to_draft/models/account_move_line.py
```diff
@@ -0,0 +1,89 @@
+# Copyright 2024 Tecnativa - Víctor Martínez
```
lmignon, line 1: @rousseldenis Is-it a file from Tecnativa or are you the author?

lmignon (changes requested):

PR #2421, opened 2026-08-03
[19.0][FIX]account_global_discount: compute analytic distributions on generated global discount lines

adrip-s73:
Global discount invoice lines are deleted and recreated whenever invoice lines or global discounts are updated. As a result, any analytic distribution entered manually on a global discount line is lost after saving.
Additionally, generated discount lines did not explicitly trigger the standard analytic distribution computation. This could cause mandatory analytic validation errors even when a matching analytic distribution model existed.

Generated global discount lines are now explicitly created as product lines, and Odoo’s standard analytic distribution computation is triggered immediately after their creation.
Before regenerating existing discount lines, their analytic distributions are temporarily preserved using the global discount and tax combination as a key. Once the lines are recreated, any previous manual distribution is restored.
This supports both scenarios:
- Automatic assignment through analytic distribution models.
- Manual analytic distributions that remain unchanged after saving.

### DIFF
--- a/account_global_discount/models/account_move.py
+++ b/account_global_discount/models/account_move.py
@@ -159,7 +159,7 @@ class AccountMove(models.Model):
                     self.company_id,
                     self.date or fields.Date.context_today(self),
                 )
-            create_method(
+            discount_line = create_method(
                 {
                     "invoice_global_discount_id": discount.id,
                     "move_id": self.id,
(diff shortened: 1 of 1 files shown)


Reyes4711-S73 (approved):
LGTM

arantxa-s73 (approved):

PR #2423, opened 2026-08-12
[19.0][FIX] account_invoice_pricelist: don't reset price_unit on vendor bills when quantity changes

dannyadair:
Supersedes #2329

### Problem
`account_invoice_pricelist` added `quantity` to `_compute_price_unit`'s
`@api.depends`. Depends merge across the inheritance chain, so quantity
changes re-fired `super()._compute_price_unit()` on **all** move types,
including vendor bills — overwriting the PO-supplied `price_unit` with
`standard_price`. Full write-up and reproducer: #2329.

### Fix
Drop `quantity` from the compute depends. For sale lines only, schedule a
`price_unit` recompute from `modified()` when quantity changes (so volume
pricelist rules still apply). Scope `_apply_pricelist_to_price_unit` to
sale moves.
Also skip that reschedule while `price_unit` is protected (same write set
an explicit price), so a user-entered unit price is not overwritten by the
deferred pricelist recompute.

### Tests
- `test_vendor_bill_pricelist_does_not_overwrite_price_unit`
- `test_write_quantity_and_price_unit_preserves_explicit_price`
- `test_quantity_only_write_still_applies_pricelist`

### DIFF
--- a/account_invoice_pricelist/models/account_move.py
+++ b/account_invoice_pricelist/models/account_move.py
@@ -134,14 +134,20 @@ class AccountMoveLine(models.Model):
         # Reschedule price_unit on sale lines when quantity changes
         # post-create. We skip `create` so explicit `price_unit` in
         # create vals isn't clobbered by the pricelist.
+        # Also skip lines whose price_unit is protected (same write also
+        # set an explicit price) — otherwise the deferred recompute
+        # overwrites the user's price with the pricelist result (often
+        # 0 for products without a matching rule) after the write ends.
         res = super().modified(fnames, create=create, before=before)
         if "quantity" in fnames and not create:
+            price_unit_field = self._fields["price_unit"]
             sale_lines = self.filtered(
                 lambda line: line.move_id.move_type
                 in ("out_invoice", "out_refund", "out_receipt")
+                and not self.env.is_protected(price_unit_field, line)
             )
             if sale_lines:
-                self.env.add_to_compute(self._fields["price_unit"], sale_lines)
+                self.env.add_to_compute(price_unit_field, sale_lines)
         return res
 
     def _apply_pricelist_to_price_unit(self):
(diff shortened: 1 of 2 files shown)


account_invoice_pricelist/models/account_move.py
```diff
@@ -121,10 +121,43 @@ def _calculate_discount(self):
...
+                self.env.add_to_compute(price_unit_field, sale_lines)
+        return res
+
+    def _apply_pricelist_to_price_unit(self):
         for line in self:
+            if line.move_id.move_type not in (
+                "out_invoice",
+                "out_refund",
+                "out_receipt",
+            ):
```
nobuQuartile, line 159:
```suggestion
           if not line.move_id.is_sale_document(include_receipts=True):
                 continue
```

account_invoice_pricelist/models/account_move.py
```diff
@@ -121,10 +121,43 @@ def _calculate_discount(self):
...
+        # Also skip lines whose price_unit is protected (same write also
+        # set an explicit price) — otherwise the deferred recompute
+        # overwrites the user's price with the pricelist result (often
+        # 0 for products without a matching rule) after the write ends.
+        res = super().modified(fnames, create=create, before=before)
+        if "quantity" in fnames and not create:
+            price_unit_field = self._fields["price_unit"]
+            sale_lines = self.filtered(
+                lambda line: line.move_id.move_type
+                in ("out_invoice", "out_refund", "out_receipt")
```
nobuQuartile, line 146:
```suggestion
                lambda line: line.move_id.is_sale_document(include_receipts=True)
```

nobuQuartile (approved):
LGTM: Functional review
Minor point for the code

AungKoKoLin1997 (commented):
LG 👍

AungKoKoLin1997 (approved):
LG 👍

PR #2426, opened 2026-08-17
[FIX] purchase_create_bill_button: exclude zero qty lines from bill

jarcosmts:
When using 'Received Quantities' billing policy, clicking 'Create Bill' added all purchase order lines to the bill, including those with qty_to_invoice = 0. This is a regression in Odoo v19 core's action_create_invoice method that removed the float_is_zero check present in v18.

Override action_create_invoice to restore the v18 filtering behavior, ensuring only lines with pending quantities are included in the bill.

### DIFF
--- a/purchase_create_bill_button/__manifest__.py
+++ b/purchase_create_bill_button/__manifest__.py
@@ -4,7 +4,7 @@
 {
     "name": "Purchase Create Bill Button",
     "summary": "Add a direct button to create bills from purchase orders",
-    "version": "19.0.1.0.0",
+    "version": "19.0.1.1.0",
     "development_status": "Production/Stable",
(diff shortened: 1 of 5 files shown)


kdominguezmts (approved):
LGTM.

The override of `action_create_invoice` correctly restores the v18 `float_is_zero` filter, so lines with `qty_to_invoice = 0` are excluded when using the Received Quantities billing policy. Section/subsection handling and invoice grouping look consistent with core.

Tests cover both the exclusion of zero-qty lines and the inclusion of pending quantities, including the v19 partial-picking / backorder wizard path.

CI is green (pre-commit, Odoo, OCB). codecov remaining red does not look blocking for this fix.

PR #2439, opened 2026-09-08
[18.0][REF] account_billing: rename _compute_tax_totals to _compute_amount

AungKoKoLin1997:
The method computes `amount_untaxed`, `amount_tax` and `amount_total`, and the model has no `tax_totals` field at all, so the name is misleading: everywhere else in Odoo (`account.move`, `sale.order`, `purchase.order`) `_compute_tax_totals` is the compute of a `tax_totals` field, and `_compute_amount` is the one computing the amounts.

Beyond the naming, it also squats the name a module adding an actual tax_totals field needs. l10n_jp_summary_invoice does exactly that, and currently has to redeclare the three amount fields only to move them off the compute it needs to override.

@qrtl QT7203

### DIFF
--- a/account_billing/__manifest__.py
+++ b/account_billing/__manifest__.py
@@ -21,5 +21,5 @@
         "report/report.xml",
     ],
     "installable": True,
-    "maintainers": ["Saran440"],
+    "maintainers": ["Saran440", "AungKoKoLin1997"],
 }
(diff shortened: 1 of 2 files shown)


AungKoKoLin1997:
@Saran440 Could you please review this PR? It is just a renaming method name to align with odoo standard.

yostashiro (approved):
This makes the method name consistent with account.move's.

@Saran440 If you agree, it’d be better to merge this sooner, before anyone extends the method. 🙏

Saran440 (approved):

nobuQuartile (approved):
LGTM

PR #2445, opened 2026-09-14
[19.0][IMP] account_invoice_discount_date: due date and discount amounts

Ricardoalso:
Show the invoice/bill list a date to actually pay by, and the early payment discount totals, instead of requiring users to open each invoice to see them.

- Add effective_due_date, computed as discount_date if an Early
  Payment Discount applies, otherwise the regular due date. It
  replaces discount_date as the list-view column so the column
  never shows blank.
- Add discount_amount_currency and discount_balance on account.move,
  mirroring the account.move.line fields of the same name, and show
  them as new list columns (hidden when there is no discount).

- [ ] Requires rebase


ivantodorovich (approved):

imlopes (approved):

account_invoice_discount_date/models/account_move.py
```diff
@@ -15,6 +15,57 @@ class AccountMove(models.Model):
...
+    @api.depends(
+        "line_ids.discount_amount_currency",
+        "line_ids.amount_currency",
+        "line_ids.discount_balance",
+        "line_ids.balance",
+    )
+    def _compute_discount_amounts(self):
+        """Sum discount-or-plain amounts of payment term lines, only when
+        at least one of them grants an Early Payment Discount"""
+        for record in self:
```
SilvioC2C: Nitpicking:
```suggestion
        for move in self:
```
Giving the variables a more descriptive name makes the code more readable 😉

account_invoice_discount_date/models/account_move.py
```diff
@@ -15,6 +15,57 @@ class AccountMove(models.Model):
...
+                record.discount_amount_currency = sum(
+                    line.discount_amount_currency or line.amount_currency
+                    for line in payment_lines
+                )
+                record.discount_balance = sum(
+                    line.discount_balance or line.balance for line in payment_lines
+                )
+            else:
+                record.discount_amount_currency = False
+                record.discount_balance = False
```
SilvioC2C: Warnings:
1- you may want to use `account.move.line.currency_id`'s method `is_zero()` to check whether `discount_amount_currency` can be considered null
2- you may need to convert amounts from the invoice lines' currency (check if you need to use `currency_id` or `company_currency_id`) to the invoice's currency (same) to prevent issues

An example:
```suggestion
            discount_amount_currency, discount_balance = 0.0, 0.0
            if any(not p.currency_id.is_zero(p.discount_amount_currency) for p in payment_lines):
                move_currency = record.currency_id
                move_company_currency = record.company_currency_id
                for pay_line in payment_lines:
                    pay_line_currency = pay_line.currency_id
                    pay_line_company_currency = pay_line.company_currency_id
                    if pay_line_currency.is_zero(pay_line.discount_amount_currency):
                        pay_line_dac = pay_line.amount_currency
                    else:
                        pay_line_dac = pay_line.discount_amount_currency
                    discount_amount_currency += pay_line_currency._convert(
                        pay_line_dac,
                        move_currency,
                        round=False,  # Let ``fields.Monetary`` do the rounding
                    )
                    if pay_line_company_currency.is_zero(pay_line.discount_balance):
                        pay_line_db = pay_line.balance
                    else:
                        pay_line_db = pay_line.discount_balance
                    discount_balance += pay_line_company_currency._convert(
                        pay_line_db,
                        move_company_currency,
                        round=False,  # Let ``fields.Monetary`` do the rounding
                    )
            record.discount_amount_currency = discount_amount_currency
            record.discount_balance = discount_balance
```

NB: this will require an update to the `@api.depends()` fields too (e.g.: `line_ids.display_type`, `currency_id`, `company_currency_id`, `line_ids.currency_id`, `line_ids.company_currency_id`).

SilvioC2C (approved):
Overall LGTM, pre-approving: please handle currencies in the compute method (and add a test about it)

account_invoice_discount_date/models/account_move.py
ivantodorovich: Isn't it safe to assume the line currency is the same to the move currency?

Standard module does pretty much the same:
https://github.com/odoo/odoo/blob/11637aa56b1ba84aba74f4259777e088d530f658/addons/account/models/account_move.py#L1211-L1214

ivantodorovich:
PR Updated:
- dropped the effective_due_date (kept only discount_date, that already matches the line-level field)
- matched field labesl to those of standard's line-level fields
- reorganized the fields in the list view

account_invoice_discount_date/models/account_move.py
SilvioC2C: I double-checked:
- the `account.move[.line].*balance` fields are linked to `account.move[.line].company_currency_id`, and it's safe to assume the move and its lines share the same currency, since `account.move.line.company_currency_id` is defined as related (and readonly) to the move's company currency (see: https://github.com/odoo/odoo/blob/11637aa56b1ba84aba74f4259777e088d530f658/addons/account/models/account_move_line.py#L60-L63)
- however, the `account.move[.line].*amount_currency` fields are linked to `account.move[.line].currency_id`, and we cannot be sure that they match:
  - both moves and move lines' `currency_id` fields are computed, but also editable
  - their computed methods may assign different currencies to a move and its move lines
  - `account.move.currency_id` has an inverse method that aligns its lines' currency to the move's currency, but it does not apply to moves with `move_type == "entry"` 
  - as a result of these behaviors (and maybe others), there might exist tax lines, CoGS lines, balance lines, rounding lines, etc, whose `currency_id` differs from their move's `currency_id`

In general, though, I would rather play it safe when it comes to amounts in currency, and always try to convert them when possible. After all, if the currency from which we're converting an amount is the same as the one we're converting the amount to, Odoo simply returns a [conversion rate of 1](https://github.com/odoo/odoo/blob/11637aa56b1ba84aba74f4259777e088d530f658/odoo/addons/base/models/res_currency.py#L273-L275) and cuts it short; not a big loss, performance-wise :smile:

Ricardoalso: Thanks for the reviews :pray: 

I did indeed take the same approach as in the standard, but I also agree that we can play it safe and convert it, even if we’re applying the same conversion. It adds a layer of security for low computational cost.
I’ll adapt the code accordingly :wink:

PR #2447, opened 2026-09-17
[18.0][FIX] sale_invoiceability_on_payment_transaction: Fix amount_to_invoice values from tests

victoralmau:
Fix `amount_to_invoice` values from tests

Related to https://github.com/odoo/odoo/commit/c48627a25ece07f075b0c1d1c5889049378c4959

Please @pedrobaeza and @carlos-lopez-tecnativa can you review it?

@Tecnativa


carlos-lopez-tecnativa (approved):

pedrobaeza (commented):
The `untaxed_amount_to_invoice` is 0, but `amount_to_invoice` is not, so this is not correct. I think the module should change, but the tests reflect the reality of what we want to achieve.

carlos-lopez-tecnativa:
> The `untaxed_amount_to_invoice` is 0, but `amount_to_invoice` is not, so this is not correct. I think the module should change, but the tests reflect the reality of what we want to achieve.

You are right. If the sale order is not paid, `amount_to_invoice` should be zero. The change in Odoo, which now uses `product_uom_qty` instead of `qty_delivered`, causes the error here because this module calls `super()` when the sale order is not paid. However, the correct approach is to adapt the code instead of changing the tests.

victoralmau:
> The `untaxed_amount_to_invoice` is 0, but `amount_to_invoice` is not, so this is not correct. I think the module should change, but the tests reflect the reality of what we want to achieve.

I understand what you're saying, but at https://github.com/odoo/odoo/commit/c48627a25ece07f075b0c1d1c5889049378c4959, only `amount_to_invoice` has been changed; therefore, the `untaxed_amount_to_invoice` field (in Odoo) is inconsistent.

victoralmau:
> > The `untaxed_amount_to_invoice` is 0, but `amount_to_invoice` is not, so this is not correct. I think the module should change, but the tests reflect the reality of what we want to achieve.
> 
> You are right. If the sale order is not paid, `amount_to_invoice` should be zero. The change in Odoo, which now uses `product_uom_qty` instead of `qty_delivered`, causes the error here because this module calls `super()` when the sale order is not paid. However, the correct approach is to adapt the code instead of changing the tests.

The problem isn't that `super` is being called; the problem is that Odoo, at https://github.com/odoo/odoo/commit/c48627a25ece07f075b0c1d1c5889049378c4959, has changed the way the `amount_to_invoice` field is calculated (without using `qty_delivered`), but the `_compute_untaxed_amount_to_invoice()` method (in `sale`) still uses `qty_delivered`, which is why there’s this inconsistency.

So are you suggesting that in this module we make the field consistent with what Odoo should have in `sale`? I don’t quite understand.

pedrobaeza:
Well, the first thing is to write in the Odoo commit to highlight the inconsistency to see what Odoo is going to do, and then act accordingly.

